PR #480, opened 2020-10-15
Rename Cloud Foundry Paketo -> Paketo

ashwin-venkatesh:
While the Paketo buildpacks project is in the CFF, their OSS name is just Paketo Buildpacks.

CC @sclevine


pearkes:
Thanks!

PR #500, opened 2020-10-15
nix: Use niv to easily pin and update to a known version of nixpkgs

arianvp:
Also adds a github action that automatically opens a PR (or updates an
existing PR) every time nixpkgs is updated.

You can comment out the Cachxi parts to get a shared binary cache for your custom pinned version of tools using https://cachix.org

More elaborate CI integration, like running git pre-commit hooks could be added.  But I left it as bare-bones as needed for now. See https://github.com/nix-dot-dev/getting-started-nix-template for inspiration.


nix/default.nix
```diff
@@ -0,0 +1,19 @@
+let
+  sources = import ./sources.nix;
+in
+import sources.nixpkgs {
+  overlays = [
+    (self: super: {
+      # niv = (import sources.niv { pkgs = self; }).niv;
```
arianvp, line 7: Could uncomment this to get the latest version of niv, instead of the one in the nixpkgs snapshot

.github/workflows/update-nix.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  update-dependencies:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2.3.3
+    - uses: cachix/install-nix-action@v11
+    # https://github.com/nmattia/niv/issues/280
+    - run: GITHUB_PATH= nix-shell --run "niv update"
+    # TODO: Up to the authors of Waypoint to set up a binary cache at cachix.org
+    # - uses: cachix/cachix-action@v6
```
arianvp, line 17: Could uncomment this to populate a shared binary cache at  waypoint.cachix.org. Highly recommended for onboarding other collegues as it reduces initial time for `nix-shell` to load

evanphx:
I definitely don't want a github action that pings us on every nixpkg update.

mitchellh:
@evanphx yes you do. 😆 (no I won’t)

Thank you! I’ll take a look at this soon. I use niv for my local nix stuff so I’m somewhat familiar but this will be educational.

arianvp:
Yeh feel free to leave the cron job out enitrely, (or maybe run it less frequently :)). I added a workflow that checks if master an pull requests provide a valid nix shell now  too. That's probably more useful.

mitchellh:
Thank you! I removed the update cron thing for now we'll do that manually, but I appreciate this a lot!

PR #502, opened 2020-10-15
Use latest puma/ruby in demo Terminal

drnic:
Always good to upgrade to new versions, even if its fake terminal output lol

### DIFF
--- a/website/pages/home/index.jsx
+++ b/website/pages/home/index.jsx
@@ -529,7 +529,7 @@ export default function HomePage() {
                     },
                     {
                       code:
-                        '[11] * Version 3.11.2 (ruby 2.6.6-p146), codename: Love Song',
+                        '[11] * Version 5.0.2 (ruby 2.7.1-p83), codename: Spoony Bard',
                       color: 'gray',
                     },
                     {


pearkes:
Unfortunately it's not fake 😱. Feel like making the companion PR over there? That one I think has more significant consequences for folks! 

https://github.com/hashicorp/waypoint-examples/blob/4ddffea619108b9e34e6b19589978ed876f2da5b/docker/ruby/Gemfile.lock#L95

pearkes:
Thanks!

drnic:
@pearkes  Yeah, I can fix the example app. Where does your ruby 2.6 come from? Your Gemfile looks like it accepts ruby 2.7 `ruby '>= 2.3.5', '< 2.8'` but in this marketing website you're seeing ruby 2.6 (before I hacked it to be 2.7)?

pearkes:
@drnic thanks for your help getting this fixed upstream!

> Update: from waypoint.hcl I see you're using pack/buildpacks -- where do I see which buildpack.io builder + included buildpacks you're using?

This is hard to find but is documented [here](https://www.waypointproject.io/docs/lifecycle/build#configuring-an-alternative-buildpack-provider). As you may know by now you can specify it explicitly, but we also have defaults.

PR #507, opened 2020-10-15
Add details for go-bindata as a build pre-req

jbayer:


mitchellh:
Wouldn't be a problem if people used Nix 😆

Issue #513, opened 2020-10-15
InvalidParameterException trying to use an existing ECS cluster via the `cluster` parameter


kylegalbraith:
**Describe the bug**
I have a static ECS cluster that was provisioned via Terraform along with the ECR repository for my container image. I then tried to use the `waypoint.hcl` file seen below and specify the name of the cluster via the `cluster` parameter. Based on [this documentation](https://www.waypointproject.io/plugins/aws-ecs#cluster), Waypoint should recognize the cluster already exists and skip trying to create it and just deploy to it.

**Steps to Reproduce**
1. Deploy the `waypoint.hcl` below with the ECS cluster, `waypoint-nextjs-cluster`, already created in your AWS account.
```hcl
project = "example-next-ecs"

app "next-ecs" {
  build {
    use "pack" {}
    registry {
      use "aws-ecr" {
        region     = "us-west-2"
        repository = "nextjs-image"
        tag        = "latest"
      }
    }

  }

  # Deploy to ECS
  deploy {
    use "aws-ecs" {
      region  = "us-west-2"
      cluster = "waypoint-nextjs-cluster"
      memory  = "512"
    }
  }
}

```

With the existing cluster already provisioned, Waypoint throws the following error when you run `up`.

```
» Building...
Creating new buildpack-based image using builder: heroku/buildpacks:18
✓ Creating pack client
✓ Building image
 │ [exporter] Reusing layer 'config'
 │ [exporter] Adding label 'io.buildpacks.lifecycle.metadata'
 │ [exporter] Adding label 'io.buildpacks.build.metadata'
 │ [exporter] Adding label 'io.buildpacks.project.metadata'
 │ [exporter] *** Images (b3e6e24ef072):
 │ [exporter]       index.docker.io/library/next-ecs:latest
 │ [exporter] Reusing cache layer 'heroku/nodejs-engine:nodejs'
 │ [exporter] Reusing cache layer 'heroku/nodejs-engine:toolbox'
 │ [exporter] Reusing cache layer 'heroku/nodejs-engine:yarn'
 │ [exporter] Reusing cache layer 'heroku/nodejs-yarn:node_modules'
✓ Injecting entrypoint binary to image

Generated new Docker image: next-ecs:latest
Creating new buildpack-based image using builder: heroku/buildpacks:18✓ Creating pack client
✓ Building image
 │ [exporter] Reusing layer 'config'
 │ [exporter] Adding label 'io.buildpacks.lifecycle.metadata'
[...]

PR #514, opened 2020-10-15
Handle case where there is only one ECS cluster with a given name

kylegalbraith:
Closes: https://github.com/hashicorp/waypoint/issues/513

This will allow you to use an existing ECS cluster with Waypoint even when there is only one ECS cluster in the AWS account.


evanphx:
Great find! Thank you!

Issue #284, opened 2020-09-29
Confirm waypoint destroy


catsby:
`waypoint destroy` worked great and was very quick. It was over before I had the chance to think to myself what the consequences were. I would like to contribute a confirmation step if others agree it would be useful, similar to `terraform destroy`.

PR #522, opened 2020-10-16
Interactively ask user before destroying app

jacksgt:
_Thanks for this really nice project! To a bit familiar with the codebase, I decided to pick up this simple issue._

To make sure users don't accidentally delete their deployment, ask for
confirmation before doing so (similar to terraform destroy).
This patch also adds a flag (-auto-approve) to override this
interactive prompt (e.g. for scripting).

Fixes https://github.com/hashicorp/waypoint/issues/284

### DIFF
--- a/internal/cli/destroy.go
+++ b/internal/cli/destroy.go
@@ -27,7 +27,7 @@ func (c *DestroyCommand) Run(args []string) int {
 	}
 
 	err := c.DoApp(c.Ctx, func(ctx context.Context, app *clientpkg.App) error {
-		if !c.flagAutoApprove {
+		if !c.flagForce {
 			// show confirmation dialog asking user for approval
 			choice, err := app.UI.Input(&terminal.Input{
 				Prompt: "Are you sure you want to destroy this app? Type 'yes' to confirm.",
--- a/internal/cli/destroy.go
+++ b/internal/cli/destroy.go
@@ -5,11 +5,11 @@ import (
 
 	"github.com/golang/protobuf/ptypes/empty"
 
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	clientpkg "github.com/hashicorp/waypoint/internal/client"
 	"github.com/hashicorp/waypoint/internal/clierrors"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
-	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 )
 
 type DestroyCommand struct {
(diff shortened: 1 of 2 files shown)


mitchellh:
So, we can't merge this yet because `app.UI.Input` actually doesn't do anything yet. lol. That's a TODO item for sure...

Beyond that, I'd say it'd be nice if this UX matched `terraform` and used `-force` instead of `-auto-approve`.

jacksgt:
> So, we can't merge this yet because `app.UI.Input` actually doesn't do anything yet. lol. That's a TODO item for sure...
:D

> Beyond that, I'd say it'd be nice if this UX matched `terraform` and used `-force` instead of `-auto-approve`.
Isn't Terraform using `-auto-approve` for `destroy`? https://www.terraform.io/docs/commands/destroy.html

akin-ozer:
@jacksgt yes latest terraform suggests "-force Deprecated: same as auto-approve." in CLI for "destroy" command.

jacksgt:
Thanks for your input. I have changed the CLI flag from `-auto-approve` to `-force`.

I was looking into implementing the interactive prompt and it seems that the lack of this feature is stemming from https://github.com/mitchellh/go-glint

However, I don't understand why the code in https://github.com/hashicorp/waypoint-plugin-sdk/blob/main/terminal/basic.go#L50 is not used instead of https://github.com/hashicorp/waypoint-plugin-sdk/blob/main/terminal/glint.go#L34 ?

internal/cli/destroy.go
```diff
@@ -27,6 +27,21 @@ func (c *DestroyCommand) Run(args []string) int {
...
+				Style:  terminal.WarningStyle,
+			})
+			if err != nil {
+				return err
+			}
+			if choice != "yes" {
+				app.UI.Output("Aborting destroy.")
+				return nil
+			}
+		}
```
briancain, line 43: Since it looks like at the moment our App UI only implements noninteractive, this condition branch should instead return an error to the user that `auto-confirm` is required for destroying the app.

```suggestion
		if !c.confirm {
			app.UI.Output("Destroying app %q requires confirmation with `-auto-approve`.", app.Ref().GetApplication(), terminal.WithWarningStyle())
			return nil
		}
```

briancain (changes requested):
Thanks @jacksgt for taking on this issue and opening a pull request! Sorry for the confusion about `force` versus `auto-approve`. Let me know if you have any questions addressing the feedback here. Sorry for taking so long to get back to you! :heart:

briancain:
I couldn't seem to get git to let me push to the original fork, so I've opened a new pull request to resolve this one! https://github.com/hashicorp/waypoint/pull/1232

Thanks again for taking the time to open a pull request and solve this issue.

Issue #538, opened 2020-10-16
Docker(builder) plugin: Allow passing the path to a Dockerfile


sharkyze:
**Is your feature request related to a problem? Please describe.**
We are using a monorepo for all of our Go services, where the Dockerfile of each service is in `./cmd/service-name/Dockerfile`. In the current implementation of the Docker builder, it does not seem possible to specify the location of the Dockerfile if it is not in the root of the project.

**Describe the solution you'd like**
Be able to set a variable with the path to the Dockerfile in the Docker builder

For example:
```hcl
build {
        use "docker" {
            dockerfile = "./cmd/api/Dockerfile"
        }
}
```

**Describe alternatives you've considered**
The suggested solution is the only solution I could come with.

**Additional context**
More info here: https://discuss.hashicorp.com/t/multiple-apps-and-build-contexts/15987

PR #539, opened 2020-10-16
builtin/docker/builder: Add dockerfile variable to BuilderConfig

sharkyze:
Closes https://github.com/hashicorp/waypoint/issues/538

Add the variable `Dockerfile` to the `BuilderConfig` of the Docker Plugin to support cases where the Dockerfile is not in the root of the project.

Please let me know if you think the variable should have another name or if you think the dockerfile name shouldn't be relative to the path of the app.

### DIFF
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 
 	"github.com/docker/cli/cli/command/image/build"
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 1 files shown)


mitchellh:
Looks great! I think in the future the only ask I'd have is to update the `Docs` function too. But I'll do that manually now.

sharkyze:
> Looks great! I think in the future the only ask I'd have is to update the `Docs` function too. But I'll do that manually now.

Great thanks. Sure I will make sure to update the `Docs` function for any future contributions.

PR #556, opened 2020-10-16
Disable cache on local fetch for downloads page

BrandonRomano:


BrandonRomano:
Actually going to just switch this to build time instead... Will open up a new PR here.

PR #559, opened 2020-10-16
Switch all releases to build time fetch

BrandonRomano:


website/pages/downloads/index.jsx
```diff
@@ -70,24 +78,54 @@ export default function DownloadsPage({ releaseData, previousVersions }) {
         brand="blue"
         productId="waypoint"
         productName={productName}
-        releases={previousVersions}
-        latestVersion={releaseData.version}
+        releases={allVersionsReleaseData}
+        latestVersion={currentVersionReleaseData.version}
```
ztanner, line 82: I think this value is undefined (check out the dropdown menu) - but we can probably use `version` here instead.

PR #575, opened 2020-10-17
docs: fix string example for aws-ecs memory

nickatsegment:
since above it is described as an `int`, as well as in the Config struct https://github.com/hashicorp/waypoint/blob/cabbaf7d25c17a9dd94b5576f597bc8a986e6e07/builtin/aws/ecs/platform.go#L1130

In my testing, strings do work though :shrug:


evanphx:
This should be changed in the ecs plugin itself because these docs are autogenerated from the plugin. It's in the example part of the platform.go file.

mitchellh:
I got the Go fix.

PR #581, opened 2020-10-17
Add support for environment variables with pack build

aranw:
This PR adds support for environment variables with the pack build phase 

Some buildpacks allow configuration via environment variables such as https://github.com/GoogleCloudPlatform/buildpacks see https://github.com/GoogleCloudPlatform/buildpacks#common-options for available environment variables

I copied the configuration style from the docker build stage so configuration looks like the following: 

```
    build {
        use "pack" {
            builder     = "gcr.io/buildpacks/builder"
            static_environment = {
                "GOOGLE_BUILDABLE" : "./cmd/webhook"
            }
        }
    }
```

I've tested it locally and can see from the output that it used my environment variable as expected 
```
 │ [builder] Running "go build -o /layers/google.go.build/bin/main ./cmd/webhook (GOCACHE=/layers/google.go.build/gocache)"
```

### DIFF
--- a/builtin/pack/builder.go
+++ b/builtin/pack/builder.go
@@ -249,7 +249,7 @@ build {
 
 	doc.SetField(
 		"static_environment",
-		"environment variables to expose to the application",
+		"environment variables to expose to the buildpack",
 		docs.Summary(
 			"these environment variables should not be run of the mill",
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
This looks great thank you!

PR #583, opened 2020-10-17
Updated options for ECS deploys

surminus:
First of all, I'd like to say how excited I am about this project! As an ECS user, while there are multiple deploy tools available, I haven't found anything that feels flexible enough to meet our requirements. The hooks feel especially beneficial. 

I'm really interested in the design choices around running an application because there are lots of opinionated tools that lack the flexibility we need. At the moment deploying into ECS with Waypoint assumes a bunch of stuff (single container on a specified port), and I'm assuming that this is mostly because it's an early release. 

I guess my main question here is: how much flexibility between the configuration and the "assumed defaults" are going to be a feature of Waypoint? Will it be possible to deploy into a very specific ECS setup, or is the point that it will be generic to allow deployments to multiple plugins?

This pull request adds a little bit of code to increase the flexibility of the ECS plugin, to:

1. Add some configuration options to the _main_ container
2. Allow running additional containers alongside

Appreciate this project is still super new, but I couldn't help myself diving in, particularly since it's something that could potentially solve a problem for us!

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -547,6 +547,21 @@ func (p *Platform) Launch(
 		},
 	}
 
+	for k, v := range p.config.Environment {
+		env = append(env, &ecs.KeyValuePair{
+			Name:  aws.String(k),
+			Value: aws.String(v),
+		})
+	}
+
+	var secrets []*ecs.Secret
+	for k, v := range p.config.Secrets {
+		secrets = append(secrets, &ecs.Secret{
+			Name:      aws.String(k),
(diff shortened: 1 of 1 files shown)


mitchellh (commented):
Thank you! I agree I think Waypoint has huge promise for ECS users. This looks really good! I just want to discuss this with some other team members before merging.

mitchellh:
Okay! This looks awesome. We're gonna change some of the HCL attribute names a bit but otherwise we're very happy with this. thank you!

surminus:
Great! Pleased I could contribute :)

PR #586, opened 2020-10-17
Add CGO_ENABLED to entrypoint

jbayer:
alpine docker images and scratch images [were not working](https://discuss.hashicorp.com/t/how-do-you-build-a-project-with-a-dockerfile-that-has-its-own-entrypoint/16086/8) until CGO_ENABLED was used in the build for `waypoint-entrypoint`.


ekrem95 (approved):

mitchellh (changes requested):
For the entry point we should set cgo to be disabled always

Issue #516, opened 2020-10-16
Nomad Plugin - allow specification of an hcl file


akio-outori:
Waypoint looks really interesting but isn't quite flexible enough for us yet (definitely understand its a day 1 release).  Of particular note for us is being able to get deployment feedback from Nomad (e.g., whether allocations of a job actually enter a running state).

One of the big items that would block adoption for us is being able to use existing Nomad job files, and I wouldn't necessarily expect (or want to wait for) abstraction of all nomad options.  Instead, what about allowing users to specify a path to an hcl file to use?

That seems like it would lower overhead for feature requests while allowing lots of custom setups to work without too much modification.

Example of one of our local dev nomad jobs below:

```
job "edge-server" {

  datacenters = ["local"]
  type        = "service"
  
  group "edge-server" {

    count = 1

    update {
      max_parallel     = 1
      auto_revert      = "true"
      min_healthy_time = "30s"
      healthy_deadline = "2m"
    }

    # scratch space for all tasks in this group, must include space
    # for all the logs (100 MB by default)
    ephemeral_disk {
      size = 200 # MB
    }

    task "edge-server" {

      driver = "docker"

      config {
        image        = "density/edge-server:0"
        network_mode = "host"

        logging {
          type = "fluentd"
          config {
            fluentd-async-connect = true
            fluentd-address       = "${attr.unique.network.ip-address}:24224"
            tag                   = "edge-server"
          }
        }
      }

      # cgroup-limited resources for the container on the host
      resources {
        cpu    = 256
        memory = 512
        network {
          port "http" {
            static = 80
          }
          port "metrics" {
            static = 8888
          }
        }
      }

      service {
        name = "edge-server"
        port = "http"
        
        check {
          type     = "http"
          port     = "http"
          path     = "/health"
          interval = "10s"
          timeout  = "2s"
        }

        check {
          type     = "script"
          name     = "check_config"
          command  = "/bin/nginx"
          args     = ["-T"]
          interval = "60s"
          timeout  = "5s"
        }
      }

    }
  }
}
```

attachmentgenie:
and as an elevated version of this, levant integration, so that we can pass variables to the nomad script

orenmazor:
I can't help but assume somebody is out there at this very moment hacking a levant plugin, and I wish I could buy that person a beer

kaykelins:
you can use exec plugin to deploy the job file

orenmazor:
[...]

PR #587, opened 2020-10-17
Add ability to pass josbpec path parameter to nomad plugin

rkrasiuk:
Addresses #516 

I had multiple questions which I thought better to discuss with maintainers rather than deciding on my own

* Should the application throw on other parameters provided together with `jobspec`?
* Should any other meta tags be set on the job that are specific to how waypoint manages the deployments? 

And, in general, any other comments appreciated.

### DIFF
--- a/builtin/nomad/platform.go
+++ b/builtin/nomad/platform.go
@@ -3,6 +3,7 @@ package nomad
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mitchellh:
Thank you for taking a stab at this! I think there are some open questions I’d love to discuss both here and internally about this. Before that, I want to say that we 100% want to support a job spec oriented plugin. That is not in question.

The question is whether this fits in the existing `use “nomad”` plugin or if we should make a separate plugin. Or, if this should support some templating so we can get the Docker image created into it for example.

Its the weekend so I just wanted to drop this by but I’ll think about this and bring it up with the team Monday.

rkrasiuk:
@mitchellh sure! thanks for a quick weekend response. i'd love `waypoint` to eventually support `levant` as well and, with that said, separate plugin, smth like `use "nomad-template"`, sounds more logical. just tell me what you decide and how i can help ; -)

evanphx:
Hi @rkrasiuk, thanks again for the PR! We're solving this problem a different way and it should be available in 0.4.0.

rkrasiuk:
@evanphx excited to see the solution. thanks for the update!

mitchellh:
I'll keep this open for now since the PR isn't quite merged, but here is the PR to watch: https://github.com/hashicorp/waypoint/pull/1299

mitchellh:
Oh this is a PR. Brain confusion. Anyways, that link to the PR is right. I will close this haha

PR #592, opened 2020-10-17
Replace Deprecated AWS session.New

alrs:
From the `godoc` for `github.com/aws/aws-sdk-go/aws/session`:
> Deprecated: Use NewSession functions to create sessions instead. NewSession has the same functionality as New except an error can be returned when the func is called instead of waiting to receive an error until a request is made.

This replaces `session.New()` with `session.NewSession()` throughout `builtin/aws`. 

I'm happy to squash this into a single commit, if that is preferred.


mitchellh:
Super clean thanks, I can squash it on my end.

PR #602, opened 2020-10-17
Fix plugin godoc copy paste

eddiezane:
This PR fixes some godoc in the plugins that was copy pasted when created.


mitchellh:
Thanks!

PR #605, opened 2020-10-18
Check for active ECS cluster

eddiezane:
As far as I can tell, when an ECS user deletes a cluster it does not actually remove all references to it from the API. The describe-clusters endpoint still returns the deleted cluster with the status of `INACTIVE`.

After deleting the cluster myself and attempting to test a different change I noticed waypoint logged that the cluster existed and skipped creation later on failing for a non existent cluster.

This PR adds a check for this status before determining if the cluster is found.

I have tested this change for both existing and deleted clusters.


mitchellh:
Thanks awesome

PR #613, opened 2020-10-18
Fix deploy docker command doc

redpanda:
I had to define the command as an array otherwise it fails during the build phase with the following error:
```
! waypoint.hcl:15,17-19: Unsuitable value type; Unsuitable value: list of string
```


mitchellh (approved):
Thank you! 👍

Issue #527, opened 2020-10-16
install/k8s: kind requires LoadBalancer 


akin-ozer:
This guide shows that if we have minimal kind cluster then we can install waypoint but we still need LoadBalancer requirement 
and we should use metallb or something like that: https://learn.hashicorp.com/tutorials/waypoint/get-started-kubernetes?in=waypoint/get-started-kubernetes

**Steps to Reproduce**
Following the guide doesn't end you up in a working state.

**Expected behavior**
Example should be complete with LoadBalancer installation or configuring waypoint to use NodePort.

ronakbanka:
@mitchellh I've written this Blog to use Waypoint with K8s inbuilt from Docker Desktop, works like a charm as LoadBalancer Type is exposed on localhost 👍 

https://medium.com/@ronakbanka/getting-started-with-hashicorp-waypoint-and-kubernetes-c818607c9b05

mitchellh:
We now detect kind and show a warning about requiring more. We're also updating the docs today.

PR #615, opened 2020-10-19
install/k8s: detect kind and show warning about additional requirements

mitchellh:
Fixes (partially) #527 

This uses a heuristic on `waypoint install` to detect Kind clusters and notes to the user that it must be setup with LoadBalancer support which it does not have out of the box nor is it trivial to setup. This will surprise users less when their installs hang.

We should update the docs too but I think this is useful to have in the CLI.

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -48,10 +48,45 @@ func (c *InstallCommand) InstallKubernetes(
 	ctx context.Context, ui terminal.UI, log hclog.Logger,
 ) (*clicontext.Config, *pb.ServerConfig_AdvertiseAddr, string, int) {
 	sg := ui.StepGroup()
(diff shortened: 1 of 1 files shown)


evanphx (approved):
This is totally fine, but we should also see if there is some API and/or config we can inspect to see if there is something configured in the cluster to handle load balancing.

We can probably also look at the events on the Service and presume that, if after a minute or so there are no updates, that the cluster doesn't have a Load Balancer handler.

mitchellh:
Agreed with both. One minute won't be enough on a load balancer, I just tried on AKS and DO and both took over 3 minutes to succeed, so I think we'd have to wait quite long there.

On detecting KinD in a better, way, totally would love to do that but couldn't find anything!

PR #619, opened 2020-10-19
fix(cli): application configuration url

### LINKED ISSUE #618: CLI: Application Configuration URL
**Describe the bug**
```bash
$ waypoint config

Usage: waypoint config SUBCOMMAND

  Manage application configuration.
  
  The config commands can be used to manage the configuration that
  Waypoint will inject into your application via environment variables.
  This can be used to set values such as ports to listen on, database URLs,
  etc.
  
  For more information see: https://waypointproject.io/docs/config

Subcommands:
    get    Get config variables.
    set    Set a config variable.
```

https://waypointproject.io/docs/config is returning 404, according to the webpage, it should be https://waypointproject.io/docs/app-config

**Steps to Reproduce**
- type `waypoint config` in your terminal
- check if you have `https://waypointproject.io/docs/config` in the returning text
- open the url in your browser to see "Page Not Found" error

**Expected behavior**
It should open Application Configuration page, which lives under https://www.waypointproject.io/docs/app-config

**Additional context**
`/config` makes more sense, but I don't know if this was a choice or just a small thing. Maybe in the long term, updating the webpage and using config instead of app-config might be a better alternative, I don't know.

gokaygurcan:
Closes #618

### DIFF
--- a/internal/cli/main.go
+++ b/internal/cli/main.go
@@ -603,7 +603,7 @@ Waypoint will inject into your application via environment variables.
 This can be used to set values such as ports to listen on, database URLs,
 etc.
 
-For more information see: https://waypointproject.io/docs/config
+For more information see: https://waypointproject.io/docs/app-config
 `,
 	},
 


mitchellh:
Thank you!

gokaygurcan:
Woah, that was quick. Thank _you_!

PR #524, opened 2020-10-16
Add `make format` command and run it on the codebase

jacksgt:
As described in CONTRIBUTING.md, everyone should run `make format` before committing and submitting their code.
This patch adds that functionality.

I also ran `make format` on the entire codebase, which fixed some minor formatting inconsistencies, mainly import order.

PR #629, opened 2020-10-19
Enable mandatory go formatting check in CI pipeline

jacksgt:
As discussed in https://github.com/hashicorp/waypoint/pull/524#issuecomment-710144890
everyone should run `make format` before submitting their code, otherwise
the pipeline will fail (now).


mitchellh:
Oh easy perfect thank you!

PR #632, opened 2020-10-19
replace env var with link

krantzinator:
Not every job that needs to notify imports the environment. Since we only use this value in the job to notify slack, we don't need to rely on the env var.


mitchellh (approved):

mitchellh:
The test failures are not important and are fixed on main now.

PR #648, opened 2020-10-20
Adding namespace config to k8s deploy/release

nsdavidson:
Adding namespace as a configuration parameter to the Kubernetes plugin allows for
dynamically setting the namespace to target without having to set a different kubeconfig
context for each.


mitchellh:
Looks great thank you!

PR #647, opened 2020-10-20
Add auth property for pulling docker image from private registry #646

rluisr:
issue #646

PR #656, opened 2020-10-20
builtin/nomad: use block for auth config and default datacenter to 'dc1'

nickethier:
This PR extends what was done in #647 by using a typed block for auth configuration instead of a map.

I also noticed that the datacenter is documented to default to `dc1` but currently does not, so I slipped in a fix for that as well.

### DIFF
--- a/builtin/nomad/platform.go
+++ b/builtin/nomad/platform.go
@@ -90,6 +90,10 @@ func (p *Platform) Deploy(
 		p.config.ServicePort = 3000
 	}
 
+	if p.config.Datacenter == "" {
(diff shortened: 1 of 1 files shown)


rluisr:
Thank you for refactoring my code.

nickethier:
Ha no problem @rluisr! Thanks for the contribution and using Waypoint! I thought having it be a typed block like this might be a better experience but didn't want to ask you to jump through the hoops 😅.

rluisr:
Don't worry about it, I definitely think waypoint is a great tool for these days. Please let me send another PR.

codyde (approved):

Issue #479, opened 2020-10-15
Add command to stop server


abtris:
Looks me as inconsistent CLI commands. I start server, no command to stop on localhost. You have to make `docker ps && docker stop` but will be nice to have command directly in `waypoint server`.

CLI for server command looks now: 

```
Usage: waypoint server SUBCOMMAND

  Server management.

  The CLI, UI, and entrypoints all communicate to a Waypoint server. A
  Waypoint server is required for logs, exec, config, and more to work.
  The recommended way to run a server is "waypoint install".

  This command contains further subcommands to work with servers.

Subcommands:
    bootstrap     Bootstrap the server and retrieve the initial auth token
    config-set    Set the server online configuration.
    install       Install the Waypoint server to Kubernetes, Nomad, or Docker
    run           Manually run the builtin server
```

I will suggest small change to:

```
Usage: waypoint server SUBCOMMAND

  Server management.

  The CLI, UI, and entrypoints all communicate to a Waypoint server. A
  Waypoint server is required for logs, exec, config, and more to work.
  The recommended way to run a server is "waypoint install".

  This command contains further subcommands to work with servers.

Subcommands:
    bootstrap     Bootstrap the server and retrieve the initial auth token
    config-set    Set the server online configuration.
    install       Install the Waypoint server to Kubernetes, Nomad, or Docker
    start           Manually run the builtin server
    stop           Manually stop the builtin server
```

**Describe alternatives you've considered**
- alternative just name can be `destroy` and stay with `run`.

**Explain any additional use-cases**
- I think cli will be clear and UX will be better

evanphx:
One reason we didn't include such a command is that it's hard to know what it means in the context of, say, kubernetes. Additionally, the `waypoint server start` command runs the waypoint server locally as a foreground process, so it would be weird if `waypoint server stop` was manipulating a waypoint server setup via `install` probably.

Maybe `waypoint server docker stop` ?

abtris:
That make sense. I think that server running outside deploying platform but that was wrong assumption.

mitchellh:
I think we're not gonna do "stop" but we are going do uninstall and upgrades. See #612

PR #660, opened 2020-10-20
Add command to stop Waypoint server in Docker

dominikbraun:
Addresses #479. 

This PR is an attempt to implement a command for stopping a Waypoint server running in Docker and basically uninstalling it. As proposed in [this comment](https://github.com/hashicorp/waypoint/issues/479#issuecomment-710738422), the command name is `waypoint server docker stop`.

The current workflow looks as follows:
1. Look for a running Waypoint container.
2. Shut it down gracefully.
3. Remove the container.
4. Check whether the `waypoint-server` volume exists, which should be the case.
5. If it does, remove it as well.

Two things came to my mind:
* Would it make sense to offer a CLI flag to preserve the volume?
* Strictly speaking, this command isn't the inverse operation of `server run` but of `server install` instead. May `server docker uninstall` be an appropriate alternative command name?

I'm happy to hear your opinions and ideas on this.

### DIFF
--- a/internal/cli/main.go
+++ b/internal/cli/main.go
@@ -17,9 +17,9 @@ import (
 	"github.com/mitchellh/cli"
 	"github.com/mitchellh/go-glint"
 
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/internal/pkg/signalcontext"
 	"github.com/hashicorp/waypoint/internal/version"
-	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


mitchellh:
So one thing is we’re currently planning on introducing `server uninstall` capabilities for all platforms in Waypoint 0.2. We just started planning this work today. So I’m wondering if we should just hold off on this... sorry! Your contribution is super welcome here but not sure if we want to merge this then undo it or move it quickly for 0.2

dominikbraun:
Merging and un-doing this would probably be of little value for the product. This is merely a draft - feel free to close it or proceed with it as you wish.

mitchellh:
Roger, we’ll discuss it. Sorry to not warn you or update the issue earlier this literally came out during discussions like 5 or 6 hours ago.

mitchellh:
Yeah we're gonna close this. Sorry! But we'll have a solution soon.

PR #661, opened 2020-10-20
Allow configuration of container listening port on ECS

codyde:
This PR enables users to configure the listening port for container workloads to deploy in ECS. This is useful for application workloads that require a specific port configured, or application deployments done via a Dockerfile. 

The default port remains 3000. This PR is similar to exposed configurations we have in Nomad, K8s, and Docker currently.


mitchellh:
LGTM. Thanks!

PR #662, opened 2020-10-21
Allow Nomad sentinel policy override on install

chuckyz:
When running `waypoint server install ...` there is no way to override Nomad sentinel policies.  This allows a simple override flag to allow folks to get running as quickly as possible.


chuckyz:
Going to close this and try to re-open with a proper email attached.

PR #665, opened 2020-10-21
Add dockerfile docs

jbayer:
I am not sure how to generate the plugin docs to see that I did this correctly, but I was able to validate that the build completes successfully.


mitchellh:
Looks good thanks, I can run the website regen.

PR #671, opened 2020-10-21
Allow Nomad sentinel policy override on install

chuckyz:
When running `waypoint server install ...` there is no way to override Nomad sentinel policies. This allows a simple override flag to allow folks to get running as quickly as possible.

### DIFF
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -14,9 +14,10 @@ import (
 )
 
 var (
-	nomadRegionF      string
-	nomadDatacentersF []string
-	nomadNamespaceF   string
+	nomadRegionF         string
+	nomadDatacentersF    []string
+	nomadNamespaceF      string
+	nomadPolicyOverrideF bool
 )
 
 // InstallNomad registers a waypoint-server job with a Nomad cluster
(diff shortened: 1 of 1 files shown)


internal/serverinstall/nomad.go
```diff
@@ -14,9 +14,10 @@ import (
 )
 
 var (
-	nomadRegionF      string
-	nomadDatacentersF []string
-	nomadNamespaceF   string
+	nomadRegionF         string
+	nomadDatacentersF    []string
+	nomadNamespaceF      string
+	nomadPolicyOverrideF bool
```
nickethier, line 20: You'll need to register the flag with the cli framework for it to be available. That can be done in the `NomadFlags` function at the end of this file.

nickethier (changes requested):
Great work! Let me know if I can help with the requested change.

internal/serverinstall/nomad.go
chuckyz: 🤦 I knew I missed something.  Thank you.

chuckyz: @nickethier let me know if that's good

nickethier (approved):

PR #674, opened 2020-10-22
Make waypoint config set command pipeable

### LINKED ISSUE #561: CLI: enhance the config command experience
* Adding the -all flag to config command which would print all the variables set in that project.(ALREADY IMPLEMENTED by config get)
````sh
➜  ✗ waypoint config get -all
  SCOPE | NAME | VALUE  
--------+------+--------
        | PORT     |  8000  
        | PASSWORD | password
        | USERNAME | username
        | .....    | .....
````

* Also maybe a flag `-raw` to get the VALUES without the tabular format in case you want to pipe them to a file(.env)
```sh
➜ ✗  waypoint config get -all -raw | tee .env
PORT=8000  
PASSWORD=password
USERNAME=username
````

* Another welcome feature would be to be able to pipe KEY/VALUE pair in the form of `KEY=VALUE` to the `config set` command
```sh
➜  ✗cat .env |  waypoint config set 
````
N/A

N/A

**Additional context**
N/A

### PR DESCRIPTION
**Pull request by dominikbraun**
Fixes #561.

This change allows the user to pass arguments to `waypoint config set` using pipes:

```
$ cat .env | waypoint config set
```

---

The linked issue also proposes a `-raw` flag for `config get`, but this flag apparently already exists:

https://github.com/hashicorp/waypoint/blob/b00f8ab1b17db429bc8242cb06a922bcdd9c8203/internal/cli/config_get.go#L137-L141

However, `waypoint config get -raw` doesn't print anything to my console. Maybe this is a separate issue.

### DIFF
--- a/internal/cli/config_set.go
+++ b/internal/cli/config_set.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: internal/cli/config_set.go [hunk 1]
```diff
@@ -25,17 +26,38 @@ func (c *ConfigSetCommand) Run(args []string) int {
...
+
+		// If there's no pipe, there are no arguments. Fail.
+		if info.Mode()&os.ModeNamedPipe == 0 {
+			_, _ = fmt.Fprintf(os.Stderr, "config set requires at least one key=value entry")
+			return 1
+		}
+
+		scanner := bufio.NewScanner(os.Stdin)
+		for scanner.Scan() {
+			configArgs = append(configArgs, scanner.Text())
```
mitchellh, line 48: Nitpick: I would change this a bit to set `configArgs := c.args` earlier up and then just append to that here. That way we don’t need the somewhat awkward `else` fall through.

mitchellh (commented):
This looks pretty good! Thank you! I requested some slight changes.

internal/cli/config_set.go
dominikbraun: Absolutely, that's a more elegant solution.

dominikbraun: So I just noticed that I'm now directly appending the config variables to `c.args` without creating a new slice `configArgs` at all - but you actually suggested to initialize `configArgs` with `c.Args` and still append the variables to `configArgs`.

Which way would you prefer?

mitchellh: I think its fine as-is since we verify `len(c.args) == 0` to guard entry into that path anyways. Thanks!

PR #680, opened 2020-10-23
Path Variables and Deferred Config Loading

mitchellh:
This is a large PR that does one major thing: deferred config loading. 

@evanphx We can talk about this over Zoom to make this easier to digest.

Previously, we decoded the entire `waypoint.hcl` in a single pass. This approach was simple, but doesn't allow us to introduce scoped HCL variables. To use scoped HCL variables, we must load each piece of the config as we need it and have the proper context.

This PR decodes different parts of the config at different times: project scope, each app scope, each stage scope, etc. Each time, it lets the caller supply more context that is available to the decoding.

The first bit of context is simple to test this: **we introduce `path.*` variables.** Configuration now has access to `path.project`, `path.app`, and `path.pwd` for safely referencing paths within their configuration.

As a side benefit, a lot more tests were added to config loading and validation to verify this works as expected.

## More Detailed Change List

* `internal/config`: deferred loading
* `internal/config`: move server configuration to `internal/serverconfig` (both client access and server)
* `internal/core`: remove a bunch of unused features that relied on config


internal/core/app.go
```diff
@@ -409,39 +311,20 @@ func (a *App) initMappers(
...
-			a.mappers = append(a.mappers, pinst.Mappers...)
-			log.Info("registered component-specific mappers", "len", len(pinst.Mappers))
-
-			// Store the closer
-			a.closers = append(a.closers, func() error {
-				pinst.Close()
-				return nil
-			})
-		}
+
```
evanphx, line 319: Right here, if `len(pint.Mappers)` is 0, we should just close the plugin immediately rather than having it linger doing nothing.

evanphx (changes requested):
One quick thing about eager closing mapperless plugins, otherwise looks good!

evanphx (approved):

PR #693, opened 2020-10-24
Add ECS config option for disabling load balancer

surminus:
The idea behind a backend service is that an external load balancer is not created for it. Backend services are may run based upon events from other AWS services, process data from an S3 bucket, or be internal only (accessible from service discovery, for example).

What it basically means is that we're not creating a load balancer. I named the property `backend_service` with the idea that in the future it may contain other properties, or potentially create an internal load balancer.

I've shamelessly stolen the naming from the competition: AWS Copilot calls it web and backend[1].

I'm aware that this is quite a complicated reshuffle, but I thought it was worth extracting the mechanisms for creation and deletion of the load balancers into functions just to make the conditional easier.

I haven't actually tested this yet, but if people think it's a good idea then I'd be happy to do this next week. I just enjoy getting my hands dirty rather than raising issues! :innocent: 

[1] https://aws.github.io/copilot-cli/docs/concepts/services/#backend-service


evanphx:
Thanks for this! The alb code needed to be extracted to a function anyway, it was too long.

I think rather than call this `backend_service` and effectively establish that there will be this future kind of service, we should just make this an option to disable the load balancer. We can introduce service roles/types in the future which will set certain options to defaults, for instance a future backend role can disable the load balancer by default.

surminus:
Yep, makes sense! I've renamed the option and reworded the commit. I'll see if I can test this in the next few days to make sure I haven't broken anything.

surminus:
@evanphx I've tested and updated this change now. Since the ReleaseManager part of the process just updates the load balancer with the new target group, I just put an early return in. 

I was also wondering if there was a specific reason why Waypoint isn't using the [UpdateService](https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#ECS.UpdateService) API. Maybe so deployments are fully in control of Waypoint rather than AWS? 

I'm mostly interested because it would be really useful to wait for the service to be considered stable as part of the release using the [WaitUntilServicesStable](https://docs.aws.amazon.com/sdk-for-go/api/service/ecs/#ECS.WaitUntilServicesStable) API, so might have a poke at that sometime if it fits with the design of the tool.

builtin/aws/ecs/platform.go
```diff
@@ -1327,6 +1384,9 @@ type Config struct {
 	// Indicate that service should be deployed on an EC2 cluster.
 	EC2Cluster bool `hcl:"ec2_cluster,optional"`
 
+	// If set to true, do not create a load balancer assigned to the service
+	DisableALB bool `hcl:"disable_alb,optional"`
```
krantzinator, line 1388: Can you add a corresponding docs.SetField for this?

krantzinator (changes requested):
Thanks for this! Can I ask you to rebase again 😅 Just the one comment otherwise.
To answer your question, we create new services rather than updating to support the Waypoint design of allowing rollbacks. I agree with you that being able to watch for stability would be useful, and will join you in thinking on how to approach that from a waypoint perspective.

surminus:
@krantzinator sorry I missed your comment! Thank you for rebasing and fixing up  ^_^

PR #696, opened 2020-10-25
Change  Docker Image used by GitLab CI/CD

abuango:
Instead of downloading the images on every run, GitLab provides regularly updated custom images for Waypoint hosted on GitLab's container registry for every version and the latest version tagged with `latest`.

### DIFF
--- a/website/content/docs/automating-execution/gitlab-cicd.mdx
+++ b/website/content/docs/automating-execution/gitlab-cicd.mdx
@@ -48,21 +48,17 @@ See the GitLab CI/CD [predefined environment variables](https://docs.gitlab.com/
 
 ```yaml
 waypoint:
-  image: docker:latest
+  # Image bootstraps the latest version of Waypoint, See https://gitlab.com/gitlab-org/waypoint-images/container_registry for more details.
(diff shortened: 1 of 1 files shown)


mitchellh:
Thank you! I'm super appreciative of GitLab hosting these images but we do have concerns from a security perspective since we can't verify the checksum matching our built binaries.

If these GitLab images are just rewrapping our officially released binaries, then this can work but I think we should at least at a download + shasum verification with our SHASUMs: https://releases.hashicorp.com/waypoint/0.1.2/waypoint_0.1.2_SHA256SUMS

If not, then I think we should put a warning here that these images are built and managed by GitLab and may not match the SHASUMs released by HashiCorp.

abuango:
Thank you for your prompt response @mitchellh. Yes, we are basically [rewrapping your officially released binaries](https://gitlab.com/gitlab-org/waypoint-images/-/blob/master/.gitlab-ci.yml), but with the possibility of some bootstrapping in the future as the technology evolves and how we might adapt it to our AutoDevOps solution, I would go with the disclaimer.

mitchellh:
Thank you that works.

PR #699, opened 2020-10-26
internal/datasource: local data should return pwd

mitchellh:
The local datasource needs to return the pwd so that plugin discovery
uses an absolute path and doesn't attempt to search by PATH (#691).

This also adds validation in internal/runner that this behavior holds.

/cc @tomhjp

### DIFF
--- a/internal/datasource/local.go
+++ b/internal/datasource/local.go
@@ -3,6 +3,8 @@ package datasource
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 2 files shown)


mitchellh:
Btw @tomhjp in case its not clear: the reason the runner has all this job download machinery is is for [remote runners](https://www.waypointproject.io/docs/roadmap#remote-runners) which aren't quite ready yet (and therefore not documented or technically "released" though a lot of the technical implementation is there).

tomhjp:
Thanks! Have tested this locally too just to confirm.

> the reason the runner has all this job download machinery is is for remote runners which aren't quite ready yet

Hah yes this did confuse me initially 😄  looks like this will enable some super cool features!

nicholasjackson (approved):

PR #702, opened 2020-10-26
Fix clone script snippet

cakejelly:
Was following the guide and noticed that `clone.sh` also requires you to specify a name


mitchellh:
Thank you!

PR #713, opened 2020-10-27
internal/{cli, config/funcs}: Wire up ability to gen func docs

### LINKED ISSUE #712: Document functions on waypointproject.io
**Describe the bug**
The functions available in HCL are not available on the website.

**Steps to Reproduce**
Look at the website, fail to see the list.

**Expected behavior**
Look at the website, see the list.

evanphx:
Closes #712

This just writes out the functions to a single partial for us to include on a content page. The functions didn't have a quick summery so I brought those over too. It's a bit too terse but it's better than nothing until we can bring over the same docs as terraform has for the stdlib HCL functions.


internal/config/funcs/doc.go
```diff
@@ -0,0 +1,76 @@
+package funcs
+
+func Docs() map[string]string {
+	return map[string]string{
+		"abs":             "return the absolute value of the given number",
```
mitchellh, line 5: I originally wanted to extract these from godocs but I see how thats very difficult, so maybe a more structured approach like this is better.

mitchellh (commented):
Yeah lots of improvement to be made here for sure, but looks good as a starting point.

internal/cli/app_docs.go
```diff
@@ -404,6 +408,80 @@ func (c *AppDocsCommand) builtinMDX(args []string) int {
...
+	addFuncs := func(fs map[string]function.Function) {
+		for k, v := range fs {
+			all[k] = v
+		}
+	}
+
+	// Add some of our functions
+	addFuncs(funcs.VCSGitFuncs("."))
+	addFuncs(funcs.Filesystem("."))
+	addFuncs(funcs.Encoding())
```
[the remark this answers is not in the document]
evanphx: the reason I didn't do that is the one that take a path, I didn't want all the code to blindly call All() and get those functions setup wrong.

internal/config/funcs/doc.go
evanphx: This was just a fast and dirty first approach, so the first cut on the website had SOMETHING for the each function.

krantzinator (approved):

briancain (approved):

mitchellh (approved):

PR #714, opened 2020-10-27
builtin/aws/ecr: Create the repository if it doesn't exist, add defaust repo name

evanphx:
This came up on discuss (https://discuss.hashicorp.com/t/trying-to-understand-the-boundaries-envisioned-for-waypoint/16037/6) and given we create everything else in the case of ECS deployments, makes sense to create the ECR repo as well.

### DIFF
--- a/builtin/aws/ecr/registry.go
+++ b/builtin/aws/ecr/registry.go
@@ -13,6 +13,7 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
+	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	"github.com/hashicorp/waypoint-plugin-sdk/docs"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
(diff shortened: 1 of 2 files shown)


mitchellh (approved):

builtin/aws/ecr/registry.go
```diff
@@ -63,18 +66,48 @@ func (r *Registry) Push(
 	}
 	svc := ecr.New(sess)
 
+	repoName := r.config.Repository
+
+	if repoName == "" {
+		log.Info("infering ECR repo name from app name")
+		repoName = "waypoint-" + src.App
```
briancain, line 73: Since we're creating an ECR repo for the user here, I think it would be great to give them a bit of extra feedback on this with the generated repo name in the step output rather than just at the info level so they know a resource has been auto-magically created for them :smile:

briancain (changes requested):

builtin/aws/ecr/registry.go
evanphx: Totally. I've updated it to use a StepGroup and give a bunch more feedback about these sorts of things.

briancain (approved):
Thanks! :sparkles: :+1:

PR #715, opened 2020-10-28
k8s/install: Add OpenShift support

jasonodonnell:
This adds support to install Waypoint and deploy applications to Red Hat's OpenShift. To install on OpenShift the CLI will try its best to figure out if this is an OpenShift cluster by checking for the existence of the `openshift` namespace. You can also explicitly tell the CLI its OpenShift by using the `-openshift` flag.

Two things needed to be done to support OpenShift:

* the `fsGroup` we set for Kubernetes is removed when installing on OpenShift. Since the Waypoint container will be run as an arbitrary UID, OpenShift will automatically assign a `fsGroup` to the Waypoint container so it can write to persistent volumes.
* the `HOME` env is set to `/data`. This tricks Waypoint into persisting its configuration file in a place it can write. Since the Waypoint container is running as a random UID there is no home directory and Waypoint tries to write to the root filesystem. Another option is we mount a memory volume to `/home/waypoint` and set `HOME` to point there but the PVC is working well enough.

### Related Changes

Since the K8s installation is getting smarter, I thought it might be good to remove the template and create the server objects in code. Due to this change the `-show-yaml` flag would no longer be supported because marshaling the structs to YAML doesn't produce valid K8s objects. I can revert this change if needed.

I found a bug on the StatefulSet: two Liveness probes were defined (TCP and HTTP). Kubernetes was silently throwing one away (gRPC probe). In the StatefulSet definition I only defined the HTTP probe.

One other slight tweak in this PR is I changed the default namespace for installation to use your current kubeconfig targeted namespace. This makes the user experience a little better because if you aren't targeting the default namespace and you don't specify one via `-namespace` flag, `kubectl` commands won't show the stuff we installed. Using the default namespace in general isn't a great idea so removing this default makes sense.


internal/cli/install.go
```diff
@@ -511,6 +551,40 @@ func (c *InstallCommand) Flags() *flag.Sets {
...
+
+		f.StringVar(&flag.StringVar{
+			Name:    "mem-request",
+			Target:  &c.config.MemRequest,
+			Usage:   "Configures the requested memory amount for the Waypoint server in Kubernetes.",
+			Default: "256Mi",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "storage-request",
```
mitchellh, line 569: I think we should prefix these all with `-k8s` if that makes sense.

mitchellh (commented):
This seems pretty good to me. I think we should hold this for 0.2.0 due to the magnitude but this seems good.

internal/cli/install.go
jasonodonnell: Good point, done!

mitchellh (approved):
I just reviewed this again and this looks great. I'm going to merge into main since that is 0.2 now and we'll tweak as we approach the upgrade/snapshot stuff prior to 0.2 release.

PR #720, opened 2020-10-28
go: bump horizon dep

evanphx:


mitchellh (approved):
As long as tests are green for these I have no problem with just merging em

PR #722, opened 2020-10-28
Update -from-project to handle local projects

### LINKED ISSUE #717: Have a parameter to specify the hcl file to be used when we do waypoint init
## Feature request

It is not possible to specify when we execute the command `waypoint init` the hcl file to be used and by consequence waypoint will pickup the file : `waypoint.hcl`.

Having the possibility to specify the file to be used would help to use waypoint with the same project but different cloud providers: docker, kubernetes, ...

` waypoint init -f my-waypoint.hcl.kubernetes`

dominikbraun:
Fixes #717.

This is some approach to solve #717. It does 3 things:
* Rename the `from` struct field for `-from-project` to `fromProject` for consistency.
* Introduce the `-from` option and store it in the "new" `from` struct field.
* Read the configuration based on `-from` instead of `configpkg.Filename`.

Happy to hear your suggestions and improvements.

### DIFF
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -210,7 +210,7 @@ func (c *baseCommand) Init(opts ...Option) error {
 
 	// If we're loading the config, then get it.
 	if baseCfg.Config {
-		cfg, err := c.initConfig(baseCfg.ConfigOptional)
+		cfg, err := c.initConfig("", baseCfg.ConfigOptional)
 		if err != nil {
 			c.logError(c.Log, "failed to load config", err)
 			return err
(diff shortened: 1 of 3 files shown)


briancain (changes requested):
Hey there @dominikbraun ! Thanks for tackling this issue and making a pull request, we appreciate it!

I think we should re-use the existing `from-project` CLI flag, rather than introducing a similar flag `from` to avoid any confusion since they would do similar things. I've added some suggested commits for that change. Additionally, we don't have to add a new parameter for these functions, we can grab the value from `baseCommand`.

Let me know if you have any questions addressing these! Thanks :heart:

briancain:
And apologies for taking so long to get back to your pull request. If you're not in a place to address this please let us know and I'll be happy to take this pull request over to see it through.

briancain (approved):

PR #723, opened 2020-10-29
Snapshot/Restore APIs 

mitchellh:
Details are in RFC [WP-039](https://docs.google.com/document/d/1Z05NZ9ZjAyZQqD9iTHeTAN00Ke6yD9_PkbTLJUsnMAk/edit#).

This is only the APIs to implement snapshot/restore. I'll implement the CLI in a future PR.

Importantly, the restore is a _stage_ restore operation currently and not a full online restore. It requires the server to be restarted to complete the restore.


evanphx (changes requested):
One minor thing about the naming of the last chunk before the trailer.

internal/server/singleprocess/state/snapshot.go
```diff
@@ -0,0 +1,462 @@
...
+		"format", header.Format.String(),
+	)
+
+	// We currently only support bolt
+	if header.Format != pb.Snapshot_Header_BOLT {
+		return db, fmt.Errorf("invalid snapshot format (got code: %d)", header.Format)
+	}
+
+	// Open a temporary file to write our raw bolt data
+	log.Info("reading bolt information and writing it into a new bolt db")
```
briancain, line 227: Same question here, maybe it's worth being extra explicit in the logs for where these paths are for debugging.

```suggestion
	log.Info("reading bolt information and writing it into a new bolt db", "bolt info temp path", ri.StageTempPath)
```

internal/server/singleprocess/state/snapshot.go
```diff
@@ -0,0 +1,462 @@
...
+
+	// Replace our file
+	log.Info("atomically replacing db file", "src", ri.StagePath, "dest", ri.DBPath)
+	if err := atomic.ReplaceFile(ri.StageTempPath, ri.DBPath); err != nil {
+		log.Error("error replacing file", "err", err)
+		return db, err
+	}
+
+	// Delete our restore data
+	log.Info("removing restore data since we're done reading it")
```
briancain, line 318:
```suggestion
	log.Info("finished reading restore data, removing now")
```

internal/server/singleprocess/state/snapshot.go
```diff
@@ -0,0 +1,462 @@
...
+	if os.IsNotExist(err) {
+		log.Debug("no restore file found, no DB restore requested")
+		return db, nil
+	}
+	if err != nil {
+		log.Error("error checking for restore file", "err", err)
+		return db, err
+	}
+
+	log.Warn("restore file found, will initiate database restore")
```
briancain, line 196: Is it worth being extra explicit here to include the file path of the restore file?

```suggestion
	log.Warn("restore file found, will initiate database restore", "path", ri.StagePath)
```

briancain (approved):
Looks good! Just left a few open questions around a few log messages to ease some debugging woes in the future, but I don't think they should block merge if the changes don't make sense :+1:

briancain (approved):

PR #739, opened 2020-11-02
netlify: fix authentication requirements

pearkes:
This is part of the fix for the issue described here: https://discuss.hashicorp.com/t/unable-to-deploy-using-waypoint-netlify-plugin/16909. The main idea here is just to allow users to configure their access_token inline. However, I'm not sure this should even be supported, given the risk of sensitive values in the Waypoint configuration. Do we have an environment variable naming pattern here? The [upstream golang API](https://github.com/netlify/open-api) doesn't support any environment variable or known file type authentication.

Beyond that, there are a couple issues that look closer to core here with this plugin.

If you run `waypoint init` with the following waypoint configuration:

```
project = "foo"

app "wp-demo" {
    build {
        use "files" {

        }
    }

    deploy {
        use "netlify" {}
    }
}
```

You'l see approximately this output:

```
 waypoint init
✓ Project "foo" and all apps are registered with the server.
✓ Plugins loaded and configured successfully
⚠️ "wp-demo" has plugins that require authentication:
- Platform "netlify"
❌ Failed to check authentication requirements!

! This step verifies that Waypoint has access to the configured systems.
  This is a best-effort check, since not all plugins support this check
  and the check can often only check that any known credentials work at
  a minimal level.

  There was an error during this step and it is shown below.

! The plugins above reported that they aren't authenticated.
! Project had errors during initialization.
  Waypoint experienced some errors during project initialization. The output
  above should contain the failure messages. Please correct these errors and
  run 'waypoint init' again.
```

I'd expect this to call the `Auth()` func, but it looks like it just validates the auth. 

In addition, when you run `up` it looks like there's an argmapper issue with the datadir:

```
waypoint up

» Building...

» Deploying...
! 1 error occurred:
  	* argument cannot be satisfied: type *anypb.Any (subtype:
  "hashicorp.waypoint.sdk.Args.DataDir.Component")
```

### DIFF
--- a/builtin/netlify/platform.go
+++ b/builtin/netlify/platform.go
@@ -56,9 +56,7 @@ func (p *Platform) Auth(
 	// just output some help text to the user.
 	if !info.Local {
 		ui.Output(
-			"Jack Pearkes needs to do this but he'll tell you to open a URL to\n" +
-				"some place and copy some token to some other place and then after\n" +
-				"all that we should be good to go.")
+			"access_token must to be set to the value of a Netlify authentication token")
 		return nil, nil
 	}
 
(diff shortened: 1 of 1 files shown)


evanphx:
@pearkes Should this PR be worked on more or closed?

mitchellh:
Let's close this, this needs work but the auth system in general needs work.

PR #752, opened 2020-11-04
internal/ceb: prevent multiple horizon agents from running in the same ceb

evanphx:
Looking through the connect and reconnect code paths, it's pretty clear that we're leaking an agent instance on every reconnect. If a user had a waypoint server that was having trouble running steady state (perhaps due to too tight of a memory budget), then it would be easy to see how an entrypoint could spawn thousands of agents accidentally.

This code just makes sure that we only run one at a time by canceling the context for the previous one if there is a context set.

### DIFF
--- a/internal/ceb/ceb.go
+++ b/internal/ceb/ceb.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"strconv"
+	"sync"
 
 	"github.com/hashicorp/go-hclog"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


briancain (approved):
:+1: CEB!!

internal/ceb/ceb.go
```diff
@@ -42,6 +43,10 @@ type CEB struct {
 	execIdx      int64
 
 	cleanupFunc func()
+
+	mu             sync.Mutex
```
mitchellh, line 47: can we constrain this to `urlAgentMu`? As an outsider reading this I'd be confused as to when `mu` needs to be held. Since we're only using it for the URL agent let's constrain it there.

mitchellh (changes requested):
Looks good! One minor request.

internal/ceb/ceb.go
evanphx: Yup! no prob.

mitchellh (approved):

PR #757, opened 2020-11-05
Templating Functions & HCL Variables for Artifacts, Deploys

mitchellh:
This introduces two major features (related):

  * A set of a `template`-prefixed HCL functions that template files or directories with the currently available HCL variables. 
  * Artifact, deploy HCL variables for registry, deploy, and release stanzas. You can use this to reference information about the result of the prior operation.

**This PR temporarily disables auth checks completely.** We will add these again at runtime in a future PR, prior to 0.2.0.

Depends on https://github.com/hashicorp/waypoint-plugin-sdk/pull/5 and will update go.mod after merging that.

## Example

```hcl
app "web" {
  build {
    use "docker" {
      dockerfile = templatefile("${path.app}/Dockerfile.prod", {
        custom_var = "hello"
      })
    }

    registry {
      use "docker" {
        image = artifact.image
        tag = gitrefpretty()
      }
    }
  }
  
  # ...
}
```


builtin/docker/copy.go
```diff
@@ -0,0 +1,52 @@
+package docker
+
+import (
+	"io"
+	"os"
+)
+
+// From: https://gist.github.com/m4ng0squ4sh/92462b38df26839a3ca324697c8cba04
```
krantzinator, line 8: User moved profiles
``` suggestion
// From: https://gist.github.com/r0l1/92462b38df26839a3ca324697c8cba04
```

krantzinator (approved):
Exciting :tada:

internal/core/eval_context.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	valBytes, ok := val.Interface().([]byte)
+	if !ok {
+		return nil
+	}
+
+	var raw map[string]interface{}
+	if err := json.Unmarshal(valBytes, &raw); err != nil {
+		return err
```
evanphx, line 30: We should probably give them a detailed error here about what parsing failed. Folks are going to pass bare strings accidentally here most likely so we should give them at first. Maybe: `errors.Wrapf(err, "Error parsing TemplateData for '%s'", key)`.

evanphx: Though maybe I'm mistaken and this would already have been validate before being put into the protobuf?

evanphx (changes requested):
This is a biggy! All looks good, though it's hard to follow some of the changes, like I wasn't sure why all the template functions don't take a basedir anymore? I must have missed something in the reading where that basedir is coming from now. Maybe it's that the base dir is being applied to path elsewhere now?

internal/core/eval_context.go
mitchellh: Yeah, the JSON is guaranteed because we encode it ourselves in the SDK based on the structure. 😄 Still, I can wrap this in the hopefully-impossible case it happens.

evanphx (approved):

Issue #719, opened 2020-10-28
Wrong 0 exit status code reported on fail to push image to private registry


rkrasiuk:
**Describe the bug**
When pushing an image to the private registry with no credentials, waypoint fails with 0 exit code.
```sh
➜ waypoint up

» Building...
✓ Initializing Docker client...
✓ Building image...
 │ Successfully tagged waypoint.local/<app name>:latest
✓ Tagging Docker image: waypoint.local/<app name>:latest => <private registry address>:<tag>
❌ Pushing Docker image...
 │ The push refers to repository [<private registry address>]
! unable to stream Docker logs to terminal: denied: Your authorization token has expired. Reauthenticate and try again.

➜ echo $?
0
```

**Steps to Reproduce**
```hcl
project = "project-name"

app "app-name" {
  build {
    use "docker" {
      buildkit = true
      disable_entrypoint = true
    }

    registry {
      use "docker" {
        image = "<private registry address>"
        tag   = gitrefhash()
      }
    } 
  }
}
```

**Expected behavior**
Expected non-zero exit code.

**Additional context**
This gives false positives on running github actions. The final action status is a success whereas the command has failed.

mitchellh:
Fixed by @krantzinator

PR #768, opened 2020-11-06
internal/core: return error if type is ErrSentinel to signal exit codes

krantzinator:
Address #719 
We were neglecting to return a non-nil error when dealing with `ErrSentinel` types, which is what we use most of the time.

### DIFF
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -275,7 +275,7 @@ func (c *baseCommand) Init(opts ...Option) error {
 // thread-safe.
 //
 // If any error is returned, the caller should just exit. The error handling
-// including messaging to the user is handling by this function call.
+// including messaging to the user is handled by this function call.
 //
 // If you want to early exit all the running functions, you should use
 // the callback closure properties to cancel the passed in context. This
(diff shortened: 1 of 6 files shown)


internal/cli/base.go
```diff
@@ -306,6 +306,9 @@ func (c *baseCommand) DoApp(ctx context.Context, f func(context.Context, *client
 		if err := f(ctx, app); err != nil {
 			if err != ErrSentinel {
 				finalErr = multierror.Append(finalErr, err)
+			} else {
+				// if we have an ErrSentinel here, we've already output info to the UI
+				return err
```
mitchellh, line 311: I think what we should do here is continue to append to finalErr, but note we got a sentinel (switch some bool to true), then at the end something like:

```go
if finalErr == nil && didErrSentinel {
  finalErr = ErrSentinal
}

return finalErr
```

The reason being: this way we attempt to run against other apps here before we exit. The idea of this function one day is that we can try multiple apps in parallel maybe (its not today).

mitchellh (changes requested):
Looks great! One suggestion.

mitchellh (approved):

krantzinator:
Oops! It looks like no changelog entry is attached to this PR. Please include a release note block in the PR body, as described in https://github.com/GoogleCloudPlatform/magic-modules/blob/master/.ci/RELEASE_NOTES_GUIDE.md:

~~~
```release-note:TYPE
Release note
```
~~~

PR #786, opened 2020-11-11
[WIP] Abstract new docs page component

jescalan:
There was a temporary docs page component in here previously as it was being developed, this PR updates it so that the properly abstracted version that will be used on all docs sites is implemented.

There is a pending change coming in that is causing the color theme to be off at the moment, and the component itself is still a canary release for early testing, which is why this PR is still marked as a draft for now.


jmfury (approved):

website/pages/style.css
```diff
@@ -1,10 +1,7 @@
 /* Global Styles */
-@import '~@hashicorp/react-global-styles/style.css';
-@import '~@hashicorp/react-global-styles/font.css';
```
jmfury, line 3: @jescalan  Already approved since the font looked fine, just curious if `font` is move into the main `style.css`? I can look myself, just something that crossed my mind

jmfury: [Ahh I see it here](https://github.com/hashicorp/mktg-global-styles/blob/main/global.css) 👍

jescalan: weirdly enough, it was actually always in there haha. i just realized it was and removed these

mitchellh (approved):
Seems fine to me. Didn’t test the preview URL though.

kylemac (approved):

PR #791, opened 2020-11-12
internal/ceb: support child process restarts on config change

mitchellh:
When config variables change, this will gracefully restart the child process to get the new config variables.

"Graceful restart" follows the same process as [Heroku](https://devcenter.heroku.com/articles/what-happens-to-ruby-apps-when-they-are-restarted): we send the child a SIGTERM and if the child doesn't exit in 30 seconds we send a SIGKILL.

This required quite a bit of refactoring, but all tests are passing and I added a test to verify we restart properly for env var changes.

### DIFF
--- a/internal/ceb/child_exec.go
+++ b/internal/ceb/child_exec.go
@@ -5,7 +5,10 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"syscall"
+	"time"
 
+	"github.com/hashicorp/go-hclog"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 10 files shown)


internal/ceb/child_exec.go
```diff
@@ -14,23 +17,127 @@ import (
...
+				log.Info("child process termination result", "err", err)
+			}
+
+			return
+
+		case cmd := <-cmdCh:
+			log.Info("child command received")
+
+			if currentCh == nil {
+				// First client is always the default, so let's wait for
```
evanphx, line 74: What does 'the default' here mean? This whole comment feels confusing, I guess we're waiting for 2 seconds or if we got a different command to run?

internal/ceb/child_exec.go
```diff
@@ -56,6 +163,66 @@ func (ceb *CEB) execChildCmd(ctx context.Context) <-chan error {
...
+			case <-time.After(30 * time.Second):
+				// Timeout, fall through to SIGKILL
+				log.Warn("graceful termination failed, will send SIGKILL")
+			}
+		}
+	}
+
+	// SIGKILL
+	log.Info("sending SIGKILL")
+	if err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL); err != nil {
```
evanphx, line 209: Worth a comment here to indicated why we're sending the negative pid value. I believe it's because you want to send it to the whole process group.

internal/ceb/config.go
```diff
@@ -77,13 +61,76 @@ func (ceb *CEB) initConfigStream(ctx context.Context, cfg *config, isRetry bool)
...
+	// Keep track of our currently executing command information so that
+	// we can diff properly to determine if we need to restart.
+	currentCmd := ceb.copyCmd(ceb.childCmdBase)
+
+	// We only init the URL service once. In the future, we can do diffing
+	// and support automatically reinitializing if the URL service changes.
+	didInitURL := false
+
 	for config := range ch {
+		if !didInitURL {
```
evanphx, line 79: I guess we don't expect the server to change the token or or address or anything for the UrlService?

evanphx (changes requested):
A couple of comments about UNIXy things mostly.

internal/ceb/config.go
mitchellh: Yeah for now. Note the comment when this variable is defined. This is just the same behavior as prior to this PR. We can add support for changing the URL service config later!

internal/ceb/child_exec.go
mitchellh: Yeah I’ll expand on this comment. I might do this in a more robust way too. But basically we always send the “default” command which is just the basic child process as invoked on the command line. But we expect to get another one quickly if we connect to the server with env vars and such. 

I might just do this with a more explicitly triggered system though so it’s not just timer crossing fingers

mitchellh: Made this better, triggered.

internal/ceb/config.go
evanphx: No problem! Just wanted to double check the assumption.

internal/ceb/child_exec.go
evanphx: If we're going to start the command and then kill it a second later, is it worth having the system wait until config arrives before starting then? 150 milliseconds isn't very long.

mitchellh: Let me just switch this out for a non-timer. I think this is giving the wrong idea here. This is really just draining the channel buffer.

mitchellh: Done! Take another look. Should be more clear.

evanphx: Ahhhhhh, I see now!

evanphx (approved):

PR #800, opened 2020-11-13
nix: move depricated nixpkgs-channels to nixpkgs

06kellyjac:
Lovely to see nix in the wild :heart: I was looking forward to something like this since @mitchellh's first nix related tweet

https://github.com/nixos/nixpkgs-channels has been depricated.
Updated niv source to use main https://github.com/nixos/nixpkgs repo.
Using the same channel/branch and rev there shouldn't be any incompatibilities.

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -13,14 +13,14 @@
     },
     "nixpkgs": {
         "branch": "nixpkgs-unstable",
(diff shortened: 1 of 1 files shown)


mitchellh:
Ah hah. Thanks!

PR #801, opened 2020-11-13
statically compile ceb to make injection easier

### LINKED ISSUE #614: Unable to inject ceb into image with nix-based waypoint build
I am trying to package waypoint 0.1.2 for the Nix package manager.
* https://github.com/NixOS/nixpkgs/pull/100994

**Describe the bug**
When using `waypoint up`, I get the following error:

```
...
 │ Successfully tagged waypoint.local/example-python:latest
⠸ Injecting Waypoint Entrypoint...
! unable to restore custom entry point binary: Asset ceb/ceb can't read by error:
  Read "ceb/ceb": gzip: invalid checksum
```
(The complete output is referenced in the context section below.)

I did spend a couple of hours learning about the go module system and go-bindata, but was not able to find the root cause. As you also use Nix, you might have a tip for me.

**Steps to Reproduce**

```
$ nix-env -f https://github.com/winpat/nixpkgs/archive/add_waypoint.tar.gz -iA waypoint
$ git clone https://github.com/hashicorp/waypoint-examples.git
$ cd waypoint-examples/docker/python
$ waypoint init
$ waypoint up
```
The `waypoint.hcl` that I've used can be found here:
* https://github.com/hashicorp/waypoint-examples/blob/c8f632147e10a5ec93e1f73a744ea986cfc364ec/docker/python/waypoint.hcl

**Expected behavior**
I would have expected that `ceb` would get successfully injected into the container.

**Additional context**

The full output with enabled verbose flags can be found here:
* https://gist.github.com/winpat/321145ab42bb5bddba7ea0adc66777b0

06kellyjac:
Resolves #614

Nix build & then injecting worked just fine.
I didn't actually use the built artifact though as I haven't had much of a go with `waypoint` yet.
Nix derivation I used: https://github.com/06kellyjac/nixpkgs/blob/07abbe37a02fa2683fb66c9169086a5f4163c793/pkgs/applications/networking/cluster/waypoint/default.nix

nixpkgs_waypoint

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ CGO_ENABLED?=0
 
 .PHONY: bin
 bin: # bin creates the binaries for Waypoint for the current platform
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./internal/assets/ceb/ceb ./cmd/waypoint-entrypoint
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags '-w -extldflags "-static"' -o ./internal/assets/ceb/ceb ./cmd/waypoint-entrypoint
 	cd internal/assets && go-bindata -pkg assets -o prod.go -tags assetsembedded ./ceb
 	CGO_ENABLED=$(CGO_ENABLED) go build -ldflags $(GOLDFLAGS) -tags assetsembedded -o ./waypoint ./cmd/waypoint
 	CGO_ENABLED=0 go build -tags assetsembedded -o ./waypoint-entrypoint ./cmd/waypoint-entrypoint


mitchellh:
Did this not work without this? I thought `CGO_ENABLED=0` forced the binary to be fully static already. We fixed bugs around this in earlier releases.

06kellyjac:
I can give it a try on my machine but thats the only important change i made to @winpat's package

---

> For simple cases where you don’t use any other cgo code it’s probably easier to just disable cgo, since the cgo code is protected with `+build cgo`:
>
> `$ CGO_ENABLED=0 go build`
>
> What if we want to use the cgo versions of the above? Or what if we want to use a cgo package such as SQLite? In those cases you can tell the C linker to statically link with `-extldflags`:
>
> `$ go build -ldflags="-extldflags=-static"`

https://www.arp242.net/static-go.html

I've had to use this flag several times to compile an actually static golang binary to work properly in a scratch container

mitchellh:
Got it, yeah I can confirm that with `CGO_ENABLED=0` on its own we're getting static binaries in the latest releases:

```
root@d90a2ef3cfaa:/# ldd /tmp/ceb
	not a dynamic executable
root@d90a2ef3cfaa:/# file /tmp/ceb
/tmp/ceb: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=8Jci_ADhTnxSpL3b3wzZ/FAQ3yKiGW6D3ujGJEu7z/u8x7m9MrNyWl7_yQFTPB/JEEbBo1Ehf0xvYO-Ft8v, not stripped
```

I haven't seen that flag necessary unless you're using CGO and want to ensure your Cgo code is statically linked. If you disable cgo completely, Go will produce statically linked binaries.

06kellyjac:
@mitchellh Tried it without and it didn't work

image

I did a couple repeat attempts and removed all my existing images just incase but still no dice

image

If you really dissagree with using this flag then I guess we'll just have to add it ourselves for the nix package because it really doesn't seem to work without it.

mitchellh:
Hm, the given error message of the invalid checksum looks like a different error.

I'm not sure why that is but I believe it has something to do with the way Nix is doing the build. I've never seen any embedded asset with `go-bindata` ever give that so I'm not super sure.

06kellyjac:
Ok using kevinburke/go-bindata `v2.33.0` instead of jteeuwen/go-bindata `a0ff2567cfb70903282db057e799fd826784d41d` didn't require `-ldflags '-w -extldflags "-static"'`

Issue #799, opened 2020-11-13
Docker builder missing `dockerfile` causes crash


orenmazor:
**Describe the bug**
If I dont specify the `dockerfile` parameter to the docker builder, I get the following error: 

```
error building image: Error response from daemon: unexpected error reading
Dockerfile: read /var/lib/docker/tmp/docker-builder999095492: is a directory
```

**Steps to Reproduce**

Simple Dockerfile:

```
FROM metabase/metabase:v0.37.1
EXPOSE 3000
```

waypoint.hcl:

```
app "myapp" {
  build {
     use "docker" {
     }
  }

  deploy {
    use "nomad" {}

  }
}

```

doing a build:

```
➜ waypoint build
✓ Initializing Docker client...
⠙ Building image...
! error building image: Error response from daemon: unexpected error reading
  Dockerfile: read /var/lib/docker/tmp/docker-builder010878143: is a directory
zsh: exit 1     waypoint build
0.28s user 3.53s system 98% cpu 3.870s total
```

adding the `dockerfile = "Dockerfile"` attribute to the builder builds the image correctly.

versions:

```
➜ waypoint version
Waypoint v0.1.4-161-geaaf62f4 (eaaf62f4)


➜ docker --version                                                                        
Docker version 20.10.0-beta1, build ac365d7
```

mitchellh:
As noted in #802, I pushed a fix for this just minutes ago!

mitchellh:
And sorry to confirm: this was when building from `main` right? I don't think this bug was ever on a released version.

orenmazor:
thanks!

and yeah I've been running off HEAD. for those fresh out of the oven commits

PR #802, opened 2020-11-16
Empty dockerfile

tekno45:
Fix for issue https://github.com/hashicorp/waypoint/issues/799#issue-742595747

IsAbs() is respecting an empty string as an absolute filepath.

checking for the docker cli default seems like an easy and low risk fix.


mitchellh:
Funny, I just pushed a fix for this to main a few minutes ago! Thank you though!

PR #803, opened 2020-11-16
Auto-scroll UI client log components

dizzyup:
Updates the streaming logs components in the UI client to support auto-scrolling and pinning/unpinning.

By default, when new lines start to overflow the logs container, the container will automatically scroll so that you remain pinned to the latest log line. If you scroll up that behavior is paused and you'll be unpinned and free to scroll up and down even as new lines are added. If you scroll back to the bottom you will become pinned again and the auto-scrolling kicks back in.

While in the unpinned mode, a button will be displayed in the top right corner - clicking the button will scroll to the bottom of the container and pin. If new lines are added while the button is visible it will display a count of how many new lines have been added.

### Demo

CleanShot 2020-11-09 at 11 04 01 2

### DIFF
--- a/ui/app/components/log-stream/index.hbs
+++ b/ui/app/components/log-stream/index.hbs
@@ -1,13 +1,30 @@
-<div class="output">
-    {{#if this.lines}}
-    {{#each this.lines as |line|}}
-    <code>{{line}}</code>
-    {{/each}}
-    {{else}}
+<div class="output-pane">
+  {{#if this.lines}}
+    <div class="output-scroll-y">
+      {{#unless isFollowingLogs}}
+        <button class="follow-logs" {{on "click" this.followLogs}}>
+          {{inline-svg "arrow-down" class="icon"}}
+          {{#if (gt this.badgeCount 0)}}
+            <div class="badge">{{this.badgeCount}}</div>
+          {{/if}}
+        </button>
+      {{/unless}}
+      <div class="output-scroll-x">
+        {{#each this.lines as |line|}}
+          <div class="log-line">
+            <code>{{line}}</code>
+          </div>
+        {{/each}}
+        <div class="output-end" {{in-viewport
+          onEnter=(action (mut isFollowingLogs) true)
+          onExit=(action (mut isFollowingLogs) false)}} {{did-update this.updateScroll this.logLines.length}}></div>
+      </div>
+    </div>
+  {{else}}
     <EmptyState>
-        <p>There are no logs to display for this app yet</p>
-        <p>Logs will be displayed when you run actions in the CLI such as <code class="inline">build</code>, <code
-                class="inline">deploy</code> or <code class="inline">release</code></p>
+      <p>There are no logs to display for this app yet</p>
+      <p>Logs will be displayed when you run actions in the CLI such as <code class="inline">build</code>,
+        <code class="inline">deploy</code> or <code class="inline">release</code></p>
     </EmptyState>
-    {{/if}}
+  {{/if}}
 </div>
--- a/ui/app/components/log-stream/index.hbs
+++ b/ui/app/components/log-stream/index.hbs
@@ -1,7 +1,7 @@
 <div class="output-pane">
   {{#if this.lines}}
     <div class="output-scroll-y">
-      {{#unless isFollowingLogs}}
+      {{#unless this.isFollowingLogs}}
         <button class="follow-logs" {{on "click" this.followLogs}}>
           {{inline-svg "arrow-down" class="icon"}}
           {{#if (gt this.badgeCount 0)}}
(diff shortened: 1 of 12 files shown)


ui/app/components/log-stream/index.hbs
```diff
@@ -1,13 +1,30 @@
-<div class="output">
-    {{#if this.lines}}
-    {{#each this.lines as |line|}}
-    <code>{{line}}</code>
-    {{/each}}
-    {{else}}
+<div class="output-pane">
+  {{#if this.lines}}
+    <div class="output-scroll-y">
+      {{#unless isFollowingLogs}}
```
gregone, line 4: `this.isFollowingLogs` ?

ui/app/components/log-stream/index.hbs
```diff
@@ -1,13 +1,30 @@
...
+        </button>
+      {{/unless}}
+      <div class="output-scroll-x">
+        {{#each this.lines as |line|}}
+          <div class="log-line">
+            <code>{{line}}</code>
+          </div>
+        {{/each}}
+        <div class="output-end" {{in-viewport
+          onEnter=(action (mut isFollowingLogs) true)
```
gregone, line 19: `this.isFollowingLogs` ? (same below)

ui/app/components/log-stream/index.ts
```diff
@@ -21,6 +26,22 @@ export default class LogStream extends Component<LogStreamArgs> {
...
   addLine(line: string) {
     this.lines = [...this.lines, line];
+    if (this.isFollowingLogs === false) {
+      this.badgeCount = this.badgeCount + 1;
+    }
+  }
+
+  @action
+  followLogs(element: any) {
+    element.target.parentElement.scroll(0, element.target.parentElement.scrollHeight);
```
gregone, line 36: This is ok, but brittle if we change the DOM structure, would `this.element.querySelector('.output-scroll-y')` be better?

ui/app/components/operation-logs/index.ts
```diff
@@ -36,6 +40,22 @@ export default class OperationLogs extends Component<OperationLogsArgs> {
...
   addLogLine(t: string, logLine: object) {
     this.logLines = [...this.logLines, { type: t, logLine: logLine }];
+    if (this.isFollowingLogs === false) {
+      this.badgeCount = this.badgeCount + 1;
+    }
+  }
+
+  @action
+  followLogs(element: any) {
+    element.target.parentElement.scroll(0, element.target.parentElement.scrollHeight);
```
gregone, line 50: this is good, but brittle if we change the DOM structure. Could we be using `this.element.querySelector('.output-scroll-y')`?

gregone (approved):
LGTM, just a few comments/questions

ui/app/components/operation-logs/index.ts
dizzyup: Agreed this should be more specific! I'm having trouble targeting the element with `querySelector` though - I get an error in the console..

`
You attempted to access the 'element' property on a Glimmer Component, but that property does not exist in Ember.js applications, it only exists in Glimmer.js apps.
`

I've just changed that `followLogs` action so that it can be run when the component is initially inserted (so the scroll position is always at the bottom to begin). For some reason it behaves differently when I use `{{did-insert}}`.

I'm a little out of my depth on JS here - would you be able to take a look and see if you can refactor so it's less brittle?

gregone: ok, so I think the best solution is probably not `querySelector` but instead [`closest('.output-scroll-y')`](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest)

ui/app/components/log-stream/index.ts
```diff
@@ -21,6 +26,26 @@ export default class LogStream extends Component<LogStreamArgs> {
...
     this.lines = [...this.lines, line];
+    if (this.isFollowingLogs === false) {
+      this.badgeCount = this.badgeCount + 1;
+    }
+  }
+
+  @action
+  followLogs(element: any) {
+    if (element.target) {
+      element.target.parentElement.scroll(0, element.target.parentElement.scrollHeight);
```
gregone, line 37:
```suggestion
      element.closest('.output-scroll-y').scroll(0, element.target.parentElement.scrollHeight);
```

ui/app/components/log-stream/index.ts
```diff
@@ -21,6 +26,26 @@ export default class LogStream extends Component<LogStreamArgs> {
...
+    } else {
+      element.parentElement.scroll(0, element.parentElement.scrollHeight);
```
gregone, line 39:
```suggestion
      element.closest('.output-scroll-y').scroll(0, element.parentElement.scrollHeight);
```

ui/app/components/operation-logs/index.ts
dizzyup: @gregone I've refactored using `.closest()` and tidied up the `element.target` conditional a bit. Thanks for the help!

gregone:
@dizzyup Thanks for getting the changes in! This looks great and ready to merge whenever you want

Issue #790, opened 2020-11-12
ecs deploy fails


aruntomar:
**Describe the bug**
Deploying to ECS fails with the following error. 
```
» Deploying...
✓ Created new ECS cluster: waypoint
✓ Created IAM role: ecr-example-nodejs
❌ IAM role not found:
! InvalidParameter: 1 validation error(s) found.
  - minimum field size of 1, GetRoleInput.RoleName.
```

**Steps to Reproduce**
I'm using the waypoint-example app
https://github.com/hashicorp/waypoint-examples/blob/main/aws-ecs/nodejs/waypoint.hcl

**Expected behavior**
Application gets deployed on ECS

**Additional context**
If i add a specific role_name as the [doc](https://www.waypointproject.io/plugins/aws-ecs) then i get the following error 

```
waypoint-examples/aws-ecs/nodejs/waypoint.hcl:24,7-16:
  Unsupported argument; An argument named "role_name" is not expected here.
```

```
deploy {
    use "aws-ecs" {
      region = "ca-central-1"
      memory = "512"
      role_name = "ecr-example-nodejs"
    }
  }
```

```
waypoint version
Waypoint v0.1.5 (084d8603)
```

krantzinator:
The variable name should be `execution_role_name`, which I'll update our documentation to reflect. You can also use `task_role_name` to assign a task IAM role; I'll add that to the documentation as well.
Thanks for bringing this up!

philippelegaldeloitte:
Hello @krantzinator ,
I am experiencing the same issue as @aruntomar.

  deploy {
    use "aws-ecs" {
      region = "eu-central-1"
      memory = "512"
    }
  }
}

yields :

» Deploying...
✓ Found existing ECS cluster: waypoint
⠙ Found existing IAM role to use: ecr-example-nodejs
! InvalidParameter: 1 validation error(s) found.
  - minimum field size of 1, GetRoleInput.RoleName.

same using :

execution_role_name = "ecr-example-nodejs" in

deploy {
    use "aws-ecs" {
      region = "eu-central-1"
      memory = "512"
      execution_role_name = "ecr-example-nodejs"
    }
  }

Is there a fix or a work around ?

Thanks

aruntomar:
@philippelegaldeloitte @krantzinator 
I think `task_role_name` is a required variable 🤔 

I tried these in sequence. 
1. Add `execution_role_name` - FAILS ❌ 
[...]

PR #824, opened 2020-11-19
Fix ECS task role name bug (issue #790)

surminus:
The `execution_role_name` is optional, but this is a required parameter, so is automatically created if it's left blank.

Setting `task_role_name` should also be optional, but since this is not required it isn't created automatically, and should not be referenced in the task definition if it's not set.

Fixes issue: https://github.com/hashicorp/waypoint/issues/790


krantzinator (approved):
Works great for both use cases. Thank you!

PR #828, opened 2020-11-19
waypoint config get: Make -raw print all variables

### LINKED ISSUE #823: Consistency of `waypoint config get` flags
The `waypoint config get` command has a `-json` and a `-raw` flag for changing the output format.

As @briancain and I discussed in #813, it might be a bit surprising for users that `-json` without any arguments prints all variables as JSON, but `-raw` without any arguments returns an error.

```shell script
$ waypoint config get -json
  {"PASSWORD":"test","USERNAME":"alice"}

$ waypoint config get -json USERNAME
  {"USERNAME":"alice"}

$ waypoint config get -raw
  # Returns an error

$ waypoint config get -raw USERNAME
  alice
```

Maybe I've overlooked something here, but I would appreciate if `-raw` behaved just like `-json`. This would also introduce the behavior suggested in #561. 

**Additional context**
Waypoint version: `Waypoint v0.1.5 (084d8603)`

dominikbraun:
Fixes #823.

This is a rather simple fix, so let me know if this is actually sufficient.

**Before this PR:**

```shell script
$ waypoint config get -raw
  # Error
```

**With this PR:**
```
$ waypoint config get -raw
  PASSWORD=test
  USERNAME=alice
```

### DIFF
--- a/internal/cli/config_get.go
+++ b/internal/cli/config_get.go
@@ -91,8 +91,10 @@ func (c *ConfigGetCommand) Run(args []string) int {
 		}
 
 		if prefix == "" {
-			fmt.Fprintf(os.Stderr, "flag '-raw' requires a named variable argument to be given")
-			return 1
+			for _, cv := range resp.Variables {
+				fmt.Printf("%s=%s\n", cv.Name, cv.Value)
+			}
+			return 0
 		}
 
 		if len(resp.Variables) == 0 {


briancain (approved):
Thanks for fixing this! Just tried it locally, looks good to me. I'm going to write up a complimentary PR just so our docs are updated to reflect the change on the website and the cli help. I'll approve this now, and merge both together once that PR is ready. Thanks again! :tada:

briancain:
Ok! Got that pull request ready to go here: https://github.com/hashicorp/waypoint/pull/829 I also updated the `-raw` flag to always print `key=value`, since that's how the `-json` flag works too.

dominikbraun:
> I also updated the -raw flag to always print key=value, since that's how the -json flag works too.

I thought about that too, but wasn't sure if that is ok in terms of backwards compability. Anyway, thanks for your quick response!

PR #829, opened 2020-11-19
Docs update for `config get raw`

briancain:
This pull request updates the docs around the new behavior of the `-raw` flag for `waypoint config get`. It also ensures that if an argument is passed, it still prints it as a `key=value` so that it behaves just like the `json` flag.

Built off of the initial commit https://github.com/hashicorp/waypoint/pull/828, can be merged once the manual backport PR is merged https://github.com/hashicorp/waypoint/pull/832

### DIFF
--- a/internal/cli/config_get.go
+++ b/internal/cli/config_get.go
@@ -107,7 +107,7 @@ func (c *ConfigGetCommand) Run(args []string) int {
 			return 1
 		}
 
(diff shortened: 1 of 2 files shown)


briancain:
Already merged https://github.com/hashicorp/waypoint/pull/828, but once this is merged I'll update the CHANGELOG.

krantzinator (approved):

PR #838, opened 2020-11-20
Use shared Terminal component

ztanner:
Uses the new `<Terminal />` component abstraction from react-components. 

/cc @jescalan to shepherd this through


jescalan:
Merged to #786 locally 💯

PR #839, opened 2020-11-20
docs/troubleshooting: Include extra help for troubleshooting deployments

briancain:
This context was taken directly from the Learn tutorials. Given that
it's useful for investigating deployed applications, and the fact that a
user might miss this context if they didn't follow the Learn guide, it
seems useful to also have this context on the docs where it's more
discoverable for people.

### DIFF
--- a/website/content/docs/troubleshooting.mdx
+++ b/website/content/docs/troubleshooting.mdx
@@ -81,7 +81,7 @@ Since you are in the deployment directory, you will observe that Waypoint
 automatically executes against the currently deployed application.
 
 From within the Docker container, validate that this is the actual application
-by listing out the directory hosting the application's compiled files.
+by listing out the directory hosting the application's compiled files. For example, if you used the `pack` builder this directory will be `/workspace`.
 
 ```shell-session
 $ cd / && ls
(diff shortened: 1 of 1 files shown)


website/content/docs/troubleshooting.mdx
```diff
@@ -58,3 +58,110 @@ $ docker volume list
...
+
+```shell-session
+$ waypoint exec /bin/bash
+```
+
+Since you are in the deployment directory, you will observe that Waypoint
+automatically executes against the currently deployed application.
+
+From within the Docker container, validate that this is the actual application
+by listing out the directory hosting the application's compiled files.
```
krantzinator, line 84: I'm not clear on what we mean by `deployment directory` here. I associate `the currently deployed application` with the `/workspace` dir which contains my application , but `exec /bin/bash` drops me into `/`. Based on the suggestion below to `cd /`, I may be experiencing slightly different behavior than you?

briancain: You're right! Maybe this behavior has changed since the Learn guide was written? I'm also dropped into `/` when I run a `waypoint exec /bin/bash`. :thinking: Maybe it's there just to ensure readers are always in the right directory so that `ls` works? I'm happy to just change it to `ls` though.

krantzinator:
```suggestion
by listing out the directory hosting the application's compiled files. For example, if you used the `pack` builder this directory will be under `/workspace`.
```

krantzinator (approved):
As discussed, once the deployment directory stuff is clarified this looks great!

Issue #745, opened 2020-11-02
aws-ecr support for AWS_DEFAULT_REGION


orenmazor:
**Is your feature request related to a problem? Please describe.**
I am currently deploying my stack to several AWS regions in several accounts.

Unfortunately, it's problematic for me that the registery settings for aws-ecr force me to hardcode the region, as such:

```
        registry {
            use "aws-ecr" {
                region = "us-east-1"
                repository = "foo"
                tag = "0xdeadbeef"
            }
        }
```

**Describe the solution you'd like**
my deploy environment is always fully defined and the AWS_DEFAULT_REGION is always set to point to the correct environment.

waypoint automatically uses the AWS environment session that is set. I'd love it if that plugin was able to use the region as well, and I wouldn't need to specify the region explicitly

this is basically a smaller subset of what #741 would help me with

**Describe alternatives you've considered**
I currently deploy via a makefile that assumes role to whatever account I need to be in. I thought about dynamically generating a waypoint.hcl per deploy but that's a smelly antipattern.

mitchellh:
I'm surprised the AWS SDK doesn't do this for us! We'll have to take a look.

orenmazor:
thanks @mitchellh! thats what I also expected, so I'm still 50/50 on whether this is just PEBKAC somehow

orenmazor:
merged MR handles this issue. thanks all!

PR #841, opened 2020-11-23
#745 environment variables to be used instead of 'region' property for aws-ecr registry

quintok:
I thought I'd give this a go as it looked like a pretty simple change and a good "first issue"

A few thoughts:
1. This logic should be centralised for all aws environment config.
2. Packer and terraform use https://github.com/hashicorp/aws-sdk-go-base which seems reasonable to migrate to.
3. This logic is too simple, it should try and lean on the SDK more (ie: load region from default profile)
4. The logic about leaning on the sdk seems similar between terraform and packer - maybe we can push that into the aws-sdk-go-base?

Finally, my dev environment is not healthy enough to check the documentation output.
After installing protobuf from homebrew and running:
```make gen/doc```

> ./vendor/proto/api-common-protos/: warning: directory does not exist.
google/rpc/status.proto: File not found.
internal/server/proto/server.proto:10:1: Import "google/rpc/status.proto" was not found or had errors.
...

### DIFF
--- a/builtin/aws/ecr/registry.go
+++ b/builtin/aws/ecr/registry.go
@@ -5,6 +5,8 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"os"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


mitchellh:
I’ll note that I do want to use the AWS SDK base lib we made but there was a reason we didn’t for 0.1 that I can’t recall. That would definitely be the more correct longer term solution.

briancain (approved):
Thanks for this! Looks good to me :+1: We'll try to use the aws sdk in the future, but for now this looks good in the short term.

PR #844, opened 2020-11-24
Config Source Settings

mitchellh:
This introduces the APIs and Entrypoint logic for configuring config sources. That is, the configuration for the plugins that synchronize configuration. 😄 For example, this is the system you'd use to configure the auth method for Vault, perhaps the Vault server address, etc.

As a plugin author, to use this you implement the standard `component.Configurable` (same as any other plugin). The initial implementation only allows primitive fields (no blocks, maps, lists) but we may be able to expand on this in the future.

The entrypoint will reconfigure the plugin: (1) on first use (2) whenever config changes (3) whenever variable requests change. The entrypoint will NOT configure the plugin if there are no dynamic vars in use of that type. Also, scenario "(3)" before we can optimize later but it was much simpler to have that logic for now.

As a user, there is no way to use this yet except directly via the API. I'll add a CLI in a follow up PR.


internal/ceb/app_config.go
```diff
@@ -51,11 +54,28 @@ func (ceb *CEB) watchAppConfig(
...
+		// We will refresh NO MATTER WHAT on this timer and prevents a
+		// flurry of config updates from preventing variable refresh.
+		if refreshCh == nil {
+			refreshCh = time.After(5 * time.Second)
+		}
+
+		// Reset our coalesce channel. Using "time.After" here "leaks"
+		// timers if we're calling this enough but they're a bunch of timers
+		// that reset relatively quickly so let's just let it happen for now.
+		coalesceCh = time.After(500 * time.Millisecond)
```
evanphx, line 77: Per the note about leaking timer, could we use a persistent `Ticker` and call `Ticker.Reset` here to boost it to 500ms?

evanphx: If you do use a persistent `Ticker` here, I'd still assign it's `.C` to `coalesceCh` so that you can set it to `nil` later on.

internal/ceb/app_config.go
```diff
@@ -79,33 +159,50 @@ func (ceb *CEB) watchAppConfig(
...
+				if !prevVarsChanged[k] {
+					prevVarsChanged[k] = v
+				}
+			}
+
+		// Case: timer fires after a period of time where we have received
+		// no other messages and we can now force a refresh.
+		case <-coalesceCh:
+			coalesceCh = nil
+			refreshNowCh := make(chan time.Time)
```
evanphx, line 184: Could we make this a global instead of a new one each time? We just want to assign a closed channel, right? Mostly trying to make the logic easier to understand.

evanphx (approved):
A couple minor bits.

krantzinator (approved):
:confetti_ball:

internal/ceb/app_config.go
mitchellh: Good idea.

internal/ceb/app_config.go
mitchellh: I avoided Tickers and Timers because I found their reset/stop logic fairly easy to mess up (the docs have all sorts of caveats on when they should be called and I've messed it up many times). I prefer to use leaky "After" unless we think it's a frequent event. In this case, I think changes to config should be infrequent enough that I'd prefer to use this since I know it works...

PR #845, opened 2020-11-24
contrib/serverinstall: Automate setup of kind+k8s with metallb

briancain:
This commit introduces a simple script that leverages `kind` to bring up
a local kubernetes cluster with metallb. This is required on linux if
you wish to install a waypoint server locally to a cluster.


mitchellh:
Can you rename it `k8s-kind` instead?

And I think the top of the README should have a paragraph or something describing why this exists. I don't want folks to think that this is required for all K8S installs.

briancain:
Makes sense! I've updated the folder name, and added a little note at the top of the readme which hopefully gives people a bit more info on why this exists :+1:

mitchellh:
Looks good, would be nice to modify the website docs somewhere to point this out. :)

briancain:
@mitchellh I think it makes sense to place the note about `contrib` near the beginning? If it becomes more complicated in the future it could have its own section, but maybe for now it's ok as a small paragraph on here: https://www.waypointproject.io/docs

contrib/serverinstall/kind-k8s/README.md
```diff
@@ -0,0 +1,97 @@
...
+  f5336320a8a3        kindest/node:v1.19.1   "/usr/local/bin/entr…"   2 minutes ago       Up 2 minutes        127.0.0.1:39077->6443/tcp   kind-control-plane
+  ```
+  * Grab container id for container named `kind-control-plane`, which is `f5336320a8a3` in this case,
+  to find its IP Address
+  * `docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' CONTAINER_ID_GOES_HERE`
+  * Update the range inside the `configs/metallb-config.yaml` file. If your
+  container IP Address was `172.18.0.4` for example, you might set the range to `172.18.0.20-172.18.0.50`.
+6) kubectl apply -f configs/metallb-config.yaml
+
+### Optional steps??
```
krantzinator, line 70: :smile: 
```suggestion
### Optional steps
```

briancain: Haha oops :+1:

mitchellh (approved):

PR #848, opened 2020-11-25
Component cleanup

gregone:
We had UI components in two parts of the directory structure:
- `ui/app/components`
- `ui/app/templates/components`

This is probably a result of changes to the colocation of component templates and js files in Ember Octane. To avoid confusion, this PR moves all the component templates into `ui/app/components`.

While I was moving the files, I also cleaned up a few Ember style things that were getting flagged by the template linter: 
- replaced some actions with `{{on ...}}` instead of `{{action ...}}` 
- used `this.` on tracked properties
- added missing `type` attribute on `button` elements

There are still a few warnings left, but I'll address those in another PR once I have my dev setup running properly


dizzyup (approved):
This is great! Some much needed cleanup and should make things easier to follow and get around 👍

PR #855, opened 2020-11-29
Stream entrypoint logs, distinguish app vs system logs, colorize

mitchellh:
This modifies the entrypoint to send its logs to the Waypoint server and make them available via `waypoint logs`.

I did this because as I was finishing up config syncing I realized its _critical_ to be able to see any errors in syncing config and its a huge PITA on most systems that aren't local Docker. 

The server API was modified to take a "source" with each log entry so we can tag app logs vs entrypoint logs vs. future other logs differently. **This is fully backwards compatible** since we just added a field and we default it to "app" scope for the zero (unset) value.

For the log streaming, we only send Info and above level logs to the stream. This currently isn't configurable. However, we still log trace level logs to the default (stderr) output. So if things are going really wrong and we need to debug further, users can still view their pod logs or container logs or whatever and grab more detailed logs. In reality for a _user_, any issues will be logged Info, Warn, or Error, so they'll show up in `waypoint logs`.

I also made it so `waypoint logs` colors the sources differently. It looks like this:

image


internal/ceb/ceb.go
```diff
@@ -61,7 +72,8 @@ type CEB struct {
 	// FIRST child command is ready to be started. This can be closed before
 	// any command is sent to childCmdCh. It indicates that the child process
 	// watcher can begin executing.
-	childReadyCh chan struct{}
+	childReadySent uint32
```
evanphx, line 75: I always worry about having a non-pointer sized value mid struct and taking it's address for `atomic` ops. At one point, this could cause issue due to pointer alignment but I'm honestly not sure anymore. I opt usually to just use, in this case, a `*uint32` directly to avoid any potential alignment issues.

briancain (approved):
:+1: looks good to me! I like the ux and color updates :art:

evanphx (approved):
One minor comment about atomic pointer alignment.

internal/ceb/ceb.go
mitchellh: Yeah I could do that to be safe.

krantzinator: We use this pattern a few other places (non-pointer sized value that is passed into `atomic.CompareAndSwap` via `&`). For consistency can we update all of them to match `*type` (ceb, runner, logbuffer, and service_entrypiont)?

mitchellh: Done! @krantzinator probably a good idea as we find em.

internal/ceb/ceb.go
```diff
@@ -66,21 +66,20 @@ type CEB struct {
 	// commands will stop the old command first. Values sent here are coalesced
 	// in case many changes are sent in a row.
 	childCmdCh chan<- *exec.Cmd
-	childInit  uint32
```
mitchellh, line 58: This just wasn't used so I deleted it.

PR #857, opened 2020-11-30
Fix #738 - Add awslog driver configuration

### LINKED ISSUE #738: Add the ability to specify the awslogs-multiline-pattern options for ECS
My app outputs logs in a specific format, prefixed with date and time. I'm currently using the [awslogs-multiline-pattern](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html) setting to make my logs appear correctly in Cloudwatch. As far as I can see we're not able to set this setting using Waypoint yet.

It would be great if Waypoint could support all the extra awslogs specific options described [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html), e.g. 'awslogs-datetime-format' and 'awslogs-multiline-pattern'.

One option would be to change the log format I'm using, but I'd rather not since the extra information I'm adding is valuable to me.

n/a

**Additional context**
n/a

quintok:
Added all the flags I could find in the driver - [Docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html#create_awslogs_logdriver_options)

I didn't move `log_group` to maintain backwards compatibility - happy to change it if desired.

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -559,6 +559,37 @@ func createSG(
 	return groupId, nil
 }
 
+func buildLoggingOptions(conf *Logging, region string, logGroup string, defaultStreamPrefix string) map[string]*string {
+
+	var streamPrefix = conf.StreamPrefix
+	if streamPrefix == "" {
+		streamPrefix = defaultStreamPrefix
+	}
+
+	result := map[string]*string {
+		"awslogs-region": aws.String(region),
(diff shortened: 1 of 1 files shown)


krantzinator (changes requested):
Looks great! I think you missed a doc field for `MultilinePattern` though.

krantzinator:
Thanks again for this, I'm porting it over to a new branch to handle the rebase. 🎉

PR #860, opened 2020-11-30
Fix path for server proto when using go generate

briancain:
Prior to this commit, go generate would run in the dir
`internal/server`, so when it went to look for the server.proto, it
would look inside
`waypoint/internal/server/internal/server/proto/server.proto`. This
commit fixes that by ensuring it goes back to the root dir to find
`server.proto`.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -11,7 +11,7 @@ import (
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 )
 
-//go:generate sh -c "protoc -I../../vendor/proto/api-common-protos -I ../.. internal/server/proto/server.proto --go_out=plugins=grpc:../.. --go-json_out=../.."
+//go:generate sh -c "protoc -I../../vendor/proto/api-common-protos -I ../.. ../../internal/server/proto/server.proto --go_out=plugins=grpc:../.."
 //go:generate mv ./proto/server.pb.json.go ./gen
 //go:generate mockery -all -case underscore -dir ./gen -output ./gen/mocks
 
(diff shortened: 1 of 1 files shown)


internal/server/server.go
```diff
@@ -11,7 +11,7 @@ import (
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 )
 
-//go:generate sh -c "protoc -I../../vendor/proto/api-common-protos -I ../.. internal/server/proto/server.proto --go_out=plugins=grpc:../.. --go-json_out=../.."
```
mitchellh, line 14: Ah, you need the `go-json-out` stuff cause that invokes `protoc-gen-go-json`

go.sum
```diff
@@ -546,6 +546,7 @@ github.com/golang-sql/civil v0.0.0-20190719163853-cb61b32ac6fe h1:lXe2qZdvpiX5WZ
 github.com/golang-sql/civil v0.0.0-20190719163853-cb61b32ac6fe/go.mod h1:8vg3r2VgvsThLBIFL93Qb5yWzgyZWhEmBwUJWevAkK0=
 github.com/golang/gddo v0.0.0-20180823221919-9d8ff1c67be5 h1:yrv1uUvgXH/tEat+wdvJMRJ4g51GlIydtDpU9pFjaaI=
 github.com/golang/gddo v0.0.0-20180823221919-9d8ff1c67be5/go.mod h1:xEhNfoBDX1hzLm2Nf80qUvZ2sVwoMZ8d6IE2SrsQfh4=
+github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b h1:VKtxabqXZkF25pY9ekfRL6a582T4P37/31XEstQ5p58=
```
briancain, line 549: Not sure if this was left over, but `go mod tidy` doesn't seem to clean it up?

mitchellh (approved):
Tested and works in my env!

PR #862, opened 2020-12-01
Add template fields to docs, rearrange docs slightly

mitchellh:
Paired with: https://github.com/hashicorp/waypoint-plugin-sdk/pull/7

This adds docs for the new template attributes that builders/registries/platforms can emit. These are called "output attributes" in the docs since we don't call them "template attributes" anywhere in the docs and that can be confusing with templating functions and so on.

I took the liberty while I was here to make some other improvements (I believe):

  * Example moved up right after the description.
  * Example is now HCL syntax highlighted.
  * Separated required from optional fields and show required first
  * Added some small help text (that is repeated since its autogenerated) at the top of sections so folks can know how to use this a bit better.
  * Renamed "variables" to "parameters". This is the word we use in the `use` docs and I think limits the confusion since we now have HCL vars and will one day have user vars and so on.

Looks like this:

CleanShot 2020-11-30 at 18 45 06@2x

### DIFF
--- a/builtin/aws/ami/builder.go
+++ b/builtin/aws/ami/builder.go
@@ -38,7 +38,7 @@ type BuilderConfig struct {
 }
 
 func (b *Builder) Documentation() (*docs.Documentation, error) {
-	doc, err := docs.New(docs.FromConfig(&BuilderConfig{}))
+	doc, err := docs.New(docs.FromConfig(&BuilderConfig{}), docs.FromFunc(b.BuildFunc()))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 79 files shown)


jennbriden:
Thank you! These are some great improvements for the docs.

krantzinator (approved):
This look so good! I support the variables -> parameters call, this is more clear.

I can't leave a suggestion on code you didn't actually edit, but we should add a period on line 596 in /k8s/platform.go `"namespace is the name of the Kubernetes namespace to apply the deployment in",`

evanphx (approved):

PR #863, opened 2020-12-01
tools: Add tools.go for bootstrapping local dev to build waypoint 

briancain:
This commit adds a simple go file to install some required dependencies
for building Waypoint.

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -7,8 +7,8 @@
 
 package tools
 
-//go:generate go install github.com/hashicorp/go-bindata
-import _ "github.com/hashicorp/go-bindata"
+//go:generate go install github.com/kevinburke/go-bindata
+import _ "github.com/kevinburke/go-bindata"
 
 //go:generate go install github.com/golang/protobuf/proto
 import _ "github.com/golang/protobuf/proto"
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -16,9 +16,6 @@ import _ "github.com/golang/protobuf/proto"
 //go:generate go install github.com/golang/protobuf/protoc-gen-go
 import _ "github.com/golang/protobuf/protoc-gen-go"
 
-//go:generate go install github.com/grpc-ecosystem/grpc-gateway
-import _ "github.com/grpc-ecosystem/grpc-gateway"
-
 //go:generate go install github.com/mitchellh/protoc-gen-go-json
 import _ "github.com/mitchellh/protoc-gen-go-json"
 
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -23,4 +23,4 @@ import _ "github.com/grpc-ecosystem/grpc-gateway"
 import _ "github.com/mitchellh/protoc-gen-go-json"
 
 //go:generate go install github.com/vektra/mockery
-import _ "github.com/vektra/mockery/v2"
+import _ "github.com/vektra/mockery"
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -16,6 +16,9 @@ import _ "github.com/golang/protobuf/proto"
 //go:generate go install github.com/golang/protobuf/protoc-gen-go
 import _ "github.com/golang/protobuf/protoc-gen-go"
 
+//go:generate go install github.com/grpc-ecosystem/grpc-gateway
+import _ "github.com/grpc-ecosystem/grpc-gateway"
+
 //go:generate go install github.com/mitchellh/protoc-gen-go-json
 import _ "github.com/mitchellh/protoc-gen-go-json"
 
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -1,0 +1,23 @@
+// +build tools
+
+// To install the following tools at the version used by this repo run:
+// $ make tools
+// or
+// $ go generate -tags tools tools/tools.go
+
+package tools
+
+//go:generate go install github.com/hashicorp/go-bindata
+import _ "github.com/hashicorp/go-bindata"
+
+//go:generate go install github.com/golang/protobuf/proto
+import _ "github.com/golang/protobuf/proto"
+
+//go:generate go install github.com/golang/protobuf/protoc-gen-go
+import _ "github.com/golang/protobuf/protoc-gen-go"
+
+//go:generate go install github.com/mitchellh/protoc-gen-go-json
+import _ "github.com/mitchellh/protoc-gen-go-json"
+
+//go:generate go install github.com/vektra/mockery
+import _ "github.com/vektra/mockery/v2"
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,14 @@ GIT_DESCRIBE=$$(git describe --tags --always --match "v*")
 GIT_IMPORT="github.com/hashicorp/waypoint/internal/version"
 GOLDFLAGS="-X $(GIT_IMPORT).GitCommit=$(GIT_COMMIT)$(GIT_DIRTY) -X $(GIT_IMPORT).GitDescribe=$(GIT_DESCRIBE)"
 CGO_ENABLED?=0
+GO_CMD?=go
+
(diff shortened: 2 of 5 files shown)


tools/tools.go
```diff
@@ -0,0 +1,26 @@
...
+import _ "github.com/hashicorp/go-bindata"
+
+//go:generate go install github.com/golang/protobuf/proto
+import _ "github.com/golang/protobuf/proto"
+
+//go:generate go install github.com/golang/protobuf/protoc-gen-go
+import _ "github.com/golang/protobuf/protoc-gen-go"
+
+//go:generate go install github.com/grpc-ecosystem/grpc-gateway
+import _ "github.com/grpc-ecosystem/grpc-gateway"
```
briancain, line 20: Quick question on this...it seems like the `mitchellh/protoc-gen-go-json` package requires this? But I don't think I'm setting it up right because it is excluding the go files :sweat_smile: 

```
==> github.com/grpc-ecosystem/grpc-gateway
package github.com/grpc-ecosystem/grpc-gateway: build constraints exclude all Go files in /home/brian/go/src/github.com/grpc-ecosystem/grpc-gateway
==> github.com/mitchellh/protoc-gen-go-json
package github.com/grpc-ecosystem/grpc-gateway/codegenerator: cannot find package "github.com/grpc-ecosystem/grpc-gateway/codegenerator" in any of:
        /usr/lib/go-1.14/src/github.com/grpc-ecosystem/grpc-gateway/codegenerator (from $GOROOT)
        /home/brian/go/src/github.com/grpc-ecosystem/grpc-gateway/codegenerator (from $GOPATH)
package github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor: cannot find package "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor" in any of:
        /usr/lib/go-1.14/src/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor (from $GOROOT)
        /home/brian/go/src/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor (from $GOPATH)
```

tools/tools.go
```diff
@@ -0,0 +1,26 @@
+// +build tools
+
+// To install the following tools at the version used by this repo run:
+// $ make tools
+// or
+// $ go generate -tags tools tools/tools.go
```
mitchellh, line 6: Does tools.go modify `go.mod` so we can pin versions? Or does that not work? 

If so, it seems like we should have `make tools` just call the `go generate` rather than have two separate implementation.

tools/tools.go
```diff
@@ -0,0 +1,26 @@
...
+import _ "github.com/golang/protobuf/protoc-gen-go"
+
+//go:generate go install github.com/grpc-ecosystem/grpc-gateway
+import _ "github.com/grpc-ecosystem/grpc-gateway"
+
+//go:generate go install github.com/mitchellh/protoc-gen-go-json
+import _ "github.com/mitchellh/protoc-gen-go-json"
+
+//go:generate go install github.com/vektra/mockery
+import _ "github.com/vektra/mockery/v2"
```
briancain, line 26: Is there a good way to make sure we get mockery version `1.1.2`? Otherwise our server.proto arguments will fail since they use the single dash flags.

mitchellh: To start I think you have to get rid of `/v2`

tools/tools.go
mitchellh: It does require that... it should just do it automatically without you having to have this import.

mitchellh:
I'm not super familiar with the "tools.go" approach so my feedback is really more questions.

tools/tools.go
briancain: Without including it, I end up with the same error saying that it's missing :thinking:

tools/tools.go
briancain: I believe it will if you run this command directly, but `make tools` turns off `GO111MODULE` before running `go generate` on this file so it doesn't end up modifying `go.mod` when you run the make target. But if you just run the command above I believe it will modify `go.mod`.

briancain: Maybe that's a nicer solution here? We could have two options that enable the module env var:

```
brian@localghost:waypoint % go generate -tags tools tools/tools.go                                                                                                                                                   ±[●][add-tools-bootstrap]
go: finding module for package github.com/vektra/mockery
go: finding module for package github.com/hashicorp/go-bindata
go: finding module for package github.com/mitchellh/protoc-gen-go-json
go: found github.com/hashicorp/go-bindata in github.com/hashicorp/go-bindata v3.0.7+incompatible
go: found github.com/mitchellh/protoc-gen-go-json in github.com/mitchellh/protoc-gen-go-json v0.0.0-20200917194518-364b693410ae
go: found github.com/vektra/mockery in github.com/vektra/mockery v1.1.2
```

tools/tools.go
briancain: Oddly enough, enabling `GO11MODULE` with go generate doesn't run into this :thinking: Maybe we go with this enabled instead for `make tools`? (related to the other comment on PR)

tools/tools.go
briancain: That seemed to do it! Lol. Thought I tried that before and it didn't work but :shrug: Looks like that was it :sweat_smile:

tools/tools.go
evanphx: If this is just running `go install` we don't even need to import them, do we? I suppose it won't include them in the waypoint binary either way. I see the code switches off modules too, so that would mean that the imports here wouldn't honor a go.sum anyway...

evanphx (changes requested):

tools/tools.go
briancain: Yeah.... initially I had turned of `GO11MODULE` since that looks to be what [vault is doing](https://github.com/hashicorp/vault/blob/f0849708a59e02025ac9808e6c1d451553af048f/Makefile#L125-L134) but I notice now they're only doing it for the `go get` steps. I'll go ahead and remove it, I think that should solve this problem anyway!

briancain: Also with that off, we probably don't need the `go get` part of this either?

evanphx: Hm, so. if you run go get with modules off, it will just download them into $GOPATH and ignore go.mod. But then when `go install` is run via `go generate`, it will use go.mod.

So I think you can skip the whole `get` phase entirely and just do `go generate` since they're referenced from `go.mod`.Does that work?

briancain: Yeah, I'll remove the get phase completely. `go generate` works just fine! I'll update the PR to just do that.

evanphx (approved):
bitmoji

krantzinator (approved):
WOMM

tools/tools.go
```diff
@@ -0,0 +1,23 @@
...
+
+// To install the following tools at the version used by this repo run:
+// $ make tools
+// or
+// $ go generate -tags tools tools/tools.go
+
+package tools
+
+//go:generate go install github.com/hashicorp/go-bindata
+import _ "github.com/hashicorp/go-bindata"
```
krantzinator, line 11: This isn't the `go-bindata` version we use. The version we use in our release builds is `github.com/kevinburke/go-bindata` https://github.com/hashicorp/waypoint/blob/da80898a086a3fc8a4f19b7180d7d198927e5355/.circleci/config.yml#L160

krantzinator (changes requested):
Sorry, coffee is still settling in :sweat_smile: The bindata repo should be updated, then this looks good!

tools/tools.go
briancain: Oooh yeah, I remember this now. Thanks!

krantzinator (approved):

PR #864, opened 2020-12-01
Add Connect item to UI header

### LINKED ISSUE #781: Setup CLI Helper in UI
We should have an additional way to set-up the CLI outside of on-boarding. Basically this step:

image

But as an option or dropdown from the navbar alongside "invite" or similar.

dizzyup:
Adds a new "Connect" button to the UI header that displays a popover containing the `context create` CLI command that is shown during the initial onboarding flow. This resolves #781.

#### New menu item for `Connect`

Connect Link

#### Displays a popover containing the `<ContextCreate>` component

Connect Popover

### DIFF
--- a/ui/app/components/actions/context.hbs
+++ b/ui/app/components/actions/context.hbs
@@ -1,0 +1,13 @@
+<span class="hint-link connect-link" {{on "click" this.toggleHint}}>
+  {{inline-svg "bolt" class="nav-icon"}}
+  Connect
+</span>
+
+{{#if this.hintIsVisible}}
+  <div class="cli-hint">
+    <p>Connect to your Waypoint instance by running the following command in the CLI</p>
+    <ContextCreate></ContextCreate>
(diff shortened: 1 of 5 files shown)


gregone (approved):

ui/app/components/actions/context.hbs
```diff
@@ -0,0 +1,13 @@
+<span class="hint-link connect-link" {{on "click" this.toggleHint}}>
+  {{inline-svg "bolt" class="nav-icon"}}
+  Connect
```
pearkes, line 3: A small nit from me here is I'm wondering if "Connect" implies it'll like somehow connect _to_ the UI. Feels a bit awkward language-wise. Could just "CLI" work? "Setup CLI" is probably the clearest but not sure if that really fits up there.

pearkes (commented):
Just some feedback on the language in the nav here.

ui/app/components/actions/context.hbs
dizzyup: Thought about this a bit - I reckon just "CLI" should work. I'll make the change and merge.

PR #867, opened 2020-12-01
internal/config: support waypoint.hcl.json for HCL-flavored JSON syntax

mitchellh:
The way we load (using the hclsimple package) will automatically use the
JSON loader if it ends in ".json", we just never accepted ".json"
before.


briancain (approved):

briancain (approved):

mitchellh:
I verified this works with this example config: https://gist.github.com/mitchellh/474622ca67013ad14d114f1850a71db6

PR #868, opened 2020-12-02
Config Sourcer Plugin Docs

mitchellh:
Paired with: https://github.com/hashicorp/waypoint-plugin-sdk/pull/8

This adds the ability to autogenerate config sourcer plugin docs. They look like the screenshot below.

**These are not the docs for "config sourcing" in general. Those are WIP in a diff branch.**

This isn't the cleanest work but its sort of how are docs are done at the moment. :) I think we can take a pass at cleaning this up in the future I think its more important to get some docs generated.

image


krantzinator (approved):
+1 for some docs!

Issue #782, opened 2020-11-10
UI Versioning and information


pearkes:
We should utilize the [`GetVersionInfo`](https://github.com/hashicorp/waypoint/blob/9f504477707e7c2884c0d0d0a1adf2ad3c8d6fcc/internal/server/proto/server.proto#L17) API to retrieve and display helpful debugging information in the UI.

pearkes:
Fixed by #869.

PR #869, opened 2020-12-02
UI Versioning and information

gregone:
https://github.com/hashicorp/waypoint/issues/782

Adding version information to the UI

Screen Shot 2020-12-07 at 17 40 07


ui/app/components/app-footer/index.hbs
```diff
@@ -0,0 +1,5 @@
+<footer id="footer">
+  {{inline-svg "hashicorp" class="icon"}}
+  <span class="copyright">&copy;{{current-year null}} HashiCorp</span>
```
pearkes, line 3:
```suggestion
  <span class="copyright">&copy; {{current-year null}} HashiCorp</span>
```
This is minor but there's usually whitespace here I think (https://demo.consul.io/ as reference).

ui/app/components/version-info/index.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="version-info">
+  <ul class="version-info-list">
+    <li>Waypoint {{this.versionInfo.version}}</li>
+    <li>API version: {{this.versionInfo.api.current}}</li>
```
pearkes, line 4: Do you think we should maybe make these versions clickable to the docs eg. the protocol version info section?

ui/app/components/version-info/index.ts
```diff
@@ -0,0 +1,23 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
+import ApiService from 'waypoint/services/api';
+import { Empty } from 'google-protobuf/google/protobuf/empty_pb';
+import { VersionInfo as info } from 'waypoint-pb';
+
+export default class VersionInfo extends Component {
+  @service api!: ApiService;
+  @tracked versionInfo: info.AsObject | undefined;
```
pearkes, line 10: I think if you assign this in the constructor you can actually not need the undefined, but maybe that would break the async behavior?

pearkes (approved):
Looks good, question and a minor suggestion.

ui/app/components/version-info/index.hbs
dizzyup: I think adding links would help give that data some context too.

Also wondering if we can shorten the API and entry point versions labels a bit - could we make them `API v1` and `Entrypoint v1`?

ui/app/styles/_footer.scss
```diff
@@ -0,0 +1,33 @@
+
+@use 'pds/tokens/color';
+@use 'pds/tokens/scale';
+@use 'pds/tokens/font';
+@use 'pds/tokens/typography' as Typography;
```
dizzyup, line 5: I don't think you need these two imports.

ui/public/images/hashicorp.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 106.86 113.14"><title>iAsset 1</title><g id="Layer_2" data-name="Layer 2"><g id="Logo"><polygon points="44.54 0 0 25.69 0 25.71 0 87.41 16.73 97.07 16.73 35.35 44.54 19.3 44.54 0"/><polygon points="62.32 0 62.32 49.15 44.54 49.15 44.54 30.81 27.8 40.47 27.8 103.44 44.54 113.12 44.54 64.11 62.32 64.11 62.32 82.33 79.05 72.67 79.05 9.66 62.32 0"/><polygon points="62.32 113.14 106.86 87.45 106.86 87.43 106.86 25.73 90.12 16.07 90.12 77.79 62.32 93.84 62.32 113.14"/></g></g></svg>
```
dizzyup, line 1: We can tidy this markup a bit and add a `currentColor` property to simplify our CSS.

```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 106.86 113.14" fill="currentColor"><polygon points="44.54 0 0 25.69 0 25.71 0 87.41 16.73 97.07 16.73 35.35 44.54 19.3 44.54 0"/><polygon points="62.32 0 62.32 49.15 44.54 49.15 44.54 30.81 27.8 40.47 27.8 103.44 44.54 113.12 44.54 64.11 62.32 64.11 62.32 82.33 79.05 72.67 79.05 9.66 62.32 0"/><polygon points="62.32 113.14 106.86 87.45 106.86 87.43 106.86 25.73 90.12 16.07 90.12 77.79 62.32 93.84 62.32 113.14"/></svg>
```

ui/app/styles/_footer.scss
```diff
@@ -0,0 +1,33 @@
...
+  color: rgb(var(--footer-text));
+  display: flex;
+  align-items: center;
+  font-size: 0.75rem;
+  margin: 0 auto;
+  width: 100vw;
+  max-width: 1600px;
+  padding: scale.$lg-3 scale.$lg-4 scale.$lg-5;
+  .icon {
+      fill: rgb(var(--footer-text));
```
dizzyup, line 22:
```suggestion
```

We can drop this and inherit from the footer with the SVG markup further down.

ui/app/styles/_footer.scss
```diff
@@ -0,0 +1,33 @@
...
+
+#footer {
+  color: rgb(var(--footer-text));
+  display: flex;
+  align-items: center;
+  font-size: 0.75rem;
+  margin: 0 auto;
+  width: 100vw;
+  max-width: 1600px;
+  padding: scale.$lg-3 scale.$lg-4 scale.$lg-5;
```
dizzyup, line 20: I'd reduce the top padding here as the #page element gives us quite a bit already.

```suggestion
  padding: scale.$sm-4 scale.$lg-4 scale.$lg-5;
```

ui/app/styles/_footer.scss
```diff
@@ -0,0 +1,33 @@
...
+  .icon {
+      fill: rgb(var(--footer-text));
+      margin: 0 scale.$sm-1 0 0;
+  }
+
+  .version-info-list {
+    display: flex;
+    justify-content: space-between;
+    li {
+      padding: 0 scale.$sm-1;
```
dizzyup, line 30: I think we'd get some more consistent horizontal spacing if we reduced this padding to only be on the right side.

```suggestion
      padding-right: scale.$base;
```

dizzyup (commented):
Made some minor styling/markup suggestions but otherwise this is great.

ui/app/components/version-info/index.ts
gregone: I initially had that, but the type checker was yelling at me because `toObject` below returns `AsObject | undefined` (it also took me a while to figure out that was the issue)

ui/app/components/version-info/index.hbs
gregone: Other option I considered: styling them in a `warning`/`error` color if minimum > current . Thoughts on that @dizzyup ?

dizzyup: Oh interesting! Perhaps if we're doing that check we could return an additional badge instead - would sit inline next to the current version and say "Update available".

gregone: Added: 
Screen Shot 2020-12-08 at 16 49 59

gregone: @dizzyup labels and version numbers are updated too

pearkes (approved):

PR #872, opened 2020-12-03
Config Sync Docs

mitchellh:
This creates the reference documentation for config syncing, including the `config` stanza.

I also fixed a couple small help-related bugs.


website/content/docs/app-config/dynamic.mdx
```diff
@@ -0,0 +1,162 @@
...
+7PMDER: [WARN]  entrypoint.config.watcher: error reading configuration values, all will be dropped: source=kubernetes
+err="unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined"
+7PMDER: [INFO]  entrypoint.child: starting child process: args=[/usr/local/bin/docker-entrypoint.sh, node, app.js]
+cmd=/usr/local/bin/docker-entrypoint.sh
+7PMDER: Example app listening on port 3000!
+```
+
+You can see the first line has an error reading configuration value. If
+you keep scrolling right, you can see the `source` is `kubernetes` and the
+`err` is shown. In this case, its telling us that it couldn't load any
```
briancain, line 130:
```suggestion
`err` is shown. In this case, it's telling us that it couldn't load any
```

briancain (approved):
Found a small typo but otherwise looks great to me! :tada: :rainbow: :scroll:

krantzinator (approved):
Love the mock troubleshooting help as well

PR #873, opened 2020-12-03
Fix slice panic in serviceName

sarendsen:
Fix slice panic when app name is 5 characters or less.


evanphx:
@sarendsen thank you for the patch! We ended up solving it a slightly different way, but thank you for helping point it out!

PR #874, opened 2020-12-03
internal/ceb: exec sessions should get all config vars

mitchellh:
This fixes an issue where `waypoint exec` sessions weren't getting
access to any of the config vars set with `waypoint config` or the
config stanza, including dynamically synced values.

This is important so that use cases like `waypoint exec rake db:migrate`
do the right thing.

### DIFF
--- a/internal/ceb/config.go
+++ b/internal/ceb/config.go
@@ -92,6 +92,11 @@ func (ceb *CEB) watchConfig(
 	// and support automatically reinitializing if the URL service changes.
 	didInitURL := false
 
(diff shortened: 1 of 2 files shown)


mitchellh:
Note: this is a regression (on main, never released). I verified in 0.1.x that config vars properly show up. 

This regression is because we changed the core config receive loop and how we build commands. 0.1.x didn't restart the running application so it never really inspected or used env vars after first launch, but it did store them for `exec`. In changing this behavior for 0.2 we broke exec sessions. This resolves that.

krantzinator (approved):

PR #880, opened 2020-12-03
AWS SSM (Parameter Store) Config Source Plugin

mitchellh:
What the title says! Docs with example shown below:

image

### DIFF
--- a/builtin/aws/ssm/config_sourcer.go
+++ b/builtin/aws/ssm/config_sourcer.go
@@ -1,0 +1,241 @@
+package ssm
+
+import (
+	"context"
+	"math"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/aws/aws-sdk-go/service/ssm"
+	awsbase "github.com/hashicorp/aws-sdk-go-base"
+	"github.com/hashicorp/go-hclog"
+	"github.com/mitchellh/mapstructure"
+	"google.golang.org/grpc/codes"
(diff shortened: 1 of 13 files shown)


builtin/aws/ssm/config_sourcer.go
```diff
@@ -0,0 +1,383 @@
...
+			if param.Name == nil || param.Value == nil {
+				// should never happen
+				log.Warn("param name or value is nil", "param", param)
+				continue
+			}
+
+			for _, k := range paramToReq[*param.Name] {
+				v, ok := values[k]
+				if !ok {
+					// skip params we don't know, since we should prepopulate it all.
```
evanphx, line 201: Could you expand on this comment? I'm not sure why we're updating known values but ignoring new ones.

evanphx (changes requested):

builtin/aws/ssm/config_sourcer.go
mitchellh: Yes, the short of it is: this should never happen. The why is: we prepopulate `values` and requests on/around line 97. If AWS is well behaved, it should only send us the things we requested. I assume that's certainly true, but we have this check here to avoid a nil exception in case it isn't true.

evanphx: AH HA! It was a question of _what_ was prepopulated that I missed.

evanphx (approved):

PR #881, opened 2020-12-03
internal/server: Block on instance deployment becoming available

### LINKED ISSUE #683: Detect workload connectivity to server during `up` and `deploy`
A lot of users get started with Waypoint installed to a local Docker or K8S cluster. When moving to real workloads (say on a real remote ECS, K8S, etc. cluster), these workloads can't communicate with the Waypoint server which makes all entrypoint-related functionality inoperative.

Detect workload connectivity during `up` and `deploy` by adding a step to wait for at least one instance to connect back to the server.

briancain:
This commit adds an extra _optional_ timeout that blocks on having a
request returned when listing instances by deployment ID. The default
general case is to not block at all, but allows the caller to define an
optional timeout and block on a deployment ID existing before returning
the requested result.

This commit does not fully fix #683, as there are more complex CLI
changes required to leverage this server change.


internal/server/singleprocess/service_instance.go
```diff
@@ -2,10 +2,13 @@ package singleprocess
 
 import (
 	"context"
+	"time"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"github.com/hashicorp/go-memdb"
```
mitchellh, line 10: Nitpick: there's a newline above this and prob shouldn't be. Our usual import styling is:

```go
import (
  STDLIB

  EXTERNAL

  INTERNAL
)
```

internal/server/singleprocess/service_instance.go
```diff
@@ -18,8 +21,43 @@ func (s *service) ListInstances(
...
+		timeout := "0"
+		if req.ConnectTimeout != "" {
+			timeout = req.ConnectTimeout
+		}
+		connectTimeout, err := time.ParseDuration(timeout)
+		if err != nil {
+			return nil, err
+		}
+
+		if connectTimeout > 0 {
```
mitchellh, line 38: I think you can change the above timeout parsing stuff and change this if to `if req.ConnectTimeout !=""`

internal/server/singleprocess/service_instance.go
```diff
@@ -18,8 +21,43 @@ func (s *service) ListInstances(
...
+			_, cancel := context.WithTimeout(ctx, connectTimeout)
```
mitchellh, line 39: you gotta assign the context here, so `ctx, cancel =` otherwise the wait below won't wait on the timeout one.

internal/server/singleprocess/service_instance.go
```diff
@@ -18,8 +21,43 @@ func (s *service) ListInstances(
...
+
+		for {
+			ws := memdb.NewWatchSet()
+			result, err = s.state.InstancesByDeployment(scope.DeploymentId, ws)
+
+			if err != nil {
+				return nil, err
+			}
+
+			if len(result) > 0 || connectTimeout == 0 {
```
mitchellh, line 51: building on the above, can change the 2nd part here to `req.ConnectTimeout == ""`

internal/server/singleprocess/service_instance_test.go
```diff
@@ -39,6 +39,7 @@ func TestListInstances(t *testing.T) {
 				Scope: &pb.ListInstancesRequest_DeploymentId{
 					DeploymentId: dep.Id,
 				},
+				ConnectTimeout: "0s",
```
mitchellh, line 42: Probably good to test this with zero value (just don't set it at all)

internal/server/proto/server.proto
```diff
@@ -1583,6 +1583,9 @@ message ListInstancesRequest {
     // the app are returned.
     Ref.Workspace workspace = 2;
   }
+
+  // Time to wait before retrying a request to connect to requested instance
+  string connect_timeout = 3;
```
mitchellh, line 1588: Nitpick: thinking about this, suggesting this field be named `wait_timeout` instead of connect since we're waiting for instances to appear (which does imply connection so 🤷 )

mitchellh (commented):
Looks great! All my comments are super nitpicky or small Go idioms. The core logic is right.

This should be testable so I think we should get a test in here. I can help pair with you on that cause I think it'll be a little sneaky. 😄

internal/server/singleprocess/service_instance.go
briancain: Hmm, when I had initially done that the compiler said `ctx` isn't used :thinking: I tried naming it to something else and I got the same error

mitchellh: Yeah you gotta use `=` and not `:=` so that it overwrites the existing value. You may have to set `var cancel context.CancelFunc` above this line so that that is declared in that case.

briancain: Ohhh right, my eyes skipped over the `:` :+1: thanks

briancain:
@mitchellh yeah, def agree we need a proper test. I wasn't 100% sure how to go about it so I'd love to pair! We can coordinate on slack

internal/server/singleprocess/service_instance.go
```diff
@@ -18,8 +20,40 @@ func (s *service) ListInstances(
...
+		// For more complex deployments, sometimes it takes a bit longer for
+		// the instance to respond. This blocks on a response when querying
+		// a deployment by id, and lets the user know it is taking a while
+		//
+		// The default case is no timeout, no blocking. Make a request and return
 
+		if req.WaitTimeout != "" {
+			connectTimeout, err := time.ParseDuration(req.WaitTimeout)
+			if err != nil {
+				return nil, err
```
mitchellh, line 32: This `err` is gonna be a Go error, we should probably wrap this in a status:

```go
status.Errorf(codes.FailedPrecondition, "Error parsing wait_timeout: %s", err)
```

We're not like super consistent about this but its a good idea when we recognize it.

mitchellh (approved):

krantzinator (approved):
nice! :taco:

Issue #672, opened 2020-10-21
Uninstall/Remove Waypoint


wahmedswl:
**Is your feature request related to a problem? Please describe.**
There should be instructions to ```Uninstall/Remove Waypoint``` fully, ```K8S``` generated stuff etc

Thanks

mitchellh:
Coming in Waypoint 0.2.

wahmedswl:
Thanks, i have to re-create k3s cluster to get rid of #669. Was not able to re-create working context and it seems the faster way ...

Thanks

Issue #612, opened 2020-10-18
waypoint reinstall/upgrade


cpswan:
**Is your feature request related to a problem? Please describe.**
If I run `waypoint install -platform=docker -accept-tos` where there is an existing install then I see:

```
⚠️ Detected existing waypoint server
! Error getting the initial token: server is already bootstrapped

  The Waypoint server has been deployed, but due to this error we were
  unable to automatically configure the local CLI or the Waypoint server
  advertise address. You must do this manually using "waypoint context"
  and "waypoint server config-set".
```

But there are time when I may want to overwrite or upgrade the existing install e.g. when I've just downloaded a newer version of the waypoint binary and the waypoint server docker image.

**Describe the solution you'd like**
When detecting new binary and docker image waypoint should ask:

```
Detected existing waypoint server, do you want to upgrade (y/n)?
```

Otherwise:

```
Detected existing waypoint server, do you want to reinstall (y/n)?
```

**Describe alternatives you've considered**
I've been manually killing and removing docker containers running waypoint server then running `waypoint install -platform=docker -accept-tos` then using `waypoint context` and `waypoint server config-set`. Seems like a lot of yak shaving that waypoint could do for me.

mitchellh:
We at the very least want a stop command which is #479 (which would include "Uninstall" in my mind). 

We're not sure yet on upgrades and so on but I'll keep this open as we think about it.

mitchellh:
Slated for 0.2.

hholst80:
I have version 2.0 but I still have the same issue? How do I completely uninstall the software so I get a fresh install experience?

hholst80:
Suggested uninstall process:

1. sudo pkill waypoint
2. docker stop waypoint-server ; docker rm waypoint-server ; docker volume rm waypoint-server
3. rm ~/.config/waypoint 

also, the current directory holds additional configuration so better start with a new directory.

briancain:
@hholst80 - Uninstalling the server will depend on what platform you've installed it to, but since you mention docker:

- docker stop waypoint-server (stop the server container from running)
- docker rm waypoint-server (remove it from docker)
- docker volume rm waypoint-server (note that this will delete any data saved by the server like the token generated during the bootstrap step... so you're ensuring that you start with a fresh volume the next time you install)
- waypoint context delete \<your current server context>

[...]

PR #882, opened 2020-12-04
internal/server: refactor install into one package

krantzinator:
Foundation to addressing #672 and #612
There's a lot of diff from moving existing code around, so I'll highlight actual changes here:
- Moved the platforms to implement an `Installer` interface, replacing the `case` setup in `InstallCommand` to set the platform
- Took `serverinstall.Config` out of `InstallCommand` and gave each `Platform` its own config
  - This was the main place I went in circles with; I wanted each Platform to have its own config, and only care about the flags it needed to. I tried out a separate `serverinstall/config` package; toyed with using `serverconfig`; moved the `InstallFlags` pieces from `InstallCommand` to `serverinstall` to separate packages and back again. I also worked on this piece with a couple teammates, experimenting with different ways to have generic flags and install them to the scoped platforms as part of init. Ultimately I decided to go with the Platform-specific flags you see here in order to overcome the various hurdles discovered.

Small pieces:
- fixed a bug in the k8s statefulset; we weren't actually using the value set in the `pull-policy` flag, and updated the flag to have an updated value of `IfNotPresent` to match k8s default (`ImagePullPolicy: apiv1.PullPolicy(c.ImagePullPolicy),`
- threw a `strings.ToLower` on the platform name for UX

Copying Evan's example of including examples in lieu of a CLI testing framework:
```
$ waypoint-dev install -platform=nomad -nomad-dc=dc1 -accept-tos
✓ Nomad allocation ready
✓ Configuring server...
...

$ waypoint-dev install -platform=docker -accept-tos
✓ Installing Waypoint server to docker
✓ Server container started!
✓ Configuring server...
...

$ waypoint-dev install -platform=kubernetes -accept-tos
✓ Creating Kubernetes resources...
✓ Kubernetes StatefulSet reporting ready
✓ Waiting for Kubernetes service to become ready..
✓ Configuring server...
...
```


internal/serverinstall/docker.go
```diff
@@ -14,14 +14,25 @@ import (
...
 	"github.com/hashicorp/waypoint/internal/serverconfig"
 )
 
-func InstallDocker(
-	ctx context.Context, ui terminal.UI, scfg *Config) (
+type DockerInstaller struct {
+	Config DockerConfig
+}
+
+type DockerConfig struct {
```
mitchellh, line 30: I think we can lowercase all these and keep them unexported (same with Nomad/K8S).

internal/serverinstall/serverinstall.go
```diff
@@ -0,0 +1,22 @@
...
+	"context"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/clicontext"
+	"github.com/hashicorp/waypoint/internal/pkg/flag"
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+type Installer interface {
```
mitchellh, line 13: Doc comment and also recommend commenting the interface methods. Take a look at some of our other interfaces we _usually_ do this.

internal/serverinstall/serverinstall.go
```diff
@@ -0,0 +1,22 @@
...
+	Install(ctx context.Context, ui terminal.UI, log hclog.Logger) (*clicontext.Config, *pb.ServerConfig_AdvertiseAddr, string, error)
+	InstallFlags(*flag.Set)
+}
+
+var Platforms map[string]Installer = map[string]Installer{
```
mitchellh, line 18: You should be able to just do this... but test it...

```suggestion
var Platforms = map[string]Installer{
```

mitchellh (changes requested):
Yeahhhhhhh this is way cleaner.

Just a couple tiny comments but this is good work.

internal/serverinstall/serverinstall.go
krantzinator: Ah right, of course!

mitchellh (approved):

briancain (approved):
:package: :tada:

PR #886, opened 2020-12-05
internal/ceb: logger should connect when config is ready, not child

mitchellh:
The end user behavioral change here is that the log stream will begin
sending to the server as soon as EntrypointConfig is established. This
means that logs will appear even if the child command is not entering
the "ready to start case", which was happening when poorly behaved
config sources were blocking.

Internally, this  changes our state tracking in the CEB to use bools and a
condition var. Previously we were using an atomic+channel. This changes the
requirement to one shared cond var + bool per state.

We aren't using a state enum because these are not mutually exclusive
states, we can have both states true at once.

**Why now?** This is one of the pain points I noticed usertesting config
syncing end to end. If the config source is buggy it could fail to start the
app, but logs also don't come through. This is part of the mitigations I 
want to put in place so users can debug this.

### DIFF
--- a/internal/ceb/ceb.go
+++ b/internal/ceb/ceb.go
@@ -46,6 +46,16 @@ type CEB struct {
 	context      context.Context
 	execIdx      int64
 
+	// stateCond and its associated locker are used to protect all the
+	// state-prefixed fields. These state fields can be watched using this
+	// cond for state changes in the CEB. Anyone waiting on stateCond should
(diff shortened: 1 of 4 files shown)


krantzinator (approved):
This is really cool, I like this a lot.

evanphx (approved):

briancain (approved):

Issue #783, opened 2020-11-10
Polling flash in UI


pearkes:
The polling of the browser UI to the server to retrieve new data is visible and causes some of the elements in the browser to "flash". This could potentially be avoided by utilizing ember-concurrency's "last result" as a cache that can be displayed as a fallback during the reload, instead of emptying out the displayed data.

pearkes:
Fixed by #896.

PR #896, opened 2020-12-09
Resolve polling flash in UI

gregone:
Issue: https://github.com/hashicorp/waypoint/issues/783 

I initially went down the rabbit hole of changing the polling pattern to avoid calling `model.refresh()`, and instead creating an Ember Concurrency task to wrap the model fetching instead. 

Then while debugging I realized the issue was probably caused by the model resolving partial data (creating "gaps" in the state as the UI refreshes). It turn out that using [`rsvp.hash`](https://api.emberjs.com/ember/release/classes/rsvp/methods/hash?anchor=hash), to wait for all promises to resolve, works just fine.

### DIFF
--- a/ui/app/routes/workspace/projects/project/app.ts
+++ b/ui/app/routes/workspace/projects/project/app.ts
@@ -5,7 +5,7 @@ import { Ref, Deployment, Build, Release, Project } from 'waypoint-pb';
 import PollModelService from 'waypoint/services/poll-model';
 import ObjectProxy from '@ember/object/proxy';
 import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
-import { resolve } from 'rsvp';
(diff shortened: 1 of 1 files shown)


mitchellh (commented):
I'm not an authority but this looks right to me. :P

pearkes (approved):
My understanding here is that `hash` will wait for all the promises to resolve before returning (rendering the route) which will resolve the flash but it's worth considering performance profile down the road. For now, totally agree this is the right trade-off.

gregone:
@pearkes definitely. the issue here (I believe) is that when `model.refresh` is called, the existing data gets clobbered if part of it is still in flight, so I'll be spending more time later to restructure the service and fetching to only update it on the controller without re-firing the route events.

PR #897, opened 2020-12-10
cli{docs}: alphabetize install flags

krantzinator:


internal/cli/install.go
```diff
@@ -182,6 +182,13 @@ func (c *InstallCommand) Run(args []string) int {
 func (c *InstallCommand) Flags() *flag.Sets {
 	return c.flagSet(0, func(set *flag.Sets) {
 		f := set.NewSet("Command Options")
+		f.BoolVar(&flag.BoolVar{
+			Name:    "accept-tos",
+			Target:  &c.flagAcceptTOS,
+			Usage:   acceptTOSHelp,
+			Default: false,
+		})
```
mitchellh, line 190: Fine for this PR but raising a flag (pun intended) we should modify modify our internal flag package to alphabetize internally.

briancain (approved):

PR #901, opened 2020-12-10
bust cache if ui/lib changes

gregone:
Files in `ui/lib` may change without a change to `yarn.lock`, causing the wrong cached assets to be built. 

This PR: 
- creates a md5 hash of `ui/lib` and outputs it to a text file 
- uses the hash of that file as an additional item on the Yarn cache key (redundant hash of a hash, I know, but Circle only allows `checksum` on a file)
- invalidates the build cache if `ui/lib` has any changes (or at least that's the plan)

The only important commit is https://github.com/hashicorp/waypoint/pull/901/commits/0537dceab6dedc47f9582f7a7a586f23ee637b63 , the other ones are whitespace changes to verify cache key behavior

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ references:
     test-results: &TEST_RESULTS_DIR /tmp/test-results
 
   cache:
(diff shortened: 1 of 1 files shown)


krantzinator (approved):
Let's try it!

PR #912, opened 2020-12-11
Flag to install runner, support installing runner with Docker

mitchellh:
This adds a flag to `waypoint install` to install a runner. This is a hidden flag for now for dev work. I need this to work on runner-related functionality more easily for this release. 😄 

The `waypoint install` command grabs an auth token for the runner and calls a new `InstallRunner` platform installer function. This way the platform can install the runner and the runner can connect & auth to the server for registration.

Only Docker support for installing runners in this PR since this is still experimental. The K8S/Nomad implementation is a noop but would definitely be done prior to a release.

### DIFF
--- a/internal/serverinstall/serverinstall.go
+++ b/internal/serverinstall/serverinstall.go
@@ -13,8 +13,36 @@ import (
 // Installer is implemented by the server platforms and is responsible for managing
 // the installation of the Waypoint server.
 type Installer interface {
-	Install(ctx context.Context, ui terminal.UI, log hclog.Logger) (*clicontext.Config, *pb.ServerConfig_AdvertiseAddr, string, error)
+	// Install expects the Waypoint server to be installed.
+	Install(context.Context, *InstallOpts) (*InstallResults, error)
+
+	// InstallFlags is called prior to Install and allows the installer to
+	// specify flags for the install CLI. The flags should be prefixed with
+	// the platform name to avoid conflicts with other flags.
 	InstallFlags(*flag.Set)
+}
+
+// InstallOpts are the options sent to Installer.Install.
+type InstallOpts struct {
+	Log hclog.Logger
+	UI  terminal.UI
+}
+
+// InstallResults are the results expected for a successful Installer.Install.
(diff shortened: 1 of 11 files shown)


internal/serverinstall/serverinstall.go
```diff
@@ -8,15 +8,67 @@ import (
...
 
 // Installer is implemented by the server platforms and is responsible for managing
 // the installation of the Waypoint server.
 type Installer interface {
-	Install(ctx context.Context, ui terminal.UI, log hclog.Logger) (*clicontext.Config, *pb.ServerConfig_AdvertiseAddr, string, error)
+	// Install expects the Waypoint server to be installed.
+	Install(context.Context, *InstallOpts) (*InstallResults, error)
+
+	// InstallRunner expects a Waypoint runner to be installed.
+	InstallRunner(context.Context, *InstallRunnerOpts) error
```
mitchellh, line 21: I had to make this a separate function call because we need to have an auth token for the runner, and to get an auth token we have to be bootstrapped, and the bootstrapped happens in the shared `waypoint install` logic _after_ `Install`. 

So the flow is: (1) p.Install (2) common logic bootstraps and gets runner token (3) call p.InstallRunner.

evanphx: TOTALLY. I almost wonder if we should call it `InstallPhaseTwo` on the off hand there could be further components. But we can always change the name WHEN we have further components.

evanphx (approved):
https://sdk.bitmoji.com/render/panel/a38e717b-9601-4351-b3a3-911b8d2d93c3-8cb64c44-6a64-4f62-9642-460dd37039d6-v1.png?transparent=1&palette=1

internal/serverinstall/serverinstall.go
mitchellh: Yeah this is a non-pluggable thing (just internally extendable) so I think we can cross that bridge later.

briancain (approved):
Looks good to me! :runner:

Issue #687, opened 2020-10-24
Port Mapping for Docker Deployments does not honour service_port definition.


AniGarikipati:
**Describe the bug**
When service_port is described in the waypoint.hcl configuration and waypoint up is run, docker still maps a random port to the desired port on container. 

Flask configuration specification 

`
if __name__ == "__main__":
    app.run(debug=True, host='0.0.0.0', port=5000)
`

Dockerfile 
```
FROM python:3.8.2
RUN pip install flask 
COPY main.py / 
ENTRYPOINT ["python"]
CMD ["main.py"]
EXPOSE 5000
```

Deploy function in waypoint hcl
```
    deploy {
        use "docker" { 
            service_port=5000
            command = ["main.py"]
            static_environment = {
                 "FLASK_ENV" : "development",
                 "DISCOVERY_DOCKER_LABEL" : "luminos"
            }
        }
    }
```

Docker ps results
```
CONTAINER ID       COMMAND                  CREATED             PORTS                              
dccf68850438        "python main.py"         3 minutes ago      0.0.0.0:32796->5000/tcp           
```

**Steps to Reproduce**
Setup a flask container using a specific port, and defined service_port in waypoint.hcl as that port. Deploy using waypoint up, and run docker ps. The port mapping does not match.

**Expected behavior**
service_port should match the port that the container is mapped to.

**Additional context**
Tested on local machine, Windows 10

fatgit:
https://www.waypointproject.io/plugins/docker#service_port

Service port is a port in the container. So you random host port mapped to corrected service (container) port.
Unfortunately, I can not see how to specify the port on the host.

Wakeful-Cloud:
I'm having this problem too (On Ubuntu 20.04). I wonder if it's as simple as changing [builtin/docker/platform.go (Line 167)](https://github.com/hashicorp/waypoint/blob/42a08ff010d3d1ba1229b3027fff6573f4312507/builtin/docker/platform.go#L167) from:
```go
{
	HostPort: "",
},
```
to:
```go
{
	HostPort: port, //"port" is declared slightly above
},
```

Currently, I'm relying on [manually editing the Docker container host port](https://stackoverflow.com/a/49371983/9280403).

1xyz:
> Because of the nature of Waypoint, implementing this isn't really possible. We're going to write up some solutions on how to achieve what you're looking for with other tools like traefik.

Can you please share the solution(s) on how to map a "non-random" port on the host to the port within the container?

PR #918, opened 2020-12-15
Fixes https://github.com/hashicorp/waypoint/issues/687

Wakeful-Cloud:
Fixes https://github.com/hashicorp/waypoint/issues/687 by specifying a host port (for Docker deployments) instead of an empty string (Which leads to random port assignment).


briancain (approved):
Thanks for the pull request!

PR #920, opened 2020-12-15
Use @hashicorp/pds-ember addon and remove in-repo lib

gregone:
This PR replaces the in-repo library with the official `@hashicorp/pds-ember` addon. 

@dizzyup Please let me know if you see things that I may have missed.


dizzyup:
@gregone This is awesome!

Style-wise, the only major thing I notice is the bottom border that now gets applied to links on hover globally - would be worth suppressing that for now - it works fine for links that are text only but it looks a bit awkward in other spots.

CleanShot 2021-01-04 at 11 59 45@2x

Looks like it's being applied by this:

CleanShot 2021-01-04 at 12 04 27@2x

I'm also getting an error transitioning to the `/releases/release`, `/deployments/deployment from the list of releases/deployments etc - something up with the breadcrumbs service:

CleanShot 2021-01-04 at 12 02 08@2x

Logs are looking a little bit odd now too (ignore the 400/500 errors there, that's just my local):

CleanShot 2021-01-04 at 12 06 17@2x

 Again just looks like a new global style that's clobbering them:

CleanShot 2021-01-04 at 12 07 35@2x

The tabs on the app pages are looking a bit off now too, because I think they have a mix of the `pds` styles and our custom ones. We should probably drop our custom styles in favour of the official ones.

CleanShot 2021-01-04 at 12 09 29@2x

dizzyup (changes requested):
Left a comment collecting a few issues I'm seeing - but overall this is excellent!

CleanShot 2021-01-04 at 12 10 38@2x
😍

gregone:
@dizzyup Thanks for the feedback and the detailed review. I've just rebased the branch to keep it up to date, and I'll be addressing these items

gregone:
@dizzyup I think I've address all of your feedback: 

- [x] removed bottom border on links
- [x] removed default blue on links
- [x] reset the `<code>` default styling
- [x] cleaned up tabs and used the addon component for those
- [x] default tab colors overriden to avoid contrast issue 

One other tiny thing I noticed and didn't address, is that there seems to be a small change to the default line height (it's now a little bigger), but i personally found that more readable, so I didn't change it back. Any thoughts on that? 

Screen Shot 2021-01-07 at 15 31 51
Screen Shot 2021-01-07 at 15 35 03

dizzyup:
@gregone Yep, this all looks great to me.

> One other tiny thing I noticed and didn't address, is that there seems to be a small change to the default line height (it's now a little bigger), but i personally found that more readable, so I didn't change it back, except for the log lines. Any thoughts on that?

Agreed, I think the new line-height feels good.

dizzyup (approved):

PR #921, opened 2020-12-15
Refactor snapshot cli commands

krantzinator:
This pulls out the core snapshot logic from the CLI into its own package, so we can use the snapshot creation in our uninstall and upgrade functions.

### DIFF
--- a/internal/clisnapshot/snapshot.go
+++ b/internal/clisnapshot/snapshot.go
@@ -1,4 +1,6 @@
-package snapshot
+package clisnapshot
+// Package clisnapshot provides access to our CLI commands to create and
+// restore snapshots
 
 import (
 	"context"
@@ -10,13 +12,11 @@ import (
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 )
 
-type Config struct {
-	// This Client points to a project's client, as set in the baseCommand
-	Client pb.WaypointClient
-}
-
-func (c *Config) WriteSnapshot(ctx context.Context, w io.Writer) error {
-	stream, err := c.Client.CreateSnapshot(ctx, &emptypb.Empty{})
+// WriteSnapshot accepts a context, WaypointClient, and io.Writer, and returns
+// an error. It uses the Client to create a snapshot and write it to the 
+// provided writer.
+func WriteSnapshot(ctx context.Context, client pb.WaypointClient, w io.Writer) error {
+	stream, err := client.CreateSnapshot(ctx, &emptypb.Empty{})
 	if err != nil {
 		return fmt.Errorf("failed to generate snapshot: %s", err)
 	}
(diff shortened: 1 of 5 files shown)


internal/clisnapshot/snapshot.go
```diff
@@ -0,0 +1,104 @@
+package snapshot
```
mitchellh, line 1: Lets add a package doc comment.

internal/clisnapshot/snapshot.go
```diff
@@ -0,0 +1,104 @@
...
+
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+type Config struct {
+	// This Client points to a project's client, as set in the baseCommand
+	Client pb.WaypointClient
+}
+
+func (c *Config) WriteSnapshot(ctx context.Context, w io.Writer) error {
```
mitchellh, line 18: I think I'd prefer just top-level functions here rather than hanging methods off a `Config`. I always find it a bit strange when a `Config` struct has methods unless its to manipulate/validate the config.

`func WriteSnapshot(ctx context.Context, client pb.WaypointClient, w io.Writer)` 

We can also use the option pattern for other args later if we wanted to. Not a huge deal to break compatibility here cause its an internal package.

internal/clisnapshot/snapshot_test.go
```diff
@@ -0,0 +1,43 @@
...
+		Client: client,
+	}
+
+	// Create a temporary directory for our test
+	td, err := ioutil.TempDir("", "test")
+	require.NoError(err)
+	defer os.RemoveAll(td)
+	path := filepath.Join(td, "fancyserver")
+
+	w, err := os.Create(path)
```
mitchellh, line 30: defer the close

internal/clisnapshot/snapshot_test.go
```diff
@@ -0,0 +1,43 @@
...
+	require.NoError(err)
+
+	err = config.WriteSnapshot(ctx, w)
+	require.NoError(err)
+
+	require.FileExists(path)
+
+	r, err := os.Open(path)
```
mitchellh, line 38: defer the close

mitchellh (changes requested):
Looks great, bikeshed like the name.

evanphx (approved):

briancain (approved):
:+1:

internal/clisnapshot/snapshot.go
```diff
@@ -1,4 +1,6 @@
-package snapshot
+package clisnapshot
+// Package clisnapshot provides access to our CLI commands to create and
+// restore snapshots
```
mitchellh, line 4: This has to go above the `package` directive for godoc to pick it up

internal/clisnapshot/snapshot.go
```diff
@@ -10,13 +12,11 @@ import (
...
-type Config struct {
-	// This Client points to a project's client, as set in the baseCommand
-	Client pb.WaypointClient
-}
-
-func (c *Config) WriteSnapshot(ctx context.Context, w io.Writer) error {
-	stream, err := c.Client.CreateSnapshot(ctx, &emptypb.Empty{})
+// WriteSnapshot accepts a context, WaypointClient, and io.Writer, and returns
+// an error. It uses the Client to create a snapshot and write it to the 
+// provided writer.
```
mitchellh, line 18: This is _totally fine_ but in the interest of sharing my viewpoints (you don't have to change this at all just sharing my opinion): you don't need to reiterate the args cause we can see that in the signature, so something like below would be enough in this case.

> WriteSnapshot requests a snapshot from the client and writes it to the provider writer.

I usually will try to add details that may be helpful that the signature doesn't show, such as:

> Cancelling the context will prematurely cancel the snapshot. This may result in partial
> writes to the writer.

In that case, knowing this function may partially write when cancelled (and that it can be canceled) _might_ be useful. Again, you don't have to add this but just showing an example of my thought process around docs/comments.

mitchellh (approved):
One thing you prob should change (the package doc) and then the rest is just informational!

internal/clisnapshot/snapshot.go
krantzinator: All good! That's helpful background on godoc strategy. It did feel repetitive.

PR #922, opened 2020-12-15
internal/server: Store URL Token in state to prevent re-registering guest account

briancain:
Prior to this pull request, if a user upgraded or brought back up a Waypoint
server, it would re-register itself with Horizon and get a new URL,
leaving the old guest account behind. This commit fixes that by storing the API
token, and skipping any registering if it exists from prior server
state.


internal/server/singleprocess/service_url.go
```diff
@@ -11,6 +11,16 @@ import (
 )
 
 func (s *service) initURLGuestAccount(acceptURLTerms bool) error {
+	// Check if API Token already exists, if so, no reason to
+	// re-register and generate a new hostname
+	apiToken, err := s.state.ServerAPITokenGet()
+	if err != nil {
+		return err
+	} else if apiToken != "" {
+		// token already set, guest account already exists
```
mitchellh, line 20: I think the success case has to set `s.urlConfig.APIToken` for this to work properly (see L52)

internal/server/singleprocess/state/server_apitoken.go
```diff
@@ -0,0 +1,27 @@
+package state
+
+import (
+	"github.com/boltdb/bolt"
+)
+
+var (
+	serverAPIToken = []byte("api_token")
```
mitchellh, line 8: Recommend renaming this and all the function calls to `url_token` and `ServerURLTokenSet` since `api_token` is relatively vague and somewhat confusing cause we also ave API tokens for Waypoint itself and this is unrelated.

mitchellh (changes requested):

internal/server/singleprocess/service_url.go
```diff
@@ -40,5 +50,9 @@ func (s *service) initURLGuestAccount(acceptURLTerms bool) error {
 	}
 
 	s.urlConfig.APIToken = accountResp.Token
+	if err := s.state.ServerURLTokenSet(accountResp.Token); err != nil {
```
briancain, line 53: Responding to your comment about setting the token in the success case...I should be doing it here for the initial check at the beginning of this function to work... is that what you meant @mitchellh ?

mitchellh: I don't think so.

I think you just need to put:

> s.urlConfig.APIToken = accountResp.Token

where my original comment is. I think all this here is fine.

briancain: Oh right, I see what you mean now! :sweat_smile: :+1:

evanphx (approved):

mitchellh (approved):

PR #931, opened 2020-12-17
builtin/aws/ecs: properly clamp the service name. Closes #873

evanphx:
The old code would panic (see #873), this new version achieves the aim of the existing code, but in a bonus move, doesn't panic at all!

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -889,6 +889,12 @@ func (p *Platform) Launch(
 			*listener.ListenerArn, *listener.LoadBalancerArn)
 	} else {
 		lbName := "waypoint-ecs-" + app.App
+
+		// Names have to be 32 characters or less, so clamp it here.
+		if len(lbName) > 32 {
+			lbName = lbName[:32]
+		}
+
 		dlb, err := elbsrv.DescribeLoadBalancers(&elbv2.DescribeLoadBalancersInput{
 			Names: []*string{&lbName},
 		})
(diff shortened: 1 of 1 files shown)


builtin/aws/ecs/platform.go
```diff
@@ -773,9 +773,14 @@ func (p *Platform) Launch(
...
 	s.Update("Registered Task definition: %s", family)
-	rand := id[len(id)-(31-len(app.App)):]
 
-	serviceName := fmt.Sprintf("%s-%s", app.App, rand)
+	serviceName := fmt.Sprintf("%s-%s", app.App, id)
+
+	// We have to clamp at a length of 32 because the Name field to CreateTargetGroup
+	// requires that the name is 32 characters or less.
+	if len(serviceName) > 32 {
+		serviceName = serviceName[:32]
```
briancain, line 782: Maybe worth logging that the name is getting trimmed here due to a limitation in `CreateTargetGroup`?

```suggestion
                L.Info("Trimming service name to 32 characters due to limitation in 'CreateTargetGroup' name field")
		serviceName = serviceName[:32]
```

briancain (approved):
Hurray for no panics! :tada:

builtin/aws/ecs/platform.go
evanphx: Nah, because the user doesn't really need to even know about this value as it's just fed to the targetgroup name.

briancain: Ahhh, cool :+1: Yeah, if they don't know about this value they probably won't care that it got trimmed

briancain (approved):
Hurray for less panics!

krantzinator (approved):
_whack_

evanphx:
> Hurray for less panics!
_____________^^^^^_______ - Syntax error detected, perhaps you meant _fewer_?

Issue #934, opened 2020-12-18
[Install][k8s] Unable to stop stucked installation process


Moep90:
**Versions**

Waypoint
```bash
$ waypoint --version
Welcome to Waypoint
Docs: https://waypointproject.io
Version: v0.2.0
```
Kubernetes:
```bash
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.0", GitCommit:"af46c47ce925f4c4ad5cc8d1fca46c7b77d13b38", GitTreeState:"clean", BuildDate:"2020-12-16T12:12:57Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.6", GitCommit:"d32e40e20d167e103faf894261614c5b45c44198", GitTreeState:"clean", BuildDate:"2020-05-20T13:08:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Describe the bug**
1. Installed PVC like described here: https://github.com/hashicorp/waypoint/issues/933
2. Installating waypoint to kubernetes: `waypoint install --platform=kubernetes -k8s-namespace=waypoint -accept-tos`
3. Trying to stop the process in CLI (NOTE: I ran this command once before but the results are the same beside the already existing message)
4. ![image](https://user-images.githubusercontent.com/3042152/102585013-e0619480-4107-11eb-8a68-3dd3e4964232.png)
5. I've to close the terminal window and open a new one

**Steps to Reproduce**
```
# pvc.yaml is from https://github.com/hashicorp/waypoint/issues/933
$ kubectl apply -f pvc.yaml
$ waypoint install --platform=kubernetes -k8s-namespace=waypoint -accept-tos
# pressing CTRL + C or CTRL + SHIFT + C to usually stop the process
```

**Expected behavior**
The installation process stops immediate after it gets killed.

**Additional context**
https://github.com/hashicorp/waypoint/issues/933

krantzinator:
Thanks for the issue. This is something we need to fix. In the meantime, you can you use `ctrl+\` to send a sigquit signal and terminate the process without completely exiting the terminal window.

krantzinator:
This fix was part of the 0.2.1 release https://github.com/hashicorp/waypoint/pull/989/files

PR #941, opened 2020-12-21
k8s install: Return error creating service/statefulset.

andrewsomething:
When attempting to install Waypoint into a namespace that does not exist, the `waypoint install` command will hang indefinitely. Errors are printed to screen, but the process never exits.

```
$ waypoint install -accept-tos -k8s-namespace=waypoint -platform=kubernetes
✓ Creating Kubernetes resources...
⠧ Waiting for Kubernetes StatefulSet to be ready...
! Error creating service namespaces "waypoint" not found
! Error creating statefulset namespaces "waypoint" not found
```

The error is checked and used to populate `ui.Output`, but it is not actually returned.

With this change, it fails correctly:

```
$ ./waypoint install -accept-tos -k8s-namespace=waypoint -platform=kubernetes
❌ Creating Kubernetes resources...
! Error creating service namespaces "waypoint" not found
! Error installing server into kubernetes: namespaces "waypoint" not found
```

Looks to be the same issue as reported in https://github.com/hashicorp/waypoint/issues/934

### DIFF
--- a/internal/serverinstall/k8s.go
+++ b/internal/serverinstall/k8s.go
@@ -217,6 +217,7 @@ func (i *K8sInstaller) Install(
 			"Error creating service %s", clierrors.Humanize(err),
 			terminal.WithErrorStyle(),
 		)
+		return nil, err
 	}
 
 	statefulSetClient := clientset.AppsV1().StatefulSets(i.config.namespace)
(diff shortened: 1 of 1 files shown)


krantzinator (approved):

krantzinator:
Thanks for this! Pretty important 😅

PR #945, opened 2021-01-05
Update UI readme 

gregone:
Adds instructions to run a local server build, and to generate the type definitions

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -45,6 +45,13 @@ accept the invalid certificate warning in your browser for this to work.
 - `ember serve local`
 - The app will be available at [http://localhost:4200](http://localhost:4200).
 
+### Generating Type Definitions after making api changes
+
+if you've made API changes in `/internal/server` and want to use those on the frontend, you'll need to generate the type definitions again: 
+
+- `go generate ./internal/server`
+- `make gen/ts`
+
 ### Code Generators
 
 Make use of the many generators for code, try `ember help generate` for more details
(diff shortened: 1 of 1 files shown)


ui/README.md
```diff
@@ -45,6 +45,21 @@ accept the invalid certificate warning in your browser for this to work.
 - `ember serve local`
 - The app will be available at [http://localhost:4200](http://localhost:4200).
 
+If you need to build the server and run it locally, you'll want to stop the existing instance, build and reinstall it in docker:
+
+- `docker stop waypoint-server; docker rm waypoint-server; docker volume prune -f`
+- `make docker/evanphx`
```
briancain, line 51: Seeing this reminds me that maybe we should have a generic make target for building the server? Maybe we could rename this to something like `make docker/server` or something in a separate PR? @evanphx @krantzinator  @mitchellh 

I've also been using the `docker/mitchellh` target because it tags the server as `dev` rather than overriding `latest`

briancain (approved):
Looks good to me! Left a comment about some potential future work for us outside of this PR...I think it is worth having a general make target for building a local server rather than the personalized ones we have :smile:

ui/README.md
briancain: I'm happy to open that PR now, maybe I'll just copy what I've been using and name it `docker/server`.

gregone: Thanks @briancain. I just reworded this

briancain: Thanks! :) Just merged that one so I think this should be good to go

PR #946, opened 2021-01-05
Add a generic build target for waypoint server container

briancain:
This pull request just copies the `docker/mitchellh` make target to something more generic for building a waypoint server. I'm happy to remove the other targets if they aren't going to be used, but don't want to mess up anyones workflow by removing them yet :smile:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,15 @@ test: # run tests
 format: # format go code
 	gofmt -s -w ./
 
+.PHONY: docker/server
+docker/server:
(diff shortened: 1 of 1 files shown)


gregone (approved):

gregone:
@briancain Thanks! I'll update https://github.com/hashicorp/waypoint/pull/945 to recommend that target instead

krantzinator (approved):

mitchellh:
Yeah I guess this is pretty generic now. Historically it had some personal bits on there. I can probably remove mine.

PR #948, opened 2021-01-06
feat: add support to env vars in hcl config files

dagi3d:
This pull request exposes the environment vars so they can be used anywhere inside the hcl file.

```
...
registry {
  use "docker" {
    image = "hello-world-${env["APP_ENV"]}"
    tag   = "1"
    local = true
  }
}
```
or

```
app "bar" {
    path = "./bar"

    labels = {
        "env": env["APP_ENV"]
    }
}
```

```
APP_ENV=staging waypoint up
```

Bear in mind that I'm pretty neophyte with Go and the exposing approach could be considered incorrect as I am not familiar with the project internals.
Also, I saw that Terraform accepts only environment variables that start with the TF_VAR_ prefix and you might want to have a similar behaviour in the long term.

### DIFF
--- a/internal/config/app_test.go
+++ b/internal/config/app_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -10,6 +11,9 @@ import (
 )
 
 func TestConfigApp_compare(t *testing.T) {
+	envVar := "test"
+	os.Setenv("APP_ENV", envVar)
+
 	cases := []struct {
 		File string
 		App  string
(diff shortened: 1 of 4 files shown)


mitchellh:
Thank you! This looks pretty clean but I think we're going to discuss whether we want to expose env vars globally like this or only in certain contexts.

orenmazor:
this would be great. I currently deploy with the exec plugin, and have as part of the command a snippet that parses `aws sts get-caller-identity` to figure out where I'm deploying to. this would let me remove all of that (I need to know so waypoint knows to start a bastion for me)

AutomationD:
@mitchellh Any word on the strategy for env vars?

evanphx:
Hi all, in the 0.4.0 release we're going to add support for variables like terraform variables. That will cover all the needs here plus more.

PR #950, opened 2021-01-07
Don't block server startup on URL service

mitchellh:
If the URL service is down, the server fails to startup. We should not behave this way.

This changes the server to attempt a blocking connection _once_. If that fails, we move to background retries. This will enable the server to come up when the URL service is down (though it may try the first blocking attempt for ~10 seconds). Additionally, logging is much improved here.

### DIFF
--- a/internal/pkg/grpcready/grpcready.go
+++ b/internal/pkg/grpcready/grpcready.go
@@ -1,0 +1,47 @@
+package grpcready
+
+import (
+	"context"
+
+	"github.com/hashicorp/go-hclog"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/connectivity"
+	"google.golang.org/grpc/status"
+)
+
+// Conn returns nil if the connection is in the ready state. If wait is
+// set, this will wait for the connection to become ready or until the context
+// is cancelled.
+//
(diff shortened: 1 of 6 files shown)


evanphx (approved):
Looks good. Got a little confused about the error checking on initURLClient initially but got it now.

briancain (approved):

internal/pkg/grpcready/grpcready.go
```diff
@@ -0,0 +1,47 @@
...
+
+		// If we're ready then we're done!
+		if s == connectivity.Ready {
+			log.Debug("connection is ready")
+			return nil
+		}
+
+		// If we have a transient error and we're not retrying, then we're done.
+		if s == connectivity.TransientFailure && !wait {
+			log.Warn("failed to connect to the server, temporary network error")
```
pearkes, line 38: It doesn't hae to be in the `grpcready` package but can we also log the connection address here? Probably the url service address either configured or default as a DNS name is the most helpful (so, likely not in this package).

mitchellh: Yeah we don't have that info here, but it is logged already now (in this PR)

pearkes: Ah great I missed it down there, sounds good.

PR #960, opened 2021-01-08
Add reminder about ui authentication command after install

gregone:


internal/cli/install.go
```diff
@@ -367,5 +367,8 @@ deployments. If this is incorrect, manually set it using the CLI command
 
 Advertise Address: %[2]s
 Web UI Address: %[3]s
+
+To launch and authenticate into the Web UI, run:
+waypoint ui -authenticate
```
briancain, line 372: Maybe this could move above here onto line 366? I kind of like the two addresses being the last thing you see in the output.

briancain (changes requested):
Oops, sorry we totally missed this one :heart: Makes sense to me, just a note for adjusting where I think it should go.

internal/cli/install.go
gregone: changed!

briancain (approved):
Nice, thanks!

PR #962, opened 2021-01-08
Git datasource auth support (basic and SSH)

mitchellh:
This completes [WP-043](https://docs.google.com/document/d/1X36B7UzfXCS9OpdVMYMEFa5faEDMIK_yEa3qh-G0YYI/edit) (HashiCorp internal).

This enables the Git datasource to clone authenticated repositories. The auth can be configured via the API or `waypoint.hcl`. In a `waypoint.hcl` it looks like this:

```hcl
project = "my-project"

runner {
  data_source "git" {
    url      = "https://github.com/hashicorp/waypoint.git"
    username = "github"
    password = "abcd1234"
    # key = file("id_rsa.key")
  }
}

app "web" {
  # ...
}
```

**Note:** we don't actually expect people to set secrets like this in the waypoint.hcl file, although it is possible. We will have alternate methods of configuring the secrets in an upcoming PR.


mitchellh:
Another important note: this switches our Git datasource from subprocessing `git` to using the `go-git` library. This is discussed in the RFC but is done because configuring auth when subprocessing is a big pain. I added many more tests to ensure our clone behavior works as expected.

internal/datasource/git.go
```diff
@@ -31,20 +36,66 @@ func (s *GitSource) ProjectSource(body hcl.Body, ctx *hcl.EvalContext) (*pb.Job_
...
+	case cfg.Username != "":
+		result.Auth = &pb.Job_Git_Basic_{
+			Basic: &pb.Job_Git_Basic{
+				Username: cfg.Username,
+				Password: cfg.Password,
+			},
+		}
+
+	case cfg.SSHKey != "":
+		result.Auth = &pb.Job_Git_Ssh{
```
evanphx, line 54: I think here we should do a little error checking that the given SSHKey can be decrypted using the given password. Because the usage of the key happens very far from the configuration, it will be hard for them to realize they used the wrong password (or no password at all) when we try to use the ssh key, but much easier if we check here and tell them.

internal/datasource/git.go
```diff
@@ -31,20 +36,66 @@ func (s *GitSource) ProjectSource(body hcl.Body, ctx *hcl.EvalContext) (*pb.Job_
...
+			Ssh: &pb.Job_Git_SSH{
+				PrivateKeyPem: []byte(cfg.SSHKey),
+				Password:      cfg.SSHKeyPassword,
+			},
+		}
+	}
```
evanphx, line 60: Should we check here that the given url and auth mechanism works?

evanphx (changes requested):
A few questions around checking that the git url and creds are correct when we configure them to prevent users having a bad time.

internal/datasource/git.go
mitchellh: Yes this is a very good idea.

internal/datasource/git.go
mitchellh: We don't really have to. The validation here isn't as bad as it appears. During `terraform init` we schedule a `ValidateOp` job and that should also validate that the cloning works properly. I think the key feedback here is that we should schedule this job as part of the UI workflow and maybe add a `waypoint project validate` CLI to the project CLI proposal as well.

evanphx: Gotcha, right. Sounds good!

evanphx (approved):

mitchellh:
Tested this and it "works": the data clones but a Docker build failed because it requires Docker-in-Docker. We need to use `img` or something similar. Another PR.

PR #967, opened 2021-01-11
POC: Nomad Exec

DingoEatingFuzz:
This is the result of an exploration into deploying exec workloads on Nomad via Waypoint.

The most interesting design decision that's worth discussing is how the Nomad plugin now maps from builtin "primitives" to a new `NomadSpec` type. In theory this will simplify writing various strategies in the Nomad plugin but even with just two options it's become a large gRPC message, so maybe it's a bad idea.


krantzinator:
@jeffwecan and @tehut may have some thoughts on this; they were just sharing their experiences around using nomad with exec in their Waypoint HashiStack today!

evanphx:
@DingoEatingFuzz I think the right way to proceed here is to write a joint Waypoint/Nomad team RFC to figure out the right direction. We can leave this open for now.

mitchellh:
I'm going to close this for now since we have no plans to work on this at the moment and I think the `nomad-jobspec` plugin in WP 0.4 makes this KIND OF possible. If we want to talk about this again, let's do it. Obviously this PR and its history will remain forever so we can always bring this back up.

PR #970, opened 2021-01-13
Support daemonless, unprivileged Docker builds with `img`

mitchellh:
This modifies the "docker" builder to support daemonless, unprivileged image builds using `img`: https://github.com/genuinetools/img

The fact we use `img` is encapsulated and not exposed to the user. This gives us the ability to use any tool in the future if we choose to go a different path. The important part is that the Docker builder does support a daemonless approach.

This is important for the runner functionality coming up so that Waypoint runners that will likely be running in containerized environments are able to perform image builds.

----------------------------------------------------------------

This PR only modifies the "docker" builder and entrypoint injection.

A future PR will modify the registry, `docker-pull`, and more`. I just didn't want to overwhelm reviewers.


internal/serverinstall/docker.go
```diff
@@ -291,7 +291,14 @@ func (i *DockerInstaller) InstallRunner(
...
 		},
-	}, nil, &network.NetworkingConfig{
+	}, &container.HostConfig{
+		// These security options are required for the runner so that
+		// Docker daemonless image building works properly.
+		SecurityOpt: []string{
+			"seccomp=unconfined",
+			"apparmor=unconfined",
+		},
+	}, &network.NetworkingConfig{
```
mitchellh, line 301: Note this is only modifying how we install the _runner_ and not the _server_.

(And the runner install stuff is currently behind a hidden experimental flag)

evanphx (approved):

Dockerfile
```diff
@@ -26,18 +30,87 @@ RUN go get github.com/kevinburke/go-bindata/...
...
+	gcc \
+	git \
+	libseccomp-dev \
+	linux-headers \
+	make
+
+RUN git clone https://github.com/mitchellh/img.git /img
+WORKDIR /img
+RUN git checkout inspect
+RUN go get github.com/go-bindata/go-bindata/go-bindata
```
krantzinator, line 53: we use `github.com/kevinburke/go-bindata/...` elsewhere

krantzinator (approved):

Dockerfile
mitchellh: Yeah, I'm using this one cause this is the one `img` requires haha. 
This is copied from their Dockerfile.

briancain (approved):
:framed_picture: :+1:

PR #972, opened 2021-01-13
feature: Uninstall server for all platforms

krantzinator:
Along with uninstall work I also:
- added a `const` for `serverName` that all platforms use
  - replaced k8s `serverName` and `serviceName` flags with consts
- made small improvements to Nomad install
- minor formatting in unrelated commands for consistency

### DIFF
--- a/internal/cli/uninstall.go
+++ b/internal/cli/uninstall.go
@@ -35,7 +35,6 @@ func (c *UninstallCommand) Run(args []string) int {
 		WithArgs(args),
 		WithFlags(c.Flags()),
 		WithNoConfig(),
-		WithClient(false),
 	); err != nil {
 		return 1
 	}
@@ -54,6 +53,8 @@ func (c *UninstallCommand) Run(args []string) int {
 	// - name the context we'll be uninstalling
 	// - generate a snapshot of the current install
 	s := sg.Add("")
+	defer func() { s.Abort() }()
+
 	contextDefault, err := c.contextStorage.Default()
 	if err != nil {
 		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
@@ -72,11 +73,8 @@ func (c *UninstallCommand) Run(args []string) int {
 		s.Update("Generating server snapshot...")
 		defer s.Abort()
 		// generate snapshot
-		config := clisnapshot.Config{
-			Client: c.project.Client(),
-		}
 		w, err := os.Create(c.snapshotFilename)
-		if err = config.WriteSnapshot(ctx, w); err != nil {
+		if err = clisnapshot.WriteSnapshot(ctx, c.project.Client(), w); err != nil {
 			fmt.Fprintf(os.Stderr, "Error generating snapshot: %s", err)
 			return 1
 		}
@@ -86,6 +84,9 @@ func (c *UninstallCommand) Run(args []string) int {
 		s.Status(terminal.StatusWarn)
 	}
 	s.Done()
+
+	// TODO: should we check if any deployments are running, and exit with 
+	// a warning to run `waypoint destroy` before proceeding?
 
 	// Uninstall
 	p, ok := serverinstall.Platforms[strings.ToLower(c.platform)]
@@ -173,7 +174,7 @@ func (c *UninstallCommand) Flags() *flag.Sets {
 		f.StringVar(&flag.StringVar{
 			Name:    "snapshot-filename",
 			Target:  &c.snapshotFilename,
-			Default: fmt.Sprintf("sever-snapshot-%d", time.Now().Unix()),
+			Default: fmt.Sprintf("waypoint-sever-snapshot-%d", time.Now().Unix()),
 			Usage:   "Filename to write the snapshot to.",
 		})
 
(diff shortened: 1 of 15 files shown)


internal/serverinstall/docker.go
```diff
@@ -319,3 +324,130 @@ func (i *DockerInstaller) InstallFlags(set *flag.Set) {
...
+	}
+
+	s.Update("Docker image %q removed", imageRef.Name())
+	s.Done()
+
+	return nil
+}
+
+func (i *DockerInstaller) UninstallFlags(set *flag.Set) {
+}
```
mitchellh, line 453: Would add a comment in the func body like `// Purposely empty, no flags`

internal/serverinstall/k8s.go
```diff
@@ -633,6 +621,203 @@ func (i *K8sInstaller) InstallFlags(set *flag.Set) {
...
+// a Kubernetes cluster
+func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
+	ui := opts.UI
+	log := opts.Log
+
+	sg := ui.StepGroup()
+	defer sg.Wait()
+
+	s := sg.Add("Inspecting Kubernetes cluster...")
+	defer s.Abort()
```
mitchellh, line 635: You gotta wrap this in a `func` or it'll abort just the first step created. You do this correctly in other implementations so probably just an easy miss. 😄

mitchellh (changes requested):
Yeah, really minor feedback this looks great!

internal/serverinstall/nomad.go
```diff
@@ -201,14 +211,14 @@ EVAL:
 	}, nil
 }
 
-// waypointNomadJob takes in a nomadConfig and returns a Nomad Job per the
+// newWaypointNomadJob takes in a nomadConfig and returns a Nomad Job per the
 // Nomad API
-func waypointNomadJob(c nomadConfig) *api.Job {
-	job := api.NewServiceJob("waypoint-server", "waypoint-server", c.region, 50)
+func newWaypointNomadJob(c nomadConfig) *api.Job {
```
briancain, line 216: I think this name should go back only because I'm using this function in Upgrades to submit the same job config to the current running Nomad job, so it might be confusing going forward since it's also the same config for updating a job.

Or maybe it could be renamed `waypointNomadJobConfig`?

internal/cli/uninstall.go
```diff
@@ -0,0 +1,203 @@
...
+
+		return 1
+	}
+
+	err = p.Uninstall(ctx, &serverinstall.InstallOpts{
+		Log: log,
+		UI:  c.ui,
+	})
+	if err != nil {
+		// point to current docs on manual server cleanup
```
briancain, line 101: Maybe a missing TODO here :smile: 

```suggestion
		// TODO: point to current docs on manual server cleanup
```

briancain (changes requested):
:tada: Looks good to me!! Just requesting a small change on a function name and then I'll be sure to do some manual testing on my end soon :) :+1:

internal/cli/uninstall.go
```diff
@@ -0,0 +1,203 @@
...
+	s := sg.Add("")
+	defer func() { s.Abort() }()
+
+	contextDefault, err := c.contextStorage.Default()
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+	s.Update("Uninstalling Waypoint server with context %q", contextDefault)
+	s.Done()
```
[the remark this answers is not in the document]
krantzinator: Good call

internal/cli/uninstall.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (c *UninstallCommand) Synopsis() string {
+	return "Uninstall the Waypoint server"
+}
+
+func (c *UninstallCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint server uninstall [options]
+	Uninstall the Waypoint server. The platform should be
```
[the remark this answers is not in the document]
krantzinator: It is too narrow! I know I had it at 80 and I have no idea when it moved to this width. Good catch.

internal/serverinstall/nomad.go
krantzinator: Oh good call. I'll just move it back to avoid those fun merge conflicts we're bound to run in to 😄

internal/cli/uninstall.go
krantzinator: Ah! I can go ahead and complete that todo now since we have our Troubleshooting page. Thanks for catching this!

mitchellh (approved):

internal/cli/uninstall.go
```diff
@@ -0,0 +1,204 @@
...
+			Name:    "platform",
+			Target:  &c.platform,
+			Default: "",
+			Usage:   "Platform to uninstall the Waypoint server from.",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "snapshot-name",
+			Target:  &c.snapshotName,
+			Default: fmt.Sprintf("waypoint-sever-snapshot-%d", time.Now().Unix()),
```
briancain, line 181: Ahhh, this is what I was talking about on my PR. I think this might be misleading in the help text, since the time will be different when the snapshot is taken? Maybe we can show the base name `waypoint-server-snapshot` here and mention in `Usage` that a timestamp will be applied when the snapshot is taken?

briancain, line 181:
```suggestion
			Default: fmt.Sprintf("waypoint-server-snapshot-%d", time.Now().Unix()),
```

krantzinator: Sure i'll update real quick to match!

briancain (approved):

PR #974, opened 2021-01-13
Daemonless, unprivileged Docker registry implementation

mitchellh:
This builds on #970 and adds daemonless Docker image push support.

This works by adding a new `location` field to the `Image` message used by the Docker plugin. Based on the location of the image, we can push it. Each Docker builder tends to have its own local registry or image format so this lets us use the correct logic for pushing.

Auth works as well.


builtin/docker/registry.go
```diff
@@ -43,102 +33,50 @@ func (r *Registry) Push(
...
+		Image: r.config.Image,
+		Tag:   r.config.Tag,
 	}
-
-	step.Done()
-
-	if r.config.Local {
-		return target, nil
+	if !r.config.Local {
+		target.Location = &Image_Registry{Registry: &empty.Empty{}}
```
evanphx, line 41: Why the logic change from the previous impl? Before we'd just return the image not doing anything. I think this will break various folks that had set the local flag.

evanphx (changes requested):
One question about the logic change for Local

builtin/docker/registry.go
mitchellh: This shouldn't break anyone since the `Location` field is new and didn't exist before. 

But I feel like the `Location` field should represent where the image is known to be. If we aren't pushing it to a remote repository, I think we should just retain what `Location` was previously set. Thoughts?

evanphx (approved):

krantzinator (approved):

mitchellh:
Merging into f-img

PR #975, opened 2021-01-13
Add instance types and instance targeting

evanphx:
This is laying the groundwork for future work. It adds a type field to instances to track (Long Running, On Demand, and Virtual).

Additionally this adds the ability to target a specific instance for an exec session.


internal/cli/exec.go
```diff
@@ -16,6 +16,71 @@ import (
...
+		}
+
+		exitCode, err = client.Run()
+		if err != nil {
+			app.UI.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+			return ErrSentinel
+		}
+
+		return nil
+	})
```
mitchellh, line 77: This is alright and all but I wonder if we should instead keep the same DoApp loop and just maybe add a couple functions we call out to to configure the `*execclient.Client`. I expect we'll have more targeting in the future (by deployment for example) and I don't want to set a precedent for just copying and pasting each time.

The actual instance lookup stuff looks fine!

internal/server/execclient/client.go
```diff
@@ -32,6 +32,10 @@ type Client struct {
 	Stdin         io.Reader
 	Stdout        io.Writer
 	Stderr        io.Writer
+
+	// If set, will cause the server to connect to this specific
+	// instance.
+	InstanceId string
```
mitchellh, line 38: I'd pull DeploymentId/DeploymentSeq down here too and just like use comments to document they're a `oneof`.

internal/server/proto/server.proto
```diff
@@ -1979,6 +1998,12 @@ message ExecStreamRequest {
 
     // Pty is set if we should allocate a PTY for this exec stream.
     PTY pty = 3;
+
+    // Instance to send the exec request to. This is indicates that
+    // the client wants an exec session to this specific instance,
+    // rather than one the server picks. Targetted instances can be
+    // any instance type, where as untargeted have to be LONG_RUNNING.
+    string target_instance_id = 4;
```
mitchellh, line 2006: Given `deployment_id` doesn't have a `target_` prefix, I'm gonna bike shed and say this one shouldn't either.

Also, unclear: if this is specified, should I _not_ specify deployment ID?

We may want to change it to this:

```proto
oneof target {
  string deployment_id = 1;
  string instance_id = 4;
}
```

(As long as you keep the same tag numbers, that is backwards compat)

internal/server/singleprocess/state/instance_exec.go
```diff
@@ -60,6 +60,36 @@ type InstanceExec struct {
 	Connected         uint32
 }
 
+func (s *State) InstanceExecCreateByTargetedInstance(exec *InstanceExec) error {
```
mitchellh, line 63: The `ByDeployment` uses `exec` just as an _out_ param, I think we should just keep following that pattern and pass instance ID as a param here. 

The implementation should set it on `InstanceExec` probably.

internal/server/singleprocess/state/instance_exec.go
```diff
@@ -77,6 +107,13 @@ func (s *State) InstanceExecCreateByDeployment(did string, exec *InstanceExec) e
 	for raw := iter.Next(); raw != nil; raw = iter.Next() {
 		rec := raw.(*Instance)
 
+		// When looking through all the instances for an exec capable instance
+		// we only consider LONG_RUNNING type instances. These are the only ones
+		// that it makes sense to send random exec sessions to.
+		if rec.Type != pb.Instance_LONG_RUNNING {
```
mitchellh, line 113: Can you add a unit test for this functionality?

mitchellh (changes requested):

internal/cli/exec.go
evanphx: Yeah, I considered that. This change was the cleanest diff but I can move it inside DoApp.

internal/server/execclient/client.go
evanphx: 👍

internal/server/proto/server.proto
evanphx: Totes fair. Also good call on the use of oneof here, that's the right move.

internal/server/singleprocess/state/instance_exec.go
evanphx: Sure.

internal/server/singleprocess/state/instance_exec.go
evanphx: Ah! I missed that one, yup.

internal/server/proto/server.proto
```diff
@@ -1646,6 +1646,25 @@ message Instance {
...
+  // Which type of instance this is
+  Type type = 5;
+
+  // Instances are one of a these types.
+  enum Type {
+    // The "traditional" instance type, a process that is running
+    // constantly for a long period of time.
+    LONG_RUNNING = 0;
+
+    // An instance that was launched in responce to a request and will
```
briancain, line 1659:
```suggestion
    // An instance that was launched in response to a request and will
```

internal/server/proto/server.proto
```diff
@@ -1646,6 +1646,25 @@ message Instance {
...
+    // disappear quickly.
+    ON_DEMAND = 1;
+
+    // An instance that is not actually running an code, but registers
```
briancain, line 1663:
```suggestion
    // An instance that is not actually running any code, but registers
```

briancain (approved):
:dart:

mitchellh (approved):

krantzinator (approved):
joining Team Approval

PR #976, opened 2021-01-13
feature: Waypoint Server Upgrade command for supported server platforms

### LINKED ISSUE #612: waypoint reinstall/upgrade

briancain:
This pull request introduces a new CLI command: `waypoint server upgrade`. For each supported server platform, it attempts to upgrade the server image in place while preserving the original database. It also by default takes a snapshot of the sever prior to upgrading in the case of any failures.

Fixes #612


briancain:
I see in PR #972 it's abstracted out the serverName and serviceName (k8s only). I'll be sure to do the same here once I can rebase off of Uninstall, or I can make the same changes here and let git handle it :smile:

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+	if !c.skipSnapshot {
+		s.Update("Taking server snapshot before upgrading")
+
+		snapshotName := fmt.Sprintf("%s-%d", defaultSnapshotName, time.Now().Unix())
+		if c.snapshotName != "" {
+			snapshotName = fmt.Sprintf("%s-%d", c.snapshotName, time.Now().Unix())
+		}
+
+		s.Update("Taking snapshot of server with name: '%s'", snapshotName)
+		w, closer, err := c.initWriter(snapshotName)
```
mitchellh, line 172: Can you explain why you made this `initWriter` helper instead of just calling `os.Create` directly? initWriter doesn't appear to do anything..

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+			s.Update("Failed to take server snapshot")
+			s.Status(terminal.StatusError)
+			s.Done()
+
+			c.ui.Output(fmt.Sprintf("Error opening output: %s", err), terminal.WithErrorStyle())
+			return 1
+		}
+
+		if closer != nil {
+			defer closer.Close()
```
mitchellh, line 183: Rather than defer this, I would do this instead across the next few lines. This ensures that the snapshot file is fsynced and written and closed right away rather than waiting for the whole upgrade process to complete (via a `defer`)

```go
err = clisnapshot.WriteSnapshot(c.Ctx, c.project.Client(), w)
w.Close()
if err !=nil {
```

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+		s.Update("Snapshot of server written to: '%s'", snapshotName)
+		s.Done()
+	} else {
+		s.Update("Server snapshot disabled on request, this means no snapshot will be taken before upgrades")
+		s.Status(terminal.StatusWarn)
+		s.Done()
+		log.Warn("Server snapshot disabled on request from user, skipping")
+	}
+
+	c.ui.Output("Waypoint server will now upgrade from version %q",
```
mitchellh, line 204: Are we able to say "to version" ?

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+			clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	s.Update("Verifying connection is valid for context %q...", ctxName)
+
+	client := pb.NewWaypointClient(conn)
+	if _, err := clientpkg.New(ctx,
```
mitchellh, line 134: I'd like to see a comment explaining why we initialize `clientpkg` if we don't use it. 

(It is important, we have to verify API compat, but I'd put a comment at least saying that)

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+		s.Update("Server snapshot disabled on request, this means no snapshot will be taken before upgrades")
+		s.Status(terminal.StatusWarn)
+		s.Done()
+		log.Warn("Server snapshot disabled on request from user, skipping")
+	}
+
+	c.ui.Output("Waypoint server will now upgrade from version %q",
+		initServerVersion)
+
+	c.ui.Output("Upgrading...", terminal.WithHeaderStyle())
```
mitchellh, line 207: Recommend putting this above the output above and turning the version info into a `WithInfoStyle()`

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+		c.ui.Output(upgradeFailHelp)
+
+		return 1
+	}
+
+	contextConfig := result.Context
+	advertiseAddr := result.AdvertiseAddr
+	httpAddr := result.HTTPAddr
+
+	originalCfg.Server.Address = contextConfig.Server.Address
```
mitchellh, line 228: Cool idea: detect if the context changed here (I think only address can change) and if so, output a warning that any other Waypoint CLI users have to update their contexts.

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+
+	resp, err = client.GetVersionInfo(ctx, &empty.Empty{})
+	if err != nil {
+		c.ui.Output(
+			"Error retrieving server version info: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle())
+		return 1
+	}
+
+	if err := c.contextStorage.Set(ctxName, originalCfg); err != nil {
```
mitchellh, line 255: Building on previous comment: only set context if the address actually changed, limits things that can fail

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+			"Error connecting to server: %s\n\n%s",
+			clierrors.Humanize(err),
+			errInstallRunning,
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+	client = pb.NewWaypointClient(conn)
+
+	resp, err = client.GetVersionInfo(ctx, &empty.Empty{})
```
mitchellh, line 247: I recommend setting the new context prior to making this call, to limit things that can fail before we persist the context.

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func (c *ServerUpgradeCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint server upgrade [options]
+
+	Upgrade Waypoint server in the current context to the latest version or the
+	server image version specified. By default, Waypoint will upgrade to server
+	version "hashicorp/waypoint:latest". Before upgrading, a snapshot of the
+	servers database will be taken in case of any upgrade failures.
```
mitchellh, line 337:
```suggestion
	server will be taken in case of any upgrade failures.
```

internal/serverinstall/docker.go
```diff
@@ -240,6 +240,201 @@ func (i *DockerInstaller) Install(
...
+			Key:   "reference",
+			Value: reference.FamiliarString(imageRef),
+		}),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(imageList) == 0 {
+		s = sg.Add("Pulling image: %s", i.config.serverImage)
```
mitchellh, line 323: Do we have to actually always pull? Otherwise `waypoint upgrade` on Docker will reuse an old `hashicorp/waypoint:latest`, right? 

We can probably follow the K8S/Nomad logic here and only always pull if the tag is `:latest`.

internal/serverinstall/k8s.go
```diff
@@ -371,6 +375,263 @@ func (i *K8sInstaller) Install(
...
+			"Error obtaining waypoint statefulset: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return nil, err
+	}
+
+	s = sg.Add("Upgrading server to %q", i.config.serverImage)
+
+	// Update pod image to requested serverImage
+	cmd := exec.Command("kubectl", "set", "image", "statefulset", serverName, fmt.Sprintf("server=%s", i.config.serverImage))
```
mitchellh, line 463: This has to be changed to use the K8S API directly, I don't think we assume `kubectl` is installed OR configured properly anywhere else [anymore].

internal/serverinstall/k8s.go
```diff
@@ -371,6 +375,263 @@ func (i *K8sInstaller) Install(
...
+		}
+
+		return true, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if waypointStatefulSet.Spec.UpdateStrategy.Type == "RollingUpdate" {
+		ui.Output("\nKuberntes is now set to upgrade waypoint server image with its\n" +
```
mitchellh, line 622:
```suggestion
		ui.Output("\nKubernetes is now set to upgrade waypoint server image with its\n" +
```

mitchellh (changes requested):

internal/cli/server_upgrade.go
briancain: Ah yeah I'll leave a note here explaining this. Although perhaps the client getting version info on 147 is enough to verify API compatibility?

internal/cli/server_upgrade.go
briancain: Yeah, so for this I had initially copied over the helpers for taking snapshots, but the original helper also allowed for it to be streamed directly in the CLI rather than saved to a file so I trimmed it down. I wasn't sure if I needed anything else so I kept the helper in place and moved on. But I don't think I do so I'll go ahead and simplify this!

internal/cli/server_upgrade.go
mitchellh: It isn't! You have to do the whole "negotiation" dance which clientpkg does. :).

internal/cli/server_upgrade.go
briancain: Let me check, the actual flag we're targeting comes from UpgradeFlags and I don't _think_ we have that available here in the CLI package. If so we should show it here :+1:

internal/cli/server_upgrade.go
briancain: Ok perfect! I'll write up a note then so it's more clear :) thanks!

internal/cli/server_upgrade.go
briancain: Yeah, good call on a warning. Given that an upgrade might happen long after the install, I'm sure the slight change in the address will be difficult to notice, so detecting that and warning is hopefully enough here.

internal/serverinstall/docker.go
briancain: Yeah, always pulling on `:latest` here is a good idea! Good catch :+1:

internal/serverinstall/k8s.go
briancain: Ah, I was afraid of that! So I noticed on our Install we use `kubectl`: https://github.com/hashicorp/waypoint/blob/6f6b1fb588c5514711bc542447b65a902e85ee60/internal/serverinstall/k8s.go#L156-L161

And was hoping maybe we could keep using it.

Otherwise, my plan was essentially to do exactly what `kubectl set image` does internally but for a single patch updating the pod image: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/set/set_image.go#L231

Seeing how `kubectl` handles updating k8s got me thinking if we need some kind of similar mechanism for updating k8s in Waypoint? Should we have an internal Patch function like this for future config updates and Upgrade just uses that here?

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,362 @@
...
+	} else {
+		s.Update("Server snapshot disabled on request, this means no snapshot will be taken before upgrades")
+		s.Status(terminal.StatusWarn)
+		s.Done()
+		log.Warn("Server snapshot disabled on request from user, skipping")
+	}
+
+	c.ui.Output("Upgrading...", terminal.WithHeaderStyle())
+
+	c.ui.Output("Waypoint server will now upgrade from version %q",
```
briancain, line 195: Oops, looks like re-arranging this removed the original PR review comment about this.

@mitchellh - So, I think the only way I can get this is by adding a `Config` to the Installer interface and exporting each platforms config? Not sure if that's a huge deal or if it's worth it for this message. Alternatively I could try to pass along the original version here to Upgrade or retrieve it again for each platform. Then at least there we'll have the version we're upgrading to easily.

mitchellh: No, not a big deal today probably not worth the effort!

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,362 @@
...
+	Upgrade Waypoint server in the current context to the latest version or the
+	server image version specified. By default, Waypoint will upgrade to server
+	version "hashicorp/waypoint:latest". Before upgrading, a snapshot of the
+	server will be taken in case of any upgrade failures.
+
+` + c.Flags().Help())
+}
+
+var (
+	defaultSnapshotName = "waypoint-server-snapshot"
```
krantzinator, line 342: Is there any reason not to have this be the default on the `snapshot-name` flag? Setting this directly as a default on the flag can clear up some of the logic on lines 157-159.

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,362 @@
...
+
+		err = clisnapshot.WriteSnapshot(c.Ctx, c.project.Client(), writer)
+		writer.Close()
+
+		if err != nil {
+			s.Update("Failed to take server snapshot")
+			s.Status(terminal.StatusError)
+			s.Done()
+
+			c.ui.Output(fmt.Sprintf("Error generating Snapshot: %s", err), terminal.WithErrorStyle())
```
krantzinator, line 180: This seems like a small rephrasing of the error message on line 176, don't think we need both? Or is this for the sake of the pattern we seem to have, where we update the Status with an error and then send the `err` itself up the chain?

internal/cli/server_upgrade.go
```diff
@@ -0,0 +1,362 @@
...
+			Name:    "platform",
+			Target:  &c.platform,
+			Default: "",
+			Usage:   "Platform to upgrade the Waypoint server from.",
+		})
+		f.StringVar(&flag.StringVar{
+			Name:    "snapshot-name",
+			Target:  &c.snapshotName,
+			Default: "",
+			Usage:   "Platform to upgrade the Waypoint server from.",
```
krantzinator, line 301: I use this wording on my `snapshot-name` flag on Uninstall:
```suggestion
			Usage:   "Filename to write the snapshot to.",
```

internal/serverinstall/docker.go
```diff
@@ -240,6 +240,201 @@ func (i *DockerInstaller) Install(
...
+		err = jsonmessage.DisplayJSONMessagesStream(resp, s.TermOutput(), termFd, true, nil)
+		if err != nil {
+			return nil, fmt.Errorf("unable to stream pull logs to the terminal: %s", err)
+		}
+		s.Done()
+		s = sg.Add("")
+	}
+
+	s.Update("Upgrading Waypoint server image from %q to %q",
+		waypointServerContainer.Image, i.config.serverImage)
```
krantzinator, line 350: nit: readability -- put each item on its own line since we're already breaking the line. i can't make a suggestion bc I can't gofmt it afterwards 😄 , but i mean:
```
(
  "Upgrading Waypoint server image from %q to %q",
  waypointServerContainer.Image,
  i.config.serverImage,
)
```

internal/serverinstall/k8s.go
```diff
@@ -263,7 +267,259 @@ func (i *K8sInstaller) Install(
...
+
+	s = sg.Add("Waiting for server to be ready...")
+	log.Info("waiting for waypoint server to become ready after image refresh")
+
+	var contextConfig clicontext.Config
+	var advertiseAddr pb.ServerConfig_AdvertiseAddr
+	var httpAddr string
+	var grpcAddr string
+
+	err = wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
```
krantzinator, line 520: i found the services to be slower than statefulsets at upgrading etc, so i'd bump this to 2 minutes just to be safe
```suggestion
	err = wait.PollImmediate(2*time.Second, 2*time.Minute, func() (bool, error) {
```

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
   is designed for you to know when and how to safely upgrade the server.
 
 Please see our [compatibility promise](/docs/upgrading/compatibility)
-for details on how to determine compatbility. This page will focus on
+for details on how to determine compatibility. This page will focus on
 how to upgrade components in both compatible and incompatible scenarios.
 
+### Server Upgrade Command
+
+As of 0.2.1, Waypoint provides a command for users to automatically upgrade
```
krantzinator, line 54: "automatically" implies hands-off to me; but the user still has to take action here and run the command, as they can't set the server to upgrade automatically on new releases.
```suggestion
As of 0.2.1, Waypoint provides a command for users to upgrade the
```

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
...
+Waypoint server in place. For each platform, Waypoint attempts to
+update the server image in place to latest by default, or to the server image
+requested from the passed in flag to the upgrade command.
```
krantzinator, line 57: confusing wording; suggestion for clarity
```suggestion
specified via a flag passed in to the upgrade command.
```

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
...
+
+~> **Note:** Due to where Waypoint server stored its database in Nomad at `/alloc`,
+the automated upgrade command cannot migrate the old servers database to a new one.
+This is only an issue if running Waypoint server 0.2.0 and older. Please follow
+the manual upgrade guide in this case.
+
+When the upgrade command is executed, the following takes place:
```
krantzinator, line 64: active tense, plus makes sense with using `Takes` on the following line
```suggestion
When the upgrade command is executed, Waypoint does the following:
```

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
...
+
+- Takes a snapshot of the current server database and save snapshot locally. The
+  snapshot file can be saved elsewhere if a new name is passed in through using
+  `-snapshot-name`, or optionally skipped all together
+- Run the upgrade for the requested platform:
```
krantzinator, line 69:
```suggestion
- Takes a snapshot of the current server database and saves snapshot locally. The
  snapshot file can be saved elsewhere if a new name is passed in using the flag
  `-snapshot-name`, or optionally skipped all together.
- Runs the upgrade for the requested platform:
```

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
...
+  - `docker` - This platform will force pull the image requested to update the
+    container in docker.
+  - `kubernetes` - This platform will use `kubectl` to set the image requested on
```
krantzinator, line 72: we'll need to remove this reference this if we are able to extricate the `kubectl` call from the statefulset update

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
...
+    the waypoint-server pod. By default, Waypoints StatefulSet is configured to
```
krantzinator, line 73:
```suggestion
    the waypoint-server pod. By default, Waypoint's StatefulSet is configured to
```

website/content/docs/upgrading/index.mdx
```diff
@@ -68,6 +106,25 @@ okay since all components continue to gracefully work while the server
 is down, and the downtime should be short. We plan on improving this
 in the future.
 
+#### Automated Upgrade
+
+1. Check any upgrade notes for the version you're upgrading to and verify
+   there are no issues that will affect your upgrade.
+
+2. Check the help text from running `waypoint server upgrade -help`
```
krantzinator, line 114:
```suggestion
2. Check the help text from running `waypoint server upgrade -help`.
```

website/content/docs/upgrading/index.mdx
```diff
@@ -41,9 +46,40 @@ considerations:
...
+  - `kubernetes` - This platform will use `kubectl` to set the image requested on
+    the waypoint-server pod. By default, Waypoints StatefulSet is configured to
+    `RollingUpdate`, so the pod will pull the requested version shortly after
+    running the Waypoint upgrade command.
+  - `nomad` - This platform will update the waypoint server job with the server
+    image requested.
+- Verifies that waypoint CLI can still connected and speak to Waypoint server
+
+If you wish to use the automated upgrade command line approach, follow the
+Standard Upgrade guide using the _Automated Upgrade_ Server section.
```
krantzinator, line 81: This wording was a little confusing for me to follow, maybe something like:
```suggestion
Standard Upgrade guide using the _Automated Upgrade_ Server Upgrade subsection.
```

krantzinator (changes requested):
😍

internal/cli/server_upgrade.go
briancain: Ah, yes! I was thinking if we left a default, we wouldn't get the timestamp and we'd overwrite previous snapshots with the same default name. But maybe our default can just be the "base name" and it operates the same as this in the case of no snapshot filename specified?

internal/cli/server_upgrade.go
briancain: Oops! Good catch on this string

website/content/docs/upgrading/index.mdx
briancain: Hm, makes sense to me. I guess it's less automatic and more _automated_. So leaving it off makes sense :+1:

website/content/docs/upgrading/index.mdx
briancain: Ah yep :+1: It was originally its own section, and it ended up making it a subsection, so nice catch!

internal/serverinstall/k8s.go
briancain: I think I can actually patch the pod itself (like how I'm deleting pods for OnDelete) rather than Statefulset, which I think makes this a bit easier. I'll give that a shot instead.

internal/cli/server_upgrade.go
krantzinator: Hm, I would rather not overwrite old snapshots automatically. I understand the point of that is to prevent someone from stacking up old snapshots, but what if they try an upgrade, it borks, they try again, and the second one also borks but they are left with a weird snapshot state now?

briancain: Ah, I think I wasn't clear enough in my previous comment! The way it works now is it doesn't overwrite previous snapshots, it uses this as a base name and appends a timestamp, so each run is a new snapshot.

I meant to say if we had set a default for the CLI, the help text would give the impression that it would save as that filename exactly, when in reality it shouldn't because we'd overwrite previous snapshots, so that's why the default value in the snapshot flag was left blank.

But maybe we set a default for the flag as the "basename" and still apply the timestamp at the end? Then update the help text to explain a timestamp is going to be appended to the default name.

internal/cli/server_upgrade.go
```diff
@@ -297,8 +298,9 @@ func (c *ServerUpgradeCommand) Flags() *flag.Sets {
 		f.StringVar(&flag.StringVar{
 			Name:    "snapshot-name",
 			Target:  &c.snapshotName,
-			Default: "",
-			Usage:   "Filename to write the snapshot to.",
+			Default: defaultSnapshotName,
+			Usage: "Filename to write the snapshot to. If no name is specified, by" +
+				"default a timestamp will be appended to the default snapshot name.",
```
krantzinator: Can you throw this change on `uninstall` too while you're merging  🙏🏻

briancain: Ah yes, good call! :+1:

krantzinator (approved):
🥇

PR #980, opened 2021-01-15
use random serial number per cert

krantzinator:
This change allows Firefox users to bypass the previous
`SEC_ERROR_REUSED_ISSUER_AND_SERIAL` error that blocked access to
the UI when a new server instance was installed or run locally.


mitchellh (approved):

briancain (approved):
:1234: :+1: :tada:

PR #982, opened 2021-01-15
feature: Add CEB Virtual instance type

evanphx:
This work lays the groundwork for the exec plugin, which will use the virtual CEB running inside a runner, with a VirtualExecHandler wired up to an exec function provided by the Deploy plugin.


mitchellh (approved):
I think this is good for now.

I think in the future after we get everything working we may want to refactor the `VirtualCEB` out into `internal/virtualceb` or something since the exported surface area of `internal/ceb` seems high and also a bit orthogonal in goals.

briancain (approved):

PR #986, opened 2021-01-17
customizable netlify client id, use Authenticate method

zdunecki:
I've resolved hardcoded Netlify OAuth client id and by the way also one of todo's related to Netlify authentication.


evanphx:
Based on state and usage, netlify is going to be remove in 0.4.

PR #989, opened 2021-01-18
internal/client: use given context so it can be cancelled

mitchellh:
While attempting to connect to a server, Ctrl-C was not working. This is
because we were using `context.Background` rather than the passed in
context when attempting to connect.

I don't know why we did this, it looks like a simple mistake from dev.
Verified that this makes Ctrl-C work properly for server connection.

Labeling for backport to 0.2.x.


krantzinator (approved):
sweet

briancain (approved):
great! :+1:

PR #990, opened 2021-01-18
internal/cli: make -platform required and do platform check earlier

mitchellh:
By moving the platform check earlier, we avoid generating a snapshot
locally on disk if there are flag validation issues.


briancain (approved):
:heavy_check_mark:

krantzinator (approved):

PR #994, opened 2021-01-19
internal/cli: colorize flags in help text and nested subcommands

mitchellh:
This modifies our CLI help output to do two things:

  1. We always highlighted commands such as "waypoint foo", but nested
     subcommands "waypoint foo bar" weren't highlighted. This fixes
     that.

  2. Flags in the help text body such as -foo and -foo-bar are now
     highlighted as well. This is only in the help text summary and not
     the flag documentation itself.

Screenshots (sorry my colors are very muted so its hard to see):

Flag highlight (see -waypoint-hcl):

CleanShot 2021-01-19 at 12 19 45@2x

Nested subcommand highlight:

CleanShot 2021-01-19 at 12 19 57@2x

### DIFF
--- a/internal/cli/help.go
+++ b/internal/cli/help.go
@@ -166,7 +166,7 @@ func (c *helpCommand) HelpTemplate() string {
 var (
 	reHelpHeader = regexp.MustCompile(`^[a-zA-Z0-9_-].*:$`)
 	reCommand    = regexp.MustCompile(`"waypoint (\w\s?)+"`)
-	reFlag       = regexp.MustCompile(`(\s|^)(-[\w-]+)(\s|$)`)
+	reFlag       = regexp.MustCompile(`(\s|^|")(-[\w-]+)(\s|$|")`)
 )
 
 const helpTemplate = `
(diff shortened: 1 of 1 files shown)


briancain (approved):
:art:

internal/cli/help.go
```diff
@@ -128,7 +165,8 @@ func (c *helpCommand) HelpTemplate() string {
 
 var (
 	reHelpHeader = regexp.MustCompile(`^[a-zA-Z0-9_-].*:$`)
-	reCommand    = regexp.MustCompile(`"waypoint \w+"`)
+	reCommand    = regexp.MustCompile(`"waypoint (\w\s?)+"`)
+	reFlag       = regexp.MustCompile(`(\s|^)(-[\w-]+)(\s|$)`)
```
krantzinator, line 169: I needed the below to get the `"-accept-tos"` bit in `install -h` to highlight.
```suggestion
	reFlag       = regexp.MustCompile(`(\s|^|")(-[\w-]+)(\s|$|")`)
```

krantzinator: We could instead remove all `"` around flags in help texts.

krantzinator: Or just `/W` and assume we won't ever have something like `$-this-is-not-a-flag#`

mitchellh: Ah yes. No this is fine with me!

PR #996, opened 2021-01-19
`waypoint project` CLI 

mitchellh:
This introduces `waypoint project list` and `waypoint project apply` CLIs.

(Note: there is no `waypoint project delete` cause we don't yet support project delete. Once we do, we'll add that.)

The "apply" CLI behaves as we described in the RFC, it merges wherever possible and everything is optional. We also have a `-waypoint-hcl` flag that can be used to specify a path to a `waypoint.hcl` file that we use to source config. Additional CLI flags can still be provided when using the HCL file mode to override values in there.

Example full CLI command to enable Git (no HCL file used):

```
waypoint project apply \
  -data-source=git \
  -git-url=https://github.com/mitchellh/waypoint-test-app.git \
  -git-auth-type=basic \
  -git-username=github \
  -git-password=abcd1234 \
  example-python
```

### DIFF
--- a/internal/cli/project_apply.go
+++ b/internal/cli/project_apply.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"io/ioutil"
+	"path/filepath"
 	"strings"
 
 	"github.com/posener/complete"
@@ -10,6 +11,8 @@ import (
 
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/internal/clierrors"
+	configpkg "github.com/hashicorp/waypoint/internal/config"
+	"github.com/hashicorp/waypoint/internal/datasource"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 )
@@ -25,6 +28,7 @@ type ProjectApplyCommand struct {
 	flagGitPassword    string
 	flagGitKeyPath     string
 	flagGitKeyPassword string
+	flagWaypointHcl    string
 }
 
 func (c *ProjectApplyCommand) Run(args []string) int {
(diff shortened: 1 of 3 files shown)


internal/cli/project_list.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	var result []string
+	for _, p := range resp.Projects {
+		result = append(result, p.Project)
+	}
+
```
krantzinator, line 38:
```suggestion

	if len(result) == 0 {
		c.ui.Output("No projects found.")
		return 0
	}
```

internal/cli/project_list.go
```diff
@@ -0,0 +1,77 @@
...
+	return "List all registered projects."
+}
+
+func (c *ProjectListCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint project list
+
+  List all registered projects.
+
+  Projects usually a single version control repository and contain
```
krantzinator, line 69:
```suggestion
  Projects usually map to a single version control repository and contain
```

internal/cli/project_list.go
```diff
@@ -0,0 +1,77 @@
...
+  exactly one "waypoint.hcl" configuration. A project may contain multiple
+  applications.
+
+  A project is registered either via the web UI, "waypoint project apply",
```
krantzinator, line 73: nit: `either` tripped me up in reading since i thought it was comparing two options
```suggestion
  A project is registered via the web UI, "waypoint project apply",
```

internal/cli/project_apply.go
```diff
@@ -0,0 +1,382 @@
...
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	// Setup our project that we're going to override
+	var proj *pb.Project
+	if resp != nil {
+		s.Update("Updating project %q...", name)
+		proj = resp.Project
```
krantzinator, line 84: For use in step confirmation
```suggestion
	var updated bool
	if resp != nil {
		s.Update("Updating project %q...", name)
		updated = true
```

internal/cli/project_apply.go
```diff
@@ -0,0 +1,382 @@
...
+		Project: proj,
+	})
+	if err != nil {
+		c.ui.Output(
+			"Error upserting project: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
```
krantzinator, line 267:
```suggestion

	if updated {
		s.Update("Project %q updated", name)
	} else {
		s.Update("Project %q created", name)
	}
```

krantzinator (changes requested):
I haven't reviewed the whole thing yet but I wanted to share these since I have to run for now. Will finish up later!

krantzinator (approved):
Looks good! As we move forward with this expanding feature work, we should heavily document where `project apply` intersects with or complements `init`. For example, running `init -update` after a `project apply` wasn't immediately clear to me.

mitchellh:
Totally, I intend to take a pass (as I always do) in the week(s) leading up to a 0.3 release to make sure we got everything covered.

evanphx (approved):

briancain (approved):

PR #998, opened 2021-01-20
Dockerfile: Ensure userid is 100 for waypoint server user

briancain:
Userid for waypoint inside waypoint server must remain at 100, otherwise
if the id changes and the server image is refreshed with a different id,
the waypoint user will fail to read boltdb with permission denied.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,9 +96,8 @@ COPY --from=builder /tmp/wp-src/waypoint-entrypoint /usr/bin/waypoint-entrypoint
 
 VOLUME ["/data"]
 
-# NOTE: userid must be 100 here. Otherwise upgrades prior to 0.2.1 will fail
-# due to user not having the proper permissions to read the server db due to
-# a different userid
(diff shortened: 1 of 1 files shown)


mitchellh (approved):

Dockerfile
```diff
@@ -96,20 +96,23 @@ COPY --from=builder /tmp/wp-src/waypoint-entrypoint /usr/bin/waypoint-entrypoint
 
 VOLUME ["/data"]
 
+# NOTE: userid must be 100 here. Otherwise upgrades prior to 0.2.1 will fail
```
krantzinator, line 99: Wouldn't changing this at any point break all upgrades from that point backward? like if we're in v0.6.2 and change this, then upgrades from 0.6.1 and prior would fail?

briancain: Yep, that's true. I'll simplify it to just say don't change it :D haha

krantzinator (approved):

PR #1000, opened 2021-01-21
serverconfig: store platform in server config

### LINKED ISSUE #893: Add Platform as a serverconfig option

In some cases, it will be useful for a Waypoint server to know what Platform it's been installed on. Like for example, including Platform as an option in a context when listing server contexts.

Include a string Platform that gets set in serverconfig when the server is installed. Likely from the `-platform` flag from `waypoint install`

n/a

Other use cases could be different operations dealing with the server that depend on what kind of platform it has been installed on.

Including a platform from serverconfig when listing server contexts:

```
brian@localghost:go % waypoint context list
   |        NAME        |  PLATFORM   |                                SERVER ADDRESS
---+--------------------+-------------+-------------------------------------------------------------------------------
   | install-1607024040 | docker     | localhost:9701
   | install-1606958260 | docker     | localhost:9701
   | install-1606955862 | docker     | localhost:9701
   | install-1607376497 | kubernetes | abcd1234.us-east-2.elb.amazonaws.com:9701

...

```

**Additional context**

briancain:
Prior to this commit, waypoint had no way internally to figure out what
kind of platform a server was, which meant it had to rely on the user to
pass the proper flag when doing any server operations on an existing
context. It also meant it was difficult to realize what a server
platform was when looking at existing context through `waypoint context
list`. This commit updates that to store the context passed in from
install.

For any previous context prior to this change, we will show `n/a` in the
`waypoint context list` output since we did not store that context platform.

```
brian@localghost:go % waypoint context list                                                                                                                                                                                           ±[main]
    |        NAME        | PLATFORM | SERVER ADDRESS
----+--------------------+----------+-----------------
    | install-1611251878 | n/a      | localhost:9701
    | install-1611251992 | docker   | localhost:9701
  * | install-1611252817 | docker   | localhost:9701
```

Fixes #893


internal/server/proto/server.proto
```diff
@@ -1162,6 +1162,8 @@ message ServerConfig {
     bool tls = 2;
     bool tls_skip_verify = 3;
   }
+
+  string platform = 4;
```
mitchellh, line 1166: can you document this?

mitchellh (approved):
This PR is good as-is. Is the plan to use this for `upgrade` and `uninstall` then to not require the `-platform` flag?

briancain:
@mitchellh - Yes! Eventually I think we can make it so that the platform flag won't be required for those commands, but we'll need users to use it at least for a while since old contexts won't have platform stored.

briancain:
Also I was thinking the upgrade/uninstall enhancement to use this can come after the initial PRs go through since it's separate.

mitchellh:
> Also I was thinking the upgrade/uninstall enhancement to use this can come after the initial PRs go through since it's separate.

Yes definitely

krantzinator (approved):
🚀

PR #1004, opened 2021-01-21
cli: more specific error msgs on install

krantzinator:
Small PR to match the error messaging we give on `uninstall` if the `-platform` flag is missing.

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -60,8 +60,17 @@ func (c *InstallCommand) Run(args []string) int {
 
 	p, ok := serverinstall.Platforms[strings.ToLower(c.platform)]
 	if !ok {
+		if c.platform == "" {
+			c.ui.Output(
+				"The -platform flag is required.",
+				terminal.WithErrorStyle(),
+			)
+
+			return 1
+		}
+
 		c.ui.Output(
-			"Error installing server into %s: invalid platform",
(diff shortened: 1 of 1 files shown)


mitchellh (approved):

internal/cli/install.go
```diff
@@ -60,8 +60,17 @@ func (c *InstallCommand) Run(args []string) int {
...
+				"The -platform flag is required.",
+				terminal.WithErrorStyle(),
+			)
+
+			return 1
+		}
+
 		c.ui.Output(
-			"Error installing server into %s: invalid platform",
+			"Error installing server into %q: invalid platform",
```
briancain, line 73: This was a bit of feedback on my upgrade PR, so I think it makes sense to do this here too:
```suggestion
			"Error installing server into %q: unsupported platform",
```

briancain (changes requested):

internal/cli/install.go
krantzinator: I _swear_ i put the `%q` in there! Must have dropped it amidst some stashes 😄 Thanks!

krantzinator: Ignore my previous comment; the git suggestion diff confused me 😄

briancain (approved):
woohoo! :+1: thanks for the update :smile:

PR #1005, opened 2021-01-22
feature: Finish exec plugin support by wiring up support in plugins and runners

evanphx:
This depends on https://github.com/hashicorp/waypoint-plugin-sdk/pull/10 as it adds the required plugin support to be able to have plugin components that can exec.

This wires the plugin support end-to-end by adding support for exec plugins as runner jobs and for the exec cli command to spawn those jobs and connect back to them.


internal/client/job.go
```diff
@@ -50,9 +50,18 @@ func (c *Project) job() *pb.Job {
 
 // doJob will queue and execute the job. If the client is configured for
 // local mode, this will start and target the proper runner.
-func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.Job_Result, error) {
+// the mon channel can be used to monitor the job status as it changes.
+// The receiver must be careful to not block sending to mon as it will block
+// the job state processing loop.
+func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI, mon chan pb.Job_State) (*pb.Job_Result, error) {
```
mitchellh, line 56: Given how many things call `doJob` and don't care about `mon`, I feel like it'd be neater to do a second `doJobMonitor` function that has this signature and have `doJob` call into that with the preserved previous signature. This is a common pattern in Go libs.

internal/client/job.go
```diff
@@ -50,9 +50,18 @@ func (c *Project) job() *pb.Job {
...
 	log := c.logger
 
+	// Be sure that the monitor is closed so the reciever knows for sure the job isn't going
+	// anymore.
+	if mon != nil {
```
mitchellh, line 61: Also, nitpick but I like it when channels end in `Ch` so I know they're channels. Its a LITTLE weird but seems like fairly idiomatic too?

internal/client/job.go
```diff
@@ -114,6 +123,7 @@ func (c *Project) queueAndStreamJob(
 	job *pb.Job,
 	ui terminal.UI,
 	jobIdCallback func(string),
+	mon chan pb.Job_State,
```
mitchellh, line 126: This is fine, cause nothing calls `queueAndStreamJob` directly except `doJob` if I recall.

internal/core/app_exec.go
```diff
@@ -0,0 +1,257 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/hashicorp/waypoint-plugin-sdk/component"
+	"github.com/hashicorp/waypoint/internal/ceb"
+	"github.com/hashicorp/waypoint/internal/config"
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+// Exec launches an exec plugin
```
mitchellh, line 21: I'd try to add a lot more detail here as it relates to CEB stuff, just so its clear this is only for _plugin based exec_ and not general CEB-enabled exec.

internal/core/app_exec.go
```diff
@@ -0,0 +1,257 @@
...
+func (a *App) Exec(ctx context.Context, id string, d *pb.Deployment) error {
+	// We need to get the pushed artifact if it isn't loaded.
+	// TODO(evanphx) I don't understand why we do this. We're mimicing what the
+	// deploy destroy code does and they load this, so we do to?
```
mitchellh, line 25: To address this comment: this is for [this feature](https://www.waypointproject.io/docs/waypoint-hcl/variables/artifact) (and similar).

internal/core/app_exec.go
```diff
@@ -0,0 +1,257 @@
...
+	_, _, err = a.doOperation(ctx, a.logger.Named("exec"), &execOperation{
+		InstanceId: id,
+		Component:  c,
+		Deployment: d,
+		Client:     a.client,
+	})
+	return err
+}
+
+// execOperation is a simple barebones operation value to conform to the
```
mitchellh, line 80: Do you need an operation? Or can you just call it directly?

internal/core/app_exec.go
```diff
@@ -0,0 +1,257 @@
...
+
+	p.log.Debug("calling plugin with session-info", "arguments", esi.Arguments)
+
+	val, err := p.app.callDynamicFunc(ctx,
+		p.log,
+		nil,
+		p.op.Component,
+		p.execer.ExecFunc(),
+		argNamedAny("deployment", p.op.Deployment.Deployment),
+		argmapper.Named("exec_info", esi),
```
mitchellh, line 181: Can also do a `argmapper.Typed` here. It doesn't make a big difference but if you don't plan to ever have multiple `ExecSessionInfo` args then its probably less confusing.

internal/server/execclient/client.go
```diff
@@ -142,7 +142,7 @@ func (c *Client) Run() (int, error) {
 
 	if ptyF != nil {
 		// We need to go into raw mode with stdin
-		if f, ok := c.Stdin.(*os.File); ok {
+		if f, ok := c.Stdout.(*os.File); ok {
```
mitchellh, line 145: This makes sense but also seems suspect. We have been using this code for some time with ptys with no problem.(even executing stuff like `vim` and resizing windows). Why this change?

internal/server/proto/server.proto
```diff
@@ -783,6 +787,17 @@ message Job {
     Deployment deployment = 1;
   }
 
+  message ExecOp {
```
mitchellh, line 790: Would comment again that this is for plugin-powered exec and not CEB

internal/server/proto/server.proto
```diff
@@ -58,6 +58,9 @@ service Waypoint {
   // ListInstances returns the running instances of deployments.
   rpc ListInstances(ListInstancesRequest) returns (ListInstancesResponse);
 
+  // FindExecInstance returns an instance that can be used for an exec session.
```
mitchellh, line 61: Actually not sure what the use case is for this, can you address that in the comment?

internal/server/singleprocess/state/instance_exec.go
```diff
@@ -97,6 +98,71 @@ func (s *State) InstanceExecCreateByTargetedInstance(id string, exec *InstanceEx
 	return nil
 }
 
+// CalculateInstanceExecByDeployment considers all the instances registered
+// to the given deployment and finds the one that is least loaded. If there
+// are no instances, returns a ResourceExhausted error. Calls to this
+// function in quick succession will return could return the same instance,
+// which is why a simple random sampling is done on all prospective instances.
+func (s *State) CalculateInstanceExecByDeployment(did string) (*Instance, error) {
+	txn := s.inmem.Txn(true)
```
mitchellh, line 107: I think this can be `false` cause its a read-only transaction, yeah?

internal/server/singleprocess/state/instance_exec.go
```diff
@@ -97,6 +98,71 @@ func (s *State) InstanceExecCreateByTargetedInstance(id string, exec *InstanceEx
 	return nil
 }
 
+// CalculateInstanceExecByDeployment considers all the instances registered
+// to the given deployment and finds the one that is least loaded. If there
+// are no instances, returns a ResourceExhausted error. Calls to this
+// function in quick succession will return could return the same instance,
+// which is why a simple random sampling is done on all prospective instances.
+func (s *State) CalculateInstanceExecByDeployment(did string) (*Instance, error) {
```
mitchellh, line 106: This should have unit tests and can surely be well unit tested.

mitchellh (changes requested):

internal/server/execclient/client.go
evanphx: The commit message on that line explains the bug. Basically it has to do with someone doing `date | waypoint exec`.

internal/core/app_exec.go
evanphx: Gotcha, thanks!

internal/core/app_exec.go
evanphx: So I was wondering that and I'd love to chat about that with you. I was mimic'ing the other functions and they all seemed to use an operation.

internal/core/app_exec.go
evanphx: Ah, gotcha.  👍🏻

internal/server/execclient/client.go
mitchellh: Hm hm yes. Okay, just verify `vim` and friends still work.

internal/client/noop.go
```diff
@@ -21,6 +21,6 @@ func (c *App) Noop(ctx context.Context) error {
 	}
 
 	// Execute it
-	_, err := c.doJob(ctx, job)
+	_, err := c.doJob(ctx, job, nil)
```
mitchellh, line 24: This is probably wrong, not compiling?

mitchellh (changes requested):
Just gotta fix the failing tests but this looks good.

internal/client/noop.go
evanphx: Oh strange! I fixed those but forgot to push the files. Should be in now.

mitchellh (approved):

PR #1007, opened 2021-01-22
Track exact reference to data downloaded for runners

mitchellh:
[WP-046](https://docs.google.com/document/d/1NdYtHsvzw4bI9PIFKWzTFDaBtJ3S3rIjNpOjlmx0lM0/edit)

This adds `DataSource.Ref` proto message type that is used to track the exact reference to the data that was used for a remote run with downloaded data. For example, this stores the commit hash for Git.

This modifies the `GetJobStream` API (user-facing API) as well to add a new `Download` event that notes that data download is complete and provides the ref information. See the proto comments for more info there. 

As usual with proto change PRs, the line diff is deceiving: it is mostly proto changes.


krantzinator (approved):

internal/datasource/datasource.go
```diff
@@ -34,7 +34,7 @@ type Sourcer interface {
 		ui terminal.UI,
 		source *pb.Job_DataSource,
 		baseDir string,
-	) (string, func() error, error)
+	) (string, *pb.Job_DataSource_Ref, func() error, error)
```
evanphx, line 37: Probably don't need to do this here, but 4 return values is right on the verge of "this should be a data type" for me.

evanphx (approved):
Just one minor note, mostly as a place for further convo about style.

internal/datasource/datasource.go
mitchellh: Big agree. Its on my backburner to refactor the `internal/datasource` package since its scope is changing a bit anyways. Lets do that in another PR, after we get all this polling and stuff in.

Issue #1013, opened 2021-01-22
cli: Provide server auth information when restarting an existing container


krantzinator:
**Is your feature request related to a problem? Please describe.**
We could do more to authenticate an existing server than we currently do. In the case of an existing server image being restarted, we should verify the token if the server is already bootstrapped. (We auto-restart an existing server container for Docker with the merging of #1009.)
The current messaging is:
```
! Error getting the initial token: server is already bootstrapped
  
  The Waypoint server has been deployed, but due to this error we were
  unable to automatically configure the local CLI or the Waypoint server
  advertise address. You must do this manually using "waypoint context"
  and "waypoint server config-set".
```

This can cause confusion for users who now thing their server may have an error, when in actuality it may be perfectly fine and ready to go.

**Describe the solution you'd like**
A message that is a warning rather than an error if we can verify that an already-bootstrapped server is authenticated with an existing context.

**Describe alternatives you've considered**
Leaving this as-is. It's _fine_ but not ideal, and not a concise message.

**Explain any additional use-cases**
If someone has a list of multiple contexts, and restarts a server image that is bootstrapped by one of the non-active contexts, providing this information could also lessen user frustration.

PR #1017, opened 2021-01-25
internal/install: Attempt to use default context for auth token 

briancain:
If a server has been restarted with the install step, the bootstrap
token step will certainly fail due to the server already being
configured. This commit fixes that by attempting to use the default
context to see if its auth token is valid for the server. If not, fail
like it was going to fail. If it's valid, go ahead and use the default
context since its the one we're looking for for the restarted server
container.

Fixes https://github.com/hashicorp/waypoint/issues/1013

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -128,8 +128,9 @@ func (c *InstallCommand) Run(args []string) int {
 		defaultCtx, err := c.contextStorage.Default()
 		if err != nil {
 			c.ui.Output(
-				"Error getting default context to use existing auth token: %s\n%s",
+				"Error getting default context to use existing auth token: %s\n\n%s\n\n%s",
 				clierrors.Humanize(err),
+				errInstallToken,
 				errInstallRunning,
 				terminal.WithErrorStyle(),
 			)
@@ -139,9 +140,10 @@ func (c *InstallCommand) Run(args []string) int {
 			defaultCtxConfig, err := c.contextStorage.Load(defaultCtx)
 			if err != nil {
 				c.ui.Output(
-					"Error loading the context %q to use existing auth token: %s\n%s",
+					"Error loading the context %q to use existing auth token: %s\n\n%s\n\n%s",
 					defaultCtx,
 					clierrors.Humanize(err),
+					errInstallToken,
 					errInstallRunning,
 					terminal.WithErrorStyle(),
 				)
@@ -154,8 +156,9 @@ func (c *InstallCommand) Run(args []string) int {
 			)
 			if err != nil {
 				c.ui.Output(
-					"Error connecting to server using existing auth token: %s\n\n%s",
+					"Error connecting to server using existing auth token: %s\n\n%s\n\n%s",
 					clierrors.Humanize(err),
+					errInstallToken,
 					errInstallRunning,
 					terminal.WithErrorStyle(),
 				)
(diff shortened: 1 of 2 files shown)


internal/cli/install.go
```diff
@@ -124,14 +124,76 @@ func (c *InstallCommand) Run(args []string) int {
...
 	tokenResp, err := client.BootstrapToken(ctx, &empty.Empty{})
 	if err != nil {
-		c.ui.Output(
-			"Error getting the initial token: %s\n\n%s",
-			clierrors.Humanize(err),
-			errInstallRunning,
-			terminal.WithErrorStyle(),
-		)
-		return 1
+		// try default context in case server was started again from install
```
mitchellh, line 127: We should probably only try this logic if `status.Code(err) == codes.PermissionDenied` yes?

My commendation would be something like:

```go
if status.Code(err) != codes.PermissionDenied {
  // The old error message and return 1
}
```

internal/cli/install.go
```diff
@@ -124,14 +124,76 @@ func (c *InstallCommand) Run(args []string) int {
...
+					errInstallRunning,
+					terminal.WithErrorStyle(),
+				)
+				return 1
+			} else {
+				// token is valid
+				log.Info("Updating context to use default context, token is valid")
+				contextConfig = defaultCtxConfig
+			}
+		} else {
```
mitchellh, line 185: Instead of this else, I'd swap the if statement and just lower the level of nesting if that makes sense.

mitchellh (changes requested):

internal/cli/install.go
briancain: Ahh yes, I was curious if we could try this specifically on that error, thanks!

internal/cli/install.go
```diff
@@ -132,15 +134,77 @@ func (c *InstallCommand) Run(args []string) int {
...
 	}
+
 	if tokenResp != nil {
 		log.Debug("token received, setting on context")
 		contextConfig.Server.RequireAuth = true
 		contextConfig.Server.AuthToken = tokenResp.Token
+	} else {
+		// try default context in case server was started again from install
+		defaultCtx, err := c.contextStorage.Default()
+		if err != nil {
```
krantzinator, line 145: We should never hit this. We overwrite the existing Default context -- or create a new one if it's been deleted -- here https://github.com/hashicorp/waypoint/blob/cfba8f4e229c46d57ddc8f9760f0567b71878f25/internal/serverinstall/docker.go#L81.
Should we move the Docker context creation down, after the `if len(containers) > 0 {` block, and just return a `nil` for an existing install, and then move the context check to line 96 here? Like:
```
contextConfig = result.Context
if contextConfig == nil {
  contextConfig, err = c.contextStorage.Default()
  if err != nil {
    ...
  }
}
```

briancain: Line 81 in `serverinstall/docker` will create it in a struct, but it doesn't save it through `contextStorage` on disk, so no default has been created or stored yet. The storage part and setting the default ends up happening here in install after the bootstrap token step right after this on 208.

We'll want this here after the platform install step, since the installer attempts to bootstrap the server after the platform install. The bootstrap step will tell us whether or not the server has been set up already, so here is where we can check to see if the context used to invoke the install command to restart the server has the right auth token. Otherwise we'd be pulling in all of that bootstrap stuff from here into the docker serverinstall platform which it shouldn't be handling.

I think we should try to keep the contextStorage operations and bootstrapping platform agonstic inside this installation function here and just let the serverinstall return the `InstallResults` struct for us to handle.

krantzinator: If I `waypoint context delete` and `docker stop waypoint-server` I make it past this error catch and end up erroring out with
```
❌ Configuring server...
! Error setting the advertise address: Authorization token is not supplied
```
Which is technically correct; the context that was created for me during the re-install of the old server does not have an authorization token in it, since I deleted the old context + auth token. So this current behavior is fine, I just wanted to make sure the error catch here will catch what we expect it to.

krantzinator (approved):

internal/cli/install.go
briancain: @krantzinator - Ah, I see what you're saying. Yeah that's a weird case :thinking: Agreed, I think we can provide a better error message in this case.

PR #1020, opened 2021-01-26
internal/{ceb,core,server/singleprocess}: Fix exec stream closure

evanphx:
This fixes a very specific edgecase where if the virtual side of the
exec handler sends an Exit message and then cancels the context in quick
succession, the server side may not see the message, only the context
cancelation. gRPC seems to prioritize the cancelation above pending
events, possibly because message sends are not guaranteed (the docs
state this).

The change then is to wait for the server side to close the stream after
the virtual side sends it's exit code. This means that the server side
is in control of seeing the messages and shutting down the stream.

We also can now simplify the handling of the ExecStream messages by
allowing server.Recv() to return any buffered messages even if the
context has already been canceled. Effectively, if we check that the
context the server stream is using has been canceled on our own, we
don't allow the server stream to return any buffered messages.

The entrypoint exec code is also simplified by having it depend only on
the execclient/exec side by checking exec.Context for readiness when
trying to relay messages back to the execclient.

### DIFF
--- a/internal/core/app_exec_test.go
+++ b/internal/core/app_exec_test.go
@@ -17,6 +17,7 @@ import (
 	serverptypes "github.com/hashicorp/waypoint/internal/server/ptypes"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
 
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	componentmocks "github.com/hashicorp/waypoint-plugin-sdk/component/mocks"
@@ -96,6 +97,10 @@ func TestAppExec_happy(t *testing.T) {
 		require.NoError(err)
 		return len(resp.Instances) == 1
 	}, 2*time.Second, 25*time.Millisecond)
+
+	// Make sure that with all the exec stream tracking we don't leak
+	// goroutines
+	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
 
 	var stderr, stdout bytes.Buffer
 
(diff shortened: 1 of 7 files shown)


internal/server/singleprocess/service_entrypoint.go
```diff
@@ -331,7 +327,6 @@ func (s *service) EntrypointExecStream(
 				errCh <- err
 				return
 			}
-			log.Trace("entrypoint event received", "event", hclog.Fmt("%#v", req.Event))
```
mitchellh, line 334: Why'd you remove this trace?

internal/server/singleprocess/service_exec.go
```diff
@@ -155,6 +155,18 @@ func (s *service) handleEntrypointExecRequest(
 				},
 			},
 		}
+	case *pb.EntrypointExecRequest_Error_:
```
mitchellh, line 158: Is this new case test-able somehow?

internal/server/singleprocess/service_entrypoint.go
evanphx: I'll add it back in.

internal/server/singleprocess/service_exec.go
evanphx: Should be able to, good call.

internal/core/app_exec_test.go
```diff
@@ -97,6 +98,10 @@ func TestAppExec_happy(t *testing.T) {
 		return len(resp.Instances) == 1
 	}, 2*time.Second, 25*time.Millisecond)
 
+	// Make sure that with all the exec stream tracking we don't leak
+	// goroutines
+	defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
```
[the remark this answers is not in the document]
evanphx: 🙌

mitchellh (approved):

briancain (approved):

PR #1021, opened 2021-01-26
UI: show destroyed resources

gregone:
This PR addresses [issue 958](https://github.com/hashicorp/waypoint/issues/958).

destroyed-deployments

### DIFF
--- a/ui/app/components/app-item/deployment.hbs
+++ b/ui/app/components/app-item/deployment.hbs
@@ -1,4 +1,4 @@
-<li class="app-item">
+<li class={{if (eq @deployment.state 4) "app-item app-item--destroyed" "app-item"}}>
   <LinkTo @route="workspace.projects.project.app.deployment" @models={{array @deployment.id}}>
     <p>
       <b class="badge badge--version">v{{@deployment.sequence}}</b>
--- a/ui/app/components/app-item/deployment.hbs
+++ b/ui/app/components/app-item/deployment.hbs
@@ -3,6 +3,11 @@
     <p>
       <b class="badge badge--version">v{{@deployment.sequence}}</b>
       <code>{{@deployment.id}}</code>
+      {{#if (eq @deployment.state 4)}}
+        <b class="badge badge--destroyed">
+          {{inline-svg "trash" class="icon"}}&nbsp;Destroyed
+        </b>
+      {{/if}}
     </p>
     <small>
       {{inline-svg (icon-for-component @deployment.component.name) class="icon"}}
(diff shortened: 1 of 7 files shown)


mitchellh (commented):
Can't comment too much on the JS side but I like this!

ui/app/components/app-item/deployment.hbs
```diff
@@ -3,6 +3,11 @@
     <p>
       <b class="badge badge--version">v{{@deployment.sequence}}</b>
       <code>{{@deployment.id}}</code>
+      {{#if (eq @deployment.state 4)}}
+        <b class="badge badge--destroyed">
+          {{inline-svg "trash" class="icon"}}&nbsp;destroyed
```
pearkes, line 8: Just a question here but something about the lowercase surprised, did you look at it with `Destroyed`?

ui/app/controllers/workspace/projects/project/app/deployments.ts
```diff
@@ -0,0 +1,27 @@
+import Controller from '@ember/controller';
+import { tracked } from '@glimmer/tracking';
+import { Deployment } from 'waypoint-pb';
+import { action } from '@ember/object';
+export default class WorkspaceProjectsProjectAppDeployments extends Controller {
+  queryParams = ['destroyed']
+
+  @tracked destroyed = false;
+
+  get hasMoreDeployments() {
```
pearkes, line 10: In general we are following a pattern of really only using controllers for query params, see here: https://github.com/hashicorp/waypoint/tree/main/ui/app/controllers

You could potentially pull this out to the app-item/deployment component?

pearkes: (so the controller is still necessary but only tracks the query param)

ui/app/templates/workspace/projects/project/app/deployments.hbs
```diff
@@ -13,6 +23,20 @@
         from the CLI</p>
     </EmptyState>
   {{/each}}
+  {{#if (and this.hasMoreDeployments (not this.destroyed))}}
```
pearkes, line 26: Ah I see why this is at the controller. Maybe this could be a new component then. Not super tied to it just trying to keep some of this stuff out of controllers given the trend there.

pearkes (commented):
A few questions, I wonder if we should limit the usage of the controller here to just the query param, but let me know what you think.

dizzyup (approved):
Looks good! I think I'd also try making the deployment ID string the same gray colour value as the "Deployed by Docker about 4 hours ago" text for those deleted rows, so they're not as prominent as the rows above/below them.

gregone:
@pearkes merging as is, I'd rather move the list views into components instead of controllers separately.

PR #1025, opened 2021-01-27
Server-side default waypoint.hcl

mitchellh:
[WP-049](https://docs.google.com/document/d/1xYIZhDczjekTk0gMMle9XuAkeqYt3ow7BLiF3r31N7o/edit#)

This adds the ability to specify the contents of a default Waypoint file with a project and store this on the Waypoint server. This waypoint.hcl contents/file will only be used if the project source itself does not have a waypoint.hcl. The user can use the `waypoint project apply` CLI to also set this.

There is one major limitation of this approach (outlined in the RFC): projects made with a server-side waypoint HCL can't be run with local operations. You can use the CLI but must use remote operations by specifying the full project/app: `waypoint up myproject/myapp`.


internal/cli/project_apply.go
```diff
@@ -254,6 +255,22 @@ func (c *ProjectApplyCommand) Run(args []string) int {
...
+			c.ui.Output(
+				"Error reading HCL file specified with the -waypoint-hcl flag: %s",
+				clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+
+			return 1
+		}
+
+		proj.WaypointHcl = bs
```
briancain, line 271: Do we do any simple HCL syntax check at all before sending it off to the server, or does the CLI just assume it's ok?

briancain (approved):

internal/cli/project_apply.go
mitchellh: A syntax check is a good idea. 

I didn't want to do a full parse cause if a user has like a `file()` call that depends on the source, it will fail. But checking actually syntax only (basic parse) is a good idea.

mitchellh: Done!

briancain (approved):

krantzinator (approved):

PR #1026, opened 2021-01-27
feature: End to end integration tests for Waypoint

briancain:
This pull request introduces a basic shell runner to execute end to end tests with Waypoint. For now, it only includes the Docker server platform, but in the future this will include Nomad and Kubernetes for end to end testing.


test-e2e/run-test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# only spin for local devs running on machine to show tests aren't frozen
+if [[ ! -v CI_ENV ]]; then
+  # shell spinner: https://www.shellscript.sh/tips/spinner/
+  spin &
+  SPIN_PID=$!
+  trap "kill -9 $SPIN_PID" `seq 0 15`
+fi
+
+go test .
```
krantzinator, line 84: can the path be dynamic so i can run `./test-e2d/run-test.sh` from root?

test-e2e/run-test.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# make tools
+# git submodule update for grpc status from api common
+# make
+
+echo "==> Pulling in waypoint-examples for test..."
+echo
+
+echo "Skipping for now"
+echo "Assuming 'waypoint-examples' repo has already been cloned into this folder"
```
krantzinator, line 60: we should do a shallow clone and delete when this part is automated; i don't personally plan to keep `waypoint-examples` nested within my `waypoint` repo.

test-e2e/docker_smoke_test.go
```diff
@@ -0,0 +1,144 @@
+package test
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+const (
+	testDir = "waypoint-examples/docker/go"
```
krantzinator, line 10: should we make the path to the waypoint-examples repo a test env var as well? that way i can smoke test against a specific branch or somesuch that i already have cloned/checked out.

briancain: Ooh good idea! 👍

test-e2e/docker_smoke_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	if !strings.Contains(stdout, "Waypoint v") {
+		t.Errorf("No version output detected:\n%s", stdout)
+	}
+}
+
+func TestWaypointDockerInstall(t *testing.T) {
+	wp := NewBinary(wpBinary, testDir)
+	stdout, stderr, err := wp.Run("install", "-platform=docker", "-accept-tos", fmt.Sprintf("-docker-server-image=%s", wpServerImage))
```
krantzinator, line 32: this collided with my current install 😅 obviously won't be a problem in CI but for running locally, does go have a "Before"-esque hook for verifying a workspace is clean?

krantzinator (approved):
This looks great, all my comments are future considerations or questions, not blockers. 🚀 🎉

test-e2e/docker_smoke_test.go
briancain: Yeah! Good idea...I was thinking we might want a pre/post tests hook for these for this sort of thing. Maybe something easy is having the first Go test just be a verification that there isn't a previously running context running or something so the tests don't fail from a collision?

krantzinator: Yeah that sounds like a good first step.

krantzinator (approved):

PR #1027, opened 2021-01-27
internal/server/state: project put should not modify applications

mitchellh:
We document that project-related APIs cannot be used for applications,
but previously we didn't preserve existing applications. If a user
called UpsertProject to modify an existing project it'd blow away all
their applications.

This couldn't be hit in the real world unless the user was using the API
directly, since `waypoint init -update` is the only thing that calls
this API and it always registers applications again.

But with future changes such as introducing server side waypoint.hcl and
future UI changes to modify projects, this would've become a big issue.

### DIFF
--- a/internal/server/singleprocess/state/project_test.go
+++ b/internal/server/singleprocess/state/project_test.go
@@ -64,6 +64,51 @@ func TestProject(t *testing.T) {
 		}
 	})
 
+	t.Run("Put does not modify applications", func(t *testing.T) {
+		require := require.New(t)
+
+		const name = "AbCdE"
+		ref := &pb.Ref_Project{Project: name}
+
+		s := TestState(t)
+		defer s.Close()
+
+		// Set
+		proj := serverptypes.TestProject(t, &pb.Project{Name: name})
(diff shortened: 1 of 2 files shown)


internal/server/singleprocess/state/project_test.go
```diff
@@ -64,6 +64,51 @@ func TestProject(t *testing.T) {
...
+
+		// Get exact
+		{
+			resp, err := s.ProjectGet(&pb.Ref_Project{
+				Project: "AbCdE",
+			})
+			require.NoError(err)
+			require.NotNil(resp)
+			require.False(resp.RemoteEnabled)
+			require.Len(resp.Applications, 1)
```
krantzinator, line 93: Given this is fixing a bug that would blow away all applications, should the test have more than 1 Applications that we ensure are still around after the Put?

krantzinator (approved):
good catch

internal/server/singleprocess/state/project_test.go
mitchellh: Sure, doesn't hurt!

PR #1035, opened 2021-01-28
Server-side waypoint.hcl needs to store the format

mitchellh:
This adds support for specifying the format (JSON vs HCL) for the server-side Waypoint.hcl.

I'm just going to merge this once tests pass since this is a minor change that is tested.


mitchellh:
The one racy test failed.

PR #1036, opened 2021-01-28
internal/server/state: CalculateInstanceExec picks least loaded

mitchellh:
This fixes an issue where if we looped over instances in the order of
most to least loaded, they would ALL be candidates. We need to reset the
candidate list when we find a lower loaded value.


evanphx (approved):
AH! Good catch.

PR #1037, opened 2021-01-28
Add a `waypoint fmt` command

### LINKED ISSUE #705: Have a waypoint fmt to format the waypoint.hcl

With a waypoint init, I get an extra set of whitespace.

I would like to have `waypoint fmt` just like a `terraform fmt` command that format the HCL content

Using `terraform fmt`

mitchellh:
Fixes #705 

This adds a `waypoint fmt` command to format `waypoint.hcl` files. This is a simplified version of `terraform fmt`: https://www.terraform.io/docs/cli/commands/fmt.html and the underlying logic is identical.


internal/cli/fmt.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (c *FmtCommand) Flags() *flag.Sets {
+	return c.flagSet(0, func(sets *flag.Sets) {
+		f := sets.NewSet("Command Options")
+
+		f.BoolVar(&flag.BoolVar{
+			Name:    "write",
+			Target:  &c.flagWrite,
+			Default: false,
+			Usage: "Overwrite the input file. If this is false, the formatted " +
```
briancain, line 106: Quick note, this should be reversed to have the opposite behavior. Always write in place, and can be disabled and write to stdout with `-write=false`

briancain (changes requested):
Works well for me! Just a quick note about the `-write` behavior to match `terraform fmt`.

briancain (approved):

internal/cli/fmt.go
```diff
@@ -0,0 +1,142 @@
...
+		WithFlags(c.Flags()),
+		WithNoConfig(),
+		WithClient(false),
+	); err != nil {
+		return 1
+	}
+
+	// If we have too many args, error immediately.
+	if len(c.args) > 1 {
+		c.ui.Output("At most one argument is expected.\n\n"+c.Help(), terminal.WithErrorStyle())
```
briancain, line 36: I noticed this should maaybe be increased if FILE is provided? So if you want to format a waypoint.hcl in a different  folder but to STDOUT with `-write=false`:

```
brian@localghost:docker % waypoint fmt angular/waypoint.hcl -write=false                                                                                                                                                                ±[●][main]
! At most one argument is expected.
```

mitchellh: Don't think this is a bug, flags in Go have to be before args, so `-write=false angular/waypoint.hcl` should work.

PR #1038, opened 2021-01-28
test-e2e/{nomad,k8s}: smoke tests for remaining server platforms

briancain:
This pull request adds the remaining supported server platforms for smoke tests. It also moves out the basic CLI test into its own file instead of living in just the docker tests.


mitchellh (approved):
We gotta get these running in CI ASAP otherwise I think these strings will get stale. :)

PR #1040, opened 2021-01-29
internal/server: send entrypoint URL config only when we get a token

mitchellh:
At the server level, this sets us up completely for any scenario where
the URL service configuration changes. However, the CEB doesn't yet
support changing URL configurations (it ignores them once it has
initialized) so we only send down the URL config once we are ready.
Future changes will just be ignored.

### DIFF
--- a/internal/server/singleprocess/service_deploy.go
+++ b/internal/server/singleprocess/service_deploy.go
@@ -55,7 +55,7 @@ func (s *service) UpsertDeployment(
 		}
 
 		// Create a context that will timeout relatively quickly
-		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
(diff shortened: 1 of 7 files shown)


internal/server/singleprocess/service_deploy.go
```diff
@@ -55,7 +55,7 @@ func (s *service) UpsertDeployment(
 		}
 
 		// Create a context that will timeout relatively quickly
-		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
mitchellh, line 58: This is an unrelated change that we want to get in as well to increase the buffer horizon has to setup the hostname.

krantzinator (approved):
works great with the new horizon changes

PR #1044, opened 2021-02-01
Fix panic from `waypoint hostname delete` with no hostname specified

gulducat:
A lil order-of-operations issue resulted in an unsightly panic.

Before:

```
❯ waypoint hostname delete

panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
... trace ...
```

After:

```
❯ ./waypoint hostname delete
! hostname required for deletion
```

### DIFF
--- a/internal/cli/hostname_delete.go
+++ b/internal/cli/hostname_delete.go
@@ -22,11 +22,11 @@ func (c *HostnameDeleteCommand) Run(args []string) int {
 		return 1
 	}
 
-	hostname := c.args[0]
 	if len(c.args) == 0 {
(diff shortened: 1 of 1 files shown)


briancain (approved):
thanks for the fix! It seems like heroku build packs are a bit broken at the moment, so that's why the integration tests are failing. Otherwise if you sign the CLA we'll be good to merge! :smile:

gulducat:
sure thing!  I did a force-push to update my author email for cla-assistant (which I'm partially responsible for operating, so really should have known to do that before making a PR 😋)

PR #1046, opened 2021-02-02
cli/exec: Fail back to old search if FindExecInstance unimplemented

briancain:
Prior to this commit, the CLI would attempt to use `FindExecInstance` on
the server when searching for deployments. If the cli made a request
against an old server version, it would fail with Unimplemented. It
would also not surface this error to the user and instead just exit.
This commit fixes that by setting the deployment ID and Sequence first,
and if the server responds Unimplmenented, it simply returns early
instead of setting the Instance ID. It also displays any errors from
searching deployments in the case that we get a different error other
than unimplemeneted or ResourceExhausted.

### DIFF
--- a/internal/cli/exec.go
+++ b/internal/cli/exec.go
@@ -102,27 +102,32 @@ func (c *ExecCommand) searchDeployments(
 		return nil, err
 	}
 
+	ec.DeploymentId = deployment.Id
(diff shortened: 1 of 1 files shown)


evanphx (approved):
Wonderful!

krantzinator (approved):
🎉

Issue #978, opened 2021-01-14
gitreftag returns an empty string 


miran248:
**Describe the bug**
`gitreftag` returns an empty string in both build and release steps.

**Steps to Reproduce**

```hcl
app "app-test" {
  build {
    use "pack" {
      builder            = "heroku/buildpacks:18"
      disable_entrypoint = true
    }

    registry {
      use "aws-ecr" {
        region     = "eu-west-1"
        repository = "app-test"
        tag        = substr(gitreftag(), 1, -1)
      }
    }
  }

  deploy {
    use "exec" {
      command = [ "echo", gitreftag() ]
    }
  }
}
```

**Expected behavior**
It should return the actual tag.

**Additional context**
`git log --pretty=oneline` returns `hash (HEAD -> branch, tag: v0.1.5, origin/branch) chore(release): 0.1.5` - `gitreftag()` should thus return `"v0.1.5"` instead of `""`. `substr` would then remove the `v` from the tag.

Both `gitrefhash` and `gitrefpretty` are working!

`macos 10.15.7`
`Waypoint v0.2.0 (360b2b3a)`

catsby:
I attempted to fix this in #1047, and while it works for me I wasn't able to figure out why the tests were passing to truly understand what the source of the bug is 🤔

PR #1047, opened 2021-02-02
Fix empty gitreftag response (#978)

catsby:
Fixes https://github.com/hashicorp/waypoint/issues/978

I was able to reproduce #978 with this config and a local git repo with 2 tags, one in the past and one on the current `HEAD` :

<details>

```hcl
project = "example-nodejs"

app "example-nodejs" {
  labels = {
    "service" = "example-nodejs"
    "env"     = "dev"
  }

  config {
    env = {
      TEST = "hiya"

      # REFTAG    = gitreftag()
      REFHASH   = gitrefhash()
      REFPRETTY = gitrefpretty()
    }
  }

  build {
    use "pack" {}
  }

  deploy {
    # use "docker" {}
    use "exec" {
      command = [ "echo", gitreftag() ]
      # command = [ "echo", gitrefhash() ]
    }
  }
  
}

```

</details>

The deploy output was always `echo ` (empty)

I dug in an found that `refTagFunc` was always hitting an `EOF` here (L157):

- https://github.com/hashicorp/waypoint/blob/298ec9289ea651bc80ca2c757bde7d325e1e70c7/internal/config/funcs/vcs_git.go#L148-L168

I dug in more and found a built-in iterator for tag references (https://github.com/go-git/go-git/blob/024d62b96ac19b40473f04931f9c12c8e24e0f7a/plumbing/object/tag.go#L339). Switching to that in this PR resolved the issue and started showing me the tag I was expecting. 

Unfortunately, I couldn't figure out why the tests pass on `main` with the time I had 🤔 

- https://github.com/hashicorp/waypoint/blob/c4d2a01aa5c13a2a0f8ab53e6420c6e50561df14/internal/config/funcs/vcs_git_test.go#L21

### DIFF
--- a/internal/config/funcs/vcs_git.go
+++ b/internal/config/funcs/vcs_git.go
@@ -153,12 +153,15 @@ func (s *VCSGit) refTagFunc(args []cty.Value, retType cty.Type) (cty.Value, erro
 	}
 
 	var tagRefStr string
-	_ = iter.ForEach(func(t *plumbing.Reference) error {
+	err = iter.ForEach(func(t *plumbing.Reference) error {
 		if t.Hash() == ref.Hash() {
 			tagRefStr = t.Name().Short()
 		}
 		return nil
 	})
+	if err != nil {
+		return cty.UnknownVal(cty.String), err
+	}
 
 	if tagRefStr != "" {
 		return cty.StringVal(tagRefStr), nil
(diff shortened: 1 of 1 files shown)


mitchellh (approved):
I'm pretty okay just merging this as-is. We have tests around this as you found that test the main happy cases so we at least know this didn't break anything obvious. I'd love to find a unit test that demonstrates your issue you had of course but I'd be fine just merging.

krantzinator (approved):
I added the label to put this on our current release branch as well 🙌🏻

briancain (approved):

PR #1049, opened 2021-02-02
cli/main: Add -version flag for CLI version

briancain:
This commit introduces a `-version` flag to `waypoint -version`. It also
overrides `-v` to print version instead of being verbose only in the
case where there is 1 arg.


krantzinator (approved):
🎸

mitchellh (approved):

briancain:
Force pushed just to fix my typo in the commit message :smile:

PR #1050, opened 2021-02-02
internal/{ceb,cli,core,runner}: Plumb (optionally dynamic) config via env to the exec plugin

evanphx:
This PR feeds configuration from the server to the exec plugin as a set of environment variables.

Unless configured as a flag in the runner, it does not provide any dynamic config variables because those require credentials to access, and it's not presumed that the runner is going to have the correct credentials.

Users that are using standalone runners and verify that their runner environment has the proper credentials can enable reading the dynamic config via the `-enable-dynamic-config` flag.

### DIFF
--- a/internal/cli/runner_agent.go
+++ b/internal/cli/runner_agent.go
@@ -15,6 +15,11 @@ import (
 
 type RunnerAgentCommand struct {
 	*baseCommand
+
+	// Indicates if exec plugins run by this runner should read dynamic
+	// config. This requires the runner to have credentials to the dynamic
+	// config sources.
+	flagDynConfig bool
 }
 
 func (c *RunnerAgentCommand) Run(args []string) int {
@@ -86,6 +91,7 @@ func (c *RunnerAgentCommand) Run(args []string) int {
 	runner, err := runnerpkg.New(
 		runnerpkg.WithClient(client),
 		runnerpkg.WithLogger(log.Named("runner")),
+		runnerpkg.WithDynamicConfig(c.flagDynConfig),
 	)
 	if err != nil {
 		c.ui.Output(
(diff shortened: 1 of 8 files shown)


internal/cli/runner_agent.go
```diff
@@ -129,7 +135,14 @@ func (c *RunnerAgentCommand) Run(args []string) int {
 }
 
 func (c *RunnerAgentCommand) Flags() *flag.Sets {
-	return c.flagSet(0, nil)
+	return c.flagSet(flagSetOperation, func(set *flag.Sets) {
+		f := set.NewSet("Command Options")
+		f.BoolVar(&flag.BoolVar{
+			Name:   "enable-dynamic-config",
```
briancain, line 141: Maybe without `enable`?

```suggestion
			Name:   "dynamic-config",
```

briancain (approved):
Looks good! :+1: just some minor nitpicks from me :smile:

internal/cli/runner_agent.go
```diff
@@ -129,7 +135,14 @@ func (c *RunnerAgentCommand) Run(args []string) int {
...
+			Target: &c.flagDynConfig,
```
briancain, line 142: Oops, didn't hit save on this comment when I hit submit on my review....

Should we be explicit and default to false here, or is unset and nil essentially false?

evanphx: Unset is false by default here, yeah.

internal/cli/runner_agent.go
evanphx: So I considered that but I prefer to have verbs in my bool flags because it helps folks understand what the value-less expression (the most common one for a bool flag) is.

briancain: Ah, makes sense! Thanks for explanation :)

krantzinator (approved):

mitchellh (approved):
Looks good to me. I definitely think this is not the right long term solution, it feels pretty weird. I'm honestly not sure if specifying a special flag or just saying "dynamic config is unsupported" is better.

Long term as we discussed, I think we need to get the target environment (i.e. the Lambda plugin spins up an ECS container) running the entrypoint and attempting to load the dynamic config there, rather than in the virtual CEB.

evanphx:
This PR got included into the #1052 mistakenly, so I'm closing it here.

PR #1052, opened 2021-02-03
Add support for AWS Lambda

evanphx:
This rounds out a couple of things core needed finish off the exec plugin work (the rest was in #1050), but mostly adds support for AWS Lambda.

At present, Lambda is setup to be able to deployed easily as an ALB TargetGroup, making it easy to use Lambdas as web targets.

This uses the new Exec Plugin support to spin up an ECS cluster that is used for the execution of the exec session, since there is no way to use Lambda itself for that functionality (at least not safely).


builtin/aws/lambda/platform.go
```diff
@@ -0,0 +1,706 @@
...
+	lamSvc := lambda.New(sess)
+	curFunc, err := lamSvc.GetFunction(&lambda.GetFunctionInput{
+		FunctionName: aws.String(src.App),
+	})
+
+	var funcarn string
+
+	// If the function exists, we update it's code rather than create a new one.
+	// Will then publish this new code to create a stable identifier for it.
+	if err == nil {
```
krantzinator, line 225: Should we specifically check for a type of `ResourceNotFoundException` on this err, and return the `err` if it's of another type?

builtin/aws/lambda/platform.go
```diff
@@ -0,0 +1,706 @@
...
+
+	// Now generate a new TargetGroup so the Lambda can be attached to an ALB easily.
+
+	svc := elbv2.New(sess)
+
+	serviceName := fmt.Sprintf("%s-%s", src.App, id)
+
+	// We have to clamp at a length of 32 because the Name field to CreateTargetGroup
+	// requires that the name is 32 characters or less.
+	if len(serviceName) > 32 {
```
krantzinator, line 354: 🎉 😄

builtin/aws/lambda/platform.go
```diff
@@ -0,0 +1,706 @@
...
+	deployment.FuncArn = funcarn
+	deployment.VerArn = verarn
+	deployment.Version = *ver.Version
+	deployment.TargetGroupArn = *ctg.TargetGroups[0].TargetGroupArn
+
+	return deployment, nil
+}
+
+// This is used by the Exec plugin to provide some helpful output
+// while we proper the exec environment.
```
krantzinator, line 390:
```suggestion
// while we prepare the exec environment.
```

builtin/aws/lambda/ecs.go
```diff
@@ -0,0 +1,381 @@
...
+
+// SetupRole creates an IAM role to use for ECS task execution. Reuses a role if there is already one.
+func (e *ecsLauncher) SetupRole(L hclog.Logger, sess *session.Session, log hclog.Logger, app *component.Source) error {
+	svc := iam.New(sess)
+
+	e.roleName = "ecr-" + app.App
+
+	log.Info("setting up IAM role")
+	L.Debug("attempting to retrieve existing role", "role-name", e.roleName)
+
```
krantzinator, line 98: As stolen from https://github.com/hashicorp/waypoint/blob/68c046a2074f16d63552dfcadcce5745c4a15197/builtin/aws/ecs/platform.go#L406-L410
```suggestion

	// role names have to be 64 characters or less, and the client side doesn't validate this.
	if len(roleName) > 64 {
		roleName = roleName[:64]
	}
	
```

krantzinator (changes requested):
Is the failing test due to the sdk update not being merged in yet?

builtin/aws/lambda/platform.go
krantzinator: I realize this is confusing placement -- this comment should really go on L265, around the `else` clause.

builtin/aws/ecr/mapper.go
```diff
@@ -0,0 +1,13 @@
+package ecr
+
+import (
+	"github.com/hashicorp/waypoint/builtin/docker"
+)
+
+// ECRImageMapper maps a ecr.Image to a docker.Image structure.
+func ECRImageMapper(src *Image) *docker.Image {
+	return &docker.Image{
```
mitchellh, line 9: Does this need the `Location` field set? Or do we not know or assume its in a local registry?

builtin/aws/lambda/platform.go
```diff
@@ -0,0 +1,706 @@
...
+				"Service": "lambda.amazonaws.com"
+			},
+			"Action": "sts:AssumeRole"
+		}
+	]
+}`
+
+// SetupRole creates an IAM role that will be used by the Lambda function. The role
+// has the basic lambda execution rules attached to it.
+func (p *Platform) SetupRole(L hclog.Logger, app *component.Source, sess *session.Session) (string, error) {
```
mitchellh, line 120: feels like this should be unexported to better convey this is an internal helper

builtin/aws/lambda/platform.go
```diff
@@ -0,0 +1,706 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		funcarn = *funcOut.FunctionArn
+
+		ui.Output("Created Lambda Function: %s", funcarn)
+	}
+
+	st := ui.Status()
```
mitchellh, line 291: If possible, I'd avoid Status and use `StepGroup` throughout this thing instead.

builtin/aws/lambda/platform.go
```diff
@@ -0,0 +1,706 @@
...
+}
+
+// DeployFunc implements component.Platform
+func (p *Platform) DeployFunc() interface{} {
+	return p.Deploy
+}
+
+// DestroyFunc implements component.Destroyer
+func (p *Platform) DestroyFunc() interface{} {
+	return p.Destroy
```
mitchellh, line 65: Do we need a `DestroyWorkspace` implementation too to clean up the other stuff you setup in here? or is it all container per-deployment?

builtin/aws/lambda/plugin.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+package aws.lambda;
+
+option go_package = "waypoint/builtin/aws/lambda";
+
+message Deployment {
```
mitchellh, line 7: Docs on these fields please (even one liners, i.e. `func_arn` is kind of obvious but `version` is not to me)

cmd/waypoint-entrypoint/main.go
```diff
@@ -37,6 +31,26 @@ func realMain() int {
 	ctx, closer := signalcontext.WithInterrupt(context.Background(), log)
 	defer closer()
 
+	if port := os.Getenv("WAYPOINT_EXEC_PLUGIN_SSH"); port != "" {
```
mitchellh, line 34: What is this?

internal/ceb/ssh.go
```diff
@@ -0,0 +1,263 @@
...
+	"syscall"
+
+	"github.com/creack/pty"
+	"github.com/gliderlabs/ssh"
+	"github.com/hashicorp/go-hclog"
+	"github.com/pkg/errors"
+	gossh "golang.org/x/crypto/ssh"
+)
+
+func RunExecSSHServer(ctx context.Context, logger hclog.Logger, sport, hostkey, key string) error {
```
mitchellh, line 21: Is this supposed to be exported or expected to be called outside this package? I feel this should be unexported.

mitchellh, line 21: Given the complexity of this (its a full SSH server), I'd love to see some unit tests on this. I think it should be easy enough to spin up one in-memory and use an SSH client to test that it works on *nix-like machines (macOS, Linux).

mitchellh (changes requested):

builtin/aws/lambda/platform.go
krantzinator: This part is what prompted my question [in Slack](https://hashicorp.slack.com/archives/C013QT1KG9W/p1612482100254500).
We create a few things as part of the workspace (e.g. Task Role) that we don't cleanup but probably should.

builtin/aws/ecr/mapper.go
evanphx: We assume it's remote, so probably need a Location, yeah.

internal/ceb/ssh.go
evanphx: It's called by the entrypoint so it has to be exported.

builtin/aws/lambda/platform.go
evanphx: Cleaned that up.

builtin/aws/lambda/platform.go
evanphx: Added it!

mitchellh (approved):

mitchellh:
Please look and fix the test failures first. :)

internal/ceb/ssh/server.go
```diff
@@ -0,0 +1,253 @@
...
+							logger.Warn("error changing window size, this doesn't quit the stream",
+								"err", err)
+						}
+					}
+				}
+			}()
+		} else {
+			logger.Debug("executing command without pty")
+
+			// You might think "Hey, why make these pipes, can't we just
```
briancain, line 150: Appreciate these kinds of comments :smile: :+1:

briancain (approved):
Rise and shine! :lab_coat: :crab:

internal/ceb/ssh/server.go
evanphx: Luckily I had hit this many years ago and when I hit it again, it didn't take long to remember. But it's EXTREMELY unclear what is happening.

PR #1053, opened 2021-02-03
Singleton jobs and job memdb fixes

mitchellh:
This introduces the concept of "singleton" jobs. You may now specify a "singleton id" when creating a job. If a job exists with that same singleton ID _and_ is queued, it will overwrite that job rather than queue a new one. We plan on using this for Git polling in the future.

This also introduces shallow copying of the job index structure when modifying it. This is required to make our MemDB usage safer. I've opted to not backport this change cause I don't know any real world bugs caused by this currently.


mitchellh:
Actually going to close this and work on more things in integration with git polling before opening. That'll result in a bunch more PRs at some point all at once but I want to test more things together first.

mitchellh:
Deleting this branch cause I pushed it up as another name.

PR #1057, opened 2021-02-05
UI: git settings

gregone:
This PR adds a new settings screen for projects. 

Currently, the happy path works (with correct user input), but there is no success/error messaging, but I will add that as soon as https://github.com/hashicorp/waypoint/pull/1206 is reviewed and merged. 

Screen Shot 2021-03-16 at 14 26 03

### DIFF
--- a/ui/app/components/app-form/project-settings.hbs
+++ b/ui/app/components/app-form/project-settings.hbs
@@ -1,0 +1,68 @@
+<form class="pds-form">
+  <div part="fields">
+    <fieldset class="pds-formFieldSet">
+      <div class="pds-formField">
+        <label for="git-source-url" class="pds-fieldName">
+          Git Source URL
+        </label>
+        <Pds::Input
+          @type="text"
+          aria-describedby="{HELP_ID} {ERROR_ID}"
+          id="git-source-url"
+          placeholder="https://github.com/mycompany/awesome-project"
+          value={{this.git.url}}
+          {{on "change" this.setGitUrl value="target.value"}}
+        />
+      </div>
+      <div class="pds-formField">
+        <label for="git-source-ref" class="pds-fieldName">
+          Git Ref
+        </label>
+        <Pds::Input
+          @type="text"
+          aria-describedby="{HELP_ID} {ERROR_ID}"
+          id="git-source-ref"
+          placeholder="main"
+          value={{this.git.ref}}
+          {{on "change" this.setGitRef value="target.value"}}
+        />
+      </div>
+      <div class="pds-formField">
+        <label for="git-source-path" class="pds-fieldName">
+          Git Source Path
+        </label>
+        <Pds::Input
+          @type="text"
+          aria-describedby="{HELP_ID} {ERROR_ID}"
+          id="git-source-path"
+          placeholder="/my-project-path"
+          value={{this.git.path}}
+          {{on "change" this.setGitPath value="target.value"}}
+        />
+      </div>
+      <div class="pds-formField">
+         <label class="pds-fieldName">
+          Automated Sync
+        </label>
+        <div class="pds-formFieldToggle">
+          <XToggle
+            @showLabels={{true}}
+            @name="git-sync-enabled"
+            @value={{@project.remoteEnabled}}
+            @onToggle={{fn (mut @project.remoteEnabled)}} as |toggle|>
+            <toggle.switch/>
+
+            <toggle.label for="git-sync-enabled">
+              <span class="pds-toggleLabel">
+                Enable automated deployment on changes
+              </span>
+            </toggle.label>
+          </XToggle>
+
+        </div>
+      </div>
+    </fieldset>
+  </div>
+
+  <Pds::Button @variant="primary" {{on "click" this.saveSettings}}>Apply</Pds::Button>
+</form>
--- a/ui/app/components/app-form/project-settings.hbs
+++ b/ui/app/components/app-form/project-settings.hbs
@@ -41,6 +41,27 @@
         />
       </div>
       <div class="pds-formField">
+        <label for="git-auth-type" class="pds-fieldName">
+          Authentication
+        </label>
+        <div class="pds-formRadioWrapper">
+          <Pds::RadioField
+            @id="git-auth-ssh"
+            name="git-auth-type"
+            {{on "input" (fn this.setAuthCase 5)}}
+          >
+            SSH
+          </Pds::RadioField>
+          <Pds::RadioField
+            @id="git-auth-basic"
+            name="git-auth-type"
+            {{on "input" (fn this.setAuthCase 4)}}
+          >
+            Username &amp; Password
+          </Pds::RadioField>
+        </div>
+      </div>
+      <div class="pds-formField">
          <label class="pds-fieldName">
           Automated Sync
         </label>
--- a/ui/app/components/app-form/project-settings.hbs
+++ b/ui/app/components/app-form/project-settings.hbs
@@ -11,7 +11,7 @@
           id="git-source-url"
           placeholder="https://github.com/mycompany/awesome-project"
           value={{this.git.url}}
-          {{on "change" this.setGitUrl value="target.value"}}
+          {{on "input" (set this "project.dataSource.git.url" (get _ "target.value"))}}
         />
       </div>
       <div class="pds-formField">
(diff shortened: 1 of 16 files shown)


krantzinator:
Looks like (Firefox) dark mode strikes again, this time with a weird orange circle behind the active selection:
image

krantzinator:
Also on Firefox, I seem to lost the headings on dark mode:
image

krantzinator:
Are all of these inputs required? Do we have default values for any of them?
I think we should note either case -- if all are required, or, if only some of them are required, which ones specifically.

gregone:
@krantzinator Updated the fields based on your feedback: 

Screen Shot 2021-03-23 at 16 16 09

Screen Shot 2021-03-23 at 16 15 59

ui/app/components/app-form/project-settings.hbs
```diff
@@ -0,0 +1,163 @@
...
+        <label for="git-source-url" class="pds-fieldName">
+          Git Source URL
+        </label>
+        <Pds::Input
+          @type="text"
+          id="git-source-url"
+          placeholder="git@github.com:mycompany/awesome-project.git"
+          value={{this.git.url}}
+          required
+          {{on "input" (set this "project.dataSource.git.url" (get _ "target.value"))}}
```
meirish, line 14: I noticed this in HCP a little bit ago - I think this syntax is pre 1.0 of the set helper 😭 - just something to keep an eye out for!

ui/app/components/app-form/project-settings.hbs
```diff
@@ -0,0 +1,163 @@
+<form class="pds-form">
+  <div part="fields">
+    <fieldset class="pds-formFieldSet">
+      <div class="pds-formField">
+        <label for="git-source-url" class="pds-fieldName">
+          Git Source URL
```
meirish, line 6: Are y'all using `ember-intl` ? I saw there was a `translations/` folder but didn't dig too much further than that. If you want to linting to be loud about it you can turn on `no-bare-strings` rule in `ember-cli-template-lint`.

ui/app/components/app-form/project-settings.hbs
```diff
@@ -0,0 +1,163 @@
...
+              </span>
+            </toggle.label>
+          </XToggle>
+
+        </div>
+      </div>
+    </fieldset>
+  </div>
+
+  <Pds::Button @variant="primary" {{on "click" this.saveSettings}}>Apply</Pds::Button>
```
meirish, line 162: If you move the save handler to the form tag at the very top with:

`<form class="pds-form" {{on "submit" this.saveSettings}}>` 

and then change this to be a `@type="submit"` button, you get the added nicety of hitting "enter" to submit the form if any of the fields are focused.

Also something to think about here: the "saving" state isn't being handled so it's possible for a user who's click-happy to trigger save more than once.

In HCP we use `ember-concurrency` tasks for all saves so that we can use the derived state to disable the save buttons too (certainly not required, but if you're using e-c otherwise, it's easier than tracking the save state manually).

ui/app/components/app-form/project-settings.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    dataSource.setGit(git);
+    ref.setDataSource(dataSource);
+    ref.setDataSourcePoll(dataSourcePoll);
+    let applist = project.applicationsList.map((app: Application.AsObject) => {
+      return new Application(app);
+    });
+    ref.setApplicationsList(applist);
+    let req = new UpsertProjectRequest();
+    req.setProject(ref);
```
meirish, line 214: /me takes a breath - wow that's quite the workout, eh?

I'm coming at this with zero experience, so if this is a bad idea please feel free to discard it 😂 - but! - have you considered initializing some of these objects in the constructor and then having an action that does the `setFoo` calls on the PB fields?  I'm not sure how that would work in the pre-filled scenario though when you're setting data from the server... I was just wondering if it's possible to reuse the proto structures instead of having the ProjectModel be a separate data structure that fills in the protos on save.

I'm not sure what's possible, and certainly not something that needs fixed in this PR, just something that struck me reading through this.

ui/app/styles/_x-toggle.scss
```diff
@@ -0,0 +1,21 @@
+:root {
+  --toggle-color-off: #{dehex(color.$ui-cool-gray-300)};
```
meirish, line 2: ha had to look up the `dehex` function - nice!

ui/app/templates/workspace/projects/project/apps.hbs
```diff
@@ -3,6 +3,7 @@
     <h1>{{@model.name}}</h1>
     <small>{{pluralize @model.applicationsList.length "application"}}</small>
   </div>
+  <LinkTo @route="workspace.projects.project.settings" @model={{@model}}><Pds::Icon @type="settings"/> Settings </LinkTo>
```
meirish, line 6: Just making sure you want to pass the whole model here and not the `id`, it looks like you're not fetching the model here: https://github.com/hashicorp/waypoint/pull/1057/files#diff-3741bb61d2053798b179535df28c0a731f67c2856d63f582e01cb90dda79ed6cR25 
but passing a whole object in a link will bypass route hooks so that may be an unwelcome surprise later !

meirish: And now that I wrote that, I'm not sure if that's still true for routes that don't have params (which I realized is true for this route) 🤔

ui/app/components/app-form/project-settings.hbs
meirish: Ah I saw you're requiring this explicitly - it looks like the 1.0 one they suggest using set with `pick` from `ember-composable-helpers`: https://github.com/pzuraq/ember-set-helper#picking-values-with-ember-composable-helpers - I'd recommend going that route and with the most recent release so you're not stuck with an upgrade later on.

ui/app/components/app-form/project-settings.ts
```diff
@@ -0,0 +1,225 @@
...
+    await this.api.client
+      .upsertProject(req, this.api.WithMeta())
+      .then(() => {
```
meirish, line 217: I think my favorite thing about `async/await` is you can use `try / catch` with this instead of `then().catch()`. It'd look like this:

```
try {
  await this.api.client.upsertProject(req, this.api.WithMeta());
  this.flashMessages.success('Settings saved');
  this.router.transitionTo('workspace.projects.project', this.project);
} catch (err) {
  this.flashMessages.error('Failed to save Settings', { content: err.message, sticky: true });
}
```

ui/app/styles/_pds-form.scss
```diff
@@ -0,0 +1,39 @@
...
+    max-width: 800px;
+  }
+
+  .pds-formField {
+    margin-bottom: scale.$lg-1;
+  }
+
+  .pds-formFieldToggle {
+    display: flex;
+    flex-direction: row;
```
meirish, line 15: I thought `row` was the default value for flex-direction - is this overriding something or will this element be placed inside something that's `column` direction?

ui/tests/unit/routes/workspace/projects/project/settings-test.ts
```diff
@@ -0,0 +1,11 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+
+module('Unit | Route | workspace/projects/project/settings', function(hooks) {
```
meirish, line 4: Seeing this made me wonder what tests for the new settings component might look like.

meirish (approved):
Hey @gregone - this looks great! I left a smattering of comments - mostly food for thought. I think the only thing I'd say you should definitely change is pulling in the newest `ember-set-helper` and use it with `pick`.

ui/app/components/app-form/project-settings.hbs
gregone: re: `pick`, yeah I saw that and hesitated, but instead mimicked what you had, to then realize I had to pin a specific version of the `get` helper 😅 . Good point, I think i'll update that so we start with the right patterns

ui/app/styles/_pds-form.scss
gregone: probably a leftover from when I experimented with layout here. removed!

ui/app/components/app-form/project-settings.hbs
gregone: done: https://github.com/hashicorp/waypoint/pull/1057/commits/f20301d89246e6ee9d57e0679570cbd513266143

ui/app/components/app-form/project-settings.hbs
gregone: done: https://github.com/hashicorp/waypoint/pull/1057/commits/6069d6b798455f60db8ddc7a702a7d670c52af2e

ui/app/components/app-form/project-settings.hbs
gregone: done!

PR #1059, opened 2021-02-05
Implement "up" as a job operation type

mitchellh:
This adds an `UpOp` job operation type that performs the `waypoint up` logic.

Previously, `waypoint up` was implemented by queueing a build, deploy, and release operation sequentially with all the glueing happening in the CLI. We need to move this back into the backend logic so that we can queue "up" from UIs, server polling, etc. and retain a consistent experience.

This also converts `waypoint up` to use the new job type so that the experience is always consistent.


evanphx (approved):
Thin Clients ™️

PR #1060, opened 2021-02-05
Persist workspace information in DB, add projects

mitchellh:
Previously, workspace information was all "realized" on server startup and via other operations and was not persisted to disk. For example, we knew what workspaces existed based on what operations were run on what applications. 

For upcoming polling work, we need to hang more information off of workspaces, such as the last data source ref that a project in a workspace used (in this PR). This information can't be efficiently realized and I took this opportunity to make workspaces more of a real data item that we persist and can query in the datastore.

There are still a variety of pitfalls to workspace layouts as they exist (we can't tell if an application is NO LONGER in use). These pitfalls still existed prior to this so this PR doesn't make things better or worse in that regard. 

**There is a breaking API change in this PR**, the `Workspace` message type changed structure (but doesn't reuse a tag). There were no previous consumers of this API however and we never documented it so I think this is okay.


evanphx (approved):
Looks good! Makes sense to store the workspaces physically so we can hang environment information on them.

PR #1063, opened 2021-02-05
Closing issue 533 prevent incomplete deployments

DerekStrickland:
This PR does 4 things.

* Upsert the deployment immediately at the beginning of deployOperation.Do so that failed deployment are represented in the state.
* Adds a top-level Rollback function to the k8s platform module.
* Performs a rollback on deploy timeouts for k8s
* Performs a rollback on user cancellation for k8s

### DIFF
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
@@ -365,26 +366,23 @@ func (p *Platform) Deploy(
 		reportedError bool
 	)
 
-	timeout := 10 * time.Minute
+	timeout := 30 * time.Second
 
-	// Setup a reusable rollback function. Could potentially become a top level plugin abstraction (e.g. RollbackFunc)
-	rollbackMessage := "Deployment cancelled by user - attempting rollback..."
-	rollback := func() {
-		// If the request gets cancelled, rollback
-		step.Update(rollbackMessage)
-		destroyErr := p.Destroy(context.Background(), log, &result, ui)
-		if destroyErr != nil {
-			step.Update(fmt.Sprintf("Deployment rollback failed with err: %+v", destroyErr))
-		}
-		// This never prints - feels like this whole thing must be racy and not dependable, but it hasn't failed yet.
-		step.Update("Rollback successful")
-	}
-
-	// Setup a cancellation func to receive done signal and rollback if user cancels
-	// TODO: This prints multiple time. Does this fire multiple times?
+	cancelMu := &sync.Mutex{}
+	pollingComplete := false
 	go func() {
 		<-ctx.Done()
-		rollback()
+		// Only rollback if polling didn't complete which indicates a user cancellation.
+		cancelMu.Lock()
+		if !pollingComplete {
+			log.Warn("deployment cancelled by user - attempting rollback...")
+			err = p.Rollback(context.Background(), log, &result, ui)
+			if err != nil {
+				log.Error(fmt.Sprintf("deployment rollback failed with err: %s", err.Error()))
+			}
+			log.Warn("deployment rollback successful")
+		}
+		cancelMu.Unlock()
 	}()
 
 	// Wait on the Pod to start
@@ -396,7 +394,7 @@ func (p *Platform) Deploy(
 
 		if time.Since(lastStatus) > 10*time.Second {
 			step.Update(fmt.Sprintf(
-				"Waiting on deployment to become available: %d/%d/%d",
+				"Waiting on deployment to become available: %d/%d/%d\n",
 				*dep.Spec.Replicas,
 				dep.Status.UnavailableReplicas,
 				dep.Status.AvailableReplicas,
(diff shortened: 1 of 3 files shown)


builtin/k8s/platform.go
```diff
@@ -367,6 +368,23 @@ func (p *Platform) Deploy(
 
 	timeout := 10 * time.Minute
 
+	cancelMu := &sync.Mutex{}
+	pollingComplete := false
```
DerekStrickland, line 372: I'm hoping there's a better way to handle this than this mutex. Open to suggestions.

builtin/k8s/platform.go
```diff
@@ -367,6 +368,23 @@ func (p *Platform) Deploy(
...
+	go func() {
+		<-ctx.Done()
```
DerekStrickland, line 374: @briancain I took a look around Waypoint as you suggested and there are examples of this approach to listening for cancellation in `ui.go` and `server.go`. I think the gist is that it just blocks this goroutine until the signal is sent and the processing continues. I didn't put this in the poll loop because it's possible the user cancels between polls and so execution would not be guaranteed. At least, that's what I observed the way I implemented it. Mabye a PEBKAC problem?

builtin/k8s/platform.go
```diff
@@ -367,6 +368,23 @@ func (p *Platform) Deploy(
...
+		// Only rollback if polling didn't complete which indicates a user cancellation.
+		cancelMu.Lock()
+		if !pollingComplete {
```
DerekStrickland, line 377: Wondering if I need to double lock...

builtin/k8s/platform.go
```diff
@@ -376,7 +394,7 @@ func (p *Platform) Deploy(
 
 		if time.Since(lastStatus) > 10*time.Second {
 			step.Update(fmt.Sprintf(
-				"Waiting on deployment to become available: %d/%d/%d",
+				"Waiting on deployment to become available: %d/%d/%d\n",
```
DerekStrickland, line 397: I'm actually wondering if adding this line break isn't masking a problem. Without it, log output is interwoven.

internal/core/app_deploy.go
```diff
@@ -150,11 +151,16 @@ func (op *deployOperation) Upsert(
 	return resp.Deployment, nil
 }
 
-func (op *deployOperation) Do(ctx context.Context, log hclog.Logger, app *App, _ proto.Message) (interface{}, error) {
+func (op *deployOperation) Do(ctx context.Context, log hclog.Logger, app *App, msg proto.Message) (interface{}, error) {
 	// Sync our config first
 	if err := app.ConfigSync(ctx); err != nil {
 		return nil, err
 	}
+	// Upsert immediately so that if deployment fails we still have a reference to it in the db
```
DerekStrickland, line 159: This is a stop-gap measure. I suspect the larger plan is to track all pipeline lifecycle events in the state, but this had the effect, in this case, of making a hung deployment more obvious to the user. In my case, I was experiencing a timeout due to a crash-loop backoff. So this was nice to have. Maybe the long term vision is something like an FSM. Not sure.

builtin/k8s/platform.go
```diff
@@ -367,6 +368,23 @@ func (p *Platform) Deploy(
 
 	timeout := 10 * time.Minute
 
+	cancelMu := &sync.Mutex{}
+	pollingComplete := false
+	go func() {
```
mitchellh, line 373: A goroutine in the plugin like this is likely unsafe. After the RPC calls end, we will terminate the plugin at some point, and there is no guarantee this goroutine cleanly exits. It is likely that that process is slow enough that this _happens_ to work, but we can't guarantee it going forward.

internal/core/app_deploy.go
mitchellh: Indeed, I have to think about this more but this change makes me feel slightly uncomfortable for the reasons you stated.

mitchellh (commented):
I didn't do a review on the code details themselves, but some high level thoughts: I'm unsure if this is a tenable, stable approach. I noted some fundamental issues around starting a goroutine from a plugin call. If you didn't run this yourself I'd claim this likely doesn't work at all! So I'm surprised to hear it does.

I think a stopgap approach may be to change the goroutine into a defer call instead. If we defer and check that we reached some end, and if we have not, roll back, then I think that would work, cause a defer will block the return from the function.

Long term, we of course want to address the comments you raised on your one core change: we need to better track full lifecycle events and handle those. But if we can get a stopgap solution for Kubernetes in I'm not against it so long as it doesn't negatively impact other things.

I'm uncertain of this PR.

mitchellh:
I just wanted to add -- since my last comment certainly came off a bit negative -- that I'm very thankful for you doing this work. 😄  I think you explored a lot of areas of how this can be resolved simply and that has a lot of value. In particular, as I said in my previous comment, I think your direction with the goroutine (but maybe as a defer instead) has promise to improve things for the K8S plugin. So, you have my gratitude there!

DerekStrickland:
> I just wanted to add -- since my last comment certainly came off a bit negative -- that I'm very thankful for you doing this work. 😄 I think you explored a lot of areas of how this can be resolved simply and that has a lot of value. In particular, as I said in my previous comment, I think your direction with the goroutine (but maybe as a defer instead) has promise to improve things for the K8S plugin. So, you have my gratitude there!

It didn't come off as negative at all! I am just grateful that you thoughtfully considered it. I'm unsure of the PR too and submitted it hoping to get feedback.  My specific concerns are these:

- I was really not confident about that goroutine so thanks for the pro-tip! Go is still newish to me, and this feedback helped clarify some things around language mechanics.
- Rolling back (or not) feels like it needs to happen as part of pipeline management. It's like the pipeline is a long-running distributed transaction, and one approach could be to manage with full rollbacks. Another might be to have checkpoints and partial rollbacks. I'm just not sure what the long-term vision is, so I wasn't sure how best to improve things for now while looking towards the future.
- It felt super iffy while I was writing it that Deploy would call Destroy or Rollback directly. I feared that some essential pipeline processing might get missed. What would have felt more natural to me would have been to have raised an event or a custom, detectable error that the pipeline could interpret to decide whether to trigger a rollback etc, but I figured at least this could start a conversation, and I could refactor.
- As stated before, I don't really think I know the ultimate vision for state management, and so it was iffy to change it. Are we shooting for one current discreet state only, or that plus keeping an event entry for all state transitions? Since I don't know, I was hesitant to make a state change. Thanks for the direct feedback on that. I'll remove it.

Based on your feedback, It seems like you are open to trying to rollback k8s _**from this plugin**_ on timeouts & user cancellation specifically, at least for now. I'll remove the Upsert, and I'll refactor to a defer and see how that pans out. Thanks again for taking the time to review.

DerekStrickland:
I tried the deferred approach but must have implemented it wrong because it never seemed to fire. Just for fun, I might tinker with this some more because...curiosity. I'm closing for now though because for all the documented reasons it seems like this really is more appropriate for the core team to think through. I'll look for another issue and ask first this time before starting :)

PR #1066, opened 2021-02-08
internal/cli: fall back to legacy "up" if server doesn't support op up

mitchellh:
This just adds back the old "up" logic from 0.2.x as a fallback. Over
time, it is possible that the UX of these two diverge but I expect that
we'll see less and less of the old behavior as people upgrade.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"strings"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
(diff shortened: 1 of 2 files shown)


evanphx (approved):
Thankfully waypoint is young enough that we'll only hit this a few places.

briancain (changes requested):
Overall looks great! I think we need to have some extra client side handling for the case when an old server doesn't respond with `Unimplemented`, i.e. before these changes.

briancain (approved):
Perfect! Thanks for the updates :smile: Legacy up works as expected

PR #1072, opened 2021-02-09
Upgrade ember to 3.24

gregone:
- upgrades Ember to 3.24 (latest Stable)
- also ran [`ember-cli-update`](https://github.com/ember-cli/ember-cli-update#usage) to ensure everything's up to date. Ran the codemods but no changes were needed


ui/config/targets.js
```diff
@@ -2,7 +2,7 @@
 
 const browsers = ['last 1 Chrome versions', 'last 1 Firefox versions', 'last 1 Safari versions'];
 
-const isCI = !!process.env.CI;
+const isCI = Boolean(process.env.CI);
 const isProduction = process.env.EMBER_ENV === 'production';
 
 if (isCI || isProduction) {
```
meirish, line 8: sidenote: you could probably reduce your UI payload size by removing this (assuming you don't support IE11).

meirish (approved):
Looks good! I had one comment about your target.js, but that's a separate concern.

ui/config/targets.js
gregone: Thanks! That's very good feedback! I'm not even sure what the Waypoint browser support policy is, but it should be fine to drop. I'll ask first and change it in another PR.

Issue #1055, opened 2021-02-04
Kubernetes Configsourcer "error retrieving config value"


coryodaniel:
**Describe the bug**

I am using the kubernetes config sourcer to set environment variables from a k8s secret generated from `kubectl create secret generic ... --from-env-file`

I am getting an error that the key in the secret does not exist. I tried changing the secret name to see if it was an issue with accessing the secret or if it was an issue with getting the value. If I change the name to a secret that _doesn't exist_ I get an error that the secret doesnt exist 👍🏼 , so waypoint is definitely finding the secret when using the right name, but it doesnt seem to be able to get the keys listed below.

I haven't set any additional config with `waypoint config source-set`.

The error: 

```
2021-02-03T17:16:42-08:00: [INFO] entrypoint: entrypoint starting: deployment_id=01EXN8H9716YG6M57DZRVC2ZX5 instance_id=01EXN9SRVWSB0R69AXME76NND3 args=[/app/bin/currency, start]
2021-02-03T17:16:42-08:00: [INFO] entrypoint: entrypoint version: full_string="Waypoint v0.2.1" version=v0.2.1 prerelease= metadata= revision=
2021-02-03T17:16:42-08:00: [INFO] entrypoint: server version info: version=v0.2.1 api_min=1 api_current=1 entrypoint_min=1 entrypoint_current=1
2021-02-03T17:16:43-08:00: [WARN] entrypoint.config.watcher: error retrieving config value: source=kubernetes key=CURRENCY_ACCOUNT_ID err="rpc error: code = NotFound desc = key not found: CURRENCY_ACCOUNT_ID"
2021-02-03T17:16:43-08:00: [WARN] entrypoint.config.watcher: error retrieving config value: source=kubernetes key=CURRENCY_ACCOUNT_KEY err="rpc error: code = NotFound desc = key not found: CURRENCY_ACCOUNT_KEY"
2021-02-03T17:16:43-08:00: [INFO] entrypoint.config.watcher: env vars changed, sending new child command
```

**Steps to Reproduce**

The `waypoint.hcl` file:

```hcl
project = "currency"

app "currency" {
  labels = {
    "service" = "currency",
    "env"     = "dev"
  }
  
  config {
    env = {
      CURRENCY_ACCOUNT_KEY = configdynamic("kubernetes", {
          name = "currency"
          key = "CURRENCY_ACCOUNT_KEY"
          secret = true      
      })

[...]

PR #1073, opened 2021-02-09
builtin/k8s: Handle secret configs from k8s

briancain:
Prior to this commit, we only handled configs that fit the ConfigMap
type. This commit updates that to also handle the case where we try to
receive secret configs when `secret = true` from configdynamic.

Fixes https://github.com/hashicorp/waypoint/issues/1055

### DIFF
--- a/builtin/k8s/config_sourcer.go
+++ b/builtin/k8s/config_sourcer.go
@@ -2,6 +2,7 @@ package k8s
 
 import (
 	"context"
+	base64 "encoding/base64"
 	"math"
 	"math/rand"
 	"sync"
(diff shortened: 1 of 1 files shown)


builtin/k8s/config_sourcer.go
```diff
@@ -126,6 +127,25 @@ func (cs *ConfigSourcer) read(
 		case *corev1.ConfigMap:
 			value, ok = d.Data[k8sReq.Key]
 
+		case *corev1.Secret:
+			var secretValue []byte
+
+			secretValue, ok = d.Data[k8sReq.Key]
+			sEnc := base64.StdEncoding.EncodeToString(secretValue)
```
mitchellh, line 134: probably want to gate this on `if ok`

builtin/k8s/config_sourcer.go
```diff
@@ -126,6 +127,25 @@ func (cs *ConfigSourcer) read(
...
+			decValue, err := base64.StdEncoding.DecodeString(sEnc)
```
mitchellh, line 135: Why do you encode and then immediately decode? Probably need to comment whats going on here.

mitchellh (changes requested):

mitchellh (approved):

PR #1074, opened 2021-02-09
internal/runner: Fix race condition around runner shutdown

evanphx:
The old code had a race condition where it was possible to shutdown the runner while a job had just started and would have continued to run.

Because of the behavior around executing the runner within the client package as a local runner inside the CLI, this bug has likely never been hit. But a busy system running a standalone runner certainly would have.

### DIFF
--- a/internal/runner/accept.go
+++ b/internal/runner/accept.go
@@ -42,9 +42,9 @@ func (r *Runner) AcceptExact(ctx context.Context, id string) error {
 var testRecvDelay time.Duration
 
 func (r *Runner) accept(ctx context.Context, id string) error {
-	r.runningMu.Lock()
+	r.runningCond.L.Lock()
 	shutdown := r.shutdown
-	r.runningMu.Unlock()
+	r.runningCond.L.Unlock()
 
 	if shutdown {
 		return ErrClosed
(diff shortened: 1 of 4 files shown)


internal/runner/accept.go
```diff
@@ -38,8 +39,14 @@ func (r *Runner) AcceptExact(ctx context.Context, id string) error {
 	return r.accept(ctx, id)
 }
 
+var testRecvDelay time.Duration
+
 func (r *Runner) accept(ctx context.Context, id string) error {
-	if r.closed() {
+	r.runningMu.Lock()
```
mitchellh, line 45: If this is just equivalent to `runnerCond.L` anyways, I think it's better to just use the locker on the cond var and not store `r.runnerMu` at all.

internal/runner/runner.go
```diff
@@ -173,13 +179,16 @@ func (r *Runner) Start() error {
...
+
+	for r.runningJobs > 0 {
+		r.runningCond.Wait()
 	}
 
-	// Wait for our jobs to complete
-	r.acceptWg.Wait()
+	r.shutdown = true
+
+	r.runningCancel()
```
mitchellh, line 191: Comment here would be good on what this does (what does this affect)

mitchellh (changes requested):

internal/runner/accept.go
evanphx: Ah! I started that but didn’t finish it. Gotcha.

mitchellh (approved):

PR #1078, opened 2021-02-10
internal/{client,runner}: start a single runner at the start of the CLI

evanphx:
This moves runners from being spawn per job to creating a single runner that runs for the lifetime of a client Project value. The reason for this change to make it possible for a server to spawn jobs on a CLI client in response to a client RPC call, rather than just when the CLI client queues the job itself.

We're going to be using this server spawned jobs function (future PR) to cleanup the API for exec and logs plugin functionality, allowing the server to decide if the plugin needs to be spawned or not, rather than the client.

### DIFF
--- a/internal/client/job.go
+++ b/internal/client/job.go
@@ -59,66 +59,27 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
 // The receiver must be careful to not block sending to mon as it will block
 // the job state processing loop.
 func (c *Project) doJobMonitored(ctx context.Context, job *pb.Job, ui terminal.UI, monCh chan pb.Job_State) (*pb.Job_Result, error) {
-	log := c.logger
-
 	// Be sure that the monitor is closed so the reciever knows for sure the job isn't going
 	// anymore.
 	if monCh != nil {
 		defer close(monCh)
 	}
 
-	// cb is used in local mode only to get a callback of the job ID
-	// so we can tell our runner what ID to expect.
-	var cb func(string)
-
 	// In local mode we have to start a runner.
 	if c.local {
-		log.Info("local mode, starting local runner")
-		r, err := c.startRunner()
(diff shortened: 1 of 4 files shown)


internal/client/job.go
```diff
@@ -59,66 +59,27 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
...
 				Id: &pb.Ref_RunnerId{
-					Id: r.Id(),
+					Id: c.activeRunner.Id(),
 				},
 			},
 		}
 	}
 
-	return c.queueAndStreamJob(ctx, job, ui, cb, monCh)
+	return c.queueAndStreamJob(ctx, job, ui, nil, monCh)
```
mitchellh, line 82: can remove this 4th arg completely from the func I think

internal/client/project.go
```diff
@@ -53,6 +62,9 @@ func New(ctx context.Context, opts ...Option) (*Project, error) {
 		}
 	}
 
+	// Used by any background goroutines that we'll spawn (like runner job processing)
+	client.bg, client.cancel = context.WithCancel(context.Background())
```
mitchellh, line 66: can we pair the names so its clear what is being canceled here? i.e. `bgCancel` instead of `cancel`? Right now we have one context so its fine, but I sometimes get confused as projects mature.

mitchellh (changes requested):
One minor change.

I think one issue still is `AcceptMany` will serialize job execution, which I think will be an issue for you. My view on how I was going to do this (why there is now `AcceptGo` or something) is that I would setup a "pool" of goroutines all blocking on `Accept` in a loop. So if you run the runner with a parallelism of N=5, I'd spin up 5 goroutines and just `for { accept() }` on them.

We may want to make that functionality a runner API though.

krantzinator (approved):

mitchellh (approved):

PR #1080, opened 2021-02-11
Move exec plugin launching to the server

evanphx:
This allows the server to launch a job itself (our first server spawned job) to start the exec plugin if a deployment plugin has the execfunc defined.

Two architectural takeaways:

1. The logic about when to use an exec plugin is now much more obvious: A deployment was created by a plugin that defined ExecFunc.
2. The CLI advertises it's local runner via grpc metadata, allowing the StartExecStream rpc the ability to see this fact, and can thusly spawn the job directly on that runner. This keeps the nature of "work spawned by a CLI happens on the CLIs runner" intact.


internal/server/singleprocess/state/instance_exec.go
```diff
@@ -98,6 +98,65 @@ func (s *State) InstanceExecCreateByTargetedInstance(id string, exec *InstanceEx
...
+func (s *State) InstanceExecCreateForVirtualInstance(ctx context.Context, id string, exec *InstanceExec) error {
+	// If the caller specified an instance id already, then just validate it.
+	if id == "" {
+		return status.Errorf(codes.NotFound, "No instance id given")
+	}
+
+	for {
+		// We have to start a new txn per iteration because we want to be able to observe
+		// the newly created record for the instance.
+		txn := s.inmem.Txn(false)
```
mitchellh, line 114: I believe you forget to `txn.Abort` clean up this.

internal/serverconfig/config.go
```diff
@@ -4,6 +4,11 @@ import (
 	"strconv"
 )
 
+// The metadata key that stores the runner id associated with a client. This is
+// used by the CLI to advertise it's local client for when the server needs to
+// spawn jobs back on that client in response to an RPC.
+const GrpcMetadataRunnerId = "waypoint-runner-id"
```
mitchellh, line 10: I think this should go in `internal/server` since that's the package that actually uses this?

internal/client/project.go
```diff
@@ -111,6 +113,18 @@ func New(ctx context.Context, opts ...Option) (*Project, error) {
 	return client, nil
 }
 
+// addMetadata adds any grpc metadata that should be include by any gRPC
+// calls that are done with returned context. This is used to inject metadata
+// about the local runner, allow the server to target jobs back to the client
+// that performed an RPC call.
+func (c *Project) AddMetadata(ctx context.Context) context.Context {
```
mitchellh, line 120: This is the only thing that's kind of suspect to me.

It feels like maybe we should do a few things instead: 

* Have an exported func on `Project` that is like `TargetRunnerId` which would return non-empty if one exists (active runner exists basically).
* Add a `WithRunnerId` func to `internal/server` so we have the yin/yang of inject/extract for context metadata.
* Have `internal/cli` do a `if TargetRunnerId() != "" { ctx = server.WithRunnerId(...) }` type thing.

I agree with what you're trying to do here, just not sure if organizationally/structurally I agree with this location.

internal/server/singleprocess/service_exec.go
```diff
@@ -51,10 +55,118 @@ func (s *service) StartExecStream(
...
+				Job: job,
+
+				// TODO unknown if this is enough time for when the request is queued
+				// by a runner-less client but a user waiting 60 seconds will get impatient
+				// regardless.
+				ExpiresIn: "60s",
+			})
+			if err != nil {
+				return err
+			}
```
mitchellh, line 130: Do we need to defer and cancel the job? If this function exits, we want the job to stop right?

mitchellh (changes requested):
Awesome! Honestly for such a big PR I have very little to say!

internal/server/singleprocess/state/instance_exec.go
evanphx: I looked and abort is a noop for read txns so I just didn't bother, since I didn't want to accumulate defer's on the stack.

internal/serverconfig/config.go
evanphx: So I didn't want to pull the whole server back into `internal/client` that's why I put it here. I can move it where ever.

internal/server/singleprocess/state/instance_exec.go
mitchellh: Ah, can you note that in the comments please?

internal/serverconfig/config.go
mitchellh: That is a really good point. How about a new `internal/server/grpcmetadata` package or something? It feels weird that the inject/extract and key are all in separate packages. :)

internal/client/project.go
evanphx: The reason it's exported like this is we do gRPC call outside the context of the project all the time, so it needed to be exported. If it were an accessor, I'd still want a code to wrap the metadata addition functionality so that it's not sprinkled everywhere we setup a context to do an RPC call.

internal/server/singleprocess/service_exec.go
evanphx: We do that below on line 140, canceling it if waiting had an error. I wasn't sure about the semantic around canceling a completed job so I only did it where I knew I wanted it canceled.

internal/client/project.go
evanphx: I don't think putting a context wrapper on the server package is the right place, I'm not sure you'd want to import it everywhere that you do an RPC call.

evanphx: Opt'd to basically do what you said originally given the refactoring into a grpcmetadata package clean it up otherwise.

mitchellh (approved):

briancain (approved):

Issue #1084, opened 2021-02-12
Repeating a release with aws-ecs improperly rewrites ALB listeners


skottler:
**Describe the bug**
Running `waypoint release -repeat` will replace the existing listener on port 443 (inferred via cert ARN) with 

**Steps to Reproduce**
Run `waypoint release -repeat` against an existing deployment with a port 443 listener. Note the previous 443 listener has been removed and the ALB is now only listening on port 80.

**Expected behavior**
The listener stays on port 443.

This bug lives in `releaser.go` and seems to interact with https://github.com/hashicorp/waypoint/issues/765. Because the listener is eagerly deleted due to #765, the conditional to verify whether there are existing listeners evals to false and then this gets called:

```
lo, err := elbsrv.CreateListener(&elbv2.CreateListenerInput{
	LoadBalancerArn: lb.LoadBalancerArn,
	Port:            aws.Int64(80),
	Protocol:        aws.String("HTTP"),
	DefaultActions: []*elbv2.Action{
		{
			ForwardConfig: &elbv2.ForwardActionConfig{
				TargetGroups: tgs,
			},
			Type: aws.String("forward"),
		},
	},
})
```

It's possible that fixing #765 will make this issue moot, but figured I'd open it since I'd noticed the behavior.

**Waypoint CLI Version:** binary built from HEAD, can also be reproduced against 0.2.x release(s).

skottler:
> It's possible that fixing #765 will make this issue moot, but figured I'd open it since I'd noticed the behavior.

Confirmed that with #765 addressed this is no longer an issue 🥳

PR #1087, opened 2021-02-12
builtin/aws/ecs: fix listener deletion check

### LINKED ISSUE #765: aws-ecs: third `up` deletes ALB
When running through the [AWS-ECS Tutorial](https://learn.hashicorp.com/tutorials/waypoint/aws-ecs?in=waypoint/docker-aws), after completing the tutorial steps I ran `waypoint up` near-simultaneously in two different terminals.  At the end when the commands finished running, the ALB had no listener (100% of traffic was being dropped).  Rerunning `waypoint up` did not restore the ALB Listener, and it required a manual fix in AWS before traffic would flow again.

 ¯\\_(ツ)_/¯   Concurrency is a hard problem, and the solutions are varied.  Couple of ideas to get the brainstorm forming:
1. Force every releaser to handle concurrency correctly
2. Add a release queue per project, with the waypoint server blocking a client/runner until it becomes active in the queue (Optional: add ability to drop non-active items from the release queue when a new release is enqueued)
3. Mix of 1/2:  Allow releasers to declare if they handle concurrency correctly.  If they do, allow concurrent runs.  If not, use the waypoint server for linearization of the releases.

Use some mutex via CI/CD runner to prevent `waypoint up` from being called concurrently.

This is likely to happen on teams with frequent commits that trigger deployments, or even as a rare scenario for small teams with relatively infrequent commits/deploys.

krantzinator:
Makes this comment true
```
If there is only 1 target group, delete the listener
```

Closes #765
Smaller implementation of #795 and closes #795

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1198,7 +1198,7 @@ func destroyALB(
 		var tgs []*elbv2.TargetGroupTuple
 
 		// If there is only 1 target group, delete the listener
-		if len(def) == 1 {
+		if len(def) == 1 && len(def[0].ForwardConfig.TargetGroups) == 1 {
 			log.Debug("only 1 target group, deleting listener")
 			_, err = elbsrv.DeleteListener(&elbv2.DeleteListenerInput{
 				ListenerArn: listener.ListenerArn,


skottler (approved):
Tried this locally and it works great and prevents the issue in #765.

Thanks @krantzinator!

skottler:
This pull request also fixes the issue described in https://github.com/hashicorp/waypoint/issues/1084.

PR #1092, opened 2021-02-12
builtin/k8s/{platform,releaser}: Support multi-port application configs

### LINKED ISSUE #821: Support multiple ports for Kubernetes deploy/release

We have a service that terminates both HTTP and HTTPS, and we would like to map a single ELB with multiple ports to that service through Waypoint. 

Kubernetes supports mapping multiple ports in a Service, whereas Waypoint only supports one. This feature request is to enable support for multiple ports in both deploy and release stages.

I'm not super familiar with HCL's flexibility, but it would be nice if the `service_port`, `node_port`, and `port` could either be a single number or an array of numbers. Having an array will map corresponding values among each other.

I've considered to not use Waypoint for this, or writing my own fork to the k8s plugin.

briancain:
This pull request introduces a new config option `ports` for the k8s platform in deploy and release. It allows users to define more than 1 listening port per application in waypoint.hcl. It more closely maps to how you would define a multi-port app in kubernetes. The old port variables will still be supported, however if they are defined along side Ports, it is considered an error.

This pull request jumps off the initial work by @rykov in https://github.com/hashicorp/waypoint/pull/846

Fixes #821


briancain:
FYI @rykov - I was able to get this working without needing to break backwards compatibility. Rather than trying to have an array of ints, it makes more sense I think to let people configure ports with the full spectrum of options that k8s allows. So I've introduced a new option Ports which is an array of map key vals that maps directly to the applications Container Port or Service Port options: https://kubernetes.io/docs/concepts/services-networking/service/

The old values are still there, and probably make sense for those who just want to expose a single port for simplicity. But for users who have more advanced needs, the Ports options should work better.

Thanks again for reporting the issue and getting the pull request started!

builtin/k8s/platform.go
```diff
@@ -578,6 +607,15 @@ deploy "kubernetes" {
 		),
 	)
 
+	doc.SetField(
+		"ports",
+		"a map of ports and options that the application is listening on",
+		docs.Summary(
+			"used to define and configure multiple ports that the application is",
+			"listening on. Can also define a 'name', 'host_port', and 'host_ip'.",
```
krantzinator, line 615: I think this would be more clear if we should all of the map keys as separate doc fields, like
```
doc.SetField("ports.port, "the port")
doc.SetField("ports.name", "name of the port")
doc.SetField("ports.host_port", "how we use the host port")
```
etc

builtin/k8s/releaser.go
```diff
@@ -91,34 +92,60 @@ func (r *Releaser) Release(
 		labelId: target.Id,
 	}
 
+	if (r.config.Port != 0 || r.config.NodePort != 0) && r.config.Ports != nil {
+		return nil, fmt.Errorf("Cannot define both 'Ports' and 'port' or 'node_port'." +
```
krantzinator, line 96: nit: is the capitalization supposed to differ across these terms?

builtin/k8s/platform.go
krantzinator: Or maybe just add `port` to your list for clarity, so like
```suggestion
			"listening on. Available keys are 'port', 'name', 'host_port', and 'host_ip'.",
```
(Also update wording on releaser doc field)

krantzinator (approved):
🎉 
This makes sense, though I think we should add clarity on the documentation.

builtin/k8s/releaser.go
briancain: Ah, nope! Ports is how it's referred to internally, but in the config it's lower case. So it should be lower case here too.

builtin/k8s/platform.go
briancain: Yeah, maybe we can document each option `ports` will support separately like the first comment suggests just so it's clear what each one does in the docs.

briancain: I'm actually not sure if `doc.SetField` can create sub docs like `ports.port` and `ports.name`. They might end up as new entries? If that's the case, I'd probably want to keep it all together and just define them inside the summary strings.

rykov:
@briancain I'm having trouble getting this to work with a `specs[0].ports.name is not set` error. Any chance you could point me to a doc, example, or a test case of how to use this? Thanks.

briancain:
Hey there @rykov - Sure thing, here's what it should look like...for example setting a single port with a name:

```hcl
project = "example-nodejs"

app "example-nodejs" {
  labels = {
    "service" = "example-nodejs",
    "env"     = "dev"
  }

  build {
    use "pack" {}
    registry {
      use "docker" {
        image = "localhost:5000/example-nodejs"
        tag   = "1"
        local = false
      }
    }
  }

  deploy {
    use "kubernetes" {
      probe_path = "/"
      ports = [
        {
          name = "http"
          port = 3000
        },
       ]
    }
  }

  release {
    use "kubernetes" {
      ports = [
        {
          node_port = 32000
        },
      ]
    }
  }
}

```

rykov:
@briancain I'm getting the following error:

```
» Deploying...
✓ Kubernetes client connected to https://... with namespace default
✓ Creating deployment...
✓ Deployment successfully rolled out!

» Releasing...
✓ Kubernetes client connected to https://... with namespace default
⠋ Updating service...
! Service "web" is invalid: [spec.ports[0].name: Required value,
  spec.ports[1].name: Required value]
```

With this configuration:

```hcl
  deploy {
    use "kubernetes" {
      probe_path = "/ping"
      replicas = 3
      ports = [
        {
          name = "http"
          port = 32020
        },
        {
          name = "https"
          port = 32021
        },
      ]
    }
  }

  release {
    use "kubernetes" {
      load_balancer = true
      ports = [
        {
          target_port = "http",
          port = 80,
        },
        {
          target_port = "https",
          port = 443,
        },
      ]
    }
  }
```

I tried a few combinations, but it seems to give a similar error when I add a second port.

briancain:
@rykov - Aha, thanks for bringing this up @rykov . Name is an optional field, but only when you define a single port. I've opened a pull request to fix this up, so it should be working in the next release: https://github.com/hashicorp/waypoint/pull/1184

PR #1093, opened 2021-02-16
Queue Project Job Op

mitchellh:
This introduces the queue project operation for queueing a job for all applications in a project.

This doesn't include testing because as I was writing tests I realized there is a big missing component to be able to unit test runners: mock plugins. That's a considerable amount of work so I commented what we need to do and I'd like to do that separately.

[WP-052](https://docs.google.com/document/d/1qbL3bPJl6s1J_F4G4hMzC3MY-2hS5LiwJ9HKkuU6Fyo/edit#)


internal/runner/accept.go
```diff
@@ -325,9 +328,7 @@ func (r *Runner) prepareAndExecuteJob(
 		if err == nil {
 			// Execute the job. We have to close the UI right afterwards to
 			// ensure that no more output is writting to the client.
-			log.Info("starting job execution")
```
mitchellh, line 328: I removed these log lines cause they're duplicated (the func that calls this does this same thing)

internal/runner/operation.go
```diff
@@ -130,7 +130,7 @@ func (r *Runner) executeJob(
 	//
 	// Note some operation types don't require downloaded data. These are
 	// not executed here but are executed in accept.go.
-	log.Info("executing operation", "type", fmt.Sprintf("%T", job.Operation))
```
mitchellh, line 133: Removed this log arg cause its now already present in all logging statements in this func (via `log.With` in caller)

PR #1094, opened 2021-02-16
Add input checking functions for both ALB and Route53 Zone ID

sshintaku:
Changed DNS, ALB, and Certificate HCL properties to "optional".  I added a function of ConfigNotify and run checks on the input in that function.  I also added unit tests for the various Listener, and Route53 API calls to ensure input is valid input.

### DIFF
--- a/builtin/aws/ecs/validator.go
+++ b/builtin/aws/ecs/validator.go
@@ -10,6 +10,34 @@ import (
 	"github.com/aws/aws-sdk-go/service/route53"
 )
 
+type MockRoute53Client struct {
+}
+
+func (c *MockRoute53Client) GetHostedZone(input *route53.GetHostedZoneInput) (*route53.GetHostedZoneOutput, error) {
+	var id string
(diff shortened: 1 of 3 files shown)


builtin/aws/ecs/validator.go
```diff
@@ -0,0 +1,56 @@
...
+
+type ALBListenerClient interface {
+	DescribeListeners(input *elbv2.DescribeListenersInput) (*elbv2.DescribeListenersOutput, error)
+}
+
+type Route53Client interface {
+	GetHostedZone(input *route53.GetHostedZoneInput) (*route53.GetHostedZoneOutput, error)
+}
+
+func IsValidArn(arn string) bool {
```
krantzinator, line 21: We don't need to export these

krantzinator (commented):
I realized I never hit submit on my feedback here, which I think is OK because we chatted in Slack a lot, but I'll go ahead and share it here for posterity 😅 
This is being addressed in #1401 though, so we can close this. Thanks for jumping in!

PR #1097, opened 2021-02-17
Add support for Logs plugin, implement AWS Lambda

evanphx:
Here we add the ability for a plugin to define `LogsFunc` and have it be used for read the logs about a Deployment.

This also includes an implementation of this functionality for AWS Lambda, reading the logs from CloudwatchLogs.

The Lambda functionality has a specific limitation, in that it's not really possible to wait and get more logs in a lightweight way. So this simple reads the current logs and nothing more.

This requires https://github.com/hashicorp/waypoint-plugin-sdk/pull/12 to be merged as well.


internal/ceb/virtual.go
```diff
@@ -333,3 +335,86 @@ func (v *Virtual) startExec(
 		}
 	}
 }
+
+// RunExec connects to the server and handles any inbound Exec requests via the
```
mitchellh, line 339: Outdated comment

internal/ceb/virtual.go
```diff
@@ -333,3 +335,86 @@ func (v *Virtual) startExec(
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case err := <-done:
+			return err
+		case lev, ok := <-logs:
+			if !ok {
+				return nil
```
mitchellh, line 392: I always find it helpful to comment what `ok` means. It is such a rarely used pattern that lots of juniors [to Go] don't realize what's happening here (and even with experience I sometimes forget the semantics). Would you mind adding some clarity here?

internal/server/proto/server.proto
```diff
@@ -923,6 +924,22 @@ message Job {
 
   message ExecResult {}
 
+  // Used to start a platform's exec function within a runner. This is only used
+  // there are no long running instances for a deployment and can fail if the
+  // platform plugin does not provide an exec function.
```
mitchellh, line 929: Do you mind adding to this plugin that callers interested in logs should use the `LogStream` (or w/e its called) API instead of this and this is only meant for internal usage.

internal/server/proto/server.proto
```diff
@@ -923,6 +924,22 @@ message Job {
...
+    // Id to assign the virtual instance created
+    string instance_id = 1;
+
+    // The deployment to create the exec session context. Ie, what
+    // application code will be available within the exec session.
+    Deployment deployment = 2;
+
+    google.protobuf.Timestamp start_time = 3;
+
+    int32 limit = 4;
```
mitchellh, line 940: Can you document fields 3 and 4?

internal/server/singleprocess/service_logs.go
```diff
@@ -1,38 +1,203 @@
...
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
+	"github.com/hashicorp/waypoint/internal/server/grpcmetadata"
+	"github.com/hashicorp/waypoint/internal/server/logbuffer"
 	"github.com/hashicorp/waypoint/internal/server/singleprocess/state"
 )
 
 // defaultLogLimitBacklog is the default backlog amount to send down.
 const defaultLogLimitBacklog = 100
 
+func (s *service) spawnLogPlugin(
```
mitchellh, line 22: I think we have some decent tests around log streams. It'd be great to add tests for this use case too. I'm not sure how easy that is actually but I'd love to talk through it given it is important behavior and also because I don't want to negatively impact our existing log style.

internal/server/singleprocess/state/app_operation.go
```diff
@@ -687,6 +717,7 @@ type listOperationsOptions struct {
 	Status        []*pb.StatusFilter
 	Order         *pb.OperationOrder
 	PhysicalState pb.Operation_PhysicalState
+	WatchSet      memdb.WatchSet
```
mitchellh, line 720: Would love to see tests around this use case for both the start and complete time indexes.

internal/server/singleprocess/state/instance_exec.go
```diff
@@ -108,39 +108,9 @@ func (s *State) InstanceExecCreateForVirtualInstance(ctx context.Context, id str
...
-
-		// Wait for the instance to show up
-		if err := ws.WatchCtx(ctx); err != nil {
-			if ctx.Err() != nil {
-				return ctx.Err()
-			}
-
-			return err
-		}
+	_, err := s.InstanceByIdWaiting(ctx, id)
```
mitchellh, line 111: Since this is only used internally, can we make it unexported for now?

mitchellh (changes requested):
Added various comments.

I'm good with the `InstanceLogs` approach for now and since it is memdb only. This makes it easy to make any changes we need in the future since there are no migration issues.

internal/ceb/virtual.go
evanphx: Yup!

internal/ceb/virtual.go
evanphx: Yeah, totally gotcha.

internal/server/proto/server.proto
evanphx: 💯

internal/server/singleprocess/service_logs.go
evanphx: Ack! Got too eager on the first day back after the weekend. Meant to write this, doing so now.

internal/server/proto/server.proto
evanphx: I gotchu

internal/server/singleprocess/state/app_operation.go
evanphx: Donzo!

internal/server/proto/server.proto
```diff
@@ -1992,6 +2015,9 @@ message GetLogStreamRequest {
 
     // Logs for a specific application in a workspace.
     Application application = 2;
+
+    // Logs for a specific instance.
+    string instance_id = 4;
```
mitchellh, line 2020: Sorry one more clarification: are these all required? Any optional/defaults?

internal/server/singleprocess/service_entrypoint.go
```diff
@@ -208,11 +208,24 @@ func (s *service) EntrypointLogStream(
 			// Read our instance record
 			instance, err := s.state.InstanceById(batch.InstanceId)
 			if err != nil {
-				return err
+				if status.Code(err) == codes.NotFound {
```
mitchellh, line 211: Can we add some logging around this so we can look in the logs to see whats happening?

mitchellh (changes requested):
One small set of changes

internal/server/singleprocess/service_entrypoint.go
evanphx: Done!

mitchellh (approved):

PR #1105, opened 2021-02-18
serverinstall/k8s: Use k8s client over execing kubectl

briancain:
This commit removes the instance of using kubectl in favor of just using
the k8s client for installing a secrets file.

### DIFF
--- a/internal/serverinstall/k8s.go
+++ b/internal/serverinstall/k8s.go
@@ -1,13 +1,11 @@
 package serverinstall
 
 import (
-	"bytes"
 	"context"
 	json "encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net"
-	"os/exec"
 	"strings"
 	"time"
 
@@ -140,11 +138,7 @@ func (i *K8sInstaller) Install(
 			return nil, err
 		}
 
-		var secretData struct {
-			Metadata struct {
-				Name string `yaml:"name"`
-			} `yaml:"metadata"`
-		}
+		var secretData apiv1.Secret
 
 		err = yaml.Unmarshal(data, &secretData)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/serverinstall/k8s.go
```diff
@@ -140,11 +138,7 @@ func (i *K8sInstaller) Install(
...
 		}
 
-		var secretData struct {
-			Metadata struct {
-				Name string `yaml:"name"`
-			} `yaml:"metadata"`
-		}
+		var secretData apiv1.Secret
 
 		err = yaml.Unmarshal(data, &secretData)
```
briancain, line 143: FYI, I'm not sure what's up with this but it isn't working exactly right. I'm testing with this config:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
stringData:
  username: hello
  password: password
```

Which works with the previous changes, but with this it looks like it's unmarshaling to nothing and `secretData` ends up empty. :thinking:

briancain: Looks like the potential issue here is we have to use a k8s client-go specific yaml parser to do this. I'll come back to this tomorrow.

briancain: Actually I have an update for this. will resolve.

internal/serverinstall/k8s.go
```diff
@@ -1,17 +1,15 @@
...
 	json "encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net"
-	"os/exec"
 	"strings"
 	"time"
 
-	"gopkg.in/yaml.v2"
+	"github.com/ghodss/yaml"
```
evanphx, line 12: Is there a reason we changed yaml packages here?

evanphx (changes requested):
Is using another yaml package intended?

internal/serverinstall/k8s.go
briancain: Yeah, I resolved the comment for how I got here on this PR, but basically the client-go package for k8s needs its own special way to unmarshall YAML. This yaml package actually seems to do a decent job for what we want without any other major code changes from what it is now, which is just calling unmarshall on the byte array into a Secrets struct. Otherwise if that isn't agreeable, there's a more complicated way to do this with the k8s api package.

evanphx (approved):

mitchellh (approved):

PR #1109, opened 2021-02-18
Polling Support

mitchellh:
This introduces support for polling and triggering an `up` on any changes in Git.

The primary architecture of this is as follows: the `Project` structure maintains the polling settings. New APIs were added to the `state` package to grab the _next_ project to poll that returns a deadline (when to poll). A new single goroutine poller is now started by the server which grabs that next project to poll, waits until the deadline, then queues a poll job. There is some watchset handling in there to detect the case that the project settings changes or projects were added/removed.

**Note, especially for @evanphx:** I'm still unhappy with the datasource interface in general and still want to take a broad pass at that. For this PR, I just kind of forced in the interface functions required to make this work. This is all internal logic though and we only have one real implementation (Git) so I think we can fix this up later without affecting any users. 

I expect this to have bugs and imperfections! I did my best to test various cases. This PR is primarily about laying the major minimum implementation for this feature.


internal/server/singleprocess/poll.go
```diff
@@ -0,0 +1,138 @@
...
+		// Outcome (3)
+		log.Trace("queueing poll job")
+		resp, err := s.QueueJob(ctx, &pb.QueueJobRequest{
+			Job: &pb.Job{
+				// SingletonId so that we only have one poll operation at
+				// any time queued per project.
+				SingletonId: fmt.Sprintf("poll/%s", p.Name),
+
+				Application: &pb.Ref_Application{
+					Project: p.Name,
```
evanphx, line 99: My read of the logic is that it's possible for a `p` that is `nil` to make it this far, based on the check for `p` being `nil` on line 40. Do we need to be guarding `p`'s `nil`-ness more?

evanphx (changes requested):
Looks good, mostly questions around some comment clarity.

internal/cli/project_apply.go
```diff
@@ -422,6 +436,21 @@ func (c *ProjectApplyCommand) Flags() *flag.Sets {
 				"if the key requires a password to decode. This is not required if " +
 				"the private key doesn't require a password.",
 		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:    "poll",
+			Target:  &c.flagPoll,
+			Default: false,
+			Usage: "Enable polling. This is only valid if a Git data source is supplied. " +
+				"This will watch the repo for changes and trigger a remote 'up'",
```
briancain, line 445:
```suggestion
				"This will watch the repo for changes and trigger a remote 'up'.",
```

internal/datasource/git.go
```diff
@@ -31,6 +32,17 @@ type GitSource struct{}
 
 func newGitSource() Sourcer { return &GitSource{} }
 
+func (s *GitSource) RefToOverride(ref *pb.Job_DataSource_Ref) (map[string]string, error) {
+	gitRef, ok := ref.Ref.(*pb.Job_DataSource_Ref_Git)
+	if !ok {
+		return nil, fmt.Errorf("ref is not a git ref: %T", ref.Ref)
```
briancain, line 38: Does the `ref` here contain a proto status code that we should be passing along?

internal/serverinstall/k8s.go
```diff
@@ -711,189 +712,235 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
...
+		); err != nil {
+			ui.Output(
+				"Error deleting Waypoint deployment: %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return err
+		}
+
+		// wait for deletion to complete
+		err = wait.PollImmediate(2*time.Second, 10*time.Minute, func() (bool, error) {
```
briancain, line 751: I've noticed we end up waiting 10 minutes for a lot of things in the k8s platform. Is this a reasonable amount of time? Is 5 minutes not enough?

briancain (changes requested):
Exciting to see this come together! I haven't stepped through everything, but I left a couple of change requests and questions for now :smile:

internal/datasource/git_test.go
```diff
@@ -248,6 +281,87 @@ func TestGitSourceGet(t *testing.T) {
...
+			},
+			nil,
+		)
+		require.NoError(err)
+		require.NotNil(newRef)
+
+		latestRef = newRef
+	})
+
+	t.Run("with latest ref", func(t *testing.T) {
```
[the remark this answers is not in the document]
mitchellh: Gooood idea.

internal/datasource/git.go
```diff
@@ -280,6 +270,112 @@ func (s *GitSource) Get(
...
+		URLs: []string{source.Git.Url},
+	})
+
+	// List our refs, equivalent to git ls-remote
+	refs, err := remote.List(&git.ListOptions{Auth: auth})
+	if err != nil {
+		return nil, err
+	}
+
+	// Find the target for HEAD
```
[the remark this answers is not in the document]
mitchellh: Actually, this might be totally broken for some cases because we're just finding HEAD for the default branch when we probably need to look for the latest commit of the given ref for the project. Whoops!

internal/runner/data.go
```diff
@@ -27,28 +27,50 @@ func (r *Runner) downloadJobData(
...
+	return sourcer.Get(ctx, log, ui, source, r.tempDir)
+}
+
+// downloadJobData takes the data source of the given job, gets the data,
+// and returns the directory where the data is stored.
+//
+// This will also return a closer function that should be deferred to
+// clean up any resources created by this. Note that the directory isn't
+// always a temporary directory (such as for local data) so callers should
+// NOT assume this and delete data. Use the returned closer.
```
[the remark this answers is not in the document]
mitchellh: I split the function so this is probably just confused. I'll take a look again.

internal/server/proto/server.proto
```diff
@@ -982,6 +1009,46 @@ message Job {
...
+  // A poll operation can be queued even if a project has polling disabled.
+  // If a project has polling enabled, a manually queued poll operation will
+  // have no effect on the poll timer or intervals.
+  message PollOp {
+    // NOTE(mitchellh): In the future I think we could have options here
+    // to poll only specific workspaces or the full project or whatever.
+    // At the time of writing this, we only support single workspace anyways.
+  }
+
+  message PollResult {}
```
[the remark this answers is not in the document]
mitchellh: Yes I did want to do that, good catch.

internal/server/singleprocess/poll.go
mitchellh: It shouldn't be because a p of nil should always result in outcome 1 or 2. But that isn't clear, and we should probably put the conditional anyways to prevent crashes and log a huge error or something.

evanphx: Yeah, it's not clear looking at it. Probably best to add some guard code just in case.

internal/serverinstall/k8s.go
mitchellh: On Azure stuff like PVCs sometimes take that long. :|

internal/datasource/git.go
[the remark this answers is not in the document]
mitchellh: I'm going to propose fixing this in a follow-up PR (before 0.3.0) if that's okay. I'll follow on right after this.

internal/server/singleprocess/poll.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		ws := memdb.NewWatchSet()
+		p, pollTime, err := s.state.ProjectPollPeek(ws)
+		if err != nil {
+			// This error really should never happen. Instead of just exiting,
+			// we log it and just sleep a minute. Hopefully someone will notice
+			// the logs. We sleep for a minute because any error that happened
+			// here is probably real bad and is gonna keep happening.
+			log.Error("error during poll queuer, sleeping 1 minute", "err", err)
```
mitchellh, line 36: Update log message to note this is a bug.

internal/server/singleprocess/poll.go
```diff
@@ -0,0 +1,148 @@
...
+		log.Debug("queued polling job", "job_id", resp.JobId)
+
+		// Mark this as complete so the next poll gets rescheduled.
+		log.Trace("scheduling next project poll time")
+		if err := s.state.ProjectPollComplete(p, time.Now()); err != nil {
+			// This should never happen so like above, if this happens we
+			// sleep for a minute so we don't completely overload the
+			// server since this is likely to happen again. We want people
+			// to see this in the logs.
+			log.Warn("error marking project polling complete", "err", err)
```
mitchellh, line 143: Report a bug message

internal/server/singleprocess/poll.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-memdb"
+
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+// runPollQueuer starts the poll queuer. The poll queuer sleeps on and
+// schedules polling operations for projects that have polling enabled.
+// This blocks and is expected to be run in a goroutine.
+func (s *service) runPollQueuer(ctx context.Context, funclog hclog.Logger) {
```
mitchellh, line 17: add comment that only one invocation must be running at any given time

internal/server/singleprocess/service.go
```diff
@@ -124,9 +130,25 @@ func New(opts ...Option) (pb.WaypointServer, error) {
...
+	go s.runPollQueuer(s.bgCtx, log.Named("poll_queuer"))
+
 	return &s, nil
 }
 
+// Close shuts down any background processes and resources that may
+// be used by the service. This should be called after the service
+// is no longer responding to requests.
+func (s *service) Close() error {
+	s.bgCtxCancel()
```
mitchellh, line 148: Add a waitgroup to wait on pollqueuer (or other bg tasks) to exit.

evanphx (approved):

PR #1110, opened 2021-02-18
builtin/k8s: Surface pod failures on deploy

briancain:
Prior to this commit, if a pod failed to pull an image, the error
wouldn't surface to the user and it would just look like the deployment
was hanging. This commit fixes that by using the `ui` to surface the
known pod failure of not being able to pull a container image.

### DIFF
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -429,7 +429,9 @@ func (p *Platform) Deploy(
 				}
 
 				if cs.State.Waiting != nil {
-					if cs.State.Waiting.Reason == "ImagePullBackOff" {
+					// TODO: handle other pod failures here
+					if cs.State.Waiting.Reason == "ImagePullBackOff" ||
+						cs.State.Waiting.Reason == "ErrImagePull" {
 						detectedError = "Pod unable to access Docker image"
 						k8error = cs.State.Waiting.Message
 					}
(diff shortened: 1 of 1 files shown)


briancain:
Just a quick gif of what this change looks like. It doesn't end up repeating itself during the poll loop and only prints once. It'll spin there for ~10 minutes until it fails with a single warning.

Peek 2021-02-18 09-59

builtin/k8s/platform.go
```diff
@@ -429,7 +429,9 @@ func (p *Platform) Deploy(
 				}
 
 				if cs.State.Waiting != nil {
-					if cs.State.Waiting.Reason == "ImagePullBackOff" {
+					// TODO: handle other pod failures here
+					if cs.State.Waiting.Reason == "ImagePullBackOff" ||
+						cs.State.Waiting.Reason == "ErrImagePull" {
```
evanphx, line 434: Didn't the old version warn in the same way but actually catch more conditions? I'm confused.

evanphx (changes requested):

builtin/k8s/platform.go
briancain: The old version only looked at `"ImagePullBackOff"`, which happens after `"ErrImagePull"`. I'm not sure if they are exclusively similar, but to me the error we're actually trying to handle here is when it fails to pull the image. I'm not sure what you mean by catching more conditions on the old version, unless there's a change prior to this PR you're talking about :smile:

evanphx: oh I can't read. Nevermind!

evanphx (approved):

krantzinator (approved):

PR #1113, opened 2021-02-19
Remove jfrog references

mdeggies:
This removes all references to `hashicorp.jfrog.io` which has been replaced with a self-hosted registry, `docker.mirror.hashicorp.services`. The registry at `hashicorp.jfrog.io` will be deprecated on March 1, 2021.


mitchellh:
Thank ya

PR #1116, opened 2021-02-19
builtin/docker: Introduce Resources map for configuring cpu, memory

### LINKED ISSUE #608: Builtin: support docker deployment resource specifications

There is currently no way to specify resource limits for apps that use docker.

Support resource limits in the docker plugin. An example deployment stanza would look like this:

```hcl
deploy {
  use "docker" {
    service_port = "8080"
    cpu = "2"
    memory = "512"
  }
}

```

N/A

It would very useful for the docker swarm community but I also believe there is value in it for people just coming off Heroku(which takes care of everything) with a single docker host who would want to prevent a rog container consuming all their host's resources.

**Additional context**
N/A

briancain:
This commit introduces the ability to configure a containers resources
limits around cpu and memory when deploying with docker.

Fixes #608

### DIFF
--- a/builtin/docker/platform.go
+++ b/builtin/docker/platform.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"fmt"
@@ -15,6 +16,7 @@ import (
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/go-connections/nat"
+	goUnits "github.com/docker/go-units"
 	"github.com/hashicorp/go-hclog"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


builtin/docker/platform.go
```diff
@@ -503,6 +525,16 @@ deploy {
 		),
 	)
 
+	doc.SetField(
+		"resources",
+		"A map of resources to configure the container with, such as memory or cpu limits.",
```
krantzinator, line 530: nit
```suggestion
		"A map of resources to configure the container with, such as memory or cpu limits",
```

krantzinator (approved):

builtin/docker/platform.go
briancain: Looking at this, it seems like some of our fields have periods, and others don't. But ultimately the docs generators adds them to the end when they get generated for the website, so maybe I'll leave this one here? Otherwise it's kind of confusing if it adds it back.

PR #1119, opened 2021-02-19
builtin/aws/ecr: use the ecr mapper

krantzinator:
This does not need to be backported; it is only on `main`.
This resolves the error `argument cannot be satisfied: type *anypb.Any` received when using the `aws-ecr` registry.


evanphx (approved):
WHOOOOOPS.

PR #1120, opened 2021-02-19
check for nil logging opts; remove endpoint

krantzinator:
This fixes two bugs.
One was our lack of a nil check on `logging` options in waypoint hcl.
The other was the aws error message we receive for `awslogs-endpoint`. I've submitted a ticket via their docs page and will update if/when I hear back. Until then, I don't want to hold up this bug fix.

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -824,26 +824,36 @@ func createALB(
 	return lbArn, tgArn, err
 }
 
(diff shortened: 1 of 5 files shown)


briancain (approved):

evanphx (approved):

krantzinator:
I'm going to look at our docs generation and see if I can get `optional` blocks to move to the `Optional` part of the page before merging this in.

briancain (approved):

PR #1124, opened 2021-02-22
Extract app config loading (incl. dynamic config) to standalone package: internal/appconfig

mitchellh:
This extracts the application config loading logic from `internal/ceb` into a standalone package `internal/appconfig`. This package handles both static and dynamic config vars and supports a much more configurable/robust interface to use. You can see the new robustness in action with the changes to the VirtualCEB implementation to use this as well!

I should've done this from the beginning, its obvious w/ hindsight 20/20. There are now THREE things that need app config: CEB, Virtual CEB (for serverless ops), and Runners. Each have or had their own semi-implementation of this. They can now share a single implementation and use options to slightly tune behavior. I haven't updated runners in this PR though.

Logic sharing aside, this is also a necessary thing to do in order to also further extract Virtual CEB into another package if we want to perhaps make a cleaner version of #1118. 👯 

~~TODO but review anyways: The standalone package is going to make it soooooo much nicer to unit test app config. I will add that prior to merging this. However, at the moment I've retained all the tests that test that behavior in `internal/ceb` so it is being tested for both static and dynamic configs.~~ This is done.


internal/appconfig/watcher.go
```diff
@@ -23,25 +24,245 @@ import (
...
+				select {
+				case <-ctx.Done():
+					// Wake up all condition vars so we wake ourself up.
+					w.currentCond.Broadcast()
+
+				case <-doneCh:
+					// Return
+				}
+			}()
+		}
```
evanphx, line 184: More a future observation that pull this functionality out to allow a condvar to integrate with channels would be good generic functionality.

internal/ceb/virtual.go
```diff
@@ -140,50 +144,49 @@ func (v *Virtual) RunExec(ctx context.Context, h VirtualExecHandler, count int)
...
+			continue
 		}
 
-		if msg.Config.EnvVars != nil {
-			static, dynamic = splitAppConfig(v.log, msg.Config.EnvVars)
+		// Get the config values, we use a iterator value of 0 here so that
+		// we will wait for at least one set of config values to resolve.
+		// In most virtual CEB cases we are only open for a single exec
+		// anyways so this resolves perfectly to our expected values.
+		env, _, err := w.Next(ctx, 0)
```
evanphx, line 160: So one thing I'm not sure of is if Next will return a value even if there are no config vars that were input. I could see a scenario where someone gets to this path without having set any config vars. I *think* the above `Update*` calls will prime it to generate an empty env to return for generation 1, but wanted to be sure.

evanphx (approved):
Just a couple questions, otherwise looks good.

internal/ceb/virtual.go
mitchellh: This is a really good point and no it won't. Let's gate on that.

PR #1128, opened 2021-02-22
Add missing Region value from Deployment

evanphx:
We need the region, it's used by the ALB to figure out where to set things up.


briancain (approved):
Nice

PR #1140, opened 2021-02-23
Honor the ALB config's port, but use the correct auto-calculated port

evanphx:
The automatically created security group is created incorrectly if we use the auto-calculated port value, but we were also not honoring the Port value that the ALB defined.


mitchellh:
Backport?

briancain (approved):

topfunky (approved):
Thanks!

PR #1142, opened 2021-02-24
internal/serverinstall: fix dropped errors

alrs:
This fixes both Kubernetes and Nomad dropped errors in `internal/serverinstall`.

### DIFF
--- a/internal/serverinstall/k8s.go
+++ b/internal/serverinstall/k8s.go
@@ -732,7 +732,14 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
 				LabelSelector: "app=" + runnerName,
 			},
 		)
+		if err != nil {
+			ui.Output(
+				"Error creating deployments watcher %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return err
 
+		}
 		// send DELETE to statefulset collection
 		if err = deploymentClient.DeleteCollection(
 			ctx,
@@ -790,7 +797,13 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
 				LabelSelector: "app=" + serverName,
 			},
 		)
-
+		if err != nil {
+			ui.Output(
+				"Error creating stateful set watcher: %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return err
+		}
 		// send DELETE to statefulset collection
 		if err = ssClient.DeleteCollection(
 			ctx,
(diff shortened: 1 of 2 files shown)


internal/serverinstall/k8s.go
```diff
@@ -721,6 +721,13 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
 			LabelSelector: "app=" + runnerName,
 		},
 	)
+	if err != nil {
+		ui.Output(
+			"Error watching deployments: %s", clierrors.Humanize(err),
```
briancain, line 726: This would actually just be an error initializing the watch variable. It doesn't watch the deployment until a bit further down

```suggestion
			"Error creating deployments watcher: %s", clierrors.Humanize(err),
```

internal/serverinstall/k8s.go
```diff
@@ -768,6 +775,9 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
 			LabelSelector: "app=" + serverName,
 		},
 	)
+	if err != nil {
+		return err
```
briancain, line 779: We should display a readable error message (like you did above) here before returning the err back. Similar to the other one, it would be an error creating a statefulset watcher.

internal/serverinstall/k8s.go
```diff
@@ -815,6 +825,9 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
...
```
briancain, line 829: Same feedback here but for PersistentVolumeClaim watcher.

internal/serverinstall/k8s.go
```diff
@@ -862,6 +875,9 @@ func (i *K8sInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error {
...
```
briancain, line 879: And here for Services watch.

internal/serverinstall/nomad.go
```diff
@@ -432,6 +432,9 @@ func (i *NomadInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error
 	s.Update("Removing Waypoint server from Nomad...")
 
 	_, _, err = client.Jobs().Deregister(serverName, i.config.serverPurge, &api.WriteOptions{})
+	if err != nil {
+		return err
```
briancain, line 436: This one should also have a readable err message before returning the err, but something like:

> "Failed to deregister waypoint server"

briancain (changes requested):
Thanks for these fixups! My main piece of feedback here is we should display a nice human readable error message prior to returning whatever error the k8s or nomad client libraries return to us similar to what you did for the deployments watcher error.

internal/serverinstall/k8s.go
alrs: Fixed.

alrs:
Hi, Brian.

> Thanks for these fixups! My main piece of feedback here is we should display a nice human readable error message prior to returning whatever error the k8s or nomad client libraries return to us similar to what you did for the deployments watcher error.

Hi, @briancain.

I've fixed the logline you've pointed out.

This package could use better logging throughout. I imitated the surrounding code when I added the `creating deployments watcher' logline.

My take is :+1: for better logging, but this is a bugfix PR from a volunteer, so product improvements should probably happen on a different pull request.

alrs:
My changes to the `k8s` code got taken out by a merge conflict, so I added some logging on the do-over.

briancain (approved):

PR #1146, opened 2021-02-25
builtin/k8s: Include user defined labels on deploy and add name/version labels

### LINKED ISSUE #966: Kubernetes plugin: Support injecting data from waypoint deploy into K8s labels
I would like to be able to talk meta information about the waypoint deploy and inject it into labels of the Kubernetes deployment. For clusters running istio expecting `app` and `version` labels, it will be helpful to have the waypoint app name as the `app` label (this can be done hard-coded for now) and the waypoint release version hash as the `version`. Right now, by default, we get a label like:

```
labels:
  name: example-app-01evrnw04bkevhzsfzk1x29n5t
  waypoint.hashicorp.com/id: 01EVRNW04BKEVHZSFZK1X29N5T
```

That label name is set here: https://github.com/hashicorp/waypoint/blob/main/builtin/k8s/platform.go#L169

I would like to be able to customize the labels on the resulting k8s deployment.

Istio can create `DestinationRule` objects using labels other than `version`, so I can use `waypoint.hashicorp.com/id` label as the target of the destination rule.

Another use case is that waypoint build information can be passed to the kubernetes deployment as well using these meta-labels. For example, the docker build could set labels on the container image that get exposed on the deployment as labels or annotations...such as buildTime, commit hash, etc.

briancain:
This pull request introduces a couple of changes and enhancements for a kubeernetes deploy:

- Includes any user defined labels to apply to the app pod on deploy from the deploy stanza in a waypoint.hcl
- Duplicates the label `name` to be `app`. This fits with how we label `waypoint-server` pod, and will fit what is expected for other services that integrate with k8s such as Istio https://istio.io/latest/docs/ops/deployment/requirements/#pod-requirements
- Duplicates the namespaced waypoint id label as `version`, because `version` is an expected key that k8s integrations expect to have (also used by Istio).

Fixes #966

### DIFF
--- a/builtin/k8s/deployment.go
+++ b/builtin/k8s/deployment.go
@@ -24,14 +24,14 @@ func (d *Deployment) newDeployment(name string) *appsv1.Deployment {
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"name": name,
+					"app": name,
 				},
 			},
 
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
-						"name": name,
+						"app": name,
 					},
 
 					Annotations: map[string]string{},
--- a/builtin/k8s/deployment.go
+++ b/builtin/k8s/deployment.go
@@ -21,6 +21,8 @@ func (d *Deployment) newDeployment(name string) *appsv1.Deployment {
 			Name: name,
 		},
 
+		// Note both name and app are included here. 'app' is expected for certain
+		// k8s integrations, where as waypoint expepcts 'name' else where for release
 		Spec: appsv1.DeploymentSpec{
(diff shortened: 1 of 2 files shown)


builtin/k8s/deployment.go
```diff
@@ -24,13 +24,15 @@ func (d *Deployment) newDeployment(name string) *appsv1.Deployment {
...
+					"app":  name,
 					"name": name,
 				},
 			},
 
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
+						"app":  name,
 						"name": name,
```
briancain, line 36: Not sure where this is happening exactly, but it seems like we use `name` some where for the local release URL to resolve so maybe we leave this as is and include app too?

krantzinator: That's a reasonable decision. I would add a comment to that effect for future-us.

krantzinator (approved):

PR #1161, opened 2021-03-04
Runner upgrade/uninstall machinery, default runner install to true

mitchellh:
This formalizes the runner management machinery: `waypoint install`, `waypoint server upgrade`, `waypoint server uninstall`, etc. RFC is [WP-053](https://docs.google.com/document/d/1im3U7iYtZLhKSjJKtqadNxJA0Ur3U30l70krz-b1WH4/edit#). 

This PR focuses on the machinery and **does not** implement Kubernetes or Nomad completely yet. I will make subsequent PRs for that. Docker is fully working since I used that for testing.

This PR also fixes a couple major bugs that manifested themselves and affect `main`:

  * Runner accept loops (in two places) handle runner deregistration and server restarts more gracefully, namely by not DDoSing the server.
  * Upgrade and uninstall do not start local servers, which prevents a local runner from starting up during these operations and DDoSing the server from the previous bug.
  * Docker uninstall/upgrade works if the container is not running, whereas previously it detected this as "not installed".

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -330,10 +330,10 @@ func (c *InstallCommand) Flags() *flag.Sets {
 		})
 
 		f.BoolVar(&flag.BoolVar{
-			Name:    "x-runner",
+			Name:    "runner",
 			Target:  &c.flagRunner,
 			Usage:   "Install a runner in addition to the server",
-			Default: false,
+			Default: true,
 			Hidden:  true,
 		})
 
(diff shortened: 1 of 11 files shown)


briancain (approved):
Looks good to me! The only major question I have here that I don't think this covers is how to handle installing runners outside of a server install (like in the case of an upgrade from an old server image where installing runners by default was not true).

mitchellh:
> Looks good to me! The only major question I have here that I don't think this covers is how to handle installing runners outside of a server install (like in the case of an upgrade from an old server image where installing runners by default was not true).

From the RFC for now:

> For users that want to add a runner to an existing installation, they must uninstall and reinstall Waypoint for now. They can restore from a snapshot after the reinstall to restore their data. This isn't an ideal behavior but it is the least intrusive to the UX of the CLI commands, I think.

But let's think about this further and maybe we can still do something before 0.3.

evanphx (approved):
dockers and kubes and contexts on my!

PR #1165, opened 2021-03-05
Update readme with dependency instructions for the gen/ts command

gregone:
We had a few undocumented dependency requirements to build the server and the API ts definitions. This PR updates the UI README to make this process more consistent. I chose to keep them in the section rather than the `Prerequisites`, as one can start ramping up on the repo without those.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -57,7 +57,7 @@ Then run the authentication steps above again.
 
 if you've made API changes in `/internal/server` and want to use those on the frontend, you'll need to generate the type definitions again: 
 
-#### required dependencies for this build step
+#### Required dependencies for build step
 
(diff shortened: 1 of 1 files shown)


ui/README.md
```diff
@@ -57,6 +57,15 @@ Then run the authentication steps above again.
 
 if you've made API changes in `/internal/server` and want to use those on the frontend, you'll need to generate the type definitions again: 
 
+#### required dependencies for this build step
```
briancain, line 60:
```suggestion
#### Required dependencies for build step
```

briancain (changes requested):
Looks good, just a small fixup for the section title!

briancain (approved):

PR #1168, opened 2021-03-08
internal/server: backoff on URL service not available

mitchellh:
This introduces an exponential backoff (up to 60s max wait) when the URL
service can't be initialized. Without this, we were spinning at nearly
100% CPU trying to reconnect, and I'm sure that's causing quite a lot of
network load on the service as well.

CleanShot 2021-03-08 at 08 56 34@2x

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/bmatcuk/doublestar v1.1.5
 	github.com/boltdb/bolt v1.3.1
 	github.com/buildpacks/pack v0.11.1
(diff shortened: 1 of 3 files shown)


krantzinator (approved):
neat

mitchellh:
I removed the backport label cause 0.2.x actually doesn't have any of this retry logic!

briancain:
@mitchellh - Ahh, oops. Makes sense :+1:

briancain (approved):

PR #1169, opened 2021-03-08
API Input Validation Foundations

mitchellh:
This brings in the foundation for API input validation.

This copies in the `internal/pkg/validationext` package that we've extracted out of our other projects. This package helps to format validation errors in proto `status` formats using `InvalidArgument` and `fieldViolations` so callers can break down individual errors. This package also adds helpers for `ozzo-validation` to work nicely with deeply nested structs and proto-style "oneof" fields.

I've implemented workspace and project validation in this PR, but stopped there so we can PR the foundation and add more validation as we go.

An example of the format is below. This should be inspectable by UIs (@gregone) to provide nicer error formatting:

CleanShot 2021-03-08 at 10 12 17@2x

### DIFF
--- a/internal/server/ptypes/workspace.go
+++ b/internal/server/ptypes/workspace.go
@@ -1,0 +1,37 @@
+package ptypes
+
+import (
+	"github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/imdario/mergo"
+	"github.com/mitchellh/go-testing-interface"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/waypoint/internal/pkg/validationext"
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+func TestGetWorkspaceRequest(t testing.T, src *pb.GetWorkspaceRequest) *pb.GetWorkspaceRequest {
+	t.Helper()
+
+	if src == nil {
+		src = &pb.GetWorkspaceRequest{}
+	}
+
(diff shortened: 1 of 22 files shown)


internal/server/ptypes/workspace.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/go-ozzo/ozzo-validation/v4"
+	"github.com/imdario/mergo"
+	"github.com/mitchellh/go-testing-interface"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/waypoint/internal/pkg/validationext"
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+func TestGetWorkspaceRequest(t testing.T, src *pb.GetWorkspaceRequest) *pb.GetWorkspaceRequest {
```
briancain, line 13: Should this func go in `workspace_test.go`?

briancain (approved):
Nice. Didn't test this fyi, but the changes make sense to me. Happy to see some validation! :heavy_check_mark:

internal/server/ptypes/workspace.go
mitchellh: Nah part of our public API is exposing helpers to get valid values, i.e. this one and `TestJob` and so on.

evanphx (approved):
🤖 . o O (/FINALLY VALIDATIONS/)

PR #1172, opened 2021-03-08
internal/datasource: Git polling respects Ref configuration

mitchellh:
The previous polling implementation only looked at HEAD. Users may
configure a separate Ref for their Git datasource, but the polling
didn't respect this.

This commit changes the polling to respect the Ref setting. The ref
setting can be in any format `git rev-parse` takes but only hash
references; caret refs and so on such as `HEAD~2` are not allowed.

### DIFF
--- a/internal/datasource/git.go
+++ b/internal/datasource/git.go
@@ -291,41 +291,89 @@ func (s *GitSource) Changes(
 		URLs: []string{source.Git.Url},
 	})
 
+	// Determine our target ref. If no Ref is specified by the user,
+	// we default to HEAD which points to whatever the default branch is.
+	targetRef := source.Git.Ref
+	if targetRef == "" {
+		targetRef = "HEAD"
+	}
+
 	// List our refs, equivalent to git ls-remote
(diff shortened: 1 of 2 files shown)


internal/datasource/git.go
```diff
@@ -291,49 +291,87 @@ func (s *GitSource) Changes(
...
-	for _, ref := range refs {
-		if ref.Name() == headTarget {
-			headRef = ref
-			break
+		// If the ref is a symbolic reference, we dereference until we find
+		// the target. An example here is HEAD may point to refs/heads/main,
+		// and so on.
+		for ref != nil && ref.Type() == plumbing.SymbolicReference {
+			ref = refMap[ref.Target()]
+		}
```
evanphx, line 349: It's unlikely, but if someone wanted to crash a waypoint server, they could configure a symbolic ref loop here and we'd be stuck forever. Maybe just throw the equiv of unix's symlink depth check in here.

evanphx (changes requested):
Minor adversarial change requested.

evanphx (approved):

PR #1176, opened 2021-03-09
Reformat CHANGELOG github issue refs

briancain:
This commit updates the changelog so that the github refs are actually
linkable/clickable to the user while on github.


briancain:
Hrm, doesn't look like any of the versions I've tried actually works here, so maybe this won't help us. I was checking by just going to `View File` to see the markdown.

PR #1178, opened 2021-03-09
Fix a couple minor bugs in SSH auth with Git

mitchellh:
This fixes two bugs that prevented SSH auth from properly working with Git:

* If no "user" is specified, use "git" (which is what every public Git host I found uses)
* We have to disable host key verification since we have no way to set that up in runners.

With these changes, I verified locally that polling setup to use SSH with GitHub works.


evanphx (approved):
Pesky users and host keys.

PR #1179, opened 2021-03-09
Introduce go-changelog to Waypoint

briancain:
This commit adds some essential template files for working with
`go-changelog`. It also includes a basic readme for how to generate the
proper CHANGELOG entries so that they can be generated with the tool.
Finally, it introduces a github workflow for checking when pull requests
forget to include a changelog file, with the option of being ignored if
the proper label is applied to the PR.

### DIFF
--- a/.changelog/README.md
+++ b/.changelog/README.md
@@ -4,6 +4,12 @@ Waypoint uses `go-changelog` to generate its changelog on release:
 
 * https://github.com/hashicorp/go-changelog
 
+To install, run the following command:
+
+```
+go get github.com/hashicorp/go-changelog/cmd/changelog-build
+```
+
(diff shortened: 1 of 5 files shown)


briancain:
Future TODO:

- Automate running `changelog-build` and inserting the output into Waypoints CHANGELOG in [waypoint-releases](https://github.com/hashicorp/waypoint-releases)

mitchellh:
I like it. I think the only suggestion I'd have is adding a make tasks perhaps or something for generating the CHANGELOG, then we can at least manually run that for now.

briancain:
Ooo, a make task is a good idea! I'll give that a go

krantzinator (approved):

PR #1182, opened 2021-03-11
remove excessive border decoration on links

gregone:
the way that the link underlines was applied changed from `text-decoration: underline` to using `border-bottom` in the latest version of the pds-addon, so our override wasn't applied anymore. This brings it back to normal. 

before: 

Screen Shot 2021-03-11 at 15 57 35

after: 

Screen Shot 2021-03-11 at 15 57 41

### DIFF
--- a/ui/app/styles/app.scss
+++ b/ui/app/styles/app.scss
@@ -96,7 +96,7 @@ ul {
   padding: 0;
 }
 
-:any-link {
+:any-link, :any-link:not(.pds-button) {
   border-bottom: none;
 }
 


dizzyup:
@gregone Good catch - also looks like that line of meta text also got changed to blue rather than the grey it previously used?

gregone:
@dizzyup  🤔 hm yeah good point. I'll fix that too

gregone:
@dizzyup updated the text back to the default color too
Screen Shot 2021-03-15 at 18 07 58

dizzyup (approved):

PR #1183, opened 2021-03-11
Allow poll settings to be configured in `runner` stanza

mitchellh:
This was a small asymmetry I noticed when working on docs: we allow every other project setting to be configured in the waypoint.hcl file but didnt allow polling. This extends support for that and updates the docs.


evanphx (approved):

briancain (approved):
:+1:

PR #1187, opened 2021-03-11
chore: adds should build script for website

BRKalow:
Adding a script to be run during the site build process to determine if we should actually build.

More information on the feature: https://vercel.com/docs/platform/projects#ignored-build-step

### DIFF
--- a/website/scripts/should-build.sh
+++ b/website/scripts/should-build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# This is run during the website build step to determine if we should skip the build or not.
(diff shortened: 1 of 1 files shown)


mitchellh:
Will this run on `main` for the tip website?

BRKalow:
Yeah, it will. In that case it'll check the diff of the `website` directory.

This is what we're currently doing, we've just added the `stable-website` check to ensure we don't miss any commits intended for production. 👍

mitchellh:
Ah gotcha gotcha, makes sense

jescalan (approved):
Looks good!

PR #1188, opened 2021-03-11
Validate in CLI and serverside that exec gets at least one arg

mitchellh:
One argument (the command to execute) is required. This validates it.

Without this, users would get exec just terminated with no message.


briancain (approved):
:heavy_check_mark:

PR #1189, opened 2021-03-11
Revert "Fixes https://github.com/hashicorp/waypoint/issues/687"

briancain:
Reverts hashicorp/waypoint#918

This is required to allow `waypoint up` to continue to work beyond the first iteration. Without this, the second `waypoint up` fails with an error that there is already a container bound for the port requested.

For now we'll go back to leaving it randomly assigned, which supports waypoints deployment model here. We have plans in a future release for supporting mutable deployments, which should hopefully allow the original issue to be resolved. /cc @Wakeful-Cloud

### DIFF
--- a/builtin/docker/platform.go
+++ b/builtin/docker/platform.go
@@ -166,7 +166,7 @@ func (p *Platform) Deploy(
 	bindings := nat.PortMap{}
 	bindings[np] = []nat.PortBinding{
 		{
-			HostPort: "",
+			HostPort: "", // this is intentionally left empty for a random host port assignment
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


builtin/docker/platform.go
```diff
@@ -166,7 +166,7 @@ func (p *Platform) Deploy(
 	bindings := nat.PortMap{}
 	bindings[np] = []nat.PortBinding{
 		{
-			HostPort: port,
+			HostPort: "",
```
briancain, line 169:
```suggestion
			HostPort: "", // this is intentionally left empty for a random host port assignment
```

evanphx (approved):

PR #1191, opened 2021-03-12
builtin/tfc: Add Terraform Cloud dynamic config sourcer

evanphx:
This PR adds the ability for Waypoint to populate configuration variables directly from Terraform Cloud. It's implemented as a config dynamic plugin, so it runs within the entrypoint binary.  This allows deployments to restart automatically when outputs within a Terraform Cloud workspace change.

An example here would be using this functionality to read a database password as an output and then have Waypoint applications automatically see any password rotations that take place and use those updated values quickly.

## Current Limitation

At present, Terraform Cloud only provides the value of outputs that are not sensitive. If you wish to use this feature, you'll need to remove the sensitive flag from any outputs. We're currently in discussion with the Terraform Cloud team about altering this behavior so that Waypoint can read all outputs.


builtin/tfc/config_sourcer.go
```diff
@@ -0,0 +1,478 @@
...
+				continue
+			}
+
+			defer resp.Body.Close()
+
+			if resp.StatusCode != 200 {
+				L.Error("error reading workspace info", "status-code", resp.StatusCode)
+
+				result.Result = &pb.ConfigSource_Value_Error{
+					Error: status.New(codes.Aborted, "fucked").Proto(),
```
briancain, line 238: We should change this to something else :sweat_smile: 
```suggestion
					Error: status.New(codes.Aborted, "failed to read workspace info").Proto(),
```

builtin/tfc/config_sourcer.go
```diff
@@ -0,0 +1,478 @@
...
+		),
+		docs.EnvVar("VAULT_SKIP_VERIFY"),
+	)
+
+	doc.SetField(
+		"refresh_interval",
+		"How often the outputs should be fetch.",
+		docs.Default("10m"),
+		docs.Summary(
+			"The format of this value is the Go time duration format. Specificly",
```
briancain, line 458:
```suggestion
			"The format of this value is the Go time duration format. Specifically",
```

builtin/tfc/config_sourcer.go
```diff
@@ -0,0 +1,478 @@
...
+		docs.Summary(
+			"This is not recommended unless absolutely necessary.",
+		),
+		docs.EnvVar("VAULT_SKIP_VERIFY"),
+	)
+
+	doc.SetField(
+		"refresh_interval",
+		"How often the outputs should be fetch.",
+		docs.Default("10m"),
```
briancain, line 456: We could make this the actual variable just to avoid any mistakes if the interval changes and the docs aren't updated?

```suggestion
		docs.Default(refreshPeriod.String()),
```

briancain (changes requested):
Awesome job! Looks great :smile: Got some minor fixup requests, but otherwise this looks legit :+1:

briancain (approved):
:tada: :tada: !

builtin/tfc/config_sourcer.go
```diff
@@ -0,0 +1,478 @@
...
+func (cs *ConfigSourcer) Documentation() (*docs.Documentation, error) {
+	doc, err := docs.New(
+		docs.FromConfig(&sourceConfig{}),
+		docs.RequestFromStruct(&reqConfig{}),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	doc.Description("Read configuration values from Terraform Cloud.")
```
mitchellh, line 373: This doesn't feel exactly right, or perhaps could be worded differently for TFC. "Read Terraform output values from Terraform Cloud." or something I dunno.

builtin/tfc/config_sourcer.go
```diff
@@ -0,0 +1,478 @@
...
+	)
+
+	doc.SetField(
+		"skip_verify",
+		"Do not validate the TLS cert presented by the Vault server.",
+		docs.Summary(
+			"This is not recommended unless absolutely necessary.",
+		),
+		docs.EnvVar("VAULT_SKIP_VERIFY"),
+	)
```
mitchellh, line 451: I think this is wrong. :)

mitchellh (changes requested):
Looks very god!

mitchellh (approved):

PR #1193, opened 2021-03-12
internal/server/singleprocess/state: Add ability to prune jobs and deployments

evanphx:
This limits deployments and jobs to 10,000 records each. The code only prunes the data from the memory indexes, it leaves in the disk.

The pruning of data is done on a regular interval rather than inline during data inserts (ie, it's like a typical garbage collection process). The pruning logic is shared between app records and jobs as well.

Pruning the data from disk would come later.

### DIFF
--- a/internal/server/singleprocess/state/app_operation.go
+++ b/internal/server/singleprocess/state/app_operation.go
@@ -100,8 +100,9 @@ func (op *appOperation) register() {
 	schemas = append(schemas, op.memSchema)
 
 	if op.MaximumIndexedRecords > 0 {
-		pruneFns = append(pruneFns, func(memTxn *memdb.Txn) (int, error) {
-			return op.pruneOld(memTxn, op.MaximumIndexedRecords)
+		pruneFns = append(pruneFns, func(memTxn *memdb.Txn) (string, int, error) {
+			cnt, err := op.pruneOld(memTxn, op.MaximumIndexedRecords)
+			return op.memTableName(), cnt, err
 		})
 	}
 }
--- a/internal/server/singleprocess/state/app_operation.go
+++ b/internal/server/singleprocess/state/app_operation.go
@@ -61,9 +61,6 @@ type appOperation struct {
 	seq map[string]map[string]*uint64
 }
 
-// By default, we allow 10000 records
-const DefaultMaximumIndexedRecords = 10000
-
 // Test validates that the operation struct is setup properly. This
 // is expected to be called in a unit test.
 func (op *appOperation) Test(t testing.T) {
--- a/internal/server/singleprocess/state/app_operation.go
+++ b/internal/server/singleprocess/state/app_operation.go
@@ -41,6 +41,13 @@ type appOperation struct {
 	// Bucket is the global bucket for all records of this operation.
 	Bucket []byte
 
+	// The number of records that should be indexed off disk. This allows
+	// dormant records to remain on disk but not indexed.
+	MaximumIndexedRecords int
+
+	// Holds how many records we've indexed at runtime.
+	indexedRecords int
+
 	// seq is the previous sequence number to set. This is initialized by the
 	// index init on server boot and `sync/atomic` should be used to increment
 	// it on each use.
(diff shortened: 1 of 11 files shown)


mitchellh (commented):
* I like only pruning the indexes. That way if we mess anything up we aren't deleting _real data_. 
* I'd introduce total record counts as package vars to avoid the iteration in jobs since that seems pretty easy. 
* For the app ops, I feel like we should just use the cursor at all times rather than `ForEach`. The existence of the secondary unmarshal, index, seq number tracking, etc. logic makes me nervous. Seems easy to introduce a bug that only exists in one or the other "mode". 
* Similar to above with jobs, I'd add functions to `*Job` that is like `IsComplete() bool` rather than reproducing that state logic over and over where its easy to introduce a bug.

evanphx:
1. No prob adding total record counts 👍 
2. I'll dry up that iteration code 👍 
3. Totes, another good drying 👍

internal/server/singleprocess/state/app_operation.go
```diff
@@ -50,6 +61,9 @@ type appOperation struct {
 	seq map[string]map[string]*uint64
 }
 
+// By default, we allow 10000 records
+const DefaultMaximumIndexedRecords = 10000
```
mitchellh, line 65: I don't think we use this anywhere.

If we do, I don't mean to get too wordy, but I think we should try to do like `DefaultAppOperationMaxIndexedRecords` (ugh such Java), but just to be clear that this is for app ops only?

If we intend this to be for _anything_ this is fine but it should probably go in a diff file.

internal/server/singleprocess/state/prune.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// pruneOld uses the types in op to scan the table indicated and prune old records.
+// The op's check function can allow the process to skip records that shouldn't be
+// pruned regardless of their age.
+func pruneOld(memTxn *memdb.Txn, op pruneOp) (int, error) {
+	op.lock.Lock()
+
+	// Easy enough, just exit if we haven't hit the maximum
+	if *op.cur < op.max {
```
mitchellh, line 25: We should protect against `op.cur == nil`. One suggestion is maybe just add a `Validate` method to `pruneOp` and use ozzo-validate or something to just do some really basic checking across the board.

internal/server/singleprocess/state/state.go
```diff
@@ -121,6 +132,39 @@ func (s *State) Close() error {
...
+		}
+
+		records += cnt
+	}
+
+	s.log.Debug("Finished pruning data",
+		"removed-jobs", jobs,
+		"removed-records", records,
+		"op-tables", len(pruneFns),
+	)
```
mitchellh, line 161: Might be useful to dump this per-function maybe as a trace so we can see prunes by table.

internal/server/singleprocess/state/prune.go
```diff
@@ -0,0 +1,85 @@
...
+	indexArgs    []interface{}
+	max          int
+	cur          *int
+	check        func(val interface{}) bool
+}
+
+// pruneOld uses the types in op to scan the table indicated and prune old records.
+// The op's check function can allow the process to skip records that shouldn't be
+// pruned regardless of their age.
+func pruneOld(memTxn *memdb.Txn, op pruneOp) (int, error) {
```
mitchellh, line 21: I know you test the callers of this, but since this is such an important func and a really genuinely nice abstraction, I'd feel better and recommend unit testing this specific func too.

mitchellh (commented):
This did end up really nice. Great job. A few non-trivial comments but this is great stuff.

internal/server/singleprocess/state/app_operation.go
evanphx: I just removed it for now since we're not using it.

internal/server/singleprocess/state/prune.go
evanphx: Added Validate(), good call.

internal/server/singleprocess/state/state.go
evanphx: That's also reported now, per table.

internal/server/singleprocess/state/prune.go
evanphx: Great call, found a bug while testing `pruneOld` in isolation.

mitchellh:
If you rebase this on main the tests should be fixed 👍

mitchellh (approved):

Issue #1192, opened 2021-03-12
Possible memory leak in WP entrypoint binary


kfh:
WP server: 0.2.3
WP client: 0.2.3

I've chosen my wording carefully because we haven't done any form of _isolated profiling_ on the entrypoint but i'm instead basing this on our observations over time as heavy users of WP in general with and without the entrypoint injected into our containers. 

This should be seen in relevance with the memory draining issue reported [here](https://github.com/hashicorp/waypoint/issues/1163). By using the _soft limit_ memory draining is only delayed. Eventually the container will be killed. 

This is a slow burner, even using the _smallest_ possible instance type on Fargate(512mb)  is takes some days before it's drained completely. But the pattern is clear, even when the apps deployed are _only idling_ there is a constant rise in memory usage. As mentioned in the linked issue we have profiled our apps using _AWS Codeguru_ extensively to make sure we are not the source of this issue. The attached image illustrate our observations:

<img width="1745" alt="entrypoint" src="https://user-images.githubusercontent.com/186453/110916018-a2e06200-8318-11eb-925e-9339ba8accdd.png">

When we _opt out of injecting the entrypoint binary_ into our containers we have not seen any memory draining issues with our WP enabled applications.

mitchellh:
While we've been very careful to try to avoid something like this, I believe it. We'll have to do some investigation on what's going on here. But can I ask are you using any features such as app config, exec, logs access, etc.?

mitchellh:
I've been running a rather unscientific test in the background as I go through my day today. I'll update this post with any findings. For this test, I deployed into Docker locally since we're testing for an entrypoint binary leak which is hopefully reproducible regardless of platform, and Docker is easy for me to setup for test.

* Idle deployment for 2 hours: no memory usage change at all
  - Unlikely we have a memory leak in this scenario
  - Steady state container RAM for me at 30 MB
* Refresh the URL service URL every second. I did this for another 2 hours.
  - Container RAM seems to go from 60 MB to up to 120 MB or so before going back down
  - Assuming this is mostly GC.  

So far I'm not seeing any leaks. That doesn't mean they don't exist, of course, its just that maybe they're not obvious to trigger yet. I'll keep my refresh loop running for the day and see what happens...

mitchellh:
[...]

PR #1194, opened 2021-03-12
cmd/waypoint-entrypoint: dump heap profile on SIGUSR1

mitchellh:
This adds functionality to dump a heap profile (in pprof format) when the entrypoint receives a SIGUSR1. The profile is written to the system temporary directory with a timestamp. The exact path that was dumped is logged. 

I hope to use this to debug #1192 and any future issues.

In the future, the team mentioned it'd be really cool if the server can request this information from an entrypoint and we can potentially use that to debug weird deployments. I think that's a good idea, but this is a simple thing we can do now and backport safely so I'm opening this PR first.

### DIFF
--- a/cmd/waypoint-entrypoint/debug.go
+++ b/cmd/waypoint-entrypoint/debug.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"runtime/pprof"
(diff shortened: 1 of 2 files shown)


evanphx (approved):

briancain (approved):
:bug: :mag:

mitchellh:
This helped us find a leak so I will merge this for now until we improve it one day later.

PR #1196, opened 2021-03-15
internal/ceb: fix dropped errors

alrs:
This fixes dropped errors in `internal/ceb` and its subpackage `internal/ceb/virtualceb`.


mitchellh (approved):
Thanks this looks good.

PR #1198, opened 2021-03-15
Expose `HasRegistry` bool to builder plugins, Docker doesn't pull if it doesn't have to

mitchellh:
The main motivation for this PR is to enable a workflow where Waypoint is using an externally built image (such as from CI) without requiring `docker` or any privileges/namespaces even in a Dockerless environment (i.e. with `img`). 

Since `img` still requires user namespaces and user namespaces aren't enabled by default on most hosted K8S platforms, this enables a working Waypoint experience if the above two conditions are met, guaranteed. 

To make this work, I've exposed a named variable `HasRegistry` that builders can opt-in for (via the typical argmapper DI stuff, see the Docker implementation in this PR) to determine if there is a registry configured.

For Docker, if the CEB injection is disabled AND there is no registry, we don't pull at all.

### DIFF
--- a/builtin/docker/pull/builder.go
+++ b/builtin/docker/pull/builder.go
@@ -142,7 +142,7 @@ type buildArgs struct {
 }
 
 // Build
-func (b *Builder) Build(args *buildArgs) (*wpdocker.Image, error) {
+func (b *Builder) Build(args buildArgs) (*wpdocker.Image, error) {
 	// Pull all the args out to top-level values. This is mostly done
 	// cause the struct was added later, but also because these are very common.
 	ctx := args.Ctx
--- a/builtin/docker/pull/builder.go
+++ b/builtin/docker/pull/builder.go
@@ -58,7 +58,7 @@ func (b *Builder) Documentation() (*docs.Documentation, error) {
 	}
 
 	doc.Description(`
-Use an existing, pre-built Docker image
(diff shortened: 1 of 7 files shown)


builtin/docker/pull/builder.go
```diff
@@ -126,15 +131,25 @@ func (b *Builder) Config() (interface{}, error) {
...
+func (b *Builder) Build(args struct {
+	// We use the struct form of arguments so that we can access named
+	// values (such as "HasRegistry").
+	argmapper.Struct
+
+	Ctx         context.Context
+	UI          terminal.UI
+	Log         hclog.Logger
+	HasRegistry bool
+}) (*wpdocker.Image, error) {
```
evanphx, line 143: While nifty, I think we should probably pull this out as a named struct. The pattern of arg structs in grpc is well known and it's less surprising than seeing an anon struct in a params listing.

evanphx (changes requested):
While I do like love pushing the envelope, probably the arg should be a named structure.

evanphx (approved):

mitchellh:
So turns out this doesn't work at all with real plugins cause we don't support named primitive types across the plugin boundary. Working on that now in waypoint-plugin-sdk.

briancain (approved):

PR #1199, opened 2021-03-15
Use accessible values for toggle theme

gregone:
The default theme used for the toggle component wasn't passing accessibility standards. This updates the theme to have a higher contrast in the default state, and use the styleguide blue for the checked state. 

A next step would possibly be to re-export the addon with a fully customized style for consumption by the `pds-addon`, or more simply have a custom theme class, but this shoudl do for now.

toggle-colors

### DIFF
--- a/ui/app/components/app-form/project-settings.hbs
+++ b/ui/app/components/app-form/project-settings.hbs
@@ -142,8 +142,13 @@
             @showLabels={{true}}
             @name="git-sync-enabled"
             @value={{this.project.dataSourcePoll.enabled}}
+            @size="small"
             @onToggle={{fn (mut this.project.dataSourcePoll.enabled)}} as |toggle|>
-            <toggle.switch/>
+            <toggle.switch
+              @onLabel='git sync on'
+              @offLabel='git sync off'
+              @name="Git Sync"
+              />
 
             <toggle.label for="git-sync-enabled">
               <span class="pds-toggleLabel">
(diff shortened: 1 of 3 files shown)


johncowen (approved):
Looks fine to me, couple of incidental questions:

1. Should the toggle be this big? I guess so, but I'm not massively aware of your overall widget styling.
2. I was just wondering why you went with the `dehex` > `rgb` > `var` instead of just using the hex code/SASS var straight.

I suppose if I wanted to nitpick I'd say maybe consider using `background-color` to overwrite, otherwise you could risk overwriting more than you think (https://developer.mozilla.org/en-US/docs/Web/CSS/background#constituent_properties), but thats just an opinion from someone who's a stickler for avoiding CSS property shorthands 😄  So feel free to discount!

dizzyup:
We could probably drop the size of the toggle a bit - @gregone do we have any other UI in Waypoint that uses the toggle? Just want to make sure if we adjust the size it still works in those spots.

I's a little while since I looked at this CSS but if memory serves the `dehex` functions were there so we have the flexibility to pass an alpha value to our theme colours. That whole convention is a bit of a workaround and not *super* elegant, but gave us the functionality we needed to support alpha levels and support light and dark mode more easily.

gregone:
@johncowen for the colors, it's indeed what @dizzyup described. I could have used the variable directly indeed, because I'm not using the transparency.

I have updated the template to use the `small` version of the toggle:

Screen Shot 2021-03-16 at 14 26 03

PR #1206, opened 2021-03-17
UI: flash messages

gregone:
Adds a notification service to the Waypoint browser UI (based on https://github.com/poteto/ember-cli-flash) to display success/error messages after user actions

notifications
("Click Me" Button not included in this PR, obvs)


gregone:
@briancain That's a good question It seems like that was a deliberate choice made by the cloud team with design (and I can think of some reasons why). I'd be in favor of moving those to the upper-right corner, but maybe I'd like to hear @meirish's thoughts before doing so

meirish:
re: positioning - I think this is where all of our other products currently have them - vault, TFC, boundary, and hcp for sure - I'm not sure about nomad or consul (there's less writing to the backend in these UIs so the pattern isn't as prominent). I think we could revisit that in the future but it would be good to align cross-product.

briancain:
FYI @gregone you'll want to rebase from `main` to fix up the go-test failures.

ui/app/services/flash-messages.ts
```diff
@@ -0,0 +1,10 @@
+import FlashMessages from 'ember-cli-flash/services/flash-messages';
+
+export default class PdsFlashMessages extends FlashMessages {
+  remove(id) {
+    let message = this.queue.findBy('id', id);
+    if (message) {
+      message._teardown();
```
meirish, line 7: I don't think you'll need this yet (but maybe with a system type).

meirish (approved):
This looks good ! We currently have our settings so that duplicate messages (keyed on the message content) aren't rendered more than once. Not sure if that's something you're wanting, but it's an option that's available - I think that's done in the app config.

ui/app/services/flash-messages.ts
gregone: Oh correct. I've updated it .

PR #1213, opened 2021-03-18
Add ability to have nested fields in docs

evanphx:
This pulls in https://github.com/hashicorp/waypoint-plugin-sdk/pull/17 and adds the rendering for the new subfields values.

It also tries up the field rendering and makes it emit "clean" markdown by default.

### DIFF
--- a/.changelog/1213.txt
+++ b/.changelog/1213.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Support and render new documentation subfields
+```
--- a/website/content/partials/components/builder-docker-pull.mdx
+++ b/website/content/partials/components/builder-docker-pull.mdx
@@ -36,7 +36,7 @@ These parameters are used in the [`use` stanza](/docs/waypoint-hcl/use) for this
 
 The image to pull.
 
-This should NOT include the tag (the value following the ':' in a Docker image). Use `tag` to define the image tag.
+This should NOT include the tag (the value following the ':' in a Docker image). Use `tag` to define the  image tag.
 
 - Type: **string**
 
--- a/website/content/partials/components/configsourcer-vault.mdx
+++ b/website/content/partials/components/configsourcer-vault.mdx
@@ -54,11 +54,11 @@ The parameters below are used with `waypoint config set-source` to configure
 the behavior this plugin. These are _not_ used in `configdynamic` calls. The
 parameters used for `configdynamic` are in the previous section.
 
-#### Required Parameters
+#### Required Source Parameters
 
 This plugin has no required source parameters.
 
-#### Optional Parameters
+#### Optional Source Parameters
 
 ##### addr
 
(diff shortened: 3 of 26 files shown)


website/content/partials/components/builder-docker-pull.mdx
```diff
@@ -36,7 +36,7 @@ These parameters are used in the [`use` stanza](/docs/waypoint-hcl/use) for this
 
 The image to pull.
 
-This should NOT include the tag (the value following the ':' in a Docker image). Use `tag` to define the image tag.
+This should NOT include the tag (the value following the ':' in a Docker image). Use `tag` to define the  image tag.
```
briancain, line 39: Not related to this PR but looks like this has an extra space defined:

https://github.com/hashicorp/waypoint/blob/99d1d67aca31349d4edf84198916e6769ed996aa/builtin/docker/pull/builder.go#L91

website/content/partials/components/configsourcer-vault.mdx
```diff
@@ -82,7 +82,7 @@ If this is not set, Vault agent will not be used. This should only be set if you
 
 The authentication method to use for Vault.
 
-This can be one of: `aws`, `kubernetes`. When this is set, configuration fields prefixed with the auth method type should be set, if required. Configuration fields prefixed with non-matching auth method types will be ignored (except for type validation). If no auth method is set, Vault assumes proper environment variables are set for Vault to find and connect to the Vault server. When this is set, `auth_method_mount_path` is required.
+This can be one of: `aws`, `kubernetes`. When this is set, configuration fields prefixed with the auth method type should be set, if required. Configuration fields prefixed with non-matching auth method types will be ignored (except for type validation).  If no auth method is set, Vault assumes proper environment variables are set for Vault to find and connect to the Vault server. When this is set, `auth_method_mount_path` is required.
```
briancain, line 85: Hrm, not sure why this has an extra space? 🤔 Maybe the new line characters are ignored?

https://github.com/hashicorp/waypoint/blob/99d1d67aca31349d4edf84198916e6769ed996aa/builtin/vault/config_sourcer.go#L451-L459

evanphx: Hm, I saw that too. I'll check it out.

briancain (approved):

website/content/partials/components/builder-docker-pull.mdx
krantzinator: Looks like the space was removed in `builtin/docker/pull/builder.go` so maybe this needs generated still?

krantzinator (approved):

.changelog/1213.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:improvement
+internal/cli: Render new documentation subfields nicely
+builtin/aws/ecs: Render logging attributes as subfields as example
```
mitchellh, line 3: can we do `plugin/aws-ecs` here? We did `builtin/` for one release but I think that was a mistake, early on we were doing plugin which I think is more useful to passers-by.

mitchellh (approved):

.changelog/1213.txt
briancain: Also I believe go-changelog expects separate entries for each improvement:

https://github.com/hashicorp/waypoint/tree/main/.changelog#multiple-entries

At least according to what Consuls docs mentioned. Seems noisey if true 🤔

evanphx: Oh, is this NOT the same as we do commit messages?

mitchellh: Don't mind either way. I think plugins are the only weird one.

evanphx: @briancain Explained they're abstract user facing components.

briancain (approved):

PR #1216, opened 2021-03-18
internal/ceb: Add ability to signal process about config changes

evanphx:
This change allows the CEB to signal a process that there are config changes. Those config changes are written to disk at a fixed path where the process can read them.

The contents of the config is the configuration specific env vars, so the process can read them and simply iterate over them and update their own environment. We could in fact release a few language libraries to do this for folks if we wanted.


mitchellh:
Going to close this PR since we discussed this 😄 and you're going a diff route.

PR #1225, opened 2021-03-23
builtin/aws/alb: clamp alb name per aws limits

krantzinator:


builtin/aws/alb/releaser.go
```diff
@@ -51,6 +51,12 @@ func (r *Releaser) Release(
 		lbName = "waypoint-" + src.App
 	}
 
+	// We have to clamp at a length of 32 because the Name field to 
+	// CreateLoadBalancer requires that the name is 32 characters or less.
+	if len(lbName) > 32 {
+		lbName = lbName[:32]
```
briancain, line 57: Maybe `log.Debug` here that we had to clamp the LB name due to the character restriction, with the original one in the message?

.changelog/1225
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
briancain, line 1: Should be `1225.txt`

krantzinator: ah thanks

builtin/aws/alb/releaser.go
krantzinator: Good idea! We do this clamping in several places -- should I add debugs to all of those in separate PRs or just in this one you think?

briancain: Hrm, I don't mind it in here if they're all grouped into the same commit! I think that makes sense.

briancain (approved):
👍🏻 looks good to me!

PR #1230, opened 2021-03-24
Specify number of deployments to retain when pruning during release

### LINKED ISSUE #988: Customize pruning count for platform

Currently, the platform will prune all deployments that exist if there are more than 2 currently out there. I want to customize this behavior in a couple ways:

* Allow for configurability of this number - say keep three previous releases out there
* Allow for configurability of the behavior based on a releaser func

In the k8s plugin, this always leaves an old version of the deployment out there unless the deployment is explicitly destroyed. I want to be able to opt into the destroy capabilities automatically based on some criteria.

* Waypoint Releaser can define a bool function that will determine if old deployments should be pruned

* Manually destroy the previous deployment using `waypoint deployment destroy <id>`

* `waypoint release -repeat` could be used to determine if the previous deployments should be marked for pruning.
* Could put more complex behavior in the destroy functions if the pruning happens based off conditions.

mitchellh:
Fixes #988 

This adds two new CLI flags to `up` and `release`: 

* `-prune` to disable pruning completely (not recommended, defaults true to prune)
* `-prune-retain` to specify how many existing deployments to retain, defaults to "2".

The `-prune-retain` flag can be set to 0 to remove all previous deploys.

This only allows setting this via the CLI (or very raw API) today. In the future we should investigate exposing this more completely in waypoint.hcl files, the UI, etc. But this gets in a minimal solution to help the folks in #988.


website/content/docs/lifecycle/release.mdx
```diff
@@ -59,3 +59,33 @@ you deploy and it is immediately available to consumers of your application.
 
 In this scenario, it is expected that either the deployment platform makes
 the deployments available to traffic or this is handled externally.
+
+## Deployment Pruning
+
+During the release stage, Waypoint will prune unreleased deployments.
+By default, Waypoint keeps the past two deployments at all times for
```
evanphx, line 66:
```suggestion
By default, Waypoint keeps just one previous deployment for
```

internal/cli/releases_create.go
```diff
@@ -163,12 +168,30 @@ func (c *ReleaseCreateCommand) Flags() *flag.Sets {
...
+			Usage:   "Prune old unreleased deployments.",
+			Default: true,
+		})
+
+		f.IntVar(&flag.IntVar{
+			Name:   "prune-retain",
+			Target: &c.flagPruneRetain,
+			Usage: "The number of unreleased deployments to keep. " +
+				"If this isn't set or is set to any negative number, " +
+				"then this will default to 2 on the server. If you want to prune " +
```
evanphx, line 191:
```suggestion
				"then this will default to 1 on the server. If you want to prune " +
```

internal/cli/up.go
```diff
@@ -191,6 +200,24 @@ func (c *UpCommand) legacyUp(
...
+			Usage:   "Prune old unreleased deployments.",
+			Default: true,
+		})
+
+		f.IntVar(&flag.IntVar{
+			Name:   "prune-retain",
+			Target: &c.flagPruneRetain,
+			Usage: "The number of unreleased deployments to keep. " +
+				"If this isn't set or is set to any negative number, " +
+				"then this will default to 2 on the server. If you want to prune " +
```
evanphx, line 217:
```suggestion
				"then this will default to 1 on the server. If you want to prune " +
```

internal/runner/operation_release.go
```diff
@@ -32,7 +32,14 @@ func (r *Runner) executeReleaseOp(
 	// ahead of time so that fails fast.
 	var pruneDeploys []*pb.Deployment
 	if op.Release.Prune {
-		log.Debug("pruning requested, gathering deployments to prune")
+		// Determine the number of deployments to keep around.
+		retain := 2
+		if op.Release.PruneRetainOverride {
+			retain = int(op.Release.PruneRetain)
```
evanphx, line 38:
```suggestion
			retain = int(op.Release.PruneRetain) + 1 // add 1 to make this the total number
```

evanphx (changes requested):
Clarify the logic and wording around it being the unrelease deployments, not total deployments.

briancain (approved):
🍇 👍🏻

evanphx (approved):

PR #1246, opened 2021-03-30
Add option for initial probe delay, timeout, and failure threshholds

frostbyte73:
If you have a project that takes more than 30 seconds to initialize, currently the only way to deploy it is by modifying the liveness initial delay seconds through kubectl. This adds the option to the kube hcl

### DIFF
--- a/.changelog/1246.txt
+++ b/.changelog/1246.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/k8s: Add new probe configuration options
+```
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -562,7 +562,7 @@ type Config struct {
 	// made to the port.
 	ProbePath string `hcl:"probe_path,optional"`
 
-	// Probe details
+	// Probe details for describing a health check to be performed against a container.
 	Probe *Probe `hcl:"probe,block"`
 
 	// Optionally define various resources limits for kubernetes pod containers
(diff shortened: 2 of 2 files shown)


briancain (commented):
Hey there @frostbyte73 - thanks for the pull request! We should definitely add the option to configure this, but I think while we're here we should support the other various probe options too like `TimeoutSeconds` and `FailureThreshhold`. Ideally, I think these options should probably be grouped under their own `probe` stanza for kubernetes and their variable config names match the values they are configuring since that's how they're configured in K8s. I think the easiest way to accomplish this would be defining a `Probe map[string]int32` which would let users define the options like:

```hcl
  deploy {
    use "kubernetes" {
      probe_path = "/"
      probe = {
        initial_delay_seconds = 30
        timeout_seconds = 2
        failure_threshhold = 3
      }
    }
  }
```

Then if those options aren't defined in the map, we can use the defaults we already have been using.

I'm happy to take this PR over to support the rest of the probe options and use your commit as a starting point, but if you want and can spare the time this week let us know.

We want to get this in soon, so if you don't have the time no worries, I'm more than happy to finish it out later this week 😄 Thanks!

frostbyte73:
@briancain Added the other probe fields - I'll leave the rest to you if you want to change any of the docs or comments.
Thanks!

briancain:
Awesome @frostbyte73 ! Looks great, thanks again for taking the time! ❤️ I'll finish this out this week. Appreciate it!

briancain (approved):
Awesome! This looks great to me now. I went ahead and fixed up the docs to use subfields. I think I won't add the other possible Probe options until we need them, since we're currently not using them at all when configuring k8s. So this looks good to go! Thanks again for the pull request 😄

PR #1249, opened 2021-03-31
waypoint/internal/server/singleprocess: fix dropped test error

alrs:
This fixes a dropped error in `waypoint/internal/server/singleprocess`.


mitchellh:
Thanks, you're finding all of these :)

PR #1253, opened 2021-04-02
UI: hcl config input from project settings

gregone:
https://user-images.githubusercontent.com/1416421/113445410-2233f380-93f6-11eb-8b3e-7f6e149618ae.mp4

### DIFF
--- a/ui/app/components/app-form/project-settings.ts
+++ b/ui/app/components/app-form/project-settings.ts
@@ -8,6 +8,10 @@ import { action } from '@ember/object';
 import { Project, UpsertProjectRequest, Job, Application } from 'waypoint-pb';
 import parseUrl from 'parse-url';
 
+const FORMAT = {
+  HCL: 0,
+  JSON: 1,
+};
 class ProjectModel {
   name: string;
   applicationsList: [];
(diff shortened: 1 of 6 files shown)


mitchellh:
Cool! This looks good. Technically, you can have a remote waypoint.hcl even without connecting it to a Git repository, but I don't really see the use case for that today so maybe this is fine for now and we'll reassess in a future version.

ui/app/components/app-form/project-settings.ts
```diff
@@ -208,10 +221,20 @@ export default class AppFormProjectSettings extends Component<ProjectSettingsArg
     dataSource.setGit(git);
     ref.setDataSource(dataSource);
     ref.setDataSourcePoll(dataSourcePoll);
+
+    if (this.serverHcl && this.project.waypointHcl) {
+      let hclEncoder = new window.TextEncoder();
+      let waypointHcl = hclEncoder.encode(this.project.waypointHcl);
+      // Hardcode hcl for now
+      ref.setWaypointHclFormat(0);
```
pearkes, line 229: Seeing `this.project.waypointHclFormat` being set to `0` up above (with the constant) should this be using that var instead of setting the value here?

pearkes (approved):
Looks great. One question but not a blocker, may be intentional for now.

ui/app/components/app-form/project-settings.ts
gregone: updated it to use a `FORMAT` constant for clarity

pearkes: Nice makes sense.

PR #1254, opened 2021-04-02
Make all deploy variables available in the DefaultReleaser setup

mitchellh:
This fixes two major bugs in core that have likely existed since 0.2.0:

First, when creating a _default releaser_, we need to ensure all the same HCL variables are available as during a `deploy {}` block render, since the deploy plugin is what gives us our default releaser.

Next, when setting up HCL variables, we wouldn't respect parent scopes and lose anything that wasn't in the innermost scope. This fixes our config parsing to properly respect all those scopes.

### DIFF
--- a/.changelog/1254.txt
+++ b/.changelog/1254.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: default releasers initialize properly when they use HCL variables
+```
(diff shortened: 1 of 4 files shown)


briancain (approved):
👍🏻 nice find! 🔍

PR #1272, opened 2021-04-07
Project name is optional for remote operations (and does not need to match, either)

mitchellh:
I noticed the UI had this little note:

image

And I didn't like it! The behavior when the projects didn't match was weird (it'd just _create_ a new project of that name on run) and this requirement to have the UI field exactly match something in the waypoint.hcl just was a poor UX.

This PR makes it so that the project field is optional for remote operations. If it is specified, the project name that invoked the operation _always takes precedence_. So even if the waypoint.hcl says `project = "foo"`, if you create a project in the UI named `my-website`, the operations and such will be associated with `my-website` and "foo" will be totally ignored.

You must still specify a `project` attribute for local operations, local `waypoint init`, etc.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -21,7 +21,7 @@ type Config struct {
 }
 
 type hclConfig struct {
-	Project string            `hcl:"project,attr"`
+	Project string            `hcl:"project,optional"`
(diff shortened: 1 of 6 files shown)


briancain (approved):
👍🏻

Issue #1264, opened 2021-04-07
ui: Support default field for git branch if none supplied


briancain:
**Is your feature request related to a problem? Please describe.**

When adding new git settings to a Waypoint project in the UI, it requests a ref to be supplied and can't be blank or use a default.

![Screenshot from 2021-04-06 11-40-29](https://user-images.githubusercontent.com/810277/113792032-ccaf5d80-96f9-11eb-92f4-87368368f235.png)

**Describe the solution you'd like**

 We should support a default value of `HEAD` if left empty so that it isn't required.

mitchellh:
Just tested HEAD (to verify it works) and it does, I think that's a good default cause it'll work no matter what the default branch is.

PR #1273, opened 2021-04-07
Default git ref to HEAD

gregone:
Addresses #1264

- defaults field value to `HEAD`
- if ref is deleted and left empty, API query sets it to `HEAD` anyway


briancain (approved):
Excellent! 🎉

Issue #1268, opened 2021-04-07
ui: git settings don't clear on new project


mitchellh:
I don't know if this is a bug, it actually worked really nicely for me cause I _did_ want to create a project pointing to the same place. But when you create a project, then create another project, the Git setting values are retained.

Bug? Not a Bug?

![image](https://user-images.githubusercontent.com/1299/113792640-3d0aae80-96fb-11eb-8f8b-ad2ddbf9aefa.png)

gregone:
definitely a bug. I had caught something similar coming from another route, but this one will be fixed by https://github.com/hashicorp/waypoint/pull/1275

PR #1275, opened 2021-04-07
refetch model on route reentry

gregone:
Fixes https://github.com/hashicorp/waypoint/issues/1268

Previously, the model would cache on the route, and not update properly, this fixes it by refetching when entering the route.


mitchellh:
I don't think this needs a changelog entry, since this never was released. Please remove that and this LGTM.

gregone:
@mitchellh yeah that makes sense 😅 . rebased and removed

mitchellh (approved):

PR #1276, opened 2021-04-07
builtin/pack: output to user if Docker is not available

mitchellh:
Pack requires a Docker daemon and we can't currently fallback to any
Dockerless solutions. This error message makes it clearer to users
earlier what is happening versus a more opaque "Docker daemon can't
connect" error that Docker provides later.


briancain (approved):
Excellent! 🎉

PR #1282, opened 2021-04-07
Add status badges for releases and deployments

### LINKED ISSUE #1269: ui: failed deployment doesn't show a failure on the app index page
When a deployment fails, the UI looks like it succeeded on the app index page, but if you click into the deployment you can clearly see that it failed. Builds show a red X when they fail so deployments probably should too?

image

See the video linked below:

https://user-images.githubusercontent.com/1299/113793234-a212d400-96fc-11eb-8038-7146f797db1c.mp4

gregone:
Fixes #1269 

Screen Shot 2021-04-07 at 21 25 58


mitchellh (approved):
Ah nice!

Issue #1262, opened 2021-04-06
ui: Can't read text input in Dark Mode


mitchellh:
See the screenshot 😄 

![image](https://user-images.githubusercontent.com/1299/113791611-e8feca80-96f8-11eb-9b0e-25dc1a36c3d9.png)

PR #1284, opened 2021-04-07
Quickfix for input contrast in dark mode

gregone:
Fixes https://github.com/hashicorp/waypoint/issues/1262 

We'll do a better fix with `pds-addon` later

Screen Shot 2021-04-07 at 22 17 09
Screen Shot 2021-04-07 at 22 17 00


mitchellh (approved):
That'll work for now. Thanks

briancain (approved):

PR #1286, opened 2021-04-07
Makefile: Use host network for pulling packages when building docker

briancain:
My docker engine seems to occasionally have issues pulling packages when building waypoint server.

```
#16 5.647 fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/community/x86_64/APKINDEX.tar.gz
#16 5.647 WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.13/main: temporary error (try again later)
```

Using my host network resolves this, so adding a `--network` flag to the `docker/server` make target. If this isn't desired we can close this, but otherwise I'll have to keep this stashed locally since that seems to be the only reliable way I can fix this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ docker/server:
 					--secret id=ssh.config,src="${HOME}/.ssh/config" \
 					--secret id=ssh.key,src="${HOME}/.ssh/config" \
 					-t waypoint:dev \
+					--network host \
 					.
 
 .PHONY: docker/evanphx


krantzinator:
My only pause is that it'd be nice to keep our local builds as close to our release style builds as possible, for reproduceability reasons. Did this start happening recently, or is it a forever kind of thing?

briancain:
Just started recently @krantzinator. This week actually. It's hit or miss when it works without it. Not sure if we'll experience the same issue in CI when it builds since it's also linux.

mitchellh:
I'm on Linux and I never have issues for what its worth...

briancain:
I'll close it for now then and keep the commit around for me 😄 hopefully a docker update will resolve it in the future

PR #1294, opened 2021-04-08
internal/pkg/epinject: preserve CMD when injecting with img

mitchellh:


Dockerfile
```diff
@@ -47,9 +47,8 @@ RUN apk add --no-cache \
 
 RUN git clone https://github.com/mitchellh/img.git /img
 WORKDIR /img
-RUN git checkout pull-config
 RUN go get github.com/go-bindata/go-bindata/go-bindata
-RUN make BUILDTAGS="seccomp noembed" && mv img /usr/bin/img
+RUN make BUILDTAGS="seccomp noembed dfrunmount dfsecrets dfssh" && mv img /usr/bin/img
```
mitchellh, line 51: This just updates to mainline img (still on my fork but rebased) and requires these tags for secret and ssh to work, although we don't use those features yet.

mitchellh:
This worked in my demo so gonna merge!

PR #1296, opened 2021-04-08
website: fix algolia indexing

BRKalow:
This is a temporary fix until we roll out some queued up package updates behind the branding rollout.


mitchellh:
Thank you!

PR #1298, opened 2021-04-08
Introduce Generations for Mutable Deployments

### LINKED ISSUE #1235: Mutable Deployment Support
Waypoint has an opinionated workflow: build, deploy, release. This opinionated workflow is what enables us (or will enable us) to build features such as per-deploy URLs, promotion, canaries, etc. 

The downside of the opinionated workflow is that it doesn't work very well with pre-existing deployment tooling that a user may have already setup. Users can't just bring their existing Nomad jobfiles, Kubernetes YAML, or other setups because most of these tools follow a pattern of updating existing resources vs. Waypoint's repave model. 

This significantly increases the burden of adopting Waypoint because users who want to "just try" Waypoint are forced to rethink how their deployment works. Additionally, at Waypoint's early age, it is common that Waypoint just doesn't support the features they need (such as all the various fields in a Kubernetes or Nomad job). 

We can't make these patterns "just work" via plugins today because Waypoint's core data model and logic has an implicit expectation that each deployment is repaved. Core data model changes are necessary to support a mutable deployment model.

Supporting existing Nomad jobfiles and features is one of our most frequently requested features both on GitHub issues and the Discuss forums. And not supporting "plain old" Nomad jobfiles sends a strange message to HashiCorp products working better together. This is something we'd like to support.

Note that we don't intend Waypoint to support any and all brownfield tooling (i.e. VM management with configuration management may be out of the question) but we do want users who are using relatively modern tooling (such as Nomad, K8S YAML, Kustomize, Helm, etc.) to be able to adopt Waypoint with relative ease.

Make the core Waypoint data model "mutable-aware" by enhancing operations to support a "generation ID". Plugins can opt-in to generating their own generation IDs. By default, Waypoint will generate a unique generation ID per operation (assuming immutable workflows as we've always done).

A "generation" represents a single set of resources that have been created or reused for an operation. Matching generation IDs mean that the same set of resources are in use. 

**Notes**
This should also address #567, #1125, and #516

mitchellh:
Fixes #1235 

This introduces the core concept of a "generation." Deployments can optionally advertise a "generation". Two operations that operate on the same generation expect that underlying resources were mutated rather than replaced. This enables plugins such as Nomad jobspec where the job specification updates a deployment in place. This will in the future enable support for raw Kubernetes apply, Helm charts, etc. as well.

The behavior of matching generations in a nutshell:

* All matching generations share the same deployment URL.
* When a generation is not new, `waypoint up` shows output to let the user know the URL will be the same and it was updated in place.
* When pruning old deployments, deployments with a matching generation with the release are ignored (since they share underlying resources).
* When releasing a deployment, it is a no-op if the generation matches the currently released deployment (since they should point to the same place).
* Generation information is available via the API directly on the operation.
* Generations are UUIDs.


internal/server/ptypes/deployment.go
```diff
@@ -0,0 +1,24 @@
...
+	// we have a generation set, we use the generation initial sequence.
+	seq := v.Sequence
+
+	// By using the generation sequence, we ensure that all deployments
+	// in a generation share the same URL.
+	if g := v.Generation; g != nil {
+		seq = g.InitialSequence
+	}
+
+	return "v" + strconv.FormatUint(seq, 10)
```
briancain, line 23: For mutable deploys with a generation ID, should we append something like `latest` after `v<generation-id>` onto the URL to show they will share the same URL for future deploys?

```
Deployment URL: https://namely-adapting-poodle--v1-latest.waypoint.run
```

Or maybe even just `latest` instead of `v1` if it never changes?

briancain: I could see where a plugin might do in-place deploys but want to increment the generation id, so maybe something like `v<gen-id>-latest` would work?

mitchellh: Generation IDs are long UUIDs though so that'd be a pretty wild URL I think.

briancain: Ahh, ok. Yeah maybe not then 😅

briancain (approved):
I had been using this after testing the nomad jobspec and k8s apply plugins. Looks good to me and works as expected.

Do you plan on opening a PR for the SDK changes before merging these?

evanphx (approved):
cb80dd8b-4a88-476a-a82b-1e1e71ade13b-8cb64c44-6a64-4f62-9642-460dd37039d6-v1

So clean!

internal/core/operation.go
```diff
@@ -54,15 +56,15 @@ func (a *App) doOperation(
...
-	hooks := op.Hooks(a)
-
 	// Init the metadata
 	msg, err := op.Init(a)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// Get our hooks
+	hooks := op.Hooks(a)
```
krantzinator, line 66: Did this need to move because of the added functionality to reinit components, or is this a "this should have been here all along" thing?

krantzinator (approved):
🥳

gregone:
Thins looks great! Just a question on this: for easier debugging, should we output the generation UUID in the CLI success message?

mitchellh:
> Thins looks great! Just a question on this: for easier debugging, should we output the generation UUID in the CLI success message?

The UUIDs aren't currently useful for anything (they don't show up in the UI, you can't look things up with them via the API, etc.) so at the moment I think no. I do think showing the UUIDs in some way for the UI, or some other visual marker or grouping that an operation is all part of the same generation, would be helpful. I think this may require some API changes but happy to do so and discuss that.

PR #1307, opened 2021-04-09
Fix GRPC TLS connect

denysvitali:
When a new context is created and `-server-tls` is set to true, but `-server-tls-skip-verify` is false (default configuration!), the TLS options are not added to the GRPC dial options, causing (in my case) the following error:

```
$ waypoint context verify
 ! Connecting with context "dev"...
! Error connecting with context "dev": grpc: no transport security set (use grpc.WithInsecure() explicitly or set
  credentials)
```

With this PR, the behavior is fixed and we're now able to connect to a TLS backend without a problem:

```
___waypoint_context_verify context verify
 + Context "dev" connected successfully.
```

### DIFF
--- a/internal/client/project.go
+++ b/internal/client/project.go
@@ -97,7 +97,7 @@ func New(ctx context.Context, opts ...Option) (*Project, error) {
 		client.activeRunner = r
 
 		// We spin up the job processing here. Anything that spawns jobs (either locally spawned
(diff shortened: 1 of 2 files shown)


mitchellh:
Great catch! I'll add a changelog entry for this and get it backported. Thank you.

denysvitali:
Thanks for the quick merge and for the awesome product! Keep up the good work 🚀

PR #1313, opened 2021-04-09
internal/core: render all output variables for destroy phases

mitchellh:
This builds on #1254 which wasn't quite complete it turns out...

We need to also ensure that we render all the variables for workspace
destruction and "destroy all" operations.

This also changes the behavior slightly so that we render the Waypoint
config using variables based on the _latest_ deploy or release.
Previously, we'd use the deploy/release we were destroying. The issue is
that the configuration we have at the moment is only the current config,
so you could get waypoint.hcl parse errors if they used variables that
didn't exist at one point or another.

Practically, I'd like to update the SDK docs to note that during a
destroy operation you only have the config for the current config rather
than a point-in-time configuration. You should persist all state you
need onto your deployment/release structure. I think that's how all
plugins work anyways today but I'll update docs there to be explicit.

I'm _not_ going to mark this for a backport because its a fairly significant change, even though the bug does exist on 0.3. Its far less likely you'd hit it on 0.3 since you'd have to be heavily using templating.

### DIFF
--- a/internal/core/app_deploy.go
+++ b/internal/core/app_deploy.go
@@ -9,6 +9,8 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/zclconf/go-cty/cty"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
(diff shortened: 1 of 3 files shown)


krantzinator (approved):
panic free!

Issue #1310, opened 2021-04-09
1st destroy operation with a Lambda and ALB fails


jbayer:
**Describe the bug**
I needed to run the `waypoint destroy` operation twice with the example Lambda project as the 1st time results in an error.

**Steps to Reproduce**
Run the Lambda example and then attempt to destroy it at the end. I encountered an error the first time. Running it again a few moments later resolves the issue.

```
$ waypoint destroy -auto-approve

» Destroying releases for application 'learn-waypoint-lambda'...

» Destroying deployments for application 'learn-waypoint-lambda'...
! Error destroying: rpc error: code = Unknown desc = ResourceInUse: Target group 'arn:aws:elasticloadbalancing:us-west-2:770746137682:targetgroup/learn-waypoint-lambda-01F2VJ9T48/c8fbab83b6ae4ee4' is currently in use by a listener or a rule
  	status code: 400, request id: 7f977192-8349-43bb-98d0-37e3fc057e9c
$ waypoint destroy -auto-approve

» Destroying shared deploy resources for application 'learn-waypoint-lambda'...
✓ Connecting to AWS
✓ Deleting Lambda function
⠼ IAM role deleted
Destroy successful!
```

**Expected behavior**
The destroy operation works the 1st time.

**Waypoint Platform Versions**
* Waypoint CLI Version: 0.3.0
* Waypoint Server Platform and Version: 0.3.0 on k8s
* Waypoint Plugin: aws-lambda

jbayer:
It also appears that the ALB and associated Security Group did not get cleaned up.

![Screen Shot 2021-04-09 at 8 49 53 AM](https://user-images.githubusercontent.com/1139532/114206869-9dbf0480-9910-11eb-80b2-d43405aa90b6.png)

PR #1315, opened 2021-04-11
[WIP] builtin/aws: graceful destroy of ALB/Lambda target group

vancluever:
This ensures that ALB-related resources get destroyed properly during
Lambda teardown by:

* Adding Destroyer to the ALB releaser.
* Adding a rudimentary waiter to the target group removal in the Lambda
destroyer.

The latter ensures that the target group can be removed successfully as in
addition to removing the dependent ALB as it can take a moment for the
ALB removal/detachment to fully register.


vancluever:
WIP: fixes #1310

Just wanted to get the WIP in, will be reviewing/cleaning up code/fixing tests shortly.

vancluever:
Whittled down the code to just the minimum needed to do the destroy. Abandoned the AWS waiters that I had in there for now as the waiters did not really seem to be adding much at all; in fact as far as I could see the waiters were passing before the target group could actually be deleted.

Still need to add tests and other things like:  1) detaching instead of deleting the ALB when the ALB was supplied externally in config versus being created, and deleting other resources created by the deploy process as mentioned by @jbayer in #1310 (in addition to addressing other detail like Waypoint flat out skipping the resources in the event of initial deletion failure).

vancluever:
Most of the resources that are in the ALB deployment should be accounted for now.

To keep track of things, I've added some rudimentary state to alb.Release. I added it on an "as-needed" basis - it may not be complete or as sophisticated as it could be.

One thing I think that this is underscoring is the need for some sort of resource lifecycle management primitives in Waypoint for situations like this. Perhaps something lightweight within the SDK or core, similar to Terraform - or maybe Terraform itself eventually!

vancluever:
> One thing I think that this is underscoring is the need for some sort of resource lifecycle management primitives in Waypoint for situations like this. Perhaps something lightweight within the SDK or core, similar to Terraform - or maybe Terraform itself eventually!

Just a note on this - this might be something that we could take advantage of the [Terraform CDK](https://github.com/hashicorp/terraform-cdk) for once Go support lands.

briancain:
Hey there @vancluever - Just a quick note, making the waypoint plugin sdk resource aware is on the roadmap for 0.4.0 Be on the lookout for an RFC soon.

vancluever:
@briancain good to know! Considering that, maybe I should just focus this on the teardown of the resources for now (just tests, etc)?

I haven't had a chance to investigate yet what's up with the fact that `waypoint destroy` seems to skip over things that it fails on destroying, as described in #1310, but is that possibly related to the lack of resource-awareness right now?

briancain:
@vancluever good question! I'm not sure why that's happening and I'm not super familiar with the lambda plugin 😅 But yeah, I imagine having our plugins be more resource aware will hopefully resolve issues like this 🙏🏻 

If adding a bit more state isn't too intrusive right now for the lambda plugin it's probably safe to keep it. We can modify it to use the plugin enhancement once that's in place.

vancluever:
@briancain I'm going to mark this ready for review shortly, main reason being is that writing acceptance tests for this has opened up another whole can of worms. See https://github.com/vancluever/waypoint/commit/c0a1dc8adcbe44f94b08299435f1d9ed08633a92 to see what I mean.

Long story short, there's a lot of work and code that we need to do to get the framework in place to set up each scenario and test it in a graceful way, and then clean up the resources afterwards. Basically like another version of the Terraform acceptance testing framework, if you will.

Definitely something I think we should work on, but I just want to ensure that this PR isn't a untied loose end for too long. 🙂

briancain:
@vancluever sounds legit! Waypoint def needs more testing, and something like that sounds useful for us. I had some pretty basic level end to end tests that _might_ make sense to live alongside some higher level acceptance tests if you're going to be running against real resources: https://github.com/hashicorp/waypoint/tree/main/test-e2e

Those tests are expected to run against a real waypoint binary and server so they're maybe a level up from the tests you shared...but I'm imagining these kinds of tests might share a similar CI pipeline separate from our main tests once we set it all up.

vancluever:
@briancain thanks for sending along those tests! Yeah I think those e2e tests are definitely more higher level as they are testing the binary as a whole and not just SDK-related components. There's probably a place for both, especially with the various kinds of CLI/client/server/runner deployment scenarios we operate under (by the way, did the website with that move? I'm having trouble finding it, but it's also the morning so 😆 )

Without going on too much of a tangent, I could definitely see a test suite residing in the SDK that targets mostly plugins, and possibly runs some part of the runner and core that are needed to manage strictly the components that are needed to tie together builds, deployments, and releases so that tests can be shipped and run in `go test`, possibly can be made debuggable, etc.

Definitely would love to explore this more 🙂

catsby:
Hello there - an alternative to this `Destroy` implementation was added in https://github.com/hashicorp/waypoint/pull/1648 using the new [Resource Manager](https://github.com/hashicorp/waypoint-plugin-sdk/tree/eda7ae600c2d69ba042ecac4d6d9333d37f6fd9d/framework/resource). I think we can close this then, unless there's a resource we're missing in the current destroy?

vancluever:
@catsby guessing that's fine. Let me know as I'm just cleaning up some old forks in my personal space, so just confirming before this code goes away completely.

catsby:
We're good to close this, thanks @vancluever !

PR #1318, opened 2021-04-12
{serverinstall/builtin}/nomad: Provide config values for memory and CPU

### LINKED ISSUE #1302: Remote runner job for Nomad clusters OOMs when doing basic tasks
Pretty simple: I tried using the new runner created by `waypoint install` as a remote runner, pulling a git repo. The runner got OOM-killed before it even finished pulling the repo down.

Not sure what the solution is - configurable resource limits? Maybe just generate some nomad jobspecs that I can modify with resource stanzas appropriate for my setup?

briancain:
This commit introduces config options to use when defining a Server and
Runner nomad job config. If no option is specified, it will use the
configured defaults for Waypoint. This commit also increases the default
CPU in MHz to 200 and Memory in MB to 600 (double the default for Nomad
tasks)

It also introduces resources stanza for nomads deploy. Unlike the install for a server and runner, it sticks to the default resource size.

Fixes #1302

### DIFF
--- a/.changelog/1318.txt
+++ b/.changelog/1318.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+plugin/nomad: Add CPU and Memory resource options for server and runner installs
+```
--- a/.changelog/1318.txt
+++ b/.changelog/1318.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-plugin/nomad: Add CPU and Memory resource options for server and runner installs
+plugin/nomad: Add CPU and Memory resource options for server and runner installs, and app deploys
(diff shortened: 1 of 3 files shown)


mitchellh:
This looks good! I think the thing we should check is if this is enough. We can run our Lambda GitOps demo through this, I can do that soon.

briancain:
@mitchellh - perfect! Thanks. Let me know and I'll update the default vars in the PR

nateberkopec:
LGTM!

krantzinator (approved):
resource config looks great!

PR #1333, opened 2021-04-14
Makefile: Add target for generating server proto

briancain:
This also makes the command for generating the server proto more
discoverable.


krantzinator (approved):
🎉

mitchellh (approved):

PR #1334, opened 2021-04-14
Add action to generate ts and open PR on changes to internal/server in main

gregone:


gregone:
After discussion with the team, closing this and moving the typescript definitions generation to Circle instead.

PR #1335, opened 2021-04-14
Project settings polish

dizzyup:
Wanted to do a little work on the form layout/design for project settings.

Also fixes a bug I noticed where we were mistakenly removing the bottom border on buttons, and a regression where we lost our tab styles in some of the onboarding steps.

#### New project page
CleanShot 2021-04-14 at 18 24 56@2x

#### Project settings page
CleanShot 2021-04-14 at 18 24 31@2x


mitchellh:
Let's also backport this puppy when it becomes time. 🐶

gregone (approved):
✨ looks great with the `card` sections! You'll need to add a changelog entry before merging

PR #1340, opened 2021-04-14
add some documentation about autocomplete install and uninstall

koesbong:
This provides some documentation about the autocomplete install and uninstall command and how it changes the shell init script

### DIFF
--- a/website/content/commands/index.mdx
+++ b/website/content/commands/index.mdx
@@ -5,9 +5,14 @@ page_title: 'Waypoint Commands (CLI)'
 
 # Waypoint Commands (CLI)
 
-Waypoint is controlled via a very easy to use command-line interface (CLI). Waypoint is only a single command-line application: `waypoint`. This application then takes a subcommand such as `artifact` or `deployment`. The complete list of subcommands is in the navigation to the left.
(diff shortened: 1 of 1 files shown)


krantzinator (approved):

website/content/commands/index.mdx
```diff
@@ -68,3 +68,5 @@ Operation Options:
       specified to the data source type being used in your configuration. This
       is used for example to set a specific Git ref to run against.
 ```
+
+Running `waypoint -autocomplete-install` will add the waypoint autocomplete capability so you can do `waypoint <tab>` on commands. Running `waypoint -autocomplete-uninstall` will remove it. Please note that this will modify your shell init script, so you will need to reload your shell.
```
mitchellh, line 72: Nitpick but we like to keep our docs (and most of our files to 80 to 100 characters wide max). Newlines in markdown aren't new paragraphs unless there are two so this keeps the lines together still. :)

briancain (approved):

website/content/commands/index.mdx
koesbong: Got it. Fixed all the other lines too and verified that it looks good on Vercel deployment.

PR #1342, opened 2021-04-15
add gen/ts job to circle

gregone:


.circleci/config.yml
```diff
@@ -376,6 +376,40 @@ jobs:
...
+            # HEAD^! is shorthand for HEAD^..HEAD (parent of HEAD and HEAD)
+            if ! git diff --quiet --exit-code HEAD^! internal/server/proto/server.proto; then
+              git config --local user.email "hashicorp-ci@users.noreply.github.com"
+              git config --local user.name "hashicorp-ci"
+              short_sha=$(git rev-parse --short HEAD)
+              go generate ./internal/server
+              make gen/ts
+              git add ui/lib
+              git commit -m "auto-updated ui/lib/* from commit ${short_sha}"
+              git push origin circle-ts-gen
```
briancain, line 406:
```suggestion
              git push origin circle-ts-gen-${short_sha}
```

In the case where lingering PRs aren't addressed this might cause some headache if they all push to the same branch.

briancain: I think you need to checkout to this branch too before comitting/pushing?

briancain (changes requested):
Just had a few fixup suggestions!

One more general question I have, it seems like this might end up duplicating the server proto generation in the cases where the generated protos have already updated in the PR that touched `internal/server/proto/server.proto`. Can this also look at `https://github.com/hashicorp/waypoint/tree/main/internal/server/gen/{server.pb.go,server.pb.json.go}` and see if it was recently updated too before it makes a PR?

.circleci/config.yml
```diff
@@ -376,6 +376,40 @@ jobs:
...
+          name: gen/ts
+          command: |
+            # check if there are any changes in internal/server/server.go
+            # if there are, we commit the ui static asset file
+            # HEAD^! is shorthand for HEAD^..HEAD (parent of HEAD and HEAD)
+            if ! git diff --quiet --exit-code HEAD^! internal/server/proto/server.proto; then
+              git config --local user.email "hashicorp-ci@users.noreply.github.com"
+              git config --local user.name "hashicorp-ci"
+              short_sha=$(git rev-parse --short HEAD)
+              go generate ./internal/server
```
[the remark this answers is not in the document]
gregone: Does `make gen/server` accomplish the same?

.circleci/config.yml
gregone: yeah, I'll need to checkout a new branch and do that. I wasn't there yet. Mostly focused on getting the gen/ts step working first

briancain: Yep! Sorry, I jumped the gun a bit on my review! ❤️

.circleci/config.yml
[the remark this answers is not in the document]
briancain: Yep, I just added this yesterday to the Makefile, should be exactly the same!

https://github.com/hashicorp/waypoint/blob/64fc38f29dc0c1901e48f4a11f8751f1fefa42bd/Makefile#L71-L73

Issue #1228, opened 2021-03-23
Input variables


krantzinator:
**Is your feature request related to a problem? Please describe.**
Enable input variables for Waypoint.

**Describe the solution you'd like**
Users should be able to input variable values via a vars file or command line arguments.

**Explain any additional use-cases**
This feature will enable other requested enhancements around variable management within Waypoint.

krantzinator:
Closed by #1548

PR #1346, opened 2021-04-16
docker: support build args

psihachina:
This commit adds the support build args:

``` build {
        use "docker" {
                dockerfile="./simple_project/Dockerfile"
                build_args=["TEST_PATH=.", "BIN_PATH_2=./.bin/app"]
	}
    }
```

### DIFF
--- a/.changelog/1346.txt
+++ b/.changelog/1346.txt
@@ -1,3 +1,3 @@
 ```release-note:feature
-plugins/docker: Add support for build args
+plugins/docker: Add support for build args for both docker and img builders
 ```
--- a/.changelog/1346.txt
+++ b/.changelog/1346.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+plugins/docker: Add support for build args
+```
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -49,6 +49,9 @@ type BuilderConfig struct {
 
 	// The name/path to the Dockerfile if it is not the root of the project
 	Dockerfile string `hcl:"dockerfile,optional"`
+
+	// Controls the passing of build time variables
+	BuildArgs []string `hcl:"build_args,optional"`
 }
 
 func (b *Builder) Documentation() (*docs.Documentation, error) {
@@ -154,6 +157,15 @@ func (b *Builder) Build(
 	}
 	cli.NegotiateAPIVersion(ctx)
 
+	buildArgs := make(map[string]*string)
+
+	if b.config.BuildArgs != nil {
+		for _, arg := range b.config.BuildArgs {
+			pair := strings.Split(arg, "=")
+			buildArgs[pair[0]] = &pair[1]
+		}
+	}
+
 	dockerfile := b.config.Dockerfile
 	if dockerfile == "" {
 		dockerfile = "Dockerfile"
@@ -199,9 +211,10 @@ func (b *Builder) Build(
 		// fail but that error message should help the user.
 		step.Update("Docker isn't available. Falling back to daemonless image build...")
 		step.Done()
+		buildArgsString := strings.Join(b.config.BuildArgs, "--build-arg ")
 		step = nil
 		if err := b.buildWithImg(
-			ctx, ui, sg, relDockerfile, contextDir, result.Name(),
+			ctx, ui, sg, relDockerfile, contextDir, result.Name(), buildArgsString,
 		); err != nil {
 			return nil, err
 		}
@@ -217,7 +230,7 @@ func (b *Builder) Build(
 		step.Done()
 		step = nil
 		if err := b.buildWithDocker(
-			ctx, ui, sg, cli, contextDir, relDockerfile, result.Name(),
+			ctx, ui, sg, cli, contextDir, relDockerfile, result.Name(), buildArgs,
 		); err != nil {
 			return nil, err
 		}
@@ -273,6 +286,7 @@ func (b *Builder) buildWithDocker(
 	contextDir string,
 	relDockerfile string,
 	tag string,
+	buildArgs map[string]*string,
 ) error {
 	excludes, err := build.ReadDockerignore(contextDir)
 	if err != nil {
@@ -314,6 +328,7 @@ func (b *Builder) buildWithDocker(
 		Dockerfile: relDockerfile,
 		Tags:       []string{tag},
 		Remove:     true,
+		BuildArgs:  buildArgs,
 	})
 	if err != nil {
 		return status.Errorf(codes.Internal, "error building image: %s", err)
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -121,6 +121,12 @@ build {
 		),
 	)
 
+	doc.SetField(
+		"buildargs",
+		"An array of strings of build-time variables passed as build-arg to docker"+
+			"or img for the build step.",
+	)
+
 	return doc, nil
 }
 
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -276,9 +276,6 @@ func (b *Builder) Build(
 
 // Translates BuildArgs from a map of key vals into a string of '--build-arg key=vals'
 func createBuildArgsString(m map[string]*string) string {
-	if m == nil {
-		return ""
-	}
 	b := []string{}
 	for key, value := range m {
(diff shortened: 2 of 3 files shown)


psihachina:
Partially solves this problem #1228

psihachina:
@krantzinator Hello. I would like to receive feedback. What do you think? Can I add something?

AutomationD:
It would be great to get this merged. @briancain is there anything on the roadmap blocking this PR to get merged?

briancain:
Nothing blocking that I know of @AutomationD 😄 I'll get this reviewed soon by myself or someone from the team @psihachina , thanks for opening the PR!

briancain:
Hey @psihachina - I recommend fixing your branch and removing the `Merge branch` commit, what you had before was fine 👍🏻

psihachina:
@briancain yes, did  stupid 😄, but I seem to have fixed it

builtin/docker/builder.go
```diff
@@ -154,6 +157,15 @@ func (b *Builder) Build(
 	}
 	cli.NegotiateAPIVersion(ctx)
 
+	buildArgs := make(map[string]*string)
+
+	if b.config.BuildArgs != nil {
+		for _, arg := range b.config.BuildArgs {
+			pair := strings.Split(arg, "=")
+			buildArgs[pair[0]] = &pair[1]
```
briancain, line 165: It looks like we build up `buildArgs` here, but later only use `b.config.BuildArgs`? Maybe we could move this right above where `buildArgs` is used since we're only needing it conditionally later on after all the fallback if-statements.

briancain (changes requested):
Hey there @psihachina ! Thanks for opening a pull request for this. Had 1 comment about moving something, and also I'd like to make sure `BuildArgs` is documented for the website. I couldn't push to your branch to update the PR because of permissions, so if you could, please add this to the `docker/builder.go` file:

```
commit 4c3fbdf0f79094805abb2a1d7dcea2b9dbfde13c (HEAD -> DOCKER-BUILD-ARGS)
Author: Brian Cain <bcain@hashicorp.com>
Date:   Mon May 3 13:11:27 2021 -0700

    Include docstring of buildargs for docker plugin on website

diff --git a/builtin/docker/builder.go b/builtin/docker/builder.go
index 6b34c051..c3e21062 100644
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -121,6 +121,12 @@ build {
                ),
        )

+       doc.SetField(
+               "buildargs",
+               "An array of strings of build-time variables passed as build-arg to docker"+
+                       "or img for the build step.",
+       )
+
        return doc, nil
 }
```

Thank you! 😄

builtin/docker/builder.go
psihachina: @briancain, you mean it 

`

    } else {
		buildArgs := make(map[string]*string)
		if b.config.BuildArgs != nil {
			for _, arg := range b.config.BuildArgs {
				pair := strings.Split(arg, "=")
				buildArgs[pair[0]] = &pair[1]
			}
		}

		// No fallback, build with Docker
		step.Done()
		step = nil
		if err := b.buildWithDocker(
			ctx, ui, sg, cli, contextDir, relDockerfile, result.Name(), buildArgs,
		); err != nil {
			return nil, err
		}
	}
`

denysvitali:
I would suggest using a `map[string] string` instead of a `[]string` for the build args 👍

builtin/docker/builder.go
```diff
@@ -284,6 +281,14 @@ func (b *Builder) Build(
 	return result, nil
 }
 
+func createKeyValuePairs(m map[string]string) string {
+	b := new(bytes.Buffer)
+	for key, value := range m {
+		fmt.Fprintf(b, "--build-arg %s=%s ", key, value)
```
denysvitali, line 287: I would strongly suggest quoting the parameters (so that `ENV_NAME="some value"` doesn't break the build) and to use an array of strings + join instead of a buffer, so that you can easily join them:
```go
strings.Join([]string{"--build-arg x=y", "--build-arg a=b"}, " ")
```

In your case:
```go
b := []string{}
for key, value := range m {
    b = append(b, fmt.Fprintf(b, "--build-arg %s=%s", key, value))
}
return strings.Join(b, " ")
```

denysvitali (changes requested):
IMHO it still needs a little fix, but other than that LGTM

denysvitali (approved):
LGTM to me now! Thanks for the changes 🚀

builtin/docker/builder.go
```diff
@@ -265,6 +281,14 @@ func (b *Builder) Build(
 	return result, nil
 }
 
+func createKeyValuePairs(m map[string]string) string {
```
briancain, line 284:
```suggestion
// Translates BuildArgs from a map of key vals into a string of '--build-arg key=vals'
func createBuildArgsString(m map[string]string) string {
```

We should have a more specific name for this func since this is what it's actually doing 😄

builtin/docker/builder.go
```diff
@@ -265,6 +281,14 @@ func (b *Builder) Build(
...
+	b := []string{}
+	for key, value := range m {
+		b = append(b, fmt.Sprintf("--build-arg \"%s=%s\"", key, value))
+	}
+	return strings.Join(b, " ")
+}
```
briancain, line 290:
```suggestion
	if m == nil {
		return ""
	}
	b := []string{}
	for key, value := range m {
		b = append(b, fmt.Sprintf("--build-arg \"%s=%s\"", key, value))
	}
	return strings.Join(b, " ")
}
```

I _think_ this join ends up adding extra whitespace to be passed to the `img` command if there are no buildargs defined. So maybe we can guard this on if BuildArgs are even defined.

builtin/docker/builder.go
```diff
@@ -213,11 +222,18 @@ func (b *Builder) Build(
 		// We set this to true so we use the img-based injector later
 		useImg = true
 	} else {
+		buildArgs := make(map[string]*string)
```
briancain, line 225: I should of noticed this the first time I reviewed this PR, but why not make the config option `BuildArgs` a `map[string]*string` to begin with? I think this would save us a translation here. Sorry I didn't notice this the first time around!

.changelog/1346.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+plugins/docker: Add support for build args
```
briancain, line 2:
```suggestion
plugins/docker: Add support for build args for both docker and img builders
```

briancain (changes requested):
Hey there @psihachina , thanks for all the updates! I left some feedback based on the latest round of updates. Let me know if you have any more questions! Otherwise I'd say this is good to go 😄

builtin/docker/builder.go
denysvitali: Are you sure?
https://play.golang.org/p/0BkcWnLhyKJ

briancain: Hrm, I had reproduced it yesterday in go playground to do this, which is why I had made the comment. Can't seem to reproduce this behavior this morning so looks like I was wrong about the extra whitespace 😄 Sorry about that @psihachina 😅

psihachina: @briancain bring it back?

briancain: Yeah you can delete the

```golang
if m == nil {
 return ""
}
```

Part. Sorry about the confusion! ❤️ 
Otherwise this PR looks good to go 😄

builtin/docker/builder.go
```diff
@@ -265,6 +274,18 @@ func (b *Builder) Build(
 	return result, nil
 }
 
+// Translates BuildArgs from a map of key vals into a string of '--build-arg key=vals'
+func createBuildArgsString(m map[string]*string) string {
+	if m == nil {
+		return ""
+	}
```
briancain, line 281:
```suggestion
```

briancain: Sorry about this! I was wrong yesterday on this feedback. Gonna add a quick suggested change and commit it for simplicity, otherwise this PR looks good to go ❤️ 👍🏻

briancain (approved):

briancain:
Thank you @psihachina !! Looks great. Appreciate you taking the time to open this PR and thanks again for responding to all the PR feedback ❤️

PR #1357, opened 2021-04-19
plugin/kubernetes-apply: Apply a directory of YAML or JSON files

mitchellh:
This introduces a `kubernetes-apply` plugin that is able to deploy any typical directory of YAML or JSON files to Kubernetes. This builds on #1298 to enable this functionality.

**Note: in the future, we'll support jsonnet via a global function.**

Full functioning example to test this here: https://github.com/hashicorp/waypoint-examples/compare/kubectl-apply-nodejs

Example waypoint.hcl:

```hcl
project = "example-nodejs"

app "example-nodejs" {
  build {
    use "pack" {}
    registry {
      use "docker" {
        image = "example-nodejs"
        tag   = "1"
        local = true
      }
    }
  }

  deploy {
    use "kubernetes-apply" {
      // Template the directory so that we process each file in the directory
      // as a template. This lets us inject the artifact from the previous step.
      path = templatedir("${path.app}/k8s")

      // This label determines what resources we own. Any that aren't present
      // in our folder that match this label will be deleted.
      prune_label = "app=myapp"
    }
  }
}
```

Example `k8s/deployment.yaml`:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deployment
  labels:
    app: myapp
spec:
  replicas: 1
  selector:
    matchLabels:
      app: myapp
  template:
    metadata:
      labels:
        app: myapp
    spec:
      containers:
      - name: myapp
        image: ${artifact.image}:${artifact.tag}
        env:
          %{ for k,v in entrypoint.env }
          - name: ${k}
            value: "${v}"
          %{ endfor }
          # Ensure we set PORT for the URL service. This is only necessary
          # if we want the URL service to function.
          - name: PORT
            value: "3000"
```


builtin/k8s/apply/platform.go
```diff
@@ -0,0 +1,297 @@
...
+
+This plugin lets you use any pre-existing set of Kubernetes resource files
+to deploy to Kubernetes. This plugin supports all the features of Waypoint.
+You may use Waypoint's [templating features](/docs/waypoint-hcl/functions/template)
+to template the resources with information such as the artifact from
+a previous build step, entrypoint environment variables, etc.
+
+### Requirements
+
+This plugin requires "kubectl" to be installed since this plugin works by
```
anubhavmishra, line 167: Maybe add a note or instructions around Kubernetes config? Say users need to have the configuration present in order for `kubectl` to access it.

anubhavmishra (dismissed):

anubhavmishra:
@mitchellh I approved it by mistake! Just wanted to leave a comment in the review. 😄

builtin/k8s/apply/platform.go
mitchellh: Good call. Fixed!

builtin/k8s/apply/platform.go
```diff
@@ -0,0 +1,302 @@
...
+// Config is the configuration structure for the Platform.
+type Config struct {
+	// The path to the job specification to load.
+	Path string `hcl:"path,attr"`
+
+	// Prune label is the label to use to destroy resources that don't match.
+	PruneLabel string `hcl:"prune_label,attr"`
+
+	// KubeconfigPath is the path to the kubeconfig file.
+	KubeconfigPath string `hcl:"kubeconfig,optional"`
```
briancain, line 146: Should we also support specifying a kube context here so that people won't have to make sure their env is right before running waypoint up?

briancain (approved):
Worked for me based on the example linked! Just had a question about an extra config var

builtin/k8s/apply/platform.go
anubhavmishra: I like that idea! If we support it, folks can specify the exact context in the config they want to target!

builtin/k8s/apply/platform.go
```diff
@@ -0,0 +1,302 @@
...
+	defer sg.Wait()
+	s := sg.Add("")
+	defer func() { s.Abort() }()
+
+	deployment := &Deployment{
+		PruneLabel: p.config.PruneLabel,
+	}
+
+	s.Update("Executing kubectl apply...")
+	cmd, err := p.cmd(ctx, s, "apply", "-f", p.config.Path, "--prune", "-l", deployment.PruneLabel)
```
RealHarshThakur, line 57: It would be nice if  `-R` flag can be included, so that if there are any sub-directories, they can be applied too. Probably something similar during `Destroy`

builtin/k8s/apply/platform.go
evanphx: Totes agree, we should have the ability to specify (or at least pick up via env) a kube context.

evanphx (changes requested):
Just needs the kube context bits!

builtin/k8s/apply/platform.go
RealHarshThakur: I'm not sure about this, but would it be a good idea to let users pass in any arguments kubectl supports. For instance, I can imagine someone wanting to supply `--server-side` flag during creation. During delete, user might want to supply the `--wait` flag.

RealHarshThakur (commented):
Hey, I'm new to the project and would like to contribute. 

Other than the comments below, is there an issue open for something like `waypoint plan`? Kubectl has  `--dry-run` which would fit there

builtin/k8s/apply/platform.go
mitchellh: Great idea. Added!

builtin/k8s/apply/platform.go
mitchellh: The `-R` flag is a good idea.

Not sure about the second one because it'd force us to be tied to `kubectl` instead of perhaps using just an API client one day (i.e. if server side apply becomes predominantly the way to go).

RealHarshThakur: Oh ok , and what about the `waypoint plan` issue? I couldn't find one for it. It might not work for all plugins but it does make sense for Kubernetes and Nomad. WDYT?

mitchellh: There are no plans currently to do that, just because it would be hard to provide the same guarantees as Nomad and Terraform.

PR #1358, opened 2021-04-19
plugin/nomad: Ensure Namespace is set if not set by user

briancain:
Since we have to pass through namespace as a ptr, if namespace is unset
then it gets dereferenced as empty string rather than nil, which means
the Nomad job api `Canonicalize()` doesn't end up setting a default
namespace. This commit fixes that by ensuring the namespace Waypoint
provides is set to a default.


krantzinator (approved):
:shipit:

PR #1364, opened 2021-04-20
internal/cli: version shows Waypoint server version

### LINKED ISSUE #1323: Add a command to show the Waypoint server version
I wasn't able to figure out how to get my current Waypoint server version. I thought `waypoint version` or `waypoint server version` might do that but they don't.

It would be great to make the `waypoint version` and `waypoint server version` commands to display the current Waypoint server version.

N/A

N/A

**Additional context**
N/A

mitchellh:
Fixes #1323 

This augments `waypoint version` to show the server version when available. This fails gracefully if the connection fails or there is no Waypoint server available. This command uses an arbitrary 2 second timeout so it doesn't block for a very long time if the connection is not working or taking a long time (unlike the CLI for other operations which will wait much longer).

image

### DIFF
--- a/.changelog/1364.txt
+++ b/.changelog/1364.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: version command now shoes the server version
+```
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -237,7 +237,7 @@ func (c *baseCommand) Init(opts ...Option) error {
 
 	// Create our client
 	if baseCfg.Client {
-		c.project, err = c.initClient()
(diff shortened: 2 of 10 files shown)


krantzinator:
Did we want to include this on the `--version` flag also, or just on the command?

mitchellh:
> Did we want to include this on the `--version` flag also, or just on the command?

Oh that does a different thing? Yes... yes... I'm gonna just fix that to call `version` so the behavior is the same.

mitchellh:
Done

krantzinator (approved):
🚢

PR #1382, opened 2021-04-22
Add ability to reference waypoint.hcl variables between eachother

evanphx:
This change adds the ability refer to other waypoint.hcl defined variables using HCL templating. Additionally, it adds a new class of variable called internal variables. These variables are not exposed to the application and are designed to be composed into other variables.

This functionally fully supports completing dynamic variables into non-dynamic ones, allowing for a better where internal dynamic variables are composed into a env variables:

```hcl
config {
  internal = {
    dbhost = configdynamic("terraform-cloud", {
       organization = "foocorp",
       workspace = "production",
       output = "dbhost"
    })
  }

  env = {
    DATABASE_URL = "postgresql://${config.internal.dbhost}"
  }
}
```


internal/appconfig/watcher.go
```diff
@@ -493,25 +497,46 @@ func configVarSortFunc(vars []*pb.ConfigVar) func(i, j int) bool {
 	}
 }
 
+// Used to just in tracking from the config split, through eval, and back
+// to exporting.
```
mitchellh, line 501: nitpick but would explain what static vs dynamic is here at a high level here and in the comment above `dynamicVar`

internal/appconfig/watcher.go
```diff
@@ -662,14 +687,32 @@ func buildAppConfig(
...
 	if len(dynamic) == 0 {
-		return static
+		return expandStaticVars(log, &ectx, staticVars)
 	}
 
+	// The way this next bit works is that any static values that referenced
+	// other static values have already been expanded before they make it this far,
+	// which means that if a static variable still contains an HCL template, it's
+	// going to reference a dynamic variable. And because dynamic variables can't
+	// reference other variables, the job is pretty easy.
```
mitchellh, line 701: Took me a sec to consider this but you're of course absolutely right... since dynamic vars can't reference anything by their nature (just fundamentally), this simplifies things a LOT. This is a really nice constraint that I didn't consider before! 👍

internal/config/app_config_sort.go
```diff
@@ -0,0 +1,182 @@
...
+// Captures the information about a variable calculated during sort to be used
+// without recalculation by the caller.
+type analyzedPair struct {
+	Pair     hcl.KeyValuePair
+	Name     string
+	Internal bool
+	Refs     []string
+}
+
+type VariableLoopError struct {
```
mitchellh, line 36: Comment on this.

Second, since its exported, I would consider putting it towards the top of the file so that people browsing the file may see it sooner but thats purely a matter of personal taste.

internal/config/app_config_sort.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func (v *VariableLoopError) Error() string {
+	return fmt.Sprintf("loop detected amongst variables: %s", strings.Join(v.LoopVars, ", "))
+}
+
+// sortVars performs a topological sort on EnvRaw and InternalRaw. The sort
+// yields the pairs in most referenced to least referenced order. Meaning
+// that the if pair X references pair R, then R will be before X in the slice.
+func (c *genericConfig) sortVars(ctx *hcl.EvalContext) ([]*analyzedPair, error) {
```
mitchellh, line 47: I think this needs unit tests. Given its a pure function that is executing a well known algorithm, I think we can cover all the edge cases and ensure this really works. I know its tested incidentally through higher level tests, but they're mostly happy path cases.

mitchellh (changes requested):
This looks really good! Relatively simple overall, at least compared to what I thought would be required.

I think the one **big** question I have is how escaping works in this scenario (i.e. I literally want `${foo}` in my value). This might just be handled by `internal/pkg/partial`. If so, I don't see any tests for it and I think we should definitely cover escaping tests at most levels.

internal/appconfig/watcher.go
krantzinator: also nitpick: confusing wording `Used to just in tracking`

evanphx: Both fixed!

internal/appconfig/watcher.go
evanphx: I realized it as I started to code it up and did a little fist pump in the air.

internal/config/app_config_sort.go
evanphx: Done!

internal/config/app_config_sort.go
evanphx: Yup, added a few. There aren't a ton of edge cases so it should cover them.

internal/config/app_config_sort.go
```diff
@@ -63,6 +65,10 @@ func (c *genericConfig) sortVars(ctx *hcl.EvalContext) ([]*analyzedPair, error)
 	}
 
 	for _, m := range maps {
+		if m.expr == nil {
```
mitchellh, line 68: Quick comment: why/when does this happen?

internal/config/app_test.go
```diff
@@ -201,6 +201,23 @@ func TestConfigApp_compare(t *testing.T) {
...
+				require := require.New(t)
+
+				vars, err := c.Config.ConfigVars()
+				require.NoError(err)
+
+				// test the static value
+				require.Len(vars, 3)
+				static, ok := vars[2].Value.(*pb.ConfigVar_Static)
+				require.True(ok)
+				require.Equal("lower(config.internal.greeting, \"FOO\")", static.Static)
```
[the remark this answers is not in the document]
evanphx: No prob! fixed.

internal/config/app_config.go
```diff
@@ -36,6 +37,8 @@ func (c *genericConfig) ConfigVars() ([]*pb.ConfigVar, error) {
 	return c.envVars()
 }
 
+var hclEscaper = strings.NewReplacer("${", "$${", "%{", "%%{")
```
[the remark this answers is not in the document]
evanphx: It's nifty!

internal/config/app_config_sort.go
evanphx: Cleanup the logic, it was only for a test but I just modified the tests to not require it.

.changelog/1382.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+config: Add ability to define internal variables and compose variables together via templating
```
mitchellh, line 2: Nitpick, we may want to turn this one into a bold heading style (see 0.3 or 0.2 or w/e changelog) but we can do that human-powered when we do a major release?

mitchellh (approved):

.changelog/1382.txt
briancain: You should be able to use markdown here and go-changelog will respect it, so adding the `*` _should_ be fine to do here if we want to bold it

mitchellh: Cool, yeah worth considering, I think this is a big feature.

internal/ceb/config.go
```diff
@@ -193,6 +195,58 @@ func (ceb *CEB) watchConfig(
...
+	"HUP":    unix.SIGHUP,
+	"INT":    unix.SIGINT,
+	"IO":     unix.SIGIO,
+	"KILL":   unix.SIGKILL,
+	"PIPE":   unix.SIGPIPE,
+	"PROF":   unix.SIGPROF,
+	"QUIT":   unix.SIGQUIT,
+	"SEGV":   unix.SIGSEGV,
+	"STOP":   unix.SIGSTOP,
+	"SYS ":   unix.SIGSYS,
```
briancain, line 238: I saw this was fixed on the other branch but maybe we can fix this here too since it's ready for merge
```suggestion
	"SYS":   unix.SIGSYS,
```

briancain (approved):

PR #1378, opened 2021-04-22
wip: update GRPC, add Deployment URL

denysvitali:
**WIP: This PR doesn't work right now because the jobs that are reported as _not implemented_ cause errors in the pipeline.**

This PR updates the protoc calls to use the new `protoc-gen-grpc`. Due to a recent change there, now the server must implement all the methods ([more information here](https://github.com/grpc/grpc-go/tree/master/cmd/protoc-gen-go-grpc#future-proofing-services)).

This whole change was done to add the Deploy URL feature partially implemented here: https://github.com/hashicorp/waypoint-plugin-sdk/pull/25

PR #1385, opened 2021-04-23
builtin: Update how builtin plugins are generated with protoc

briancain:
This pull request changes the generate command to protoc. It no longer
supports -go_opt. This commit fixes it by fixing up all the protoc
commands to work just like how the server.proto works by only specifying
a go_out option.

I'm not sure why this hasn't come up yet, but it is possible I've some how bumped
to a new version that removed `--go_opt`. Since our server proto doesn't use it and
compiles fine on the version of protoc, I figured these should behave the same way.

It also includes a Makefile target so that it's easier to discover how to generate
plugin protos.


builtin/aws/alb/plugin.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.23.0
-// 	protoc        v3.11.2
+// 	protoc-gen-go v1.25.0
+// 	protoc        (unknown)
```
denysvitali, line 4: Why is the version `unknown` here?

denysvitali:
I would suggest looking at #1378 or https://github.com/hashicorp/waypoint-plugin-sdk/pull/25 (the GRPc / Protobuf part).

I still think there is something wrong with your repos. There is a new (unfortunately uncompatible) version of `protoc-gen-go` that now requires to implement an `Unimplemented<Service Name>Server` and doesn't accept anymore the `--go_opt=plugins:grpc` but uses `--go-grpc_opt` instead.

https://github.com/grpc/grpc-go/blob/master/cmd/protoc-gen-go-grpc/README.md#future-proofing-services

Is there a reason why you haven't switched to that yet? I have to admit that I tried it on the waypoint-plugins-sdk repo, but it's not really straight forwards and involves a lot of changed files (making the review difficult), but I really feel this is something that is needed.

builtin/aws/alb/plugin.pb.go
briancain: Good question. I have protoc installed on Linux, but can't seem to find out why my generated version always ends up `(unknown)` 🤔

evanphx:
I think we need to figure out the versions in play because I don't have the issue illustrated here. I'm worried about something NEEDING --go-opt...

denysvitali:
I mean, the `--go_opt=plugins:grpc` seems to be incompatible in the newer versions and `--go-grpc_out` must be used instead. 😅

briancain:
I'm going to close this pull request for now. I think we need to figure out a better way to ensure this project can keep everyone on the same version of protoc and protoc-gen-go before this PR happens.

PR #1387, opened 2021-04-26
feat: add deployment URL

denysvitali:
This PR, which has https://github.com/hashicorp/waypoint-plugin-sdk/pull/27 as a pre-requisite, adds support for Deployment URLs.

With this PR, it is now possible to show in the CLI the deployment URLs provided by the plugins, without the need of Horizon (URL Service).

**Warning:** The PR contains a `go.mod` entry that points to our fork since the PR in [https://github.com/hashicorp/waypoint-plugin-sdk](waypoint-plugin-sdk) hasn't been merged yet.


builtin/aws/lambda/deployment.go
```diff
@@ -4,4 +4,8 @@ import (
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 )
 
+func (d *Deployment) URL() string {
+	return ""
```
mitchellh, line 8: I think this should be optional so these no-ops don't have to be implemented (this is more of SDK feedback, but repeating it here) and also so we don't break any previous plugins.

builtin/k8s/deployment.go
```diff
@@ -45,4 +45,8 @@ func (d *Deployment) newDeployment(name string) *appsv1.Deployment {
 	}
 }
 
+func (d *Deployment) URL() string {
+	return d.Url
```
mitchellh, line 49: Maybe I'm being dense, but where are all these being populated?

internal/cli/deployment_create.go
```diff
@@ -96,7 +100,7 @@ func (c *DeploymentCreateCommand) Run(args []string) int {
 		// Output
 		app.UI.Output("")
 		switch {
-		case releaseUrl != "":
+		case releaseUrl != "" || deployUrl != "":
```
mitchellh, line 103: I think we need to add a `releaseUrl != ""` check in the output below so that we don't output an empty release URL.

internal/core/app_deploy.go
```diff
@@ -135,32 +135,34 @@ type deployOperation struct {
 	// Set by init
 	autoHostname pb.UpsertDeploymentRequest_Tristate
 
-	// component is initialized and recycled at various points in the deployment
+	// Component is initialized and recycled at various points in the deployment
 	// operation as we have more templating information for the config.
-	component *Component
+	Component *Component
```
mitchellh, line 140: Is there a reason behind this change related to adding a deployment URL?

If there is not a reason, OR its not related to deployment URLs, let's table this for another issue to discuss. This feels unrelated to this PR.

internal/server/server.go
```diff
@@ -13,7 +13,7 @@ import (
 
 //go:generate sh -c "protoc -I../../vendor/proto/api-common-protos -I ../.. ../../internal/server/proto/server.proto --go_out=plugins=grpc:../.. --go-json_out=../.."
 //go:generate mv ./proto/server.pb.json.go ./gen
-//go:generate mockery -all -case underscore -dir ./gen -output ./gen/mocks
+//go:generate mockery --all --case underscore --dir ./gen --output ./gen/mocks
```
mitchellh, line 16: This appears unrelated to this PR. Please remove.

krantzinator:
Thanks for taking a swing at this! I'm going to leave a couple of small comments here while the sdk side is sorted out.

internal/core/app_deploy.go
krantzinator, line 140: I think this change was auto-generated somehow, as I don't see a functional reason for it. Can you revert this?

internal/server/proto/server.proto
```diff
@@ -1854,6 +1854,9 @@ message Deployment {
   // id is the unique ID for this deployment
   string id = 1;
 
+  // url is the URL to the Deployment
+  string url = 18;
```
mitchellh, line 1858: This is fine, but as a bigger comment I think we should rethink our data model slightly around URLs. Right now it is unclear what the difference is between `deploy.Url` and `deploy.Preload.Url` and both MIGHT be empty or both MIGHT NOT be empty or ONE may be empty, etc. Actionably, I think we should:

* Have more detailed comments.
* Have a canonical URL field that has SOME URL if one exists (may still be empty if none exist)
* Clearly delineate between URL service URL and Platform URL

internal/core/app_deploy.go
denysvitali: This change was made to be in line with Release. In there the `Component` is exported, I'm not sure why - but I wanted to make it consistent :P

mitchellh (changes requested):
The idea here is A+, no complaints, I think we should support this. I think the implementation side needs a bit of work. There's one big comment in there around data modeling.

I think you're using a different version of Mockery which caused changes in all the generated mocks. Please use the pinned version we have in our Makefile. Thanks!

internal/server/gen/mocks/waypoint_server.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v1.1.2. DO NOT EDIT.
+// Code generated by mockery 2.7.4. DO NOT EDIT.
```
krantzinator, line 1: Could you put the mockery version change in a different PR, or maybe at least a different commit? There are a couple of places this is pinned, so it would be nice to capture the whole change at once.

internal/server/server.go
denysvitali: Will do :)

internal/server/proto/server.proto
denysvitali: I didn't found any documentation around `deploy.Preload.Url`. I think it's supposed to be an Horizon pre-defined URL before the application is even deployed, is that correct?

internal/server/gen/mocks/waypoint_server.go
krantzinator: Can disregard this comment in favor of Mitchell's review :)

builtin/aws/lambda/deployment.go
denysvitali: I'm not sure how we can make it optional to be honest, what would you suggest in this case?

mitchellh: No problem! I think a separate interface would be the way to go (in the SDK):

```go
type DeploymentWithURL interface {
  Deployment

  URL() string
}
```

builtin/k8s/deployment.go
denysvitali: It's not being populated in fact. This is just an empty URL at the moment, waiting to be implemented:
https://github.com/hashicorp/waypoint/pull/1387/files/1dcacbb891bf4f181adfd3c952a96373073648be#diff-1b0cb4abccfe10f5d0dcb1e343487a8da3a3da4c7b06dd4e86be69b5456681f1R10

internal/core/app_deploy.go
denysvitali: Unrelated, sorry. It was just a change to be consistent with:
https://github.com/swisscom/waypoint/blob/feature/deployment-url/internal/core/app_release.go#L152

internal/server/gen/mocks/waypoint_server.go
denysvitali: You're right Rae, I will remove this from the PR since it's out of scope

internal/server/proto/server.proto
mitchellh: Yes, because our data model only understands URL service URLs currently. :) That's the big conceptual shift in this PR I think is that we're now saying deployments may have a platform-provided URL. But the data model doesn't quite reflect that and I think it should.

denysvitali: How would you like to approach this?

Do you have any proposed change?

denysvitali:
Since the CI checks are failing (due to the fact that https://github.com/hashicorp/waypoint-plugin-sdk/pull/27 is still not merged), I temporarily added a replacement in `go.mod`, so that the tests are passing.

Please, do not merge if the SDK change isn't merged first.

denysvitali:
@mitchellh / @briancain can you please re-start the integration test? I can't do it from my side

krantzinator:
Reran the failed workflow

denysvitali:
Ping @krantzinator @mitchellh or @briancain , do you know what's wrong with the `vendor` folder check?

Btw, rebased on top of main and good to go :)

krantzinator:
@denysvitali looks like you need to run a `go mod tidy`

denysvitali:
Thanks, tests are OK now ;)

.changelog/1387.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+internal: Add support for Deployment URLs
```
mitchellh, line 2: We don't add internal changes to the changelog, recommendation below:

```suggestion
core: platform plugins may now advertise deployment-specific URLs
```

denysvitali: I have no clue on why it is not working for me (using Nix as well). A double hypen fixes the issue for me, but as we already discussed this should be addressed in another PR. Sorry, it might have slipped into my commit.

internal/server/server.go
briancain: I believe Mockery version 2.x+ uses double hyphens but we're on an older version so we use single.

denysvitali: But does the nix environment use a newer one?  Because in my nix shell, the default (single hypen) wasn't working IIRC

briancain: I'm not sure, I don't use it 😄 But when I was first getting started I just installed the latest Mockery and this was an issue I ran into, rolling back to `1.1.2` fixed it for me.

https://github.com/hashicorp/waypoint/blob/7d6229c3167209b80459e67bdc669c32465aceb0/tools/tools.go#L22-L23

briancain: Looks like it should technically be installing the right version.... https://github.com/hashicorp/waypoint/blob/7d6229c3167209b80459e67bdc669c32465aceb0/shell.nix#L61-L62

denysvitali: Really, really strange. I'll have to double check my environment since shell.nix reports the 1.x version too 😮

denysvitali: Reverted the change, sorry!

denysvitali:
Uh. Just noticed that after some changes it is not working anymore (deployment.Url is not filled). Let me re-work on this.

denysvitali:
Okay, so let me unpack the new changes.

### `Deployment.Url` is now used only internally

Since we want to avoid having redundant fields, `Deployment.Url` will now be used only internally. The CLI and the UI will only consider `Deployment.Preload.DeployUrl`

We use `setDeploymentUrlIfNeeded` to write `d.Preload.DeployUrl` from `d.Url` if the latter exists.
This only happens when the client requests some `Deployment` resources (e.g: `GetDeployment` and `ListDeployments`), exactly like the URL service invocation works.

### The CLI now displays deployment URLs too

Kind of self-explanatory. Here is an example:

```
$ waypoint deployment list -workspace=develop
     | ID |   PLATFORM   |  DETAILS   |    STARTED     |   COMPLETED    |                   URL                   |     HEALTH
-----+----+--------------+------------+----------------+----------------+-----------------------------------------+-----------------
  ✖  | 12 | cloudfoundry | artifact:3 | 9 seconds ago  | 8 seconds ago  |                                         | Unknown status
  🚀 | 11 | cloudfoundry | artifact:3 | 14 minutes ago | 14 minutes ago | xy-api-v11.internal.swisscom.ch | Unknown status
  ✔  | 10 | cloudfoundry | artifact:3 | 50 minutes ago | 50 minutes ago | xy-api-v10.internal.swisscom.ch | Unknown status
```

### The UI now shows Deployment URLs too

Previously a button "Visit Deployment" was shown - now the full URL is shown, exactly like for the Releases

Screenshot 2021-06-09 at 20 58 58


### The Deployment view now shows `v10` instead of `#10`

Before my changes, this part was inconsistent: v0 was displayed in the general view but `#10` was shown when the deployment was clicked.

Screenshot 2021-06-09 at 20 59 40

denysvitali:
Friendly ping :)

@briancain

internal/cli/deployment_list.go
```diff
@@ -281,6 +281,7 @@ func (c *DeploymentListCommand) Run(args []string) int {
 					details[0],
 					startTime,
 					completeTime,
+					b.Preload.DeployUrl,
```
briancain, line 284: I think given this bug: https://github.com/hashicorp/waypoint/issues/1608
We should only be showing the deploy URL with the `verbse` flag. Otherwise I think deployment list might be pretty difficult to format and look nice until that is fixed.

briancain (changes requested):
Thanks for the ping @denysvitali - sorry this took a while to get a final review!

I had a small comment about the deploy URLs in `waypoint deployment list`, and I think it would be best to separate the UI changes in a follow-up PR. Then we can get the core functionality in with this at least and get a proper UI review on the follow-up PR.

FYI I am around today (pacific time in the US) but I am on vacation next week 🏖️ so someone else from the team can help out and review next week on any updates or on the UI PR 😄

denysvitali:
I'm solving #1608 by adding a new URL flag (to display it), because one is probably not interested in seeing all the additional details but might want to see the deployment URL:

```
$ waypoint deployment list -workspace=develop
     | ID |   PLATFORM   |   DETAILS   |  STARTED   | COMPLETED  | HEALTH  
-----+----+--------------+-------------+------------+------------+---------
  🚀 | 64 | cloudfoundry | artifact:34 | 1 week ago | 1 week ago | n/a     
  ✔  | 63 | cloudfoundry | artifact:33 | 1 week ago | 1 week ago | n/a    
```


```
$ waypoint deployment list -workspace=develop -V
     | ID |   PLATFORM   |              DETAILS               |  STARTED   | COMPLETED  | HEALTH  
-----+----+--------------+------------------------------------+------------+------------+---------
  🚀 | 64 | cloudfoundry | artifact:34                        | 1 week ago | 1 week ago | n/a     
     |    |              | artifact.product:the-product    
     |    |              | build.product:the-product       
     |    |              | deployment.product:the-product  
  ✔  | 63 | cloudfoundry | artifact:33                        | 1 week ago | 1 week ago | n/a     
     |    |              | artifact.product:the-product    
     |    |              | build.product:the-product       
     |    |              | deployment.product:the-product  
```


```
$ waypoint deployment list -workspace=develop -u
     | ID |   PLATFORM   |   DETAILS   |  STARTED   | COMPLETED  | HEALTH |                   URL                    
-----+----+--------------+-------------+------------+------------+--------+------------------------------------------
  🚀 | 64 | cloudfoundry | artifact:34 | 1 week ago | 1 week ago | n/a    | product-1-c1f94b83.xyz.swisscom.ch  
  ✔  | 63 | cloudfoundry | artifact:33 | 1 week ago | 1 week ago | n/a    | product-1-1c414a16.xyz.swisscom.ch  
```

denysvitali:
Test are failing because of a network issue:

> Error making request.
> Error: connect ETIMEDOUT 140.82.113.4:443
>    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1144:16)

krantzinator:
If a plugin hasn't implemented deployment URLs, what is the expected behavior? I'm not seeing anything listed in this output for e.g. the nomad plugin. I would expect it would show the default at that point?
```
$ waypoint deployment list -u
     | ID | PLATFORM |               DETAILS               |   STARTED    |  COMPLETED   |      HEALTH      | URL  
-----+----+----------+-------------------------------------+--------------+--------------+------------------+------
  🚀 |  1 | nomad    | image:8cc9f30                       | 1 minute ago | 1 minute ago | ✔ - 1 minute ago |      
```

denysvitali:
> If a plugin hasn't implemented deployment URLs, what is the expected behavior?

This is a very good question! We could eventually fall back to the `b.Preload.Url` (which should be the Hashicorp Horizon URL), if available, or return N/A if none of them are available.

What do you think?

krantzinator:
> > If a plugin hasn't implemented deployment URLs, what is the expected behavior?
> 
> This is a very good question! We could eventually fall back to the `b.Preload.Url` (which should be the Hashicorp Horizon URL), if available, or return N/A if none of them are available.
> 
> What do you think?

That makes sense to me!

denysvitali:
@krantzinator Fixed as agreed

internal/cli/deployment_list.go
```diff
@@ -267,16 +276,30 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
+			}
+
+			if c.flagUrl {
+				theUrl := "N/A"
+				if b.Url != "" {
+					theUrl = b.Url
+				} else if b.Preload.DeployUrl != "" {
+					theUrl = b.Url
+				}
+				columns = append(columns, theUrl)
```
krantzinator, line 298: quick bug fix and a couple of syntactical preferences to match the repo's style
```suggestion
				url := "n/a"
				if b.Url != "" {
					url = b.Url
				} else if b.Preload.DeployUrl != "" {
					url = b.Preload.DeployUrl
				}
				columns = append(columns, url)
```

krantzinator:
Looking good! Thanks for sticking with this! I gave one last suggestion, and then it looks like we need to do a rebase and regenerate the protobufs. Let me know if you'd like me to do that. 🙌🏻

internal/cli/deployment_list.go
denysvitali: Sounds good! I tend not to use `url` because it conflicts with the `net.url` package, but since we're not using it here all good!

denysvitali:
> Let me know if you'd like me to do that. 🙌🏻

Yes, please do it!

krantzinator:
@denysvitali can you give me access to the forked repo so I can edit in this PR rather than making a new branch?

denysvitali:
@krantzinator : No clue on why GitHub doesn't let me give you guys permissions over the branch itself, so I added you as a writer to our repository :)
The invite is still pending though

krantzinator:
@denysvitali I got it! Just waiting on the docker server build to finish re-caching the new go mod stuff so I can give it a quick spin before merging :)

krantzinator (approved):

PR #1389, opened 2021-04-26
Update internal GCP Auth to use new SignJWT endpoint

catsby:
Updates Waypoints' internal GCP auth to use GCP's IAM Service Account Credential endpoint for signing JWTs, as the IAM endpoint versions are deprecated and being turned off (not totally sure when but speculation/inference is around July 1, 2021). See https://cloud.google.com/iam/docs/migrating-to-credentials-api for more information on the deprecation and migration. 

I'm honestly not sure where this is used and there isn't a directly related test file associated with this, so I have not tested other than to ensure the project still compiles.

See also for more backstory:

- https://github.com/hashicorp/vault-plugin-auth-gcp/issues/100
- https://github.com/hashicorp/go-gcp-common/pull/4
- https://github.com/hashicorp/vault-plugin-auth-gcp/pull/108
- https://github.com/hashicorp/vault/pull/11473

Note: I added the `pr/no-changelog` label because I didn't feel this change warranted it, but let me know if you'd prefer one added

### DIFF
--- a/.changelog/1389.txt
+++ b/.changelog/1389.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/vault: Use Service Account Credentials API for GCP SignJWT endpoint
+```
--- a/builtin/vault/internal/auth/gcp/gcp.go
+++ b/builtin/vault/internal/auth/gcp/gcp.go
@@ -16,7 +16,7 @@ import (
 	"github.com/hashicorp/vault/api"
 	"github.com/hashicorp/vault/sdk/helper/parseutil"
 	"golang.org/x/oauth2"
-	iam "google.golang.org/api/iam/v1"
+	iam "google.golang.org/api/iamcredentials/v1"
 
 	"github.com/hashicorp/waypoint/builtin/vault/internal/auth"
 )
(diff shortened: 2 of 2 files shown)


briancain (commented):
Hey @catsby ! Thanks.

I think we'll also want to update 👇🏻 too?

https://github.com/hashicorp/waypoint/blob/1a46a877f37f407b1a76e0bf6fdd258223fbfeee/builtin/google/cloudrun/platform.go#L154-L157

I could see this needing a CHANGELOG entry only because in the future, older Waypoints will probably fail once the endpoint is deprecated? It might be nice in that sense to remember when the update to the new endpoint usage was released.

catsby:
Hey @briancain - I _believe_ the usage in `google/cloudrun/platform.go` is correct; the URL change from `projects/%s/serviceAccounts/%s` to `projects/-/serviceAccounts/%s` only pertains to the Service Account Credentials API. The [API to test permissions](https://cloud.google.com/iam/docs/testing-permissions) seems to be correct. The changes required only matter for signing JWTs, and I'm not finding spots in the Waypoint code where that method is otherwise called. Let me know what you think, maybe I'm missing something?

krantzinator (approved):
ran through a release to GCP with this and it works 🎉

briancain (approved):
👍🏻 looks good!

PR #1395, opened 2021-04-28
Add ability to write config into files

evanphx:
This adds a file block to `config {}` which indicates paths and the contents to be stored at those paths within the application. This data is written by the entrypoint.

This is draft while we discuss how to expose the signal selector configuration.


.changelog/1395.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+config: Add ability write configuration values as files rather than environment variables.
```
krantzinator, line 2: Juuuust sliding in a little changelog fixup :disappears:
```suggestion
config: Add ability to write configuration values as files rather than environment variables.
```

internal/ceb/config.go
```diff
@@ -112,12 +113,12 @@ func (ceb *CEB) watchConfig(
 	// env stores the currently known list of environment vars we set on the
 	// child. We need to store this since we want to launch all exec sessions
 	// with the latest/current view on env vars too.
```
piotrromanowski, line 115: I think we need to update this to reflect the changes made. It now contains any updated configurations to the application (envvars & udpatedFiles). Same on line 174

internal/config/app_config_sort.go
```diff
@@ -112,6 +115,7 @@ func (c *genericConfig) sortVars(ctx *hcl.EvalContext) ([]*analyzedPair, error)
 				Name:     key,
 				Refs:     refs,
 				Internal: m.internal,
+				Path:     m.path,
```
piotrromanowski, line 118: Is this preferred for it's explicitness?

    Path: m.prefix == "config.file."

Would also work here right?

internal/config/app_config.go
```diff
@@ -64,6 +72,32 @@ func (c *genericConfig) envVars() ([]*pb.ConfigVar, error) {
...
+			config["file"] = cty.MapVal(file)
+			ctx.Variables["config"] = cty.MapVal(config)
+
+		} else {
+			env[pair.Name] = val
+
+			// Because of the nature of the hcl map type, we have to rebuild these
+			// each time we modify them.
+			config["env"] = cty.MapVal(env)
+			ctx.Variables["config"] = cty.MapVal(config)
```
piotrromanowski, line 97: nit: at first glance I was expecting this last line in these conditions to be different since they're in their own blocks. It might be read clearer to place it once after the conditions instead

internal/ceb/config.go
```diff
@@ -247,6 +262,27 @@ var sigMap = map[string]os.Signal{
...
+		if err != nil {
+			log.Error("error writing application file", "error", err, "path", fc.Path)
+		} else {
+			log.Info("wrote application file to disk", "path", fc.Path)
+		}
+	}
+
+	if cfg.FileRewriteSignal != "" {
+		if sig, ok := sigMap[strings.ToUpper(cfg.FileRewriteSignal)]; ok {
+			ceb.childSigCh <- sig
```
piotrromanowski, line 279: Would we still want to continue with this if we errorerd on line 270?

piotrromanowski (commented):
I realize this is just a draft, but I figured I'd try my hand at contributing somehow to the waypoint project. Feel free to ignore 😅

internal/ceb/foo.txt
```diff
@@ -0,0 +1 @@
+via config
```
krantzinator, line 1: Should this be in a test dir?

internal/appconfig/watcher.go
```diff
@@ -298,10 +324,15 @@ func (w *Watcher) watcher(
 	refreshNowCh := make(chan time.Time)
 	close(refreshNowCh)
 
-	// prevSent is flipped to true once we update our first set of compiled
+	// prevFilesSent is flipped to true once we update our first set of compiled
```
krantzinator, line 327:
```suggestion
	// prevEnvSent is flipped to true once we update our first set of compiled
```

krantzinator (approved):

internal/server/proto/server.proto
```diff
@@ -2223,6 +2234,10 @@ message ConfigVar {
   // It exists only to be referenced by other variables.
   bool internal = 8;
 
+  // Indicates that this is actually be written as a file, with the name
+  // field being the filename.
+  bool name_is_path = 9;
```
mitchellh, line 2239: Any reason you didn't use a `oneof` on name instead? It seems like you're effectively making that happen here but using a bool instead feels weird. In my mind I expected something like the below....

```proto
oneof name {
  string env = 1;
  string path = 2
}
```

internal/server/proto/server.proto
```diff
@@ -2443,6 +2471,9 @@ message EntrypointConfig {
   // be nil if the user is running an old enough server so always nil-check this.
   DeploymentInfo deployment = 4;
 
+  // The signal to send the application when config files change.
+  string file_change_signal = 6;
```
mitchellh, line 2475: Nitpick: I'd move this line up below the config_sources entry since this is config related. Also consider a `config_` prefix like `config_file_change_signal` or just `config_file_signal`

internal/server/proto/server.proto
```diff
@@ -102,6 +102,9 @@ service Waypoint {
   // each variable.
   rpc GetConfig(ConfigGetRequest) returns (ConfigGetResponse);
 
+  // Set a metadata configuration item on a project or application.
+  rpc SetMetadata(MetadataSetRequest) returns (google.protobuf.Empty);
```
mitchellh, line 106: I feel like we should use `UpsertProject` and `UpsertApplication` for this and not insert a new API. This is the API that "edit project/application" uses in the UI for example and as a caller it feels weird that some fields are edited via an "update project" API and some are via a "set metadata with a oneof" API.

mitchellh (changes requested):

internal/appconfig/watcher.go
```diff
@@ -244,6 +246,26 @@ func (w *Watcher) notify(
...
+			cebFunctions[k] = v
+		}
+	}
+
+	// Add some of our functions
+	addFuncs(funcs.Filesystem())
+	addFuncs(funcs.Encoding())
+	addFuncs(funcs.Datetime())
+	addFuncs(funcs.Jsonnet())
+}
```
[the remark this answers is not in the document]
evanphx: Oh yes! We discussed that too.

internal/appconfig/watcher.go
```diff
@@ -424,11 +455,15 @@ func (w *Watcher) watcher(
...
 			log.Trace("refreshing app configuration")
-			newEnv := buildAppConfig(ctx, log,
+			newEnv, newFiles := buildAppConfig(ctx, log,
 				w.plugins, static, dynamic, dynamicSources, prevVarsChanged)
 
 			sort.Strings(newEnv)
 
+			sort.Slice(newFiles, func(i, j int) bool {
+				return newFiles[i].Path < newFiles[j].Path
+			})
```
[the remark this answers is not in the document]
evanphx: Totes.

internal/server/proto/server.proto
evanphx: Certainly could! That will ripple out far and wide.

internal/server/proto/server.proto
evanphx: Ok, that's fair.

internal/server/proto/server.proto
evanphx: Sure, no prob.

internal/server/proto/server.proto
mitchellh: Yeah... it would. We can punt it to a future branch experiment.

internal/appconfig/watcher.go
```diff
@@ -372,7 +403,7 @@ func (w *Watcher) watcher(
 		// Case: caller sends us a new set of variables
 		case newVars := <-w.inVarCh:
 			// If the variables are the same as the last set, then we do nothing.
-			if prevSent && w.sameAppConfig(log, prevVars, newVars) {
+			if prevEnvSent && w.sameAppConfig(log, prevVars, newVars) {
```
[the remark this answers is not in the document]
evanphx: You're totally right, it should include `prevFileSent` there.

internal/config/app_config_sort.go
evanphx: It would, but it's better to have been explicit about the choice.

internal/ceb/foo.txt
evanphx: WHOOPS.

internal/server/proto/server.proto
evanphx: 👍

internal/appconfig/watcher.go
```diff
@@ -68,9 +70,9 @@ type Watcher struct {
 	// currentCond is used to lock and notify updates for currentEnv.
 	currentCond *sync.Cond
 
-	// currentEnv is the list of current environment variable values for
+	// currentFiles is the current environment variables and application config files for
```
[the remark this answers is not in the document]
evanphx: Yup, it was. Thanks!

internal/appconfig/watcher.go
```diff
@@ -265,6 +287,10 @@ func (w *Watcher) watcher(
 	// this to compare and prevent unnecessarilly restarting the command.
 	var prevEnv []string
 
+	// prevFiles keeps track of the last set of files we computed. We do
+	// this to compare and prevent unnecessarilly restarting the command.
```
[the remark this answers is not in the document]
evanphx: Thanks!

internal/ceb/config.go
evanphx: Yup, updated!

internal/config/app_config.go
evanphx: Yeah, I've cleaned this up so it's clearer.

internal/ceb/config.go
evanphx: Totes right, I've fixed that. Thanks!

internal/server/proto/server.proto
```diff
@@ -2258,6 +2273,19 @@ message ConfigGetResponse {
   repeated ConfigVar variables = 1;
 }
 
+message MetadataSetRequest {
```
mitchellh, line 2276: Don't need this anymore I think.

briancain (approved):

internal/server/singleprocess/state/metadata.go
```diff
@@ -0,0 +1,25 @@
+package state
+
+import (
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+// GetFileChangeSignal checks the metadata for the given application and its
+// project, returning the value of FileChangeSignal that is most relevent.
+func (s *State) GetFileChangeSignal(scope *pb.Ref_Application) (string, error) {
```
mitchellh, line 9: Should this be in metadata.go?

internal/server/singleprocess/state/metadata_test.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+	serverptypes "github.com/hashicorp/waypoint/internal/server/ptypes"
+)
+
+func TestMetadata(t *testing.T) {
```
mitchellh, line 12: Old test name

mitchellh (changes requested):

internal/server/proto/server.proto
evanphx: 👍

internal/server/singleprocess/state/metadata.go
evanphx: Good call, moved.

internal/server/singleprocess/state/metadata_test.go
evanphx: Moved the test!

mitchellh (approved):

PR #1401, opened 2021-05-03
Properly validate the ECS ALB options. Fixes #1070

### LINKED ISSUE #1070: aws/ecs existing load balancer cannot be used
**Describe the bug**
cannot use existing  application load balancer in AWS.

When I try to deploy with "waypoint deploy" and this configuration 

```hcl
project = "backend"

app "backend" {

  build {
    use "docker" {}

    registry {
      use "aws-ecr" {
        region     = "eu-central-1"
        repository = "backend-dev"
        tag        = "latest"
      }
    }
  }

  deploy {
    use "aws-ecs" {
      region  = "eu-central-1"
      memory  = 512
      subnets = ["subnet-xxxx", "subnet-xxxxx"]
      alb {
        listener_arn = "arn:aws:elasticloadbalancing:eu-central-xxxxxx"
        zone_id      = "ZoneID"
        domain_name  = "api.test.com"
        certificate  = "cert_arn"
      }
      service_port = 3000
    }
  }

}
```

I get 

```
 waypoint deploy                            

» Deploying...
! When using an existing listener, Route53 setup is not available

```

due to this error in the code:

properties are required properties by default but in this if condition it fails because it doesn`t accept those properties:

https://github.com/hashicorp/waypoint/blob/a400665719d58d936c36cabd4a5beac8b0c91892/builtin/aws/ecs/platform.go#L213

evanphx:
Since we've got a more complex relationship between the values that can be modeled in HCL tags, this introduces using ozzo to validate parts of the config struct.

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -17,12 +17,14 @@ import (
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/route53"
+	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	"github.com/hashicorp/waypoint-plugin-sdk/docs"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/builtin/aws/utils"
 	"github.com/hashicorp/waypoint/builtin/docker"
+	"github.com/hashicorp/waypoint/internal/pkg/validationext"
 )
 
 type Platform struct {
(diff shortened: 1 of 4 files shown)


builtin/aws/ecs/platform.go
```diff
@@ -34,6 +36,42 @@ func (p *Platform) Config() (interface{}, error) {
 	return &p.config, nil
 }
 
+// ConfigSet is called after a configuration has been decoded
+// we can use this to validate the config
+func (p *Platform) ConfigSet(config interface{}) error {
```
briancain, line 41: Is this function called automatically through the SDK or something internal?

builtin/aws/ecs/platform.go
```diff
@@ -1350,13 +1388,13 @@ func (p *Platform) Destroy(
 
 type ALBConfig struct {
 	// Certificate ARN to attach to the load balancer
-	CertificateId string `hcl:"certificate"`
+	CertificateId string `hcl:"certificate,optional"`
```
briancain, line 1391: Should these be `*string` to tell unset versus empty string? 🤔

builtin/aws/ecs/platform.go
evanphx: Yup! It's something the SDK supports, we just didn't use it before in ECS.

builtin/aws/ecs/platform.go
evanphx: I considered that, yeah, but opt'd to leave them as is because there are folks using `""` to mean not set atm so we'd have to check for `set but empty` anyway.

krantzinator (approved):

briancain (approved):

krantzinator (approved):

PR #1402, opened 2021-05-03
Remove old bindata_ui.go file

catsby:
The file `/internal/server/bindata_ui.go` was removed in https://github.com/hashicorp/waypoint/pull/118, but I believe accidentally re-added in https://github.com/hashicorp/waypoint/pull/119 . The current file is located at [/internal/server/gen/bindata_ui.go](https://github.com/hashicorp/waypoint/blob/f710fe10ee4a29f6fbd020656a8ebd3ad124d14d/internal/server/gen/bindata_ui.go)


krantzinator (approved):
😅

briancain (approved):
Makes sense...should this be backported too for the next release?

PR #1404, opened 2021-05-03
Change the AWS log level to match the waypoint log level. Fixes #1319

### LINKED ISSUE #1319: Match AWS go client logs with Waypoint log level

When looking through Waypoint logs that make use the AWS client (like lambda or ecs), Waypoint doesn't end up exposing more details logs from the AWS client in use when doing various AWS operations like creating resources or making API requests to AWS.

Have Waypoint configure its AWS client for each platform to match the log level of Waypoint when requested, i.e. running `waypoint up -vvv` would turn on TRACE level logging with the AWS client function calls. We should also allow for more detailed AWS logging through the runner and server as well, either by default, or through a config option.

Only enable detailed AWS logging if an env var or config option is enabled, rather than automatically adjusting based on the requested log level of Waypoint.

Detailed AWS logs will help us more easily determine what's going on when trying to triage a bug with AWS plugins and see what's happening underneath the hood

evanphx:
I also considered setting aws's log function to be a proxy for hclog.Logger, but didn't because there are no levels on it, so it's best it shows up as just opaque output for the human to parse visually.

### DIFF
--- a/.changelog/1404.txt
+++ b/.changelog/1404.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+platform/aws: Set the AWS log level along with the waypoint log level
+```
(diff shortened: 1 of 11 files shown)


briancain (approved):
Excellent! 🌳

PR #1412, opened 2021-05-04
UI: status checks

gregone:
e.g: 
Screen Shot 2021-05-04 at 18 18 12

Only adding the status check components for now. Adding them to the UI and replacing the existing badges will be done later


ui/app/components/status-badge/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        return 'down';
+        break;
+      case 2:
+        return 'ready';
+        break;
+      case 1:
+        return 'alive';
+        break;
+      case 0:
+        return 'unknown';
```
briancain, line 34: Forgive me for not knowing typescript (I think??) but is there a `case default` or something where we return unknown always?

briancain: The API should behave this way too, but maybe in the case where something goes wrong the UI could handle it the same way

ui/translations/en-us.yaml
```diff
@@ -93,3 +93,20 @@ state:
     success: 'Released successfully'
     error: 'Release encountered an error'
     pending: 'Release pending'
+
+status_badge:
+  partial:
+    text: 'Partial'
+    tooltip: 'Application is partially running'
```
briancain, line 100:
```suggestion
    tooltip: 'Application is partially available'
```

ui/translations/en-us.yaml
```diff
@@ -93,3 +93,20 @@ state:
...
+  down:
+    text: 'Down'
+    tooltip: 'Application is down'
+  ready:
+    text: 'Up'
+    tooltip: 'Application is running'
```
briancain, line 106:
```suggestion
    tooltip: 'Application is ready'
```

Minor nit, but we have both "Ready" and "Alive" states where both are technically "running". So maybe we could be more specific here?

briancain (commented):
Sweet, looking good Greg! 😄 Excited to see this all hooked up and working once I get the server side work ready to go 👍🏻 🙂

ui/app/components/status-badge/index.ts
gregone: @briancain You're right I should have a default case too

PR #1415, opened 2021-05-04
plugin/ecs: Allow memory_reservation to be set on the main container

### LINKED ISSUE #1163: Memory reservation feature in aws-ecs plugin not exposed through task definition
WP server: 0.2.3
WP client: 0.2.3

This isn't exactly a bug but more a missing feature ~~that is critical~~ for applications deployed on AWS Fargate. I'll explain why.

We have over the last couple of weeks started load testing our applications which all are running on Fargate. A common pattern has been that our instances are _drained of memory_ even though our apps are using very little memory. We have profiled _extensively_ using _AWS Codeguru_ and found  nothing that indicates a _memory leak_ on our side. After turning our attention to the surrounding _ecs container_  we have seen that it is in fact the container that is using _almost all memory_ allocated to the instance which in turn leads to the container beeing killed by the Fargate process who is supervising it. 

After looking into the AWS task documentation and [ _memory settings_](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_memory) specifically we have found the _memoryReservation_ feature which controls how much memory is allocated to the _main container_. This is of now not exposed through the WP task definition though it is available when using [_sidecars_.](https://www.waypointproject.io/plugins/aws-ecs#sidecar-memory_reservation).

By manually editing the _memoryReservation_, which is a _soft limit_ on the main container, we have seen an ~~drastical~~ improvement for all our Fargate deployed applications. Memory usage for whole instances are now in line with what we expected.

Adjusting the soft limit only delayed the problem. The main source of this issue is mentioned [here](https://github.com/hashicorp/waypoint/issues/1192).

evanphx:
Closes #1163

This also enforces the memory and memory_reservation settings on the additional containers.

### DIFF
--- a/.changelog/1415.txt
+++ b/.changelog/1415.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+plugin/ecs: Allow memory_reservation to be set on primary container
+```
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -71,6 +71,24 @@ func (p *Platform) ConfigSet(config interface{}) error {
 		}
 	}
 
+	err := utils.Error(validation.ValidateStruct(c,
+		validation.Field(&c.Memory, validation.Required, validation.Min(4)),
+		validation.Field(&c.MemoryReservation, validation.Min(4), validation.Max(c.Memory)),
+	))
+	if err != nil {
+		return err
+	}
+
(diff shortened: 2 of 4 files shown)


briancain (approved):
Nice ❇️ 💾 !

PR #1426, opened 2021-05-05
internal/cli: don't use conn if error during accept

mitchellh:
This prevents a crash that could happen since during the error case conn
is nil.


briancain (approved):
🏃🏻

PR #1441, opened 2021-05-06
Fix the order of known log entries returned by the GetLogStream api

evanphx:


internal/server/logbuffer/logbuffer.go
```diff
@@ -191,11 +205,7 @@ type Reader struct {
 	closed  uint32
 }
 
-// Read returns a batch of log entries, up to "max" amount. If "max" isn't
-// available, this will return any number that currently exists. If zero
-// exist and block is true, this will block waiting for available entries.
-// If block is false and no more log entries exist, this will return nil.
-func (r *Reader) Read(max int, block bool) []Entry {
+func (r *Reader) readTimedEntries(max int, block bool) []timedEntry {
```
briancain, line 208: Func comment here?

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"io"
+	"time"
+)
+
+// LogMerge can combine multiple log streams into one stream. It presumes
+// each stream emits entries in time order, and then weaves the entries
+// together to create a total time ordered stream.
+type LogMerge struct {
```
mitchellh, line 11: Given the name of the package is `logbuffer` and has "log" in it, I think this type should probably be named `Merger` or something. This is I think a tao of idiomatic Go or Effective Go or w/e.

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,153 @@
...
+	Time() time.Time
+
+	// Value returns the log entry itself. This allows time addition
+	// wrappers to return their original value.
+	Value() interface{}
+}
+
+// LogMergeInput is value that returns TimedEntry's for LogMerge
+// to weave together.
+type LogMergeInput interface {
```
mitchellh, line 29: Going along with above, bikeshed but perhaps `MergeReader` makes more sense here, so you're composing Reader types similar to `io` and `Reader` happens to also implement `MergeReader`.

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,153 @@
...
+	Next() (TimedEntry, error)
+}
+
+// NewLogMerge creates a new LogMerge, with the stream generated
+// from the given inputs.
+func NewLogMerge(inputs ...LogMergeInput) (*LogMerge, error) {
```
mitchellh, line 35: `NewMerger` ?

Also, perhaps no error return value here.

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,153 @@
...
+	lm.inputs = inputs
+
+	lm.heads = lm.makeHeads()
+
+	return &lm, nil
+}
+
+// AddInput adds a new input to the LogMerge. NOTE: inputs can
+// not be removed at this time.
+func (l *LogMerge) AddInput(input LogMergeInput) {
```
mitchellh, line 46: This isn't used currently, do we need this or should we remove it?

internal/server/logbuffer/merge.go
evanphx: `type ThatWhichMerges struct .... ` 🤣 

Totes, I'll make this Merger, that's a better name.

internal/server/logbuffer/logbuffer.go
evanphx: Done!

internal/server/logbuffer/merge.go
evanphx: Totes agree!

internal/server/logbuffer/merge.go
evanphx: mmhmm

internal/server/logbuffer/merge.go
evanphx: We don't! It was used at one point.

internal/server/logbuffer/logbuffer.go
```diff
@@ -71,6 +73,18 @@ func New() *Buffer {
 //
 // This is safe for concurrent access.
 func (b *Buffer) Write(entries ...Entry) {
+	// We could ask Entry if it supports time, but using ingest time is
+	// pretty standard as well. So we wrap each entry in a timed entry
+	// with the time it arrived, so that later LogMerge can make sense
```
piotrromanowski, line 78:
```suggestion
	// with the time it arrived, so that later Merger can make sense
```

internal/server/logbuffer/logbuffer.go
```diff
@@ -223,6 +236,38 @@ func (r *Reader) Read(max int, block bool) []Entry {
...
+	// Convert back to the raw Entry's because that's what the API expects.
+	entries := make([]Entry, len(timedEntries))
+	for i, te := range timedEntries {
+		entries[i] = te.entry
+	}
+
+	return entries
+}
+
+// Next returns the next entry. This is used by LogMerge to merge multiple
```
piotrromanowski, line 258:
```suggestion
// Next returns the next entry. This is used by Merger to merge multiple
```

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+	// Time is the associated time of the value. This is used to sort
+	// the entry against other entries from other inputs.
+	Time() time.Time
+
+	// Value returns the log entry itself. This allows time addition
+	// wrappers to return their original value.
+	Value() interface{}
+}
+
+// MergeReader is value that returns TimedEntry's for LogMerge
```
piotrromanowski, line 27:
```suggestion
// MergeReader is value that returns TimedEntry's for Merger
```

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+// to weave together.
+type MergeReader interface {
+	Next() (TimedEntry, error)
+}
+
+// NewMerger creates a new LogMerge, with the stream generated
```
piotrromanowski, line 33:
```suggestion
// NewMerger creates a new Merger, with the stream generated
```

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+	entries[bestIdx] = nil
+
+	return best, bestInput
+}
+
+// ReaderEntry is returned by ReadNext. It provides access to the TimedEntry
+// that is next as well as the input that generated the entry. This
+// type is important because it allows the caller to figure out the context
+// of the entry from the input. Because LogMerge is going to effectively
```
piotrromanowski, line 120:
```suggestion
// of the entry from the input. Because Merger is going to effectively
```

internal/server/logbuffer/logbuffer.go
```diff
@@ -71,6 +73,18 @@ func New() *Buffer {
...
+	// We could ask Entry if it supports time, but using ingest time is
+	// pretty standard as well. So we wrap each entry in a timed entry
+	// with the time it arrived, so that later LogMerge can make sense
+	// of the total order across multiple buffers.
+	timedEntries := make([]timedEntry, len(entries))
+	for i, ent := range entries {
+		timedEntries[i] = timedEntry{
+			ingestTs: time.Now(),
+			entry:    ent,
+		}
```
piotrromanowski, line 85: Hmmm, What if we provided a constructor for timedEntry so that we didn't have to provide the current time each time

    func NewTimedEntry(entry Entry) timeEntry {
        return &timedEntry{
            ingestTs: time.Now(),
            entry: entry
       }
   }

making this callsite a little cleaner and not requiring the consumer to provide the current time each time

      timedEntries[i] = NewTimedEntry(ent)

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+// NewMerger creates a new LogMerge, with the stream generated
+// from the given inputs.
+func NewMerger(readers ...MergeReader) *Merger {
+	var lm Merger
+	lm.readers = readers
+
+	lm.heads = lm.makeHeads()
+
+	return &lm
```
piotrromanowski, line 41: nit: could we simplify this to 

      return &Merger{
               readers: readers,
               heads: make([]TimedEntry, 0, len(readers)),
     }

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	ent := (*t)[0]
+
+	*t = (*t)[1:]
+
+	return ent, nil
+}
+
+// Create a slice to be used by refillEntries and findNext
```
piotrromanowski, line 61: Is this still being planned to be used in these functions? Or is this outdated documentation?

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for i := 0; i < count; i++ {
+		pop := l.refillEntries(heads)
+		if pop == 0 {
+			break
+		}
+
+		ent, ip := l.findNext(heads)
+		out = append(out, ReaderEntry{TimedEntry: ent, Reader: ip})
+	}
```
piotrromanowski, line 143: Hmmm, have we thought about moving the sorting to the constructor to do it upfront and then save on cost during the reads?

piotrromanowski (commented):
Still wrapping my head around this patch, but this is cool!

internal/server/logbuffer/logbuffer.go
evanphx: There isn't a reason too, it's completely internal.

internal/server/logbuffer/merge.go
evanphx: This is the comment for makeHeads. Not sure what you mean?

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+// TimedEntry is the interface each input returns entries in
+type TimedEntry interface {
+	// Time is the associated time of the value. This is used to sort
+	// the entry against other entries from other inputs.
+	Time() time.Time
+
+	// Value returns the log entry itself. This allows time addition
+	// wrappers to return their original value.
+	Value() interface{}
```
piotrromanowski, line 24: nit: `Value() Entry` may be clearer here (Even though it's also an empty interface{})

internal/server/logbuffer/merge.go
evanphx: Which sorting? This is the core of the algorithm and it has to happen here.

internal/server/logbuffer/merge.go
evanphx: Could!

internal/server/logbuffer/merge.go
piotrromanowski: Ahh I misinterpretted this as "refilEntries and findNext will use this function"

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return pop
+}
+
+// Find the entry in entries with earliest time, returning the entry and the
+// input that generated it.
+func (l *Merger) findNext(entries []TimedEntry) (TimedEntry, MergeReader) {
+	var (
+		best      TimedEntry
+		bestIdx   int
```
piotrromanowski, line 92: nit: maybe `earliest` and `earliestIndex` would be clearer?

internal/server/logbuffer/merge.go
piotrromanowski: Sorry no current sorting. I meant not having to do the lookup for earliest Entries each time you call Next() and instead move it to be done upfront once

internal/server/logbuffer/logbuffer.go
piotrromanowski: got it, make sense!

internal/server/logbuffer/merge.go
evanphx: The reason it's not Entry is if it were, then the implementors would have to import logbuffer to satisfy the interface.

internal/server/logbuffer/merge.go
evanphx: I feel like best is clearer.

internal/server/logbuffer/merge.go
evanphx: Because there isn't the ability to know to move it, that's how the algorithm works. There isn't visibility into which one will be earliest until all readers have had a chance to contribute a value.

internal/server/logbuffer/merge.go
piotrromanowski: Oh, I missed that, good point!

internal/server/singleprocess/service_logs.go
```diff
@@ -305,14 +304,160 @@ func (s *service) GetLogStream(
...
+	for _, record := range records {
+		r := record.LogBuffer.Reader(backlog)
+		readerToInstance[r] = record
+
+		readers = append(readers, r)
+	}
+
+	lm := logbuffer.NewMerger(readers...)
+
+	lines := make([]*pb.LogBatch_Entry, 60)
```
piotrromanowski, line 358: I'm Curious, how did we choose "60" as the line log batches value here? Is this something worth documenting?

internal/server/singleprocess/service_logs.go
```diff
@@ -305,14 +304,160 @@ func (s *service) GetLogStream(
...
+
+	// Read out all the log entries from LogMerge. This never blocks waiting
+	// for new log entries, it will simply let each reader output all known
+	// entries and then loop exits.
+	for {
+		entries, err := lm.ReadNext(60)
```
piotrromanowski, line 364: This log mergers usage ended up being:
https://ih1.redbubble.net/image.1401903701.4969/flat,750x,075,f-pad,750x1000,f8f8f8.jpg

piotrromanowski (commented):
I haven't gone through all the tests but otherwise this looks good to me! :) Thanks for bearing with me as I was wrapping my head around the algorithm for fetching next entries

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Value returns the log entry itself. This allows time addition
+	// wrappers to return their original value.
+	Value() interface{}
+}
+
+// MergeReader is value that returns TimedEntry's for Merger
+// to weave together.
+type MergeReader interface {
+	Next() (TimedEntry, error)
```
mitchellh, line 30: Another Nitpick... `Next` feels a bit too vague. Perhaps `NextTimedEntry` here. Its a nitpick but the motivation is looking at the godoc for `Reader` it feels weird that Next returns a TimedEntry and Read doesn't. I think it'd be clearer to do this and make it easier for existing types to implement this (since `Next` might reasonably be implemented for OTHER things). 

Just a thought.

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+	var lm Merger
+	lm.readers = readers
+
+	lm.heads = lm.makeHeads()
+
+	return &lm
+}
+
+// TimedEntries is a convience type of TimedEntry's that provides
+// the LogMergeInput interface.
```
mitchellh, line 45: Old interface

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+
+	ent := (*t)[0]
+
+	*t = (*t)[1:]
+
+	return ent, nil
+}
+
+// Create a slice to be used by refillEntries and findNext
+func (l *Merger) makeHeads() []TimedEntry {
```
mitchellh, line 62: This is only called by `New`, why don't we just do this directly in new without making a function for it?

internal/server/logbuffer/merge.go
```diff
@@ -0,0 +1,146 @@
...
+// shuffle the values that are put into it, the caller is going to have to deal
+// with entries appearing in any order and the input provides critical context.
+type ReaderEntry struct {
+	TimedEntry
+	Reader MergeReader
+}
+
+// ReadNext returns a slice of InputEntrys that are next in total time order.
+// The result might be fewer than count values, depending on what is available.
+func (l *Merger) ReadNext(count int) ([]ReaderEntry, error) {
```
mitchellh, line 130: To make this more similar to `Reader`, should this just be `Read` as well?

internal/server/singleprocess/service_logs.go
mitchellh: I'd also pull this out into a function-local `const`, since we use it twice it appears.

internal/server/logbuffer/merge.go
evanphx: Sure!

internal/server/logbuffer/merge.go
evanphx: Ah true, as an interface it's a bit to vague.

internal/server/singleprocess/service_logs.go
```diff
@@ -305,14 +313,161 @@ func (s *service) GetLogStream(
...
+			})
+		}
+	}
+
+	// Step 2: startup background forwarders for all the readers we spawned above.
+
+	// We lock around this section because if one of the launched goroutines exits
+	// very quickly, we might still be adding the others when it does exit.
+	{
+		instanceSetLock.Lock()
```
mitchellh, line 432: This is the only thing that bothers me about the stream changes, is that we _almost kind of_ duplicate this logic twice (here and just below). I didn't think about it too hard, but is there not some common logic we can extract into a func to setup the log forwarding or something and call that in both places? The juggling of the `instanceSetLock` in particular in two places kind of makes me nervous.

mitchellh (commented):
Finally got around to reviewing the API changes. Just one comment, thanks for adding tests too!

internal/server/singleprocess/service_logs.go
evanphx: So I refactored this code a few different ways to try and deal with that lock and their were all either confusing and didn't generally improve the understanding. Because of the bifurcated nature of the types of readers, this ended up being the version that, while did require some mutex manipulation duplication, was the cleanest to read and understand.

mitchellh: Fair enough!

mitchellh (approved):

briancain (approved):
👍🏻 📝

PR #1446, opened 2021-05-07
cli/up: Only print inplace deploy message if generation id exists

briancain:
This commit fixes a UI output bug where Waypoint would declare the
deployment was mutable and deployed in place for immutable deploys due
to the inplace check not looking for generation id existing.

I apparently missed the up cli fix. For reference this was already fixed in deploy/release https://github.com/hashicorp/waypoint/pull/1366

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -58,6 +58,7 @@ func (c *UpCommand) Run(args []string) int {
 		// inplace is true if this was an in-place deploy. We detect this
 		// if we have a generation that uses a non-matching sequence number
 		inplace := result.Deploy.Deployment.Generation != nil &&
+			result.Deploy.Deployment.Generation.Id != "" &&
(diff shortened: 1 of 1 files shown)


krantzinator (approved):
Should this chunk of logic be in a little helper func somewhere? If we have to fiddle with it again, we can consider moving it at that time.

mitchellh (approved):

briancain:
> Should this chunk of logic be in a little helper func somewhere? If we have to fiddle with it again, we can consider moving it at that time.

Hrm, it definitely could. I'm not sure where though! If we had a util package for CLI then it could go there, but I don't think we do yet. Like for example we just made the clisnapshot helpers its own package completely....but I don't see a good place for something like this to go 🤔 But yeah def, if a third PR is required in another place for this check we should make a util or something for this 👍🏻

PR #1447, opened 2021-05-07
Poll EC2 AutoScaling activity to check for failures and cancellations

### LINKED ISSUE #1386: EC2 Deploy waits indefinitely
**Describe the bug**
I'm attempting to use the aws-ec2 plugin to deploy a single instance to a non-default VPC. The AWS I'm working in doesn't have a default VPC. Everything seems to work but CLI waits indefinitely with the message: "Waiting for the first instance to start..."

**Steps to Reproduce**

I'm running the following via CLI:
```
waypoint up -app ec2-web -vvv
```

Here's what my waypoint.hcl app looks like:
```
app "ec2-web" {
  build {
    use "aws-ami" {
      region = "us-west-2"
      name = "ubuntu/images/hvm-ssd/ubuntu-focal-20.04-arm64-server-20210315"
    }
  }

  deploy {
    use "aws-ec2" {
      count {
        desired = 1
        min = 1
        max = 2
      }
      instance_type = "t3.nano"
      key = "beefcake"
      region = "us-west-2"
      service_port = 8080
      subnet = "subnet-89ef61d3"
//      security_groups = ["sg-37d22f44"]
    }
  }

  release {
    use "aws-alb" {
      count = 1
      subnets = ["subnet-89ef61d3"]
    }
  }
}
```

**Expected behavior**
An EC2 instance is deployed.

applicable:

* Waypoint CLI Version: 0.3.1
* Waypoint Server Platform and Version: docker
* Waypoint Plugin: aws-ec2

**Additional context**
I've confirmed via AWS console that certain things are getting created such as:
* Auto Scaling Group
* Security Group
* Launch Configuration
* Target Group

The VPC in question has 3 subnets of which just 1 is public. That subnet is specified above in the deploy and release stanzas.

Any guidance would be greatly appreciated.

Thanks!

catsby:
When the EC2 plugin launches an AutoScaling Group, waypoint polls until it detects at least 1 instance is running. If there is a misconfiguration (as seen in [#1386](https://github.com/hashicorp/waypoint/issues/1386#issuecomment-834624477)) and the instances cannot launch, waypoint polls indefinitely. 

In this PR we periodically check the ASG activities endpoint and exit the deployment step if  we find any cancellations or failures. 

Example output of #1386 with this patch applied: 

https://user-images.githubusercontent.com/61721/117514176-de825b80-af58-11eb-9ad7-fac8d9cb964a.mov

Fixes #1386


builtin/aws/ec2/platform.go
```diff
@@ -243,7 +246,36 @@ func (p *Platform) Deploy(
...
+			var lastStatusCode, lastStatusMessage string
+			if len(dsa.Activities) > 0 {
+				lastStatusCode = aws.StringValue(dsa.Activities[0].StatusCode)
+				lastStatusMessage = aws.StringValue(dsa.Activities[0].StatusMessage)
+			}
+
+			if lastStatusCode == autoscaling.ScalingActivityStatusCodeFailed ||
+				lastStatusCode == autoscaling.ScalingActivityStatusCodeCancelled {
+				return nil, status.Errorf(codes.FailedPrecondition, "error setting up autoscaling group. Last status code (%s): \n Message: %s", lastStatusCode, lastStatusMessage)
+			}
```
briancain, line 275: Would it make sense to put an else branch here and log the `lastStatusCode` at a DEBUG/TRACE level? Or does the AWS client already give us that in its logs 🤔

briancain (approved):
Looks good to me @catsby ! 🎉

builtin/aws/ec2/platform.go
catsby: We discussed internally and decided against the mentioned `else` clause here. Resolving this, thanks

PR #1458, opened 2021-05-11
Add context inspect (and better context help)

### LINKED ISSUE #1107: `path` command or subcommand for Waypoint artifacts
If you're flipping between OSes in managing Waypoint, you may not know the default paths for configs, etc.

Either a command (e.g. `waypoint path context`) or artifact-specific subcommand (e.g. `waypoint context path`)

using `find` or my favorite search engine to ask the default config path for a specific OS flavor

evanphx:
This adds a new `context inspect` command that helps users discover what the context actually is.

Additionally, it enhances the help text for `waypoint context` to include the main config path, since it's common that a user is going to be interested in that value if they're looking at the context help.

Fixes #1107

### DIFF
--- a/internal/cli/context_inspect.go
+++ b/internal/cli/context_inspect.go
@@ -1,0 +1,98 @@
+package cli
+
+import (
+	"encoding/json"
+
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/pkg/flag"
+	"github.com/posener/complete"
+)
+
+type ContextInspectCommand struct {
+	*baseCommand
+
+	flagJson bool
+}
+
+func (c *ContextInspectCommand) Run(args []string) int {
+	// Initialize. If we fail, we just exit since Init handles the UI.
+	flagSet := c.Flags()
+	if err := c.Init(
+		WithArgs(args),
+		WithFlags(flagSet),
+		WithNoConfig(),
+		WithClient(false),
+	); err != nil {
+		return 1
+	}
+
+	def, err := c.contextStorage.Default()
+	if err != nil {
+		def = "<unknown>"
+	}
+
+	if c.flagJson {
+		data, err := json.MarshalIndent(map[string]interface{}{
+			"config_path":     c.homeConfigPath,
+			"default_context": def,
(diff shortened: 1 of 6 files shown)


internal/cli/context_inspect.go
```diff
@@ -0,0 +1,98 @@
...
+	return c.Flags().Completions()
+}
+
+func (c *ContextInspectCommand) Synopsis() string {
+	return "Output current context info."
+}
+
+func (c *ContextInspectCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint context inspect <name>=<value>
```
briancain, line 93: What is `name` and `value` here? Maybe this was accidentally left in since I don't think this CLI command takes an argument?

briancain: It could be nice if if accepted a context name though so you wouldn't have to switch the current context to see the context info:

```
Usage: waypoint context inspect <context_name>
```

briancain (changes requested):
Looks good! 🎉  Gave it a shot locally:

```
brian@localghost:kind-k8s λ waypoint context inspect

» Context Info:
      config path: /home/brian/.config/waypoint
  default context: install-1620771637
```

Otherwise had a question about the doc string and potentially an enhancement to support an argument `context_name` for inspect.

internal/cli/context_inspect.go
briancain: Actually maybe it's `[context_name]` I don't remember the accepted way to convey optional arguments for CLI doc help strings 😅

briancain: This might only be worthwhile if we end up showing more information with `inspect` now that I've written all of that out loud....

Since we're only showing default context and config path I don't think that makes sense, so I think just fixing up the doc string is good 👍🏻

evanphx: Oh bad copypasta, just need to remove it.

evanphx: I added the ability to inspect a specific context too.

izaaklauer (approved):

briancain (approved):
👍🏻 nice! perfect :)

PR #1461, opened 2021-05-12
Default the platform on upgrade to the one in the context

### LINKED ISSUE #1233: Use platform defined on Server for Upgrade/Uninstall commands if no -platform flag is provided

As of https://github.com/hashicorp/waypoint/pull/1000 and https://github.com/hashicorp/waypoint/pull/1129, Waypoint Server now stores what kind of platform its installed to via the install command. For upgrades and uninstalls (and potentially other CLI commands that require `-platform`), we should instead look and use the platform option stored on the Server and local context file rather than requiring the user to repeat themselves with the flag.

The easiest way to accomplish this first is to have each command look at the current context used to invoke the command, and see if there's a platform option stored. If there is not one, it might be an old context that was created prior to the feature. In this case it should error out requesting the user provide the platform flag.

Not doing this and requiring the platform flag as we do today

evanphx:
Fixes #1233

### DIFF
--- a/internal/cli/server_upgrade.go
+++ b/internal/cli/server_upgrade.go
@@ -47,41 +47,6 @@ func (c *ServerUpgradeCommand) Run(args []string) int {
 		return 1
 	}
 
-	// Error handling from input
-
-	// NOTE(briancain): This check will look different once https://github.com/hashicorp/waypoint/issues/1233
-	// is implemeneted. This func should initially look for the platform
-	// saved in the context and if none was included, fall back to look for an
-	// included platform flag like we do below
-	if !c.confirm && c.platform == "" {
-		c.ui.Output(confirmReqMsg, terminal.WithErrorStyle())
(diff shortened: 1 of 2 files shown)


internal/cli/server_upgrade.go
```diff
@@ -236,6 +236,10 @@ func (c *ServerUpgradeCommand) Run(args []string) int {
 
 	// We update the context config if the server addr has changed between upgrades
 	if originalCfg.Server.Address != contextConfig.Server.Address {
+		// Update the plaform here, basically to upgrade an older context that didn't
```
evanphx, line 239:
```suggestion
		// Update the platform here, basically to upgrade an older context that didn't
```

briancain (approved):
Looks good! Works as expected in my testing 👍🏻

izaaklauer (approved):

PR #1466, opened 2021-05-12
website: bring nested-node in scope for sub-pages

### LINKED ISSUE #1462: NestedNode fails to render properly
**Describe the bug**
The change made in #1214 doesn't seem to work. From markdown like this:

```
##### <NestedNode>logging.region</NestedNode>
```

The following html shows up in my browser:

Screen Shot 2021-05-11 at 5 37 11 PM

**Steps to Reproduce**
Use `<NestedNode>`

**Expected behavior**
The NestedNode is expanded and rendered by the website tooling.

kylemac:
closes #1462  (hopefully)

@evanphx - my initial test in `/content/docs/index.mdx` showed things were working but then I realized I hadn't put the `<NestedNode>` component in scope for `/commands` or `/plugins`. This fixes that issue but let me know if thats not exactly the issue you were observing.

### DIFF
--- a/website/pages/commands/[[...page]].jsx
+++ b/website/pages/commands/[[...page]].jsx
@@ -8,6 +8,9 @@ import {
 const NAV_DATA_FILE = 'data/commands-nav-data.json'
 const CONTENT_DIR = 'content/commands'
 const basePath = 'commands'
+import NestedNode from 'components/nested-node'
+
+const additionalComponents = { NestedNode }
 
 export default function DocsLayout(props) {
   return (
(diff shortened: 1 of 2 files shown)


evanphx:
This doesn't seem to work, I still get the same HTML as shown in #1462. Could the markdown be confusing things?

evanphx:
Could the issue be that `<NestedNode>` is used in an .mdx partial?

evanphx:
Ok, I did some testing and `<NestedNode>` works `content/plugins/aws-ecs.mdx` fine, but doesn't work on the partials that are included into there.

evanphx (changes requested):
Doesn't seem to work in partials still.

kylemac:
@evanphx ah okay. partials are a different story. one of the limitations of partials is that we don't yet support MDX components. that said, lemme dig in a little bit and see if there's a workaround for this

evanphx:
@kylemac We're going to close this because we need it on partials. When/if that comes back around, we'll make a new PR or reopen this one.

PR #1475, opened 2021-05-15
Pack Plugin support for non-default process types.

### LINKED ISSUE #598: Pack Plugin support for non-default process types.
Buildpacks support multiple types of processes beyond the default, which is web. Today the Waypoint pack plugin does not support that.

Be able to specify a different Buildpack image process type using the pack plugin for a Waypoint deployment.

**Additional context**
Here is [buildpacks documentation](https://buildpacks.io/docs/app-developer-guide/run-an-app/#non-default-process-type) on how to specify a launch process, including non-default ones.

Tylerian:
### Context
Buildpacks support multiple types of processes beyond the default, which is `web`.

### Problem
Today the Waypoint `pack` plugin does not support that.

Fixes #598

### DIFF
--- a/builtin/pack/builder.go
+++ b/builtin/pack/builder.go
@@ -52,6 +52,9 @@ type BuilderConfig struct {
 
 	// Files patterns to prevent from being pulled into the build.
 	Ignore []string `hcl:"ignore,optional"`
+	
+	// Process type that will be used when setting container start command.
+	ProcessType string `hcl:"process_type,optional" default:"web"`
 }
 
 const DefaultBuilder = "heroku/buildpacks:18"
@@ -133,12 +136,12 @@ func (b *Builder) Build(
 		AppPath:    src.Path,
 		Env:        b.config.StaticEnvVars,
 		Buildpacks: b.config.Buildpacks,
-
 		ProjectDescriptor: project.Descriptor{
 			Build: project.Build{
 				Exclude: b.config.Ignore,
 			},
 		},
+		DefaultProcessType: b.config.ProcessType,
 	}
 
 	err = client.Build(ctx, bo)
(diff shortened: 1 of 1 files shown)


builtin/pack/builder.go
```diff
@@ -52,6 +52,9 @@ type BuilderConfig struct {
 
 	// Files patterns to prevent from being pulled into the build.
 	Ignore []string `hcl:"ignore,optional"`
+	
+	// Process type that will be used when setting container start command.
+	ProcessType string `hcl:"process_type,optional" default:"web"`
```
briancain, line 57: Why default to `"web"`? I think we could set no default, update `ProcessType` to be `*string`, so if the config option is unset it just passes nil through. Then I think the `pack` package can handle the default type there instead 🤔

briancain (changes requested):
Thanks for opening a pull request on Waypoint! I've got a question about the default option, otherwise I think this change makes sense 😄

builtin/pack/builder.go
Tylerian: > Why default to "web"?

I made `web` the default value because that's what the `pack` package [is using itself](https://github.com/buildpacks/pack/blob/216a4d1de720d51a76ccfa0d397b26150263b494/internal/build/lifecycle_execution.go#L23).

> Then I think the pack package can handle the default type there instead 🤔

The idea of letting the actual `pack` package decide the default `ProcessType` sounds good to me, but bear in mind that it does so for [`PlatformAPI` versions greater than `0.4`](https://github.com/buildpacks/pack/blob/216a4d1de720d51a76ccfa0d397b26150263b494/internal/build/lifecycle_execution.go#L380).

evanphx: Setting it to web by default here is fine but i'm unsure what is reading that struct tag to populate it. Is that a feature of HCL I'm unaware of?

evanphx:
@Tylerian Could you add a .changelog entry for this change? Check out https://github.com/hashicorp/waypoint/pull/1461/files for an example.

PR #1477, opened 2021-05-17
Convert Docker to use new SDK Resource Manager

mitchellh:
This converts the Docker plugin to use the new [resource manager library](https://github.com/hashicorp/waypoint-plugin-sdk/pull/29) in the SDK. This will improve error handling and makes the code a bit more structured. There is no other functionality changes in this PR.


krantzinator (approved):

izaaklauer (approved):

builtin/docker/platform.go
```diff
@@ -349,12 +452,12 @@ func (p *Platform) pullImage(cli *client.Client, log hclog.Logger, ui terminal.U
 
 		// if we have images do not pull
 		if len(sum) > 0 {
+			s.Done()
```
briancain, line 455: I _think_ Evans recent work with stepgroups is going to fix this, but I noticed  this line never resolves as Done:

Screenshot from 2021-05-18 07-50-02

briancain: We could add an update here that's like:

```suggestion
			s.Update("Docker image %q up to date!", in)
			s.Done()
```

briancain (changes requested):
NIce! Looks good, I just had a request to update some of the output 😄 But otherwise works well for me 🎉 🐳

briancain (approved):
🐳 🎉

PR #1487, opened 2021-05-17
plugin/docker: Report on deploy status

briancain:
This commit introduces the implementation of the Status function for
docker on deploy. It queries the current deployed containers health and
looks for if it was configured using dockers built in health checks. If
not, the function attempts to make some basic assumptions for
determining the health of the container.

### DIFF
--- a/builtin/docker/platform.go
+++ b/builtin/docker/platform.go
@@ -212,6 +212,8 @@ func (p *Platform) Status(
 	s.Update("Finished building report for Docker platform")
 	s.Done()
 
+	// NOTE(briancain): Replace ui.Status with StepGroups once this bug
+	// has been fixed: https://github.com/hashicorp/waypoint/issues/1536
 	st := ui.Status()
 	defer st.Close()
 
--- a/builtin/docker/platform.go
+++ b/builtin/docker/platform.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
@@ -209,9 +208,6 @@ func (p *Platform) Status(
 	result.TimeGenerated = ptypes.TimestampNow()
 	log.Debug("status report complete")
 
-	// TODO(briancain): remove me GH #1469
-	time.Sleep(500 * time.Millisecond)
-
 	// update output based on main health state
 	s.Update("Finished building report for Docker platform")
 	s.Done()
@@ -227,9 +223,6 @@ func (p *Platform) Status(
 	} else {
 		st.Step(terminal.StatusError, fmt.Sprintf("Container %q is reporting not ready!", containerInfo.Name))
 	}
-
-	// TODO(briancain): remove me GH #1469
-	time.Sleep(500 * time.Millisecond)
 
 	return &result, nil
 }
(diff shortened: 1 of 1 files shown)


builtin/docker/platform.go
```diff
@@ -95,6 +104,136 @@ func (p *Platform) resourceManager(log hclog.Logger) *resource.Manager {
...
+
+	// TODO(briancain): remove me GH #1469
+	time.Sleep(500 * time.Millisecond)
+
+	// update output based on main health state
+	s.Update("Finished building report for Docker platform")
+	s.Done()
+
+	st := ui.Status()
+	defer st.Close()
```
evanphx, line 220: Per the discussion last week, don't use Status, use the same stepgroup you have here and set the step's status.

builtin/docker/platform.go
```diff
@@ -95,6 +104,136 @@ func (p *Platform) resourceManager(log hclog.Logger) *resource.Manager {
...
+	if result.Health == sdk.StatusReport_READY {
+		st.Step(terminal.StatusOK, fmt.Sprintf("Container %q is reporting ready!", containerInfo.Name))
+	} else if result.Health == sdk.StatusReport_PARTIAL {
+		st.Step(terminal.StatusWarn, fmt.Sprintf("Container %q is reporting partially available!", containerInfo.Name))
+	} else {
+		st.Step(terminal.StatusError, fmt.Sprintf("Container %q is reporting not ready!", containerInfo.Name))
+	}
+
+	// TODO(briancain): remove me GH #1469
+	time.Sleep(500 * time.Millisecond)
```
evanphx, line 232: Again, should be fixed, please retest after rebase!

evanphx (changes requested):

builtin/docker/platform.go
briancain: Ah yeah, switching back from Status to StepGroups, I remember now why I tried using the `ui.Status` directly. Every time I do something like this:

```golang
s.Update("some text to be printed!!")
s.Status(terminal.StatusOk)
```

The text shows up as just `ok` and not `"some text to be printed!!"`. Additionally if I do something like:

```golang
s.Update("some text to be printed!!")
s.Status(terminal.StatusWarn)
```

It just prints `warn` in a green status ok format.

builtin/docker/platform.go
```diff
@@ -95,6 +104,136 @@ func (p *Platform) resourceManager(log hclog.Logger) *resource.Manager {
...
+		result.Health = sdk.StatusReport_ALIVE
+	} else {
+		result.Health = sdk.StatusReport_PARTIAL
+	}
+
+	result.TimeGenerated = ptypes.TimestampNow()
+	log.Debug("status report complete")
+
+	// TODO(briancain): remove me GH #1469
+	time.Sleep(500 * time.Millisecond)
```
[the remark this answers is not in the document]
briancain: It looks like it! Gave it a couple of tries and it printed right every time 🎉

builtin/docker/platform.go
briancain: Quick note from our video sync: This code should be working as mentioned above, but the behavior described sounds like a bug. The right move is to still move away from Status, but we should address the output bug too.

briancain:
Rebased off `main` now instead of the core status report impl. branch!

krantzinator (approved):
Woooo 🚀

PR #1488, opened 2021-05-17
Feature: Implement core handling of component.Status

briancain:
This pull request introduces a new feature for gathering and reporting on the
status of deployed and released applications from Waypoint. Build off of https://github.com/hashicorp/waypoint-plugin-sdk/pull/23,
it introduces a new operation for handling the status report for a given platform.

Still left todo in future pull requests:

- Executing a status report query on an interval (like the git poller). There's some refactoring to make polling more generic and I didn't want to include that in this already large PR, so that work will come after this.
- Implementing gathering status reports for the builtin supported plugins. Docker is already done and covered in a subsequent PR https://github.com/hashicorp/waypoint/pull/1487
- Update `waypoint deployment list` UX to account for status report of deployment


internal/core/app_status_report.go
```diff
@@ -0,0 +1,221 @@
...
+// Builds a status report on the given deployment
+func (a *App) StatusReport(
+	ctx context.Context,
+	deployTarget *pb.Deployment,
+	releaseTarget *pb.Release,
+) (*pb.StatusReport, *sdk.StatusReport, error) {
+	// Query the artifact
+	var artifact *pb.PushedArtifact
+	if deployTarget != nil && deployTarget.Preload != nil &&
+		deployTarget.Preload.Artifact != nil {
```
piotrromanowski, line 31: I think we can avoid this last `deployTarget.Preload.Artifact != nil` check and get the same result right?

internal/core/app_status_report.go
```diff
@@ -0,0 +1,221 @@
...
+	c, err := a.createStatusReporter(ctx, &evalCtx)
+	if status.Code(err) == codes.Unimplemented {
+		c = nil
+		err = nil
+	}
+	if err != nil {
+		a.logger.Error("error creating component in platform", "error", err)
+		return nil, nil, err
+	}
+	defer c.Close()
```
piotrromanowski, line 69: Maybe I'm wrong, but if `c` is assigned nil on line 62 then is it possible to seg fault here if c is being de-refenced? (I didn't check to see if c is actually being derefenced, just thought I'd ask)

internal/core/app_status_report.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, err
+	}
+
+	log.Debug("initializing status report plugin")
+	// Potential bug here with k8s apply plugin
+	// Works with docker and k8s ok
+	c, err := componentCreatorMap[component.PlatformType].Create(ctx, a, hclCtx)
+	if err == nil {
+		// We have a status reporter configured, use that.
+		return c, nil
```
piotrromanowski, line 114: nit: I think it's usually preferred to have the "happy path" as the last return

https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88
Also mentioned by Dave Cheney https://dave.cheney.net/2020/02/23/the-zen-of-go

internal/core/app_status_report.go
briancain: No, we need the extra check because deployTarget might not of been preloaded with the passed through protobuf message, meaning just checking on Artifact will cause a panic with Preload being nil. You'll find a similar check in app release.

piotrromanowski: right, I'm saying we could keep `deployTarget != nil && deployTarget.Preload != nil` and leave off the last check because if it's nil, then `var artifact *pb.PushedArtifact` will just be assigned nil anyway. Does that make sense?

briancain: Ah, I see what you're saying. I had read that comment as the opposite. Yeah makes sense 👍🏻

internal/server/proto/server.proto
```diff
@@ -2150,6 +2181,117 @@ message Release {
...
+message StatusReport {
+
+  // The application that this build is part of.
+  Ref.Application application = 1;
+
+  // The workspace that this exists in
+  Ref.Workspace workspace = 2;
+
+  // the deployment id that this status report was generated on
+  string deployment_id = 3;
```
mitchellh, line 2260: This probably has to be a oneof cause this may also be a release ID right?

mitchellh (changes requested):
I think one major comment I have that would require some reshuffling (but not too bad): should status reports be slightly more separated from the things they can report on?

Right now, you have `core.StatusReport` and it takes a deployment and a release, but really that's a oneof (you set one or the other). If we wanted to add support for status reporting say... an artifact or some future custom workflow element (doesn't exist today) it'd be a bit of a slog.

One place that is particularly squishy is the HCL context setup, cause we have to be aware of every context to set things up in as we introduce them (i.e. releases are different from deployments and that's already a bug I commented on now).

A different path I could see being taken:

1. Add specific `ReleaseStatusReport` and `DeployStatusReport` functions to core. These would be responsible for using some shared logic to setup the HCL context and component.
2. These call into an unexported `statusReport` function that takes just the component and target.

In this world, adding a new component type is hopefully not too difficult and the unique component logic would be done by the component type file.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,214 @@
...
+				"error", err)
+			return nil, nil, err
+		}
+
+		artifact = resp
+	}
+
+	var evalCtx hcl.EvalContext
+	evalCtx.Variables = map[string]cty.Value{}
+	if err := evalCtxTemplateProto(&evalCtx, "artifact", artifact); err != nil {
```
mitchellh, line 54: If we're targeting a release we actually do need the deployment added here.

See my bigger comment though.

internal/cli/deployment_create.go
```diff
@@ -85,6 +97,25 @@ func (c *DeploymentCreateCommand) Run(args []string) int {
 			}
 
 			releaseUrl = releaseResult.Release.Url
+
+			// NOTE(briancain): Because executeReleaseOp returns an initialized struct
+			// of release results, we need this deep check here to really ensure that a
+			// release actually happened, otherwise we'd attempt to run a status report
+			// on a nil release
+			if releaseResult != nil && releaseResult.Release != nil &&
+				releaseResult.Release.Release != nil {
```
evanphx, line 106: I believe the crash was fixed with the bump to argmapper 0.2 so double check the depth of checking we need here.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+		c = nil
+		err = nil
+	}
+	if err != nil {
+		a.logger.Error("error creating component in platform", "error", err)
+		return nil, nil, err
+	}
+
+	a.logger.Debug("starting status report operation")
+	statusReporter, ok := c.Value.(component.Status)
```
evanphx, line 49: `c` can be nil here from the Unimplemented block, which will crash I believe?

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+	hclCtx := evalCtx.NewChild()
+	if _, err := a.deployEvalContext(ctx, hclCtx); err != nil {
+		return nil, nil, err
+	}
+
+	c, err := a.createStatusReporter(ctx, hclCtx, component.PlatformType)
+	if status.Code(err) == codes.Unimplemented {
+		c = nil
+		err = nil
+	}
```
evanphx, line 42: Should this just return an empty status report right away? It looks like the code below here doesn't deal with `c` being nil.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+	hclCtx := evalCtx.NewChild()
+	if _, err := a.deployEvalContext(ctx, hclCtx); err != nil {
+		return nil, nil, err
+	}
+
+	c, err := a.createStatusReporter(ctx, &evalCtx, component.ReleaseManagerType)
+	if status.Code(err) == codes.Unimplemented {
+		c = nil
+		err = nil
+	}
```
evanphx, line 80: Ditto with the code in DeployStatusReport.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var status *sdk.StatusReport
+	if result != nil {
+		status = result.(*sdk.StatusReport)
+	}
+
+	return msg.(*pb.StatusReport), status, nil
```
evanphx, line 123: Probably best to guard this type cast with the checked version and return an error if msg was not a `*pb.StatusReport` in case there is a bad plugin, we don't want it to crash.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+		},
+
+		LoadDetails: pb.Deployment_ARTIFACT,
+	})
+	if status.Code(err) == codes.NotFound {
+		// TODO: fix panic. Should this be an error though?
+		resp = nil
+		err = nil
+		a.logger.Warn("deployment not found, will attempt status report regardless",
+			"deployment_id", r.DeploymentId)
```
evanphx, line 186: Best solution to the deref crash of `resp` below is probably just return and not set the variables here.

internal/runner/operation_status_report.go
```diff
@@ -0,0 +1,62 @@
...
+	app, err := project.App(job.Application.Application)
+	if err != nil {
+		return nil, err
+	}
+
+	op, ok := job.Operation.(*pb.Job_StatusReport)
+	if !ok {
+		// this shouldn't happen since the call to this function is gated
+		// on the above type match.
+		panic("operation not expected type")
```
evanphx, line 25: Best to return an error rather than panic, even though it's not expected. Unexpected things happen everyday.

internal/server/singleprocess/state/status_report.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+var statusReportOp = &appOperation{
+	Struct: (*pb.StatusReport)(nil),
+	Bucket: []byte("statusreport"),
+
+	MaximumIndexedRecords: 1, // Only store the "latest" status report
```
evanphx, line 11: Should we also be pruning the records out of boltdb then too?

evanphx (changes requested):
A few things around some panics and structure.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+
+func (a *App) createStatusReporter(
+	ctx context.Context,
+	hclCtx *hcl.EvalContext,
+	componentType component.Type,
+) (*Component, error) {
+	log := a.logger
+
+	log.Debug("initializing status report plugin")
+	c, err := componentCreatorMap[componentType].Create(ctx, a, hclCtx)
```
briancain, line 219: Double check that this returns something even if Status isn't implemented.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,340 @@
...
+
+	switch t := op.Target.(type) {
+	case *pb.Deployment:
+		op.DeployTarget = t
+	case *pb.Release:
+		op.ReleaseTarget = t
+	}
+
+	if op.DeployTarget != nil && op.DeployTarget.Deployment != nil {
+		args = append(args, argNamedAny("target", op.DeployTarget.Deployment))
```
briancain, line 322: Move these checks inline to the switch above on 314

internal/server/singleprocess/state/status_report.go
briancain: RE: video review, this is doing the right thing by storing 1 report.

internal/core/app_status_report.go
briancain: From the testing now, it looks like you're right in that even if Status is unimplemented, the component creator still returns something and doesn't raise an error. We do get a component back, so this conversion doesn't crash. I'll still clean this up to handle this scenario properly where `c` ends up nil

internal/runner/operation_status_report.go
briancain: Leaving a note here, during our sync I noted that our other operations all have this pattern, so we decided to leave it in and have plans in the future to fix up all of the operations to yank out this pattern.

internal/core/app_status_report.go
briancain: I resolved this for now by returning early when we get an Unimplemented error so that should prevent this here.

internal/cli/deployment_create.go
briancain: Another note here...we talked about this in our sync. Argmapper handles nil better and doesn't panic anymore but I think it's still worth guarding against this since there was no release at all. Otherwise we will still attempt a status report and return an empty report, so this saves us some extra func calls.

internal/core/app_status_report.go
```diff
@@ -0,0 +1,356 @@
...
+	var statusReport *sdk.StatusReport
+	if result != nil {
+		statusReport = result.(*sdk.StatusReport)
+	}
+
+	reportResp, ok := msg.(*pb.StatusReport)
+	if !ok {
+		return nil,
+			status.Errorf(codes.FailedPrecondition, "unsupported status report response returned from plugin")
+	} else {
```
mitchellh, line 128: No need for this `else` block cause the if above returns.

internal/server/proto/server.proto
```diff
@@ -2242,6 +2271,112 @@ message Release {
...
+  Status status = 5;
+
+  // id is the unique ID for this status report.
+  string id = 6;
+
+  // StatusReport is the raw report object encoded directly from the plugin.
+  google.protobuf.Any status_report = 7;
+
+  // Resources that make up the report. The root most report is expected
+  // to be the overall health associated with the report.
```
mitchellh, line 2362: "root most report" is what? the zero index element? Perhaps clarify this comment a bit.

mitchellh (approved):

PR #1490, opened 2021-05-18
docker: support build context

psihachina:
This commit adds the context support:

``` 
    build {
        use "docker" {
            dockerfile = "../../projects/hamster/Dockerfile"
            build_args = ["PROJECT_PATH=./projects/hamster"]
            context = "../../"
        }
    
    }
```

### DIFF
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -52,6 +52,9 @@ type BuilderConfig struct {
 
 	// Controls the passing of build time variables
 	BuildArgs map[string]*string `hcl:"build_args,optional"`
+
+	// Controls the passing of build context
+	Context string `hcl:"context,optional"`
 }
 
 func (b *Builder) Documentation() (*docs.Documentation, error) {
@@ -127,6 +130,11 @@ build {
 			"or img for the build step.",
 	)
 
+	doc.SetField(
+		"context",
+		"Build context path",
+	)
+
 	return doc, nil
 }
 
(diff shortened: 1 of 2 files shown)


evanphx:
Could you talk more about your use case for specifying the directory that that the Dockerfile is read from?

psihachina:
For example, a mono repository with such a structure:
image

AutomationD:
It would be great to have this, as it would be super useful to have shared libs in a monorepo structure, for example.

builtin/docker/builder.go
```diff
@@ -180,9 +188,18 @@ func (b *Builder) Build(
...
-	contextDir, relDockerfile, err := build.GetContextFromLocalDir(src.Path, dockerfile)
-	if err != nil {
-		return nil, status.Errorf(codes.FailedPrecondition, "unable to create Docker context: %s", err)
+	var contextDir string
+
+	if b.config.Context == "" {
+		contextDir, relDockerfile, err = build.GetContextFromLocalDir(src.Path, dockerfile)
+		if err != nil {
+			return nil, status.Errorf(codes.FailedPrecondition, "unable to create Docker context: %s", err)
+		}
```
evanphx, line 197: Since this is duplicated with the other branch of the if, could you pull it out and just have the `path` passed in be controlled by the presence of `b.config.Context`?

evanphx (changes requested):

evanphx:
@psihachina Looks good! Could you rebase on main to pickup the conflict? Then I'll be ready to merge.

evanphx (approved):

PR #1504, opened 2021-05-19
Very minor typo in a comment

izaaklauer:


briancain (approved):
Makes sense 👍🏻 Since it's just a doc comment I'd say probably not worth it, but generally this would probably need to come along with a proto gen update, but it's a small typo so probably not worth it 😄

PR #1509, opened 2021-05-19
internal/server: Variables added to the proto for project and job

krantzinator:
This is the base level foundation for supporting storing variable values on the server. In talking with @gregone, it would be best to merge this in now so that he can work on supporting this in the UI without the proto gens getting too far behind `main`.
Note: the changelog should come when the feature is complete, so I have marked `no-changelog` for now.


internal/server/proto/server.proto
```diff
@@ -287,6 +287,53 @@ message Application {
...
+  }
+
+  // Source tracks where a value came from for more helpful error messaging.
+  // This is optional. If unset, we will give a generic error message.
+  oneof source {
+    SourceCLIArg cliarg = 4;
+    SourceCLIFile file = 5;
+    SourceEnvVar envvar= 6;
+    SourceRepoFile repofile = 7;
+    SourceServer server = 8;
```
mitchellh, line 307: Just a nitpick on this...

First, I'd remove the `Source` prefix. We don't do that for any of our other `oneof`.

Next, I think we separate multiple words with underscoring? I'm not SURE but I think so. So `cli_arg`, `env_var`, etc.

However, I think some of those don't need to be so long, consider: `cli`, `file`, `env`, `vcs`, `server`.

internal/server/proto/server.proto
```diff
@@ -287,6 +287,53 @@ message Application {
...
+  }
+
+  // SourceCLIArg is set if a variable value is set via a -var flag
+  message SourceCLIArg {}
```
mitchellh, line 311: You may be planning on adding fields later, but you can just use `google.protobuf.Empty` instead of creating an empty message. if we ever introduce structure later, this is completely backwards compatible since protobuf if structural/tag-based and doesn't care about the name of a thing.

internal/server/proto/server.proto
```diff
@@ -287,6 +287,53 @@ message Application {
...
+
+  // SourceCLIFile is set if a variable value is set via a -var-file flag
+  message SourceCLIFile {
+    string fileName = 1;
+    bytes hclPos = 2;
```
mitchellh, line 316: What do you expect these `bytes` to be?

In my mind, we'd probably want to recreate the `hcl.Pos` and/or `hcl.Range` types in proto: https://pkg.go.dev/github.com/hashicorp/hcl/v2#Pos And use that structurally. It should be relatively easy to get these out of HCL.

internal/server/proto/server.proto
krantzinator: Nice, thanks!

internal/server/proto/server.proto
krantzinator: Hmmm yeah I think those are good. The comments are there to be more explicit and that + the names make sense.

internal/server/proto/server.proto
krantzinator: I was expecting the `bytes` to be the Pos/Range data, yes, and abstractly was assuming I'd convert those to the HCL type once I got the bytes out of the proto message. By "recreate in proto" do you mean create a new message type? Cause that sounds good too.

internal/server/proto/server.proto
```diff
@@ -332,6 +390,9 @@ message Project {
   // change.
   string file_change_signal = 8;
 
+  // Variable values stored on the server. These values are set in the UI.
+  repeated Variable variables = 9;
```
mitchellh, line 394: I would modify the comment to add that these are still a work in progress, experimental, and so far do nothing so that if we make a releaes without variables complete, we aren't misleading folks who might be reading this.

mitchellh (approved):
One small comment change.

internal/server/proto/server.proto
```diff
@@ -691,6 +752,13 @@ message Job {
   DataSource data_source = 6;
   map<string, string> data_source_overrides = 7;
 
+  // variables store the key/value pairs of parsed variables; the parse prior to
+  // running a job only verifies syntax correctness. Verifying type checks and
+  // the presence of required values will both need to be done in the job's 
+  // validation
+  repeated Variable variables = 9;
+  // map<string, string> variables = 9;
```
briancain, line 760: Should this be removed?
```suggestion
```

briancain (approved):
Looks great! 🎉 📨

internal/server/proto/server.proto
krantzinator: Yes :D

PR #1526, opened 2021-05-21
Install the correct version of mockery.

izaaklauer:
Without specifying a mockery package, nothing gets installed.

Without injecting the version via ldflags, mockery prepends the wrong version number (v1.0.0) to all of the mocks that it generates. I do not love hard-coding 1.1.2 in here, but this seems unlikely to change frequently and I don't see another path. Bash trickery inside `go:generate` seems to not work and would be a bit hacky anyway.

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -19,5 +19,5 @@ import _ "github.com/golang/protobuf/protoc-gen-go"
 //go:generate go install github.com/mitchellh/protoc-gen-go-json
 import _ "github.com/mitchellh/protoc-gen-go-json"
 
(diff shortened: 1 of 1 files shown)


briancain (approved):
Looks good to me! 👍🏻

krantzinator (approved):
sweet

krantzinator:
Regarding the hardcoding: that's fine, because if we ever updated mockery we'd have to update the pinned versions in the nix shell and re-run a `gen/server` to update all the mocks anyway, so lots of updates to be had all around.

PR #1529, opened 2021-05-22
ListWorkspaces API supports filtering by project/app

mitchellh:
This adds filtering capabilities to `ListWorkspaces` to list workspaces only in use by a specific project or app. I added this to better support our frontend team as they implement workspaces in the UI. 

@hashicorp/waypoint-frontend: **Note** that ListWorkspaces only returns workspaces that _already exist_. Workspaces are only created when they're used. So even the "default" workspace will only exist once a single operation has been done. For the UI purposes, you may want to default the empty result to add "default" since that will be the default workspace used.

This point is also important for **creating workspaces**. You'll notice there is no "Create workspace" API today. Workspaces are automatically created the first time they're used. 

Please let me know if you need any different behavior 😄


gregone (approved):
Looks Good! Thanks for adding that to the API

izaaklauer (approved):

briancain (approved):

PR #1533, opened 2021-05-24
UI: Use Mirage’s DB/ORM/Factory in tests and dev

jgwhite:
## What is this?

This change updates our UI development/testing stack to use [Mirage’s full DB/ORM/factory system](https://miragejs.com/docs/main-concepts/orm/) to manage fixture data.

Where previously we generated protobufs in `mirage/services/*.ts`, now we’ll manage fixtures using Mirage’s declarative API.

This gives us more control over fixture data, both in tests and in development mode.

## What does it look like?

Here’s what it looks like to create fixtures on-the-fly:

https://user-images.githubusercontent.com/34030/119402444-3a7d0c00-bcdd-11eb-8dec-268f8557e6f3.mp4

For an illustration of how this will aid our testing, see the diffs for our acceptance tests. Note particularly that we are no longer asserting on values that appear in global fixtures, and can instead declare *exactly* the data a particular tests needs:

**Before:**

```ts
await page.visit();

// Currently no way to seed past the default in mirage/services/builds.ts
assert.equal(page.buildList.length, 4);
```

**After:**

```ts
let project = this.server.create('project', { name: 'microchip' });
let application = this.server.create('application', { name: 'wp-bandwidth', project });
this.server.createList('build', 4, 'random', { application });

await page.visit();

assert.equal(page.buildList.length, 4);
```

## How does this work with grpc-web?

1. We add `toProtobuf` and in some cases `toProtobufRef` methods to our Mirage models.
2. In our mock gRPC handlers we retrieve the appropriate models from Mirage’s DB and then convert them to Waypoint Protobufs using aforementioned methods.
3. We introduce [`encode`](https://github.com/hashicorp/waypoint/blob/ui-use-full-mirage-orm/ui/mirage/helpers/protobufs.ts#L4-L28) and [`decode`](https://github.com/hashicorp/waypoint/blob/ui-use-full-mirage-orm/ui/mirage/helpers/protobufs.ts#L30-L50) functions for working with the protobuf wire format.

## Why are there so many TODOs? Why doesn’t `mirage/models` cover every type?

The exercise of mapping Mirage models to Waypoint Protobufs turns out to be a good way to familiarize oneself with the API and relationships between messages and entities in the system. Rather than presenting this work as “complete” I think it’s better to merge the groundwork presented here and then bring in the whole UI team to keep improving it together.

## How should I verify/review this change?

Something like this:

1. Check out the branch
2. `yarn ember serve`
3. Visit [http://localhost:4200/tests](http://localhost:4200/tests)
4. Verify everything’s green
5. Visit [http://localhost:4200](http://localhost:4200)
6. Verify everything is behaving more-or-less as it was before
7. Try playing around with the fixtures in dev and test
8. Report anything that breaks in a mysterious way

### DIFF
--- a/ui/mirage/models/project.ts
+++ b/ui/mirage/models/project.ts
@@ -1,0 +1,31 @@
+import { Model, hasMany } from 'ember-cli-mirage';
+import { Project, Ref } from 'waypoint-pb';
+
+export default Model.extend({
+  applications: hasMany(),
+
+  toProtobuf(): Project {
+    let result = new Project();
+
+    result.setApplicationsList(this.applications.models.map((a) => a.toProtobuf()));
+    // TODO: result.setDataSource(...)
+    // TODO: result.setDataSourcePoll(...)
+    // TODO: result.setExtension(...)
+    result.setFileChangeSignal(this.fileChangeSignal);
+    result.setName(this.name);
+    result.setRemoteEnabled(this.remoteEnabled);
+    result.setWaypointHcl(this.waypointHcl);
+    result.setWaypointHclFormat(Project.Format.HCL);
+
+    return result;
+  },
+
+  toProtobufRef(): Ref.Project {
+    let result = new Ref.Project();
+
+    // TODO: result.setExtension(...)
+    result.setProject(this.name);
+
+    return result;
+  },
+});
--- a/ui/mirage/factories/application.ts
+++ b/ui/mirage/factories/application.ts
@@ -1,0 +1,12 @@
+import { Factory, trait } from 'ember-cli-mirage';
+import faker from '../faker';
+
+export default Factory.extend({
+  simple: trait({
+    name: 'simple-application',
+  }),
+
+  'with-random-name': trait({
+    name: () => `wp-${faker.hacker.noun()}`,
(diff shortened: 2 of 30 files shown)


ui/mirage/models/project.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    result.setApplicationsList(this.applications.models.map((a) => a.toProtobuf()));
+    // TODO: result.setDataSource(...)
+    // TODO: result.setDataSourcePoll(...)
+    // TODO: result.setExtension(...)
+    result.setFileChangeSignal(this.fileChangeSignal);
+    result.setName(this.name);
+    result.setRemoteEnabled(this.remoteEnabled);
+    result.setWaypointHcl(this.waypointHcl);
+    // TODO: result.setWaypointHclFormat(...)
```
gregone, line 18: looks like this one is done :)

gregone (approved):
Looks Great! Thank you for tackling that!

ui/mirage/models/project.ts
jgwhite: Good spot!

Issue #1506, opened 2021-05-19
Replace `{{inline-svg }}` icons with `<Pds::Icon/>` component


gregone:
**Is your feature request related to a problem? Please describe.**
Since the migration to [Structure](https://github.com/hashicorp/structure) , the Waypoint Browser UI renders icons using the `inline-svg` helper, and in some places, uses the new `<Pds::Icon />` component. We should standardize to using the new pattern. 

**Describe the solution you'd like**
Replace all icon instances currently using `inline-svg` with `Pds::Icon`

PR #1534, opened 2021-05-24
UI: Updated SVGs to utilize Structure Icons component

sabrinako:
#1506

### DIFF
--- a/ui/app/components/app-item/deployment.hbs
+++ b/ui/app/components/app-item/deployment.hbs
@@ -5,12 +5,12 @@
       <code>{{@deployment.id}}</code>
       {{#if (eq @deployment.state 4)}}
         <b class="badge badge--destroyed">
-          {{inline-svg "trash" class="icon"}}&nbsp;Destroyed
+          <Pds::Icon @type="trash" class="icon" />&nbsp;Destroyed
         </b>
       {{/if}}
     </p>
     <small>
-      {{inline-svg (icon-for-component @deployment.component.name) class="icon"}}
+      <Pds::Icon @type={{icon-for-component @deployment.component.name}} class="icon" />
       <span>{{if (eq @deployment.status.state 1) 'Deploying' 'Deployed'}} by
         <b>{{component-name @deployment.component.name}}</b>
         {{#if (eq @deployment.status.state 1)}}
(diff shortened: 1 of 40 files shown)


jgwhite:
Added `pr/no-changelog` to this because it shouldn’t have any user-facing impact and therefore doesn’t require a changelog entry.

ui/app/components/app-item/deployment.hbs
```diff
@@ -22,9 +22,9 @@
     </small>
   </LinkTo>
   {{#if (and @deployment.preload.deployUrl (not-eq @deployment.state 4))}}
-    <ExternalLink href="https://{{@deployment.preload.deployUrl}}" class="button button--secondary">
+    <ExternalLink href="https://{{@deployment.preload.deployUrl}}" class="button button--secondary button--external-link">
```
jgwhite, line 25: As a future change, perhaps we should make `<ExternalLink>` apply `button--external-link` automatically 🤔

ui/app/components/header/index.hbs
```diff
@@ -1,16 +1,17 @@
...
   <nav>
     <ExternalLink href="https://waypointproject.io/docs">
-      {{inline-svg "docs" class="nav-icon"}}
+      <Pds::Icon @type="docs" class="nav-icon" />
       Documentation
     </ExternalLink>
     {{#if this.canLogout}}
       <hr />
       <Actions::Context />
+      <hr />
```
jgwhite, line 14: Nice touch!

ui/app/styles/_page.scss
```diff
@@ -675,20 +680,22 @@
...
-        height: scale.$lg-14;
-
         path {
           fill: color.$black;
           stroke: $brand-color;
         }
+
+        > .pds-icon__svgWrapper {
+          display:flex;
+        }
```
jgwhite, line 698: If there’s time, we could file an issue on the structure repo explaining why we needed to do this and suggest some ways to fix it at-source. It’s not a bug exactly but it would be better if we didn’t have to reach into the component like this.

jgwhite (approved):
Fabulous! Pixel-perfect throughout as far I can tell.

All my comments are suggestions for follow-up work.

As noted in Slack, let’s remove `ember-inline-svg` as part of this.

ui/app/styles/_page.scss
sabrinako: I just checked and it looks like some change I made later on fixed the need to add this. I removed this styling and we don't have to file the issue to the structure repo! 🙂

PR #1535, opened 2021-05-25
Generate TS API Bindings for variables and workspaces

gregone:
Generates TS Bindings for the new API changes


jgwhite (approved):
Looks good to me 👍

Issue #1521, opened 2021-05-20
internal/server/singleprocess: Flaky tests


krantzinator:
**Describe the bug**
Flaky tests in `internal/server/singleprocess`

**Steps to Reproduce**
Flakes from:
- ~`TestServiceRunnerJobStream_cancel`: [one](https://app.circleci.com/pipelines/github/hashicorp/waypoint/4393/workflows/3c705bad-3280-4e46-91ab-9c22c1636ee5/jobs/31799) and [two](https://app.circleci.com/pipelines/github/hashicorp/waypoint/5710/workflows/1a5d72db-0c69-4873-b52b-9ba984ef6d03/jobs/43406)~ Fixed by #1538 
- `TestServiceGetLogStream_depPlugin`: [one](https://app.circleci.com/pipelines/github/hashicorp/waypoint/4771/workflows/8ca9236e-eb93-4b79-8896-52431556b9f6/jobs/35491) and [two](https://app.circleci.com/pipelines/github/hashicorp/waypoint/5786/workflows/a602c260-80fd-4db5-b944-218852e3bd98/jobs/44248)
- [`TestServiceEntrypointExecStream_invalidSessionId`](https://app.circleci.com/pipelines/github/hashicorp/waypoint/2907/workflows/174cbbb0-caba-489c-8f20-c5e7658e59f0/jobs/20495)

**Expected behavior**
Less flakes.

**Waypoint Platform Versions**
All.

mitchellh:
So the `fault` errors are actually potentially bugs in BoltDB. They seem to pop up in _any_ test and the stack traces are deep in Bolt. I would separate those from this issue cause thats a bigger discussion of if we want to switch to the etcd fork since they maybe have fixed this...

The others are probably real flakes that are our fault. 😄

krantzinator:
Updated to remove the `fault` error tests, and will put that info in our planning doc instead.

evanphx:
Given the merger of #1538, we're going to presume that these are fixed.

PR #1538, opened 2021-05-26
Fixes job cancellation bug hidden behind race condition

izaaklauer:
Fixed the issue that caused these failures in CI:

- https://app.circleci.com/pipelines/github/hashicorp/waypoint/5710/workflows/1a5d72db-0c69-4873-b52b-9ba984ef6d03/jobs/43406
- https://app.circleci.com/pipelines/github/hashicorp/waypoint/4393/workflows/3c705bad-3280-4e46-91ab-9c22c1636ee5/jobs/31799

This was a tricky one to track down. From the failed tests, we can see that the client was receiving duplicate cancellation events. I was able to replicate the failed test by adding a sleep [here](https://github.com/hashicorp/waypoint/blob/35a9d6f2577957aeaa2c847fa6b2a53d295864d0/internal/server/singleprocess/service_runner.go#L281-L284).

Here’s what I think is happening:

There are three goroutines at play here.  One watching for DB changes and adding them to a channel ([A](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L245-L274)), one listening for client events and adding them to a channel ([B](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L279-L318)), and a big one pulling off the channels and doing work ([C](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L322-L379)). 

In the [final phase of this test](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner_test.go#L260-L273), the client [sends a completion event onto the stream](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner_test.go#L261-L265). [B gets it](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L284), and [passes it to C synchronously](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L302) via an unbuffered channel. [C handles it](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L341), which eventually results in a DB update on the job (marking its state as Success), and then the race begins. 

On one side, B is going to return, which will trigger its[ deferred `cancel()`](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L280), which will cause C to return and the test to pass. On the other side, the update that C just made to the job is going to get picked up by [A’s watchset](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L266), which will result in one more [job update](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L260) for [C to process](https://github.com/hashicorp/waypoint/blob/2359bec96add1281a9f478cf93a612d6ee07a5ab/internal/server/singleprocess/service_runner.go#L345) before `cancel()` comes around, which  will exercise the bug. 

Generally, the memdb watchset path is slower than cancelling the context, so C never sees the final “Success” update to the job, and we don’t exercise the bug.

The bug is that we were treating the completion event as a cancellation event with an updated timestamp. It looks like `*timestamppb.Timestamp`’s aren’t nicely comparable with `==` - they return not equal even if they have the same underlying time values. Converting them to `time.Time` first does the trick.

Addresses one of the problems in #1521

### DIFF
--- a/internal/server/singleprocess/service_runner.go
+++ b/internal/server/singleprocess/service_runner.go
@@ -351,7 +351,7 @@ func (s *service) RunnerJobStream(
 			// the cancel time changes. The cancel time only changes if multiple
 			// cancel requests are made.
 			if job.CancelTime != nil &&
-				(lastJob == nil || lastJob.CancelTime.AsTime() != job.CancelTime.AsTime()) {
+				(lastJob == nil || !lastJob.CancelTime.AsTime().Equal(job.CancelTime.AsTime())) {
 				// The job is forced if we're in an error state. This must be true
 				// because we would've already exited the loop if we naturally
 				// got a terminal event.
--- a/internal/server/singleprocess/service_runner.go
+++ b/internal/server/singleprocess/service_runner.go
@@ -3,7 +3,6 @@ package singleprocess
 import (
 	"context"
 	"io"
-	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
@@ -278,10 +277,7 @@ func (s *service) RunnerJobStream(
 	// so we can exit properly on client side close.
 	eventCh := make(chan *pb.RunnerJobStreamRequest, 1)
 	go func() {
-		defer func() {
-			time.Sleep(2 * time.Second)
-			cancel()
-		}()
+		defer cancel()
 
 		for {
 			log.Trace("waiting for job stream event")
--- a/internal/server/singleprocess/service_runner.go
+++ b/internal/server/singleprocess/service_runner.go
@@ -3,6 +3,7 @@ package singleprocess
 import (
 	"context"
 	"io"
+	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/go-memdb"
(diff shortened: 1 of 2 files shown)


internal/server/singleprocess/service_runner.go
```diff
@@ -351,7 +351,7 @@ func (s *service) RunnerJobStream(
 			// the cancel time changes. The cancel time only changes if multiple
 			// cancel requests are made.
 			if job.CancelTime != nil &&
-				(lastJob == nil || lastJob.CancelTime != job.CancelTime) {
+				(lastJob == nil || lastJob.CancelTime.AsTime() != job.CancelTime.AsTime()) {
```
evanphx, line 354: Best to use `time.Equal` here if we're converting them to `time.Time` values.

evanphx (changes requested):

evanphx (approved):
:shipit:

krantzinator (approved):
NICE

briancain (approved):
🔥 🐛

Issue #1540, opened 2021-05-26
Couple of minor misaligned icons


jgwhite:
**Describe the bug**

Just noticed a couple of places where icons are misaligned following the conversion to `<Pds::Icon>`

**Steps to Reproduce**

1. Visit the new project page (i.e. [https://localhost:9702/default/new](https://localhost:9702/default/new))
2. Observe the following:
   <img width="411" alt="image" src="https://user-images.githubusercontent.com/34030/119649493-fbf16980-be22-11eb-8417-e597a1be4910.png">
3. Visit an empty app (i.e. one for which you have only run  `waypoint init`)
4. Observe the following:
   <img width="143" alt="image" src="https://user-images.githubusercontent.com/34030/119649762-4c68c700-be23-11eb-9cf9-41b9f79a2e87.png">

PR #1544, opened 2021-05-26
UI: Fixed misaligned icons

sabrinako:
#1540 
**How the noted buggy icons look now with the fixes**
Screen Shot 2021-05-26 at 1 21 34 PM
Screen Shot 2021-05-26 at 1 21 21 PM

### DIFF
--- a/ui/app/styles/_code.scss
+++ b/ui/app/styles/_code.scss
@@ -45,9 +45,12 @@
     }
 
     .icon {
+      display: flex;
+      align-items: center;
+      justify-content: center;
       width: 12px;
       height: 12px;
-      font-size: scale.$base;
+      font-size: 12px;
     }
   }
 }
--- a/ui/app/styles/_page.scss
+++ b/ui/app/styles/_page.scss
@@ -380,12 +380,6 @@
         }
       }
     }
-
-    .pds-icon {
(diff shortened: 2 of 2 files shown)


ui/app/styles/_page.scss
```diff
@@ -380,12 +380,6 @@
         }
       }
     }
-
-    .pds-icon {
```
koesbong, line 384: should this be turned into a mixin so it's reusable?

ui/app/styles/_code.scss
```diff
@@ -45,9 +45,12 @@
     }
 
     .icon {
+      display: flex;
+      align-items: center;
+      justify-content: center;
       width: 12px;
       height: 12px;
-      font-size: scale.$base;
+      font-size: 12px;
```
gregone, line 53: not just related to this PR, but I'd encourage using `rem` units or size variables from the `scale` module instead of `px`

gregone (approved):

ui/app/styles/_code.scss
sabrinako: I went with the 12px because the `font-size` of `copyable-code` was also 12px, but will definitely try to do that more!

PR #1547, opened 2021-05-26
plugin/k8s: Add StatusFunc to Deploy and Release

briancain:
This pull request introduces a status report forwarder for the k8s platform. It inspects the deployed Pods and its containers to determine the overall health of a deployment or release.

Relies on https://github.com/hashicorp/waypoint-plugin-sdk/pull/32

### DIFF
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -658,7 +658,7 @@ func (p *Platform) Status(
 	defer sg.Wait()
 
 	step := sg.Add("Gathering health report for Kubernetes platform...")
-	defer step.Abort()
+	defer func() { step.Abort() }() // Defer in func in case more steps are added to this func in the future
 
 	csInfo, err := p.getClientset()
 	if err != nil {
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/golang/protobuf/ptypes"
 	"github.com/hashicorp/go-hclog"
 	"github.com/mitchellh/mapstructure"
 	corev1 "k8s.io/api/core/v1"
@@ -20,8 +21,10 @@ import (
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	"github.com/hashicorp/waypoint-plugin-sdk/docs"
 	"github.com/hashicorp/waypoint-plugin-sdk/framework/resource"
+	sdk "github.com/hashicorp/waypoint-plugin-sdk/proto/gen"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/builtin/docker"
+	"github.com/hashicorp/waypoint/internal/clierrors"
 )
 
 const (
@@ -68,6 +71,10 @@ func (p *Platform) Auth() error {
 
 func (p *Platform) ValidateAuth() error {
 	return nil
+}
+
+func (p *Platform) StatusFunc() interface{} {
+	return p.Status
 }
 
 // DefaultReleaserFunc implements component.PlatformReleaser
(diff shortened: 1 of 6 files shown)


evanphx (approved):
This looks good but I wonder if we might have a more release-oriented status in the future. Atm because a release uses a deployment, we're showing the same status twice. Not an issue for now, but we should dig into K8 more and figure out if there is some release/service specific status we could be showing.

briancain:
@evanphx Agreed...was thinking the same thing after finishing this up! I thought about reported on the Service itself that we release to, but it doesn't look like there's much the k8s api tells us about it compared to the pods themselves.

krantzinator (approved):
🎉

builtin/k8s/status_report_util.go
```diff
@@ -0,0 +1,81 @@
...
+	var result sdk.StatusReport
+	result.External = true
+	resources := make([]*sdk.StatusReport_Resource, len(podList.Items))
+
+	// Report on most recently observed status of a deployments pod
+	for _, pod := range podList.Items {
+		// Overall Pod Health
+		podStatus := pod.Status
+		result.HealthMessage = podStatus.Message
+		result.Health = podPhaseToHealth(podStatus.Phase)
```
izaaklauer, line 72: It looks like `result` will carry the Health and HealthMessage of whatever the last pod was in your slice. I'm not sure exactly what the StatusReport adjectives mean, but I think you could declare the whole `result` as:

`StatusReport_READY` if all pods are ready

`statusReport_ALIVE` if all pods are in alive or ready, but not all are ready

`statusReport_DOWN` if all pods are down

`statusReport_UNKNOWN` if all pods are unknown? any pods? not sure no this one.

`statusReport_PARTIAL` for any mix that includes a down or an unknown?

builtin/k8s/status_report_util.go
```diff
@@ -0,0 +1,81 @@
...
+	return &resourceHealth
+}
+
+// Translate a Pod Phase into a Waypoint Health Status
+func podPhaseToHealth(
+	phase corev1.PodPhase,
+) sdk.StatusReport_Health {
+	var healthResult sdk.StatusReport_Health
+
+	switch phase {
```
izaaklauer, line 41: I think that the pod conditions have a closer mapping our statuses than the pod phase. For example, the `PodReady` condition is documented as indicating the pod has passed health checks and can take requests, but the `Running` phase only implies that the pod has been bound to a node and the containers have been created.

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

builtin/k8s/status_report_util.go
briancain: Oops, good catch!

builtin/k8s/status_report_util.go
briancain: That makes sense! I'll base the report on that for pods instead. Agreed ConditionStatus looks a lot better.

briancain: Hrm, actually looking at this, Pod Conditions don't really describe any failure states... I see "Ready", "Initialized", "PodScheduled", and "ContainersReady" as [possible conditions](https://pkg.go.dev/k8s.io/api/core/v1#PodConditionType), where as the pod phase describes the full potential state a pod could be in: https://pkg.go.dev/k8s.io/api/core/v1#PodPhase

builtin/k8s/platform.go
```diff
@@ -641,6 +648,74 @@ func (p *Platform) Destroy(
...
+	ctx context.Context,
+	log hclog.Logger,
+	deployment *Deployment,
+	ui terminal.UI,
+) (*sdk.StatusReport, error) {
+	sg := ui.StepGroup()
+	defer sg.Wait()
+
+	step := sg.Add("Gathering health report for Kubernetes platform...")
+	defer step.Abort()
```
mitchellh, line 661: I recommend using `defer func() { step.Abort() }()` in case we ever add other steps.

builtin/k8s/releaser.go
```diff
@@ -268,6 +276,80 @@ func (r *Releaser) Destroy(
...
+	ctx context.Context,
+	log hclog.Logger,
+	release *Release,
+	ui terminal.UI,
+) (*sdk.StatusReport, error) {
+	sg := ui.StepGroup()
+	defer sg.Wait()
+
+	step := sg.Add("Gathering health report for Kubernetes platform...")
+	defer step.Abort()
```
mitchellh, line 289: Same feedback here.

mitchellh (approved):
Looks good, one easy comment.

builtin/k8s/status_report_util.go
briancain: Ok, got that fixed up now!

builtin/k8s/status_report_util.go
```diff
@@ -0,0 +1,132 @@
...
+		result.HealthMessage = "all pods are reporting down"
+	} else if unknown == len(podList.Items) {
+		result.Health = sdk.StatusReport_UNKNOWN
+		result.HealthMessage = "status of all pods cannot be determined"
+	} else if alive == len(podList.Items) {
+		result.Health = sdk.StatusReport_ALIVE
+		result.HealthMessage = "all pods are reporting alive"
+	} else {
+		result.Health = sdk.StatusReport_PARTIAL
+		result.HealthMessage = "all pods are reporting a mixed statusu"
```
izaaklauer, line 128: typo "statusu"

briancain: Oof, thanks 👍🏻

builtin/k8s/status_report_util.go
```diff
@@ -0,0 +1,132 @@
...
+		result.Health = sdk.StatusReport_DOWN
+		result.HealthMessage = "all pods are reporting down"
+	} else if unknown == len(podList.Items) {
+		result.Health = sdk.StatusReport_UNKNOWN
+		result.HealthMessage = "status of all pods cannot be determined"
+	} else if alive == len(podList.Items) {
+		result.Health = sdk.StatusReport_ALIVE
+		result.HealthMessage = "all pods are reporting alive"
+	} else {
+		result.Health = sdk.StatusReport_PARTIAL
```
izaaklauer, line 127: Question here: "PARTIAL" implies to me "Partially Ready". With this logic, if 4/5 are down and 1/5 is unknown, we'll get PARTIAL.

Just double-checking that we really want PARTIAL to mean any non-homogeneous set of underlying statuses.

izaaklauer: Ah, also just found this in the RFC: `PARTIAL = 4; // Some resources in deployment are OK, others are not OK
`

PR #1548, opened 2021-05-27
core: add support for input variables

krantzinator:
Adds support for input variables.

Docs will be in a separate PR.

Input variables are currently available within any stanzas in the waypoint.hcl; they are not available for top-level objects such as `project` parameter.

A couple of videos showing functionality, combined with #1658

https://user-images.githubusercontent.com/8461333/124027691-28b81400-d9c1-11eb-9b5e-8fec9cf384cd.mp4

https://user-images.githubusercontent.com/8461333/124027688-2786e700-d9c1-11eb-9062-5ede40730460.mp4


internal/core/project.go
```diff
@@ -60,6 +60,11 @@ type Project struct {
 	// overrideLabels are the labels specified via the CLI to override
 	// all other conflicting keys.
 	overrideLabels map[string]string
+
+	// variables are set on the job directly so that the runner
+	// can access this set of values along with any server- or vcs-stored
+	// values for final evaluation
+	variables []*pb.Variable
```
mitchellh, line 67: This comment seems off or I'm just misunderstanding this. It seems like the core needs the variables fully rendered so that it can parse the configuration? i.e. they should be fully merged/flattened/whatever at this point.

internal/config/variables.go
```diff
@@ -0,0 +1,420 @@
...
+// them.
+const VarEnvPrefix = "WP_VAR_"
+
+// Collect values from flags, local var files, and env vars
+func CollectInputVars(vars map[string]string, files []string) ([]*pb.Variable, hcl.Diagnostics) {
+	var diags hcl.Diagnostics
+	ret := []*pb.Variable{}
+
+	// First we'll deal with environment variables, since they have the lowest
+	// precedence.
```
mitchellh, line 202: Does precedence matter? Since we're retaining the source, do we calculate precedence later? Or does the order of `ret` actually matter?

internal/config/variables.go
```diff
@@ -0,0 +1,420 @@
...
+			Source: "server",
+			Value:  val,
+			Expr:   expr,
+		})
+	}
+
+	return diags
+}
+
+func (variables *Variables) SortPrecedence() ([]*pb.Variable, error) {
```
mitchellh, line 395: Also, this seems super important and should be tested. This is like, _the_ logic behind what values are used, I think.

mitchellh (changes requested):
Awesome @krantzinator! There's a ton of good stuff here.

I think my only major issue is with `internal/config/variables.go`. The rest of how you're parsing it, how you're using it, seems right. The _style_ of `internal/config/variables.go` feels a bit weird. To define that a bit more, it feels (1) not strictly idiomatic Go (2) not Waypoint-ish Go style (3) feels like it conflates some knowledge across abstraction boundaries. 

I tried to step back to think "okay, what are we doing here big picture" and it seems like what you're achieving is:

* Parse variable definitions from an HCL config (and maybe defaults)
* Load variable values from env vars
* Load variable values from files
* Load variable values from flags
* Sort by precedence

That stands out to me as each should be their own function. Right now you have `CollectInputVars` that seems to do both env and file, and `CollectRemote` which seems to assign values to definitions? (That one isn't clear to me). 

Order of calling those functions doesn't seem to matter since we sort by precedence later. So I feel like `internal/cli` can just call the env and file load manually and separately, and the runner can call the appropriate functions as well, then sort/merge.

There's also enough here that it _could_ make sense to put all these functions in a package like `internal/config/vars`. Cause I think you're probably expanding the exported surface area of the `internal/config` package by quite a lot and that's usually a smell to me to extract a package. 

My recommendation would be to take a pass at the types/functions with this lens. If you want to hop on a Zoom happy to do that too.

internal/config/variables.go
```diff
@@ -0,0 +1,420 @@
...
+		{
+			Name: "type",
+		},
+		{
+			Name: "description",
+		},
+	},
+}
+
+func (c *Config) DecodeVariableBlocks(variables *Variables) hcl.Diagnostics {
```
mitchellh, line 83: This feels like it shouldn't be a separate thing and should always just happen on config parse. Any reason to make this separate that I'm missing? 

My thought would it happens always on `Load` and there is either a struct field or a function to return the `*Variables` type.

krantzinator:
Thanks @mitchellh ! I'll respond with thoughts to your overall here, and then take a look at the piece-specific suggestions.

>The style of internal/config/variables.go feels a bit weird. To define that a bit more, it feels (1) not strictly idiomatic Go (2) not Waypoint-ish Go style (3) feels like it conflates some knowledge across abstraction boundaries.

I do agree with this. I am struggling to see the entire picture here, so a lot of these functions are in the vein of "make it work, make it right, make it fast" mentality; and they are all currently in the "make it work" stage with me anticipating quite a bit of refactoring as I add other pieces and come back to consolidate. "Not strictly idiomatic Go" is also definitely a thing 😄 My Ruby/JavaScript brain background comes to the forefront when I'm working out a new architecture.

>Right now you have CollectInputVars that seems to do both env and file, and CollectRemote which seems to assign values to definitions? (That one isn't clear to me).

This isn't clear to me either 😅 This is still in the "Make it work" part of my brain. Definitely want to refactor this once I've got a couple more pieces ironed out.

>Order of calling those functions doesn't seem to matter since we sort by precedence later. So I feel like internal/cli can just call the env and file load manually and separately, and the runner can call the appropriate functions as well, then sort/merge.

This is my plan 🙌🏻 .

>There's also enough here that it could make sense to put all these functions in a package like internal/config/vars. Cause I think you're probably expanding the exported surface area of the internal/config package by quite a lot and that's usually a smell to me to extract a package.

That's how I started, and then I pulled it back into config because it felt like a too-early abstraction. I agree it feels right now to put it in its own package, or at least most of it with maybe one or two links between config and config/vars.

>My recommendation would be to take a pass at the types/functions with this lens. If you want to hop on a Zoom happy to do that too.

Sure thing.

internal/config/variables.go
krantzinator: For sure. This was the point where I stopped and decided to push up what I had. This function is only half done; I think I noted that where I call it, rather than as a comment here, so that's on me. I'm not totally sure how I want this function to work yet.

internal/cli/base.go
```diff
@@ -74,6 +77,13 @@ type baseCommand struct {
 	// flagLabels are set via -label if flagSetOperation is set.
 	flagLabels map[string]string
 
+	// flagVars sets values for defined input variables
+	flagVars map[string]string
+
+	// flagVarFile is a HCL or JSON file setting one or more values
+	// for defined input variables
+	flagVarFile string
```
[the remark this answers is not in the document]
krantzinator: Done.

internal/core/project.go
krantzinator: Yes, comment fixed.

internal/config/variables.go
krantzinator: I went back and forth on this a bit. The order of `ret` does matter, currently. The intention with `source` was to use it primarily for error messaging, though it could also be used to build out a specific precedence sort. I'm not tied to either solution, but having things evaluated in precedence-order is what I've got at the moment.

internal/config/variables.go
```diff
@@ -0,0 +1,420 @@
...
+			Value:  &pb.Variable_Str{Str: val},
+			Source: &pb.Variable_Cli{},
+		})
+	}
+
+	return ret, diags
+}
+
+// Collect values from server and remote-stored wpvars file
+func (variables *Variables) CollectInputValRemote(files []*hcl.File, serverVars []*pb.Variable) hcl.Diagnostics {
```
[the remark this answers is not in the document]
krantzinator: This function has been renamed twice now 😄 And changed quite a bit over the last refactor. It now returns a new slice at least 🎉

internal/config/variables.go
krantzinator: Ok as mentioned in an earlier comment, I've decided to take the sort of precedence out of an explicit function call, and evaluate values in precedence order. Again, not tied to either implementation (but FYI this function doesn't exist anymore)

internal/config/variables.go
krantzinator: This is now in Config and makes a call to the variables package for individual block parsing. It returns a `map[string]*variables.Variable`.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+		diags = append(diags, hclDiags...)
+		if f == nil || f.Body == nil {
+			return nil, diags
+		}
+	}
+
+	// Before we do our real decode, we'll probe to see if there are any
+	// blocks of type "variable" in this body, since it's a common mistake
+	// for new users to put variable declarations in wpvars rather than
+	// variable value definitions.
```
evanphx, line 476: This is great diagnostics work!

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+			if eq == -1 {
+				// Seems invalid, so we'll ignore it.
+				continue
+			}
+
+			name := raw[:eq]
+			rawVal := raw[eq+1:]
+
+			ret = append(ret, &pb.Variable{
+				Name:   name,
```
evanphx, line 203: Should we force these to lower case? Or just expect that folks will do `WP_VARS_name`? I don't recall which terraform does.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+		}
+
+		// We have to specify the three different simple types we support -- string,
+		// bool, number -- when doing the below evaluation of hcl expressions
+		// because of our translation to-and-from protobuf format.
+		// While cty allows us to parse all simple types as LiteralValueExpr, we
+		// have to first translate the pb values back into cty values, thus
+		// necessitating a separate case statement for each simple type
+		var expr hclsyntax.Expression
+		switch pbv.Value.(type) {
```
evanphx, line 284: This is a minor nit, you could use `switch sv := pbv.Value.(type)` here and then below, you can use `sv` as the type that was detected, so like on line 287, you could have `sv` instead of `pbv.Value(*pb.Variable_Hcl)`.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+
+	// Check working directory (vcs or local) for *.auto.wpvars(.json) files
+	var varFiles []string
+	if files, err := ioutil.ReadDir(wd); err == nil {
+		for _, f := range files {
+			name := f.Name()
+			if !isAutoVarFile(name) {
+				continue
+			}
+			varFiles = append(varFiles, wd+"/"+name)
```
evanphx, line 363:
```suggestion
			varFiles = append(varFiles, filepath.Join(wd, name))
```

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+			Source: &pb.Variable_Cli{},
+		})
+	}
+	return ret, diags
+}
+
+// EvalInputValues evaluates the provided variable values and validates their
+// types per the type declared in the waypoint.hcl for that variable name.
+// The order in which values are evaluated corresponds to their precedence, with
+// higher precedence values overwriting lower precedence values.
```
evanphx, line 235: I'd add a note about `vs` and that it is there to make sure that all the needed variables are populated (and that normally it comes from observing all the `variable` decl blocks in a waypoint.hcl file).

internal/core/project.go
```diff
@@ -124,7 +128,11 @@ func NewProject(ctx context.Context, os ...Option) (*Project, error) {
 
 	// Initialize all the applications and load all their components.
 	for _, name := range opts.Config.Apps() {
-		appConfig, err := opts.Config.App(name, nil)
+		// Set input variables for applications and components
+		evalCtx := config.EvalContext(nil, p.dir.DataDir()).NewChild()
+		config.AddVariables(evalCtx, &p.variables)
```
evanphx, line 133: It's a little odd to use pointers to maps, and since you just deref the pointer in `AddVariables` I'd probably just change `AddVariables` to take the map without a pointer.

evanphx (changes requested):
Looks great! A few little odds and ends.

.changelog/1548.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+core: Add support for input variables
```
mitchellh, line 2: Just a note that this is surely a candidate for **bold feature** (that we don't support automating yet). So no action you have to do @krantzinator but just for us to know when we cut a release and build the changelog.

internal/config/config.go
```diff
@@ -132,14 +142,45 @@ func Load(path string, opts *LoadOptions) (*Config, error) {
...
+		InputVariables: vs,
 	}, nil
 }
 
 // HCLContext returns the eval context for this configuration.
 func (c *Config) HCLContext() *hcl.EvalContext {
 	return c.ctx.NewChild()
 }
+
+func DecodeVariableBlocks(body hcl.Body) (map[string]*variables.Variable, hcl.Diagnostics) {
```
mitchellh, line 158: This needs a doc comment.

mitchellh: Also, I don't think this needs to be exported? I can't see anything else calling it.

internal/config/eval_context.go
```diff
@@ -95,6 +96,11 @@ func finalizeContext(ctx *hcl.EvalContext) *hcl.EvalContext {
 	return ctx
 }
 
+func AddVariables(ctx *hcl.EvalContext, vs *variables.InputValues) *hcl.EvalContext {
```
mitchellh, line 99: doc comment

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+	// listed in descending precedence order for ease of reference
+	sourceCLI     = "cli"
+	sourceFile    = "file"
+	sourceEnv     = "env"
+	sourceVCS     = "vcs"
+	sourceServer  = "server"
+	sourceDefault = "default"
+
+	// Prefix for collecting variable values from environment variables
+	varEnvPrefix = "WP_VAR_"
```
mitchellh, line 31: all these above here probably want to be `const`, and not var.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+	Type cty.Type
+
+	// Description of the variable
+	Description string
+
+	// The location of the variable definition block in the waypoint.hcl
+	Range hcl.Range
+}
+
+type HclVariable struct {
```
mitchellh, line 72: doc comment

internal/config/config.go
mitchellh: Also, should this be in the `variables` package? (not rherotical, real question)

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+	Value  cty.Value
+	Source string
+	Expr   hcl.Expression
+	// The location of the variable value if the value was provided from a file
+	Range hcl.Range
+}
+
+// DecodeVariableBlock validates each part of the variable block,
+// building out a defined *Variable
+func DecodeVariableBlock(block *hcl.Block) (*Variable, hcl.Diagnostics) {
```
mitchellh, line 96: if we move `DecodeVariableBlocks` (plural) into this package, then this could become unexported.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+	return &v, diags
+}
+
+// SetJobInputValues collects values set via the CLI (-var, -varfile) and
+// local env vars (WP_VAR_*) and translates those values to pb.Variables. These
+// pb.Variables can then be set on the job for eventual parsing on the runner,
+// after the runner has decoded the variables defined in the waypoint.hcl.
+// All values are set as protobuf strings, with the expectation that later
+// evaluation will convert them to their defined types.
+func SetJobInputValues(vars map[string]string, files []string) ([]*pb.Variable, hcl.Diagnostics) {
```
mitchellh, line 176: `Set` prefix here makes it sound like this is setting some state. It made me scared you were using a global var somewhere! But that isn't the case. 

I think it'd be more idiomatic to simply drop the `Set`. 

It is also confusing because you actually have a thing called `InputValues` and this seemingly has nothing to do with it; it returns variables. So maybe this would make more sense to be named `LoadVariableValues` or something.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+		})
+	}
+	return ret, diags
+}
+
+// EvalInputValues evaluates the provided variable values and validates their
+// types per the type declared in the waypoint.hcl for that variable name.
+// The order in which values are evaluated corresponds to their precedence, with
+// higher precedence values overwriting lower precedence values.
+func EvalInputValues(
```
mitchellh, line 236: Similarly, a bit confusing, cause it feels like you're not evaluating the input values here, you're evaluating variables that turns into values. Should this be named differently?

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+
+	for v, def := range vs {
+		iv[v] = def.Default
+	}
+
+	for _, pbv := range pbvars {
+		variable, found := vs[pbv.Name]
+		if !found {
+			// TODO krantzinator: this should be a DiagWarning rather than a DiagErr
+			// What to do with a warning diag type?
```
mitchellh, line 251: I think error is fine. It was annoying that this became an error in TF because so much workflow was already built on it being a warning, but since we haven't shipped anything yet, might as well start super strict.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+			source = sourceCLI
+		case *pb.Variable_File_:
+			source = sourceFile
+		case *pb.Variable_Env:
+			source = sourceEnv
+		case *pb.Variable_Vcs:
+			source = sourceVCS
+		case *pb.Variable_Server:
+			source = sourceServer
+		}
```
mitchellh, line 275: Another way to do this would be a `map[reflect.Type]string` then you could `source := sourceMap[reflect.TypeOf(pbv.Source)]`. Not a big deal, something we can clean up later, this certainly works.

mitchellh, line 275: I recommend a default case here that logs or errors, otherwise `source` is empty string and who knows how that cacades.

internal/config/variables/variables.go
mitchellh: I'd say this isn't a nit and is pretty important, this gets you much better type safety imo.

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+					"details.",
+			})
+		}
+	}
+
+	return iv, diags
+}
+
+// LoadVCSFiles loads any *.auto.wpvars(.json) files in the source repo
+func LoadVCSFiles(wd string) ([]*pb.Variable, hcl.Diagnostics) {
```
mitchellh, line 351: Naming bikeshedding again, I'd drop VCS from this. The variables package itself doesn't care if this is a VCS repo or not. It's just loading files from a working directory. I'd rename this to like `LoadAutoFiles` or something.

internal/client/project.go
```diff
@@ -227,6 +228,14 @@ func WithClientConnect(opts ...serverclient.ConnectOption) Option {
 	}
 }
 
+// WithVars sets variable values from flags and local env on any operations.
+func WithVars(m []*pb.Variable) Option {
```
mitchellh, line 232: For core you use `WithVariables`. I don't care which it is, but I'd be consistent if you abbreviate variables or not.

internal/runner/operation.go
```diff
@@ -118,6 +118,23 @@ func (r *Runner) executeJob(
...
+	// Here we'll load our values from the server/UI, as well as VCS.
+	// The order values are added to our final pbVars slice in the order
+	// of precedence
+	pbVars := resp.Project.GetVariables()
+	vcsVars, diags := variables.LoadVCSFiles(wd)
+	if diags.HasErrors() {
+		return nil, diags
+	}
+	pbVars = append(pbVars, vcsVars...)
+	pbVars = append(pbVars, job.Variables...)
```
mitchellh, line 130: I'd reorder this a bit to:

```go
	vcsVars, diags := variables.LoadVCSFiles(wd)
	if diags.HasErrors() {
		return nil, diags
	}

	pbVars := resp.Project.GetVariables()
	pbVars = append(pbVars, vcsVars...)
	pbVars = append(pbVars, job.Variables...)
```

This is a nit pick but I find it cleaner to bunch together the accumulation if possible.

internal/server/proto/server.proto
```diff
@@ -320,6 +320,9 @@ message Variable {
 
   oneof value {
     string str = 2;
+    bool bool = 9;
+    int64 num = 10;
+    // Unimplemented
```
mitchellh, line 325: is this a future todo? should you mark it `TODO`?

internal/config/variables/variables.go
```diff
@@ -0,0 +1,531 @@
...
+	Name        string         `hcl:",label"`
+	Default     cty.Value      `hcl:"default,optional"`
+	Type        hcl.Expression `hcl:"type,optional"`
+	Description string         `hcl:"description,optional"`
+}
+
+// InputValues are used to store values collected from various sources.
+// Values are added to the map in precedence order, and then used to
+// create the final map of cty.Values for config hcl context evaluation.
+type InputValues map[string]*InputValue
```
mitchellh, line 82: This is a later comment after I read the rest of this PR, I want to bikeshed this again and note if we want to name this just `Values` and rename `InputValue` to just `Value`. 

I originally read `InputValue` and thought "what other type of value is there?" And figured I'd run into one but didn't. Is there a reason for the additional qualifier?

mitchellh (changes requested):
Great work! This is very clean and there's certainly a lot here. I didn't have any fundamentally "broken" feedback, just lots of polish points.

internal/config/config.go
krantzinator: Ah I think I forgot to unexport it when I moved it from `variables` to `config`. To your question, my thinking was that this is a thing that `config` is doing to decode things in the `config`, and then this `config` function calls the necessary `variables` function.
But I don't have strong feelings about it, so if in go-isms it makes more sense to put it in `variables` I'm down with that.

internal/config/variables/variables.go
krantzinator: I will take the credit for stealing this idea 😄

internal/config/variables/variables.go
krantzinator: Both Terraform and Packer expect folks to copy the case exactly, so I think we should not force to lower.

internal/config/variables/variables.go
krantzinator: Done!

internal/config/variables/variables.go
krantzinator: Good call! Notes added.

internal/core/project.go
krantzinator: Done!

.changelog/1548.txt
krantzinator: 10-4

internal/config/eval_context.go
krantzinator: done

internal/config/variables/variables.go
krantzinator: done

internal/config/variables/variables.go
krantzinator: done

internal/config/variables/variables.go
krantzinator: Makes sense, done!

internal/client/project.go
krantzinator: Set both to `WithVariables`

internal/runner/operation.go
krantzinator: I like it, done

internal/server/proto/server.proto
krantzinator: This was a historic `TODO` I forgot to remove because I didn't mark it `TODO` so it didn't show up in my search 😅

internal/config/variables/variables.go
krantzinator: Cool cool, TODO removed

internal/config/variables/variables.go
krantzinator: Hmmm yeah I think that makes sense. The returned set of pb.Variables could theoretically be used for anything, not specifically the job, so that doesn't need to be in the func name. `LoadVariableValues` is as good as anything.

internal/config/variables/variables.go
krantzinator: I'm going to log it. If there's an error here, it's because our code isn't setting a source correctly somewhere, and I don't want to break a user's ability to use this if we mess up a code path.

krantzinator: Oh no that's way better. Thanks.

internal/config/variables/variables.go
krantzinator: Sure, how's `EvalInputVariables`?

internal/config/variables/variables.go
krantzinator: I named it this way under the assumption we might expand variable types later and have other types of values. But there is no immediate reason to name it so specifically, and so simpler is good for now. I'll name it Values/Value.

internal/config/config.go
krantzinator: Given your other comment, I'll go ahead and move this into `variables`.

internal/config/variables/variables.go
krantzinator: done!

internal/config/variables/variables.go
krantzinator: Maybe just `EvaluateVariables`

krantzinator:
FYI: Once this is approved I'm going to rebase and pull the protobuf gen commit to the front, since I anticipate this having a hard time backporting in general, but definitely on that protobuf gen. Having that at the end to just throw away and re-gen on `release/0.4.x` will be much nicer.

internal/server/proto/server.proto
```diff
@@ -320,6 +320,8 @@ message Variable {
 
   oneof value {
```
mitchellh, line 321: Docs on all of these would be good, particularly the `hcl` one :)

mitchellh (commented):
Looking reallllllll close. Just one small comment.

evanphx (approved):
Looks great!

internal/server/proto/server.proto
krantzinator: Oooh documenting this pointed me to a bug. This will be a minute.

krantzinator: OK, that was an embarrassing miss but yay for code reviews 🎉

denysvitali:
Hey! Thanks for the PR!

I was wondering if this change takes into consideration this simple use case that I'm facing, and if it's not too late it would be awesome to have it implemented before this gets merged:

Currently, the [Internal Values](https://www.waypointproject.io/docs/app-config/internal) can be used to have a better interpolation in the `env` block of the `config` stanza. Unfortunately `config.internal.XYZ` variables do not work outside the `config` stanza and are therefore a limitation as there is no proper way to use them as variables inside the `waypoint.hcl` file.

It would be nice to have something like your `variable` implementation, that can be used internally, kind of what you are doing with them having a `default` value, but without the possibility (or with a less obvious possibility) to change it externally.

As an example, we would like to be able to differentiate our hostnames depending on where we deploy our workload, and for that we use a `jsonnet` that could eventually work as follows:

```hcl
env {
  MY_URL = file(jsonnetfile("${path.app}/env_urls.jsonnet", {ext_vars: {env = "${workspace.name}"}}))
}
```

While this could ideally work, unfortunately it requires now to specify multiple times the same `file(jsonnetfile(...)))` call over and over again. We would like to be able to have a way to use proper variables inside the `waypoint.hcl` file, so that we can simply use something like:

```hcl
variables {
  env_url = jsonnetfile("${path.app}/env_urls.jsonnet", {ext_vars: {env = "${workspace.name}"}})
}
env {
  MY_URL = "https://${vars.env_url.some_key}/index.html"
  MY_SECOND_URL = "https://${vars.env_url.some_other_key}/"
}
```

This will definitely give us (and other people) more flexibility without being forced to use the `config` stanza (which only works if you are using the Waypoint Entrypoint, which in our case is not used).

---

I seem to understand that with your PR, this could be eventually be solved as follows:

```hcl
variable "env_urls" {
  default = jsonnetfile("${path.app}/env_urls.jsonnet", {ext_vars: {env = "${workspace.name}"}})
  type = bool
} 
```

Unfortunately this will also expose the variable to the frontend and will most likely cause some problems since it's intended to be an "internal variable"

krantzinator:
@denysvitali This sounds out of scope of this current PR. Input variables as implemented here do not support functions as values, so unfortunately the last example you give will not work.
It sounds a bit like you want support for a different type of variable system? When you mentioned:
> `variable` implementation, that can be used internally, kind of what you are doing with them having a default value, but **without the possibility** (or with a less obvious possibility) **to change it externally.**

that makes me think the problem you are wishing to solve is different from the problem this particular feature is targeting. The entire point of input variables is to allow for values to be changed externally 🙂

mitchellh:
@krantzinator @denysvitali This sounds more like you want [local values similar to Terraform](https://www.terraform.io/docs/language/values/locals.html). This is something that @krantzinator and I have talked about building off of this work in the future. 

Note on internal variables not being available outside of `config`: Evan and I talked about that annoyance for exactly the use case you ran into, but the reason we did this is because config variables are especially special in that they are "rendered" or "realized" or "loaded" (all synonymous in this case) at a different time to everything else.

mitchellh (approved):
I think we're ready.

denysvitali:
@mitchellh Yes, that's exactly what I need! Thanks for the clarification!

Do you want me to open a feature request for it?

mitchellh:
Yes please.

PR #1550, opened 2021-05-27
server: Add API call for getting latest status report

briancain:
Include a latest status report API endpoint to get the last generated report for an application


internal/server/gen/server.pb.go
```diff
@@ -1,13 +1,14 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.26.0
+// 	protoc-gen-go v1.25.0
```
gregone, line 3: Nit: should this version number be locked and not changed?

gregone (approved):
Looks good, thanks!

internal/server/gen/server.pb.go
briancain: Ahhhhh 😭 that was adjusted back on my machine because of https://github.com/hashicorp/waypoint-plugin-sdk/pull/32 😕

briancain: Fixed it

krantzinator: Do we need to update the proto version on the SDK to match what we're using here? That would be my bad -- I should have done both at once when I upped the version.

krantzinator (approved):

internal/server/gen/server.pb.go
briancain: We should, but when we do it's going to break proto gen for anyone on the older version. There's a go import path issue that needs to be fixed (I have the fix ready for when we are ready to bump the version)

PR #1557, opened 2021-05-28
fix: adapt doc build, add to shell.nix

denysvitali:
This fix corrects the way `make gen/doc` works on newly cloned repositories (that are missing the `doc/` folder) and adds an entry for `protoc-gen-doc` in the `shell.nix` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ tmp/
 /waypoint
 /waypoint-entrypoint
 /pkg
+/doc
 
 # System-specific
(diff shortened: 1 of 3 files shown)


mitchellh:
This works for me, looks good.

mitchellh:
https://github.com/hashicorp/waypoint/commit/2a92eba09ff6ceb820b0a410248c45637fb563b6

Not sure why GitHub isn't registering that I merged this, but I did (via the command line)

Issue #1497, opened 2021-05-18
Improve lifecycle output across plugins using Resource Manager


briancain:
**Is your feature request related to a problem? Please describe.**

Pre-0.4.0 Waypoint, much if not all of the CLI output and UX for Build Deploy Release rests on the Plugin authors shoulders. As Waypoints plugins get more complicated, and as third party developers write their own plugins, it will become difficult to ensure the output of Waypoint is clear, understandable, and familiar. Having some consistent messages that always are printed no matter the plugin will help Waypoint feel like one cohesive tool.

A basic example of what Waypoint does today is how the CLI will print the beginning of each lifecycle (i.e. `Building...` `Deploying...` `Releasing...`). What happens after that is very unstructured and up to the plugin to decide what is shown.

**Describe the solution you'd like**

To keep Waypoint feeling familiar and understandable no matter what plugin a user might be using, Waypoint should have some guaranteed output during the lifecycle operations that give users a sense of what is happening and how the deployment or release is going.

With the new work to introduce a Resource Manager to plugins for managing the resources it creates for a given lifecycle, Waypoint should use that information to output what its doing to the user in a way that's useful and to build confidence that Waypoint is doing expected work for a deployment or release.

By showing information on what resources have been created and managed, and unifying output across plugins to show a deployment or release was successful will raise confidence for the user that the lifecycle operation succeeded. It should also help or aid in triaging failures if the output for managing resources or deployment/releases is always surfaced in a clear way.

**Explain any additional use-cases**

When a user decides to try out a different plugin, or write their own, the plugin should feel familiar and have output that is easily understandable. This familiarity should come "free" without requiring the plugin author to take on that burden. Instead, they can focus on output that's relevant to their specific plugin.

**Additional context**

briancain:
Closing this for now. We did some work to improve the general output of the core plugins. I don't see this being tackled to its fullest any time soon.

PR #1561, opened 2021-06-01
builtin/{docker/k8s/nomad}: Add advisory message on non-ready reports

briancain:
This commit adds some additional wordage in the case where a deployment
or release report was generated and its health is anything other than
READY. In some cases, a report that was generated immediately after the
deployment or release will be unhealthy given slow startup times, so this
warning adds some extra context for what the report means.

As an aside, this is one of the reasons why https://github.com/hashicorp/waypoint/issues/1497 will be very useful! This status output based on a StatusFuncs returned report could live in a single spot rather than duplicated across all plugins.

### DIFF
--- a/builtin/docker/platform.go
+++ b/builtin/docker/platform.go
@@ -220,10 +220,17 @@ func (p *Platform) Status(
 	st.Update("Determining overall container health...")
 	if result.Health == sdk.StatusReport_READY {
 		st.Step(terminal.StatusOK, fmt.Sprintf("Container %q is reporting ready!", containerInfo.Name))
-	} else if result.Health == sdk.StatusReport_PARTIAL {
-		st.Step(terminal.StatusWarn, fmt.Sprintf("Container %q is reporting partially available!", containerInfo.Name))
 	} else {
-		st.Step(terminal.StatusError, fmt.Sprintf("Container %q is reporting not ready!", containerInfo.Name))
+		if result.Health == sdk.StatusReport_PARTIAL {
+			st.Step(terminal.StatusWarn, fmt.Sprintf("Container %q is reporting partially available!", containerInfo.Name))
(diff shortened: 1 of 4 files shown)


krantzinator (approved):

izaaklauer (approved):

builtin/docker/platform.go
```diff
@@ -220,10 +220,17 @@ func (p *Platform) Status(
...
+		if result.Health == sdk.StatusReport_PARTIAL {
+			st.Step(terminal.StatusWarn, fmt.Sprintf("Container %q is reporting partially available!", containerInfo.Name))
+		} else {
+			st.Step(terminal.StatusError, fmt.Sprintf("Container %q is reporting not ready!", containerInfo.Name))
+		}
+
+		// Extra advisory wording to let user know that the deployment could be still starting up
+		// if the report was generated immediately after it was deployed or released.
+		st.Step(terminal.StatusWarn, "Waypoint detected that the current deployment "+
+			"is not ready, however your\napplication might be available or still starting up.")
```
evanphx, line 233: I think in this case I'd probably nix the `\n` and just have it be a long line. I don't want plugin authors to get in the habit of trying to control the output by adding newlines to step statuses, etc.

evanphx (changes requested):

builtin/docker/platform.go
briancain: Sounds good, maybe I'll adjust it to be similar to the long message we print for kind clusters for the serverinstall:

https://github.com/hashicorp/waypoint/blob/e26abbc526af9ad049fb89206d8dc6f228154d0e/internal/serverinstall/k8s.go#L1471-L1478

evanphx (approved):

PR #1564, opened 2021-06-01
Add support for installing Waypoint server to AWS ECS 

catsby:
This PR adds initial support for installing the Waypoint server into AWS ECS as a service. The installation borrowed heavily from the previous impelemention of ECS applications created by waypoint, and has a similar architecture. 

## UX

```console
$ waypoint install --platform=ecs -accept-tos -ecs-cluster waypoint-server 
✓ Networking setup
✓ Created new ECS cluster: waypoint-server
✓ EFS ready
✓ Found existing IAM role to use: waypoint-server-execution-role
✓ Created CloudWatchLogs group to store logs in: waypoint-server-logs
✓ Network load balancer created
✓ Created ECS Service (waypoint, cluster-name: waypoint-server)
✓ Waiting for target group to be healthy...
✓ Service launched!
✓ Server installed and configured!
✓ Installing runner...
✓ Found existing IAM role to use: waypoint-server-execution-role
✓ Created CloudWatchLogs group to store logs in: waypoint-runner-logs
✓ Runner service created

The CLI has been configured to connect to the server automatically. This
connection information is saved in the CLI context named "install-1622563419".
Use the "waypoint context" CLI to manage CLI contexts.

The server has been configured to advertise the following address for
entrypoint communications. This must be a reachable address for all your
deployments. If this is incorrect, manually set it using the CLI command
"waypoint server config-set".

To launch and authenticate into the Web UI, run:
waypoint ui -authenticate

Advertise Address: <NLB-public-dns>.elb.us-west-2.amazonaws.com:9701
Web UI Address: https://<NLB-public-dns>.elb.us-west-2.amazonaws.com:9702

```

## Infrastructure used
 
The main components installed are detailed below:

- EFS Filesystem: used for persisting the Waypoint server database
- EC2 Security Group: authorizing traffic in to the cluster, as well as enabling access to the EFS FileSystem from the Cluster
- CloudWatch Log Groups: a log group is created for each of the Waypoint server and runner services
- IAM Execution role
- Network load balancer: required to enable both HTTP and gRPC connections*
- Load balancing target groups: a target group is created for each of the Waypoint server and runner services
- ECS Cluster: a cluster dedicated to running the Waypoint server and runner
- ECS server service: runs the Waypoint server task
- ECS runner service: runs the default Waypoint runner task
- ECS Task Definitions: for both the server and runner

------

\* A Network Load Balancer is required because user-supplied TLS certificates are not yet supported


internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2190 @@
...
+	defaultNLBName           = "waypoint-server-nlb"
+)
+
+var ulid string
+
+func init() {
+	// generate a ULID for use in FileSystem creation and unique tag value for
+	// this installation
+	ulid, _ = component.Id()
+}
```
evanphx, line 49: This feels a bit odd, could we move this inside Install and have it be local instead of global?

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2190 @@
...
+
+type ECSInstaller struct {
+	config ecsConfig
+}
+
+type ecsConfig struct {
+	// serverImage is the image/tag of the Waypoint server to use. Default is
+	// hashicorp/waypoint:latest
+	serverImage string `hcl:"server_image,optional"`
```
evanphx, line 58: Pretty sure that private data members can't be written/read by HCL, so this won't work.

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2190 @@
...
+		})
+	}
+
+	def := ecs.ContainerDefinition{
+		Essential: aws.Bool(true),
+		Command: []*string{
+			aws.String("runner"),
+			aws.String("agent"),
+			aws.String("-vvv"),
+			aws.String("-liveness-tcp-addr=:1234"),
```
evanphx, line 2037: 1234?

evanphx (changes requested):

briancain (changes requested):
Generally looks good! Nice job 😄 

The other major comment here that I have is instead of using lots of `Status` outputs, we should use StepGroups for things like this for example:

These messages have green checks but read like Waypoint is still checking on them

```
✓ Waiting for EFS mount targets to become available...
```

```
✓ Installing runner...
```

Once it becomes available or installed we should update the stepgroup refresh the line and say the waiting or installing has completed.

Another example, we can combine some of these messages into stepgroups so we aren't printing too much output on install:

```
✓ Examining existing CloudWatchLogs groups...
✓ Created CloudWatchLogs group to store logs in: waypoint-server-logs
```

These kinds of messages should be combined into a single stepgroup message too.

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2190 @@
...
+		Default: "waypoint-server",
+	})
+	set.StringVar(&flag.StringVar{
+		Name:    "region",
+		Target:  &i.config.region,
+		Usage:   "Configures the region specific things.",
+		Default: "us-west-2",
+	})
+	set.StringSliceVar(&flag.StringSliceVar{
+		Name:   "subnets",
```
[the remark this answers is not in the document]
briancain: I rushed to post these during our sync but for the outside observer, we made a decision early on for serverinstall flags to always be namespaced with the platform they're for, so that's the reasoning here 😄

internal/serverinstall/ecs.go
catsby: As we discussed, I removed the `ulid` value completely and changed how we do tags in bcc87175

internal/serverinstall/ecs.go
catsby: Made them public in 65a54924

internal/serverinstall/ecs.go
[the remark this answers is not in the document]
catsby: I updated the names to have prefixes in 815954cb

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2190 @@
...
+		Services: []*string{aws.String(runnerName)},
+	})
+	if err != nil {
+		return false, err
+	}
+
+	return len(services.Services) > 0, nil
+}
+
+func (i *ECSInstaller) InstallFlags(set *flag.Set) {
```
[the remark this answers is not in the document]
catsby: we discussed this internally and are going to leave things as they are for now

internal/serverinstall/ecs.go
catsby: We discussed this internally, and are going to leave it for now. I borrowed this from the [k8s install](https://github.com/hashicorp/waypoint/blob/afb837e3f2e77ca102926fbddcefa114bdbf4971/internal/serverinstall/k8s.go#L999-L1002)

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2190 @@
...
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+	"github.com/hashicorp/waypoint/internal/serverconfig"
+)
+
+const (
+	defaultRunnerLogGroup = "waypoint-runner-logs"
+	defaultServerLogGroup = "waypoint-server-logs"
+
+	defaultTaskCPU     = 512
+	defaultTaskMemory  = 1024
```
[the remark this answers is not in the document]
catsby: Made these configurable in cfbc923a

.changelog/1564.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:feature
+New server platform - ECS: Use `--platform=ecs` to install waypoint server to
```
briancain, line 2:
```suggestion
serverinstall/ecs: New server platform - ECS: Use `--platform=ecs` to install waypoint server to
```

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2177 @@
...
+	// - ELB Network Load Balancer
+	// - EFS File System
+
+	s.Update("Deleting ECS resources...")
+	if err := deleteEcsResources(ctx, sess, resources); err != nil {
+		return err
+	}
+	s.Done()
+
+	s = sg.Add("Deleting Cloud Watch Log Group resources...")
```
briancain, line 591: Maybe instead of a new stepgroup we can just make this an `Update` so it just is all 1 line of output when these resources get deleted? If not, maybe add an `Update` after the deletion to change `Deleting` to `Deleted` 🤔

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2177 @@
...
+		if *r.ResourceType == "AWS::ECS::Cluster" {
+			clusterArn = *r.ResourceArn
+		}
+	}
+	s.Update("Deleting ECS resources...")
+	if err := deleteEcsCommonResources(ctx, sess, clusterArn, resources); err != nil {
+		return err
+	}
+	s.Done()
+	s = sg.Add("Deleting Cloud Watch Log Group resources...")
```
briancain, line 990: Maybe same comment here, could combine these into an Update and finish with `"Deleted ECS resources"`

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2177 @@
...
+	set.StringVar(&flag.StringVar{
+		Name:    "ecs-cluster",
+		Target:  &i.config.Cluster,
+		Usage:   "Configures the Cluster to install into.",
+		Default: "waypoint-server",
+	})
+	set.StringVar(&flag.StringVar{
+		Name:    "ecs-region",
+		Target:  &i.config.Region,
+		Usage:   "Configures the region specific things.",
```
briancain, line 1034:
```suggestion
		Usage:   "Configures which AWS region to install into.",
```

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2177 @@
...
+	set.StringVar(&flag.StringVar{
+		Name:    "ecs-server-image",
+		Target:  &i.config.ServerImage,
+		Usage:   "Docker image for the Waypoint server.",
+		Default: defaultServerImage,
+	})
+	set.StringVar(&flag.StringVar{
+		Name:    "ecs-region",
+		Target:  &i.config.Region,
+		Usage:   "Configures the region specific things.",
```
briancain, line 1085:
```suggestion
		Usage:   "Configures which AWS region to install into.",
```

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2177 @@
...
+		Usage:   "Configures the requested memory amount for the Waypoint server task in ECS.",
+		Default: "1024",
+	})
+}
+
+func (i *ECSInstaller) UninstallFlags(set *flag.Set) {
+	set.StringVar(&flag.StringVar{
+		Name:    "ecs-cluster",
+		Target:  &i.config.Cluster,
+		Usage:   "Configures the Cluster to upgrade.",
```
briancain, line 1107:
```suggestion
		Usage:   "Configures the Cluster to uninstall.",
```

internal/serverinstall/ecs.go
```diff
@@ -0,0 +1,2177 @@
...
+		Default: "waypoint-server",
+	})
+	set.StringVar(&flag.StringVar{
+		Name:    "ecs-region",
+		Target:  &i.config.Region,
+		Usage:   "Configures the region specific things.",
```
briancain, line 1113:
```suggestion
		Usage:   "Configures which AWS region to uninstall from.",
```

briancain (approved):
Looks good to me! I just ran through the code and left some suggestions around output and flag doc strings. 🎉

.changelog/1564.txt
catsby: I thought about this one and felt that the package name (`serverinstall`) was too niche to be immediately obvious to causal readers what it meant. I can accept this if you feel strongly about it though

internal/serverinstall/ecs.go
catsby: Agreed, done in 6b212967

internal/serverinstall/ecs.go
catsby: Agreed, done in 6b212967

.changelog/1564.txt
briancain: Hmnm good point. We do need something here. In the past I've used `serverinstall`. I don't know if `plugin/ecs` makes sense, but maybe `install/ecs`?

evanphx (approved):
Looks good!

PR #1567, opened 2021-06-02
Feature: status reporting for ALB releases

izaaklauer:
This implements status reports for the aws alb releaser plugin. It waits briefly for the release's target group to finish initializing, then reports on the health of the targets.

A few caveats:
- For applications that are slow to start (and most ec2 deployments), the initial status check will return unhealthy or partial. 
- I'm not looking at target group health. Currently not reporting on the state of the ALB, security group, DNS records, or any other resources at play during a release.
- This assumes we only have one target group associated with each release. When we implement more complex deployment strategies that require having multiple target groups behind the same ALB, this will need to be re-thought.
- Lambda function target groups have health checks disabled. Rather than present state UNKNOWN for all lambda releases, I'm mapping disabled to healthy. If that disabled state shows up in any other scenarios, we may need to revisit this.
- I'm not nil-checking before dereferencing pointers returned by the aws api, in keeping with our existing patterns.
 
Tested with deployments from ec2 and lambda.


builtin/aws/alb/releaser.go
```diff
@@ -304,9 +315,145 @@ func (r *Releaser) Release(
 	return &Release{
 		Url:             "http://" + hostname,
 		LoadBalancerArn: *lb.LoadBalancerArn,
+		TargetGroupArn:  target.Arn,
+		Region:          target.Region,
```
izaaklauer, line 319: I needed the region to build the right client. [Other plugins](https://github.com/hashicorp/waypoint/blob/main/builtin/aws/ec2/platform.go#L95) have this on their config. Not sure which pattern we prefer.

builtin/aws/alb/releaser.go
```diff
@@ -304,9 +315,145 @@ func (r *Releaser) Release(
...
 
+func (r *Releaser) Status(
+	ctx context.Context,
+	log hclog.Logger,
+	release *Release,
+	ui terminal.UI,
+) (*sdk.StatusReport, error) {
+
+	if release.Region == "" {
+		log.Debug("Region is not available for this release. Unable to determine status.")
```
izaaklauer, line 331: I have a feeling that we might hit this case in the upgrade path, where someone upgrades a server then tries to promote a deployment that didn't encode the region?

krantzinator (approved):
I haven't fully tested this, but the basics look good.

builtin/aws/alb/releaser.go
krantzinator: This would be good to test, sounds like a correct assumption.

builtin/aws/alb/releaser.go
```diff
@@ -304,9 +315,145 @@ func (r *Releaser) Release(
...
+		})
+	}
+
+	// If AWS registers targets slowly or incrementally, we may report an artificially low number of total targets.
+	totalTargets := len(targetHealthDescriptions)
+
+	if healthyCount == totalTargets {
+		report.Health = sdk.StatusReport_READY
+		msg := fmt.Sprintf("All %d targets are healthy.", totalTargets)
+		report.HealthMessage = msg
```
briancain, line 441:
```suggestion
		report.HealthMessage =  fmt.Sprintf("All %d targets are ready.", totalTargets)
		st.Step(terminal.StatusOk, report.HealthMessage)
```

builtin/aws/alb/releaser.go
```diff
@@ -304,9 +315,145 @@ func (r *Releaser) Release(
...
+	} else if healthyCount > 0 {
+		report.Health = sdk.StatusReport_PARTIAL
+		msg := fmt.Sprintf("%d/%d targets are healthy.", healthyCount, totalTargets)
+		report.HealthMessage = msg
+		st.Step(terminal.StatusWarn, msg)
+	} else {
+		report.Health = sdk.StatusReport_DOWN
+		msg := fmt.Sprintf("All targets are unhealthy, however your application might be available or still\nstarting up.")
+		report.HealthMessage = msg
+		st.Step(terminal.StatusWarn, msg)
```
briancain, line 451:
```suggestion
		report.HealthMessage =  "All targets are unhealthy, however your application might be available or still starting up."
		st.Step(terminal.StatusWarn, report.HealthMessage)
```

builtin/aws/alb/releaser.go
```diff
@@ -304,9 +315,145 @@ func (r *Releaser) Release(
...
+	report.Resources = []*sdk.StatusReport_Resource{}
+
+	healthyCount := 0
+	for _, tgHealth := range targetHealthDescriptions {
+
+		targetId := *tgHealth.Target.Id
+
+		var health sdk.StatusReport_Health
+
+		switch *tgHealth.TargetHealth.State {
```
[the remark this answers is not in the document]
briancain: Maybe leave a note above the switch?

briancain (changes requested):

builtin/aws/alb/releaser.go
```diff
@@ -304,9 +315,145 @@ func (r *Releaser) Release(
...
+	startTime := time.Now().Unix()
+	for {
+		if targetHealthDescriptions != nil {
+			sleepDuration := time.Second * time.Duration(targetGroupInitializationPollingIntervalSeconds)
+			log.Debug("Sleeping %0.f seconds to give the following target group time to initialize:\n%s", sleepDuration.Seconds(), release.TargetGroupArn)
+			time.Sleep(sleepDuration)
+		}
+
+		if startTime+int64(targetGroupInitializationTimeoutSeconds) <= time.Now().Unix() {
+			return nil, fmt.Errorf("timed out after %d seconds waiting for the following target group to initialize:\n%s", time.Now().Unix()-startTime, release.TargetGroupArn)
```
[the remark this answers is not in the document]
briancain: Or even better just mark the status report as Unknown and break out of the loop completely and then the health message could be that the initialization timed out after `x` seconds.

builtin/aws/alb/releaser.go
```diff
@@ -12,14 +11,21 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	"github.com/hashicorp/waypoint-plugin-sdk/docs"
+	sdk "github.com/hashicorp/waypoint-plugin-sdk/proto/gen"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/builtin/aws/utils"
+	"time"
```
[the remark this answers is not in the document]
krantzinator: I have a plugin that auto-arranges these for me. I'm actually not sure if it's the base Go extension or something I added separately.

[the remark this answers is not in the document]
briancain: Same... mine auto-formats!

briancain (approved):
Code looks good to me! 🎉

PR #1570, opened 2021-06-02
Bug: adding release or deployment ID to the status report

izaaklauer:
I don't think this is the *right* way to wire this information through, but it seems to work for now.

### DIFF
--- a/internal/core/app_status_report.go
+++ b/internal/core/app_status_report.go
@@ -334,7 +334,9 @@ func (op *statusReportOperation) Do(
 	}
 	realMsg.ResourcesHealth = resourcesHealth
 
-	// Add the deployment/release ID to the report. TODO: wire the resource ID through explicitly
+	// Add the deployment/release ID to the report.
+	// TODO: this is a stopgap solution - we should wire resource ID information into here in a more generic way
+	// rather than continue to append to this switch case.
 	switch op.Target.(type) {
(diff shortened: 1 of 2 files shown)


internal/core/app_status_report.go
```diff
@@ -333,6 +334,16 @@ func (op *statusReportOperation) Do(
 	}
 	realMsg.ResourcesHealth = resourcesHealth
 
+	// Add the deployment/release ID to the report. TODO: wire the resource ID through explicitly
+	switch op.Target.(type) {
```
mitchellh, line 338: You want to use a type switch with an assignment here: https://golang.org/doc/effective_go#type_switch

That'll clean things up a lot for you.

briancain (changes requested):
Just a quick change on the error return!

mitchellh:
Also @briancain is there no way to test this? This seems like a good test candidate.

briancain:
@mitchellh Yep, we should update the tests too for this and be more explicit as to what values are returned. Right now we're mostly just checking that something was returned: 

https://github.com/hashicorp/waypoint/blob/main/internal/core/app_status_report_test.go#L124-L126

If you don't mind @izaaklauer - we should also add a check to match deploy/release id in the report.

briancain (approved):

PR #1594, opened 2021-06-03
Feature: CLI list deployments shows health for each entry

izaaklauer:
Previously, running `waypoint deployment list` would attach the most recent health check of any kind to the most recent deployment. With this change, we display the health of each deployment listed.

Example:
Screen Shot 2021-06-03 at 6 48 31 PM

In writing this, I realize that our ListStatusReports endpoint lists _all_ status reports for all resources, with no way to filter to a specific resource or recency limit. We'll probably need to enhance that pretty soon as we build up more report types and volume.

Also it's a big protobuf diff, but it should get us back to the the code produced by the version of protoc locked in by go.mod.


internal/core/app_status_report.go
```diff
@@ -345,6 +344,9 @@ func (op *statusReportOperation) Do(
 		return nil, status.Errorf(codes.FailedPrecondition, "unsupported status operation type")
 	}
 
+	// Add the time generated to the outer status report
+	realMsg.TimeGenerated = report.TimeGenerated
```
izaaklauer, line 348: This feels like something the UI will likely want to include (cc @jgwhite ), so I added it to the top level StatusReport. It also means we don't need to dig into the lower-level status report in the CLI at all anymore.

jgwhite: Yes, this is going to be super helpful in the UI too.

internal/cli/deployment_list.go
```diff
@@ -177,28 +165,29 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
-						statusReportComplete = "?"
-					}
-					if t, err := ptypes.Timestamp(statusReport.TimeGenerated); err == nil {
-						statusReportComplete = fmt.Sprintf("%s - %s", statusReportComplete, humanize.Time(t))
+			// Add status report information if we have any
+			statusReportComplete := "Unknown Status"
+			for _, statusReport := range statusReportsResp.StatusReports {
+				if deploymentTargetId, ok := statusReport.TargetId.(*pb.StatusReport_DeploymentId); ok {
+					if deploymentTargetId.DeploymentId == b.Id {
+						switch statusReport.Health.HealthStatus {
```
jgwhite, line 173: Happy to see we’re switching on the outer `HealthStatus` rather than the SDK’s `Health` enum.

internal/server/proto/server.proto
```diff
@@ -2387,6 +2387,9 @@ message StatusReport {
     // Resource id
     string id = 4;
   }
+
+  // the time when this report was generated
+  google.protobuf.Timestamp time_generated = 10;
```
jgwhite, line 2392: Small consistency thing: elsewhere in the schema we seem to use `_time` as a suffix rather than a prefix. i.e. `Workspace.active_time`, `Status.start_time`, `Status.complete_time`, `Job.complete_time`. I wonder if it might be worth following suit here, even if it doesn’t match the SDK quite as well.

```suggestion
  google.protobuf.Timestamp generated_time = 10;
```

jgwhite (commented):
Works great!

image

One small comment on naming consistency, but otherwise looks good to me.

internal/cli/deployment_list.go
```diff
@@ -10,19 +10,15 @@ import (
...
-	sdk "github.com/hashicorp/waypoint-plugin-sdk/proto/gen"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	clientpkg "github.com/hashicorp/waypoint/internal/client"
 	"github.com/hashicorp/waypoint/internal/clierrors"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 	serversort "github.com/hashicorp/waypoint/internal/server/sort"
+	"github.com/posener/complete"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
krantzinator, line 21: These should move back up to line 13-15

internal/core/app_status_report.go
```diff
@@ -345,6 +344,9 @@ func (op *statusReportOperation) Do(
 		return nil, status.Errorf(codes.FailedPrecondition, "unsupported status operation type")
 	}
 
+	// Add the time generated to the outer status report
+	realMsg.GeneratedTime = report.TimeGenerated
```
krantzinator, line 348: :chefs-kiss:

krantzinator (commented):
Excited to have this! I had a couple questions and one nit.

internal/cli/deployment_list.go
```diff
@@ -177,28 +165,29 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
-						statusReportComplete = "✖"
-					case sdk.StatusReport_PARTIAL:
-						statusReportComplete = "●"
-					case sdk.StatusReport_UNKNOWN:
-						statusReportComplete = "?"
-					}
-					if t, err := ptypes.Timestamp(statusReport.TimeGenerated); err == nil {
-						statusReportComplete = fmt.Sprintf("%s - %s", statusReportComplete, humanize.Time(t))
+			// Add status report information if we have any
+			statusReportComplete := "Unknown Status"
```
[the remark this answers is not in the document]
briancain: Agreed, that sounds better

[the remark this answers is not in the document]
izaaklauer: Agreed. 
Screen Shot 2021-06-04 at 11 10 37 AM

internal/cli/deployment_list.go
```diff
@@ -112,22 +108,14 @@ func (c *DeploymentListCommand) Run(args []string) int {
 		sort.Sort(serversort.DeploymentCompleteDesc(resp.Deployments))
 
 		// get status reports
-		statusReportResp, err := client.GetLatestStatusReport(ctx, &pb.GetLatestStatusReportRequest{
+		statusReportsResp, err := client.ListStatusReports(ctx, &pb.ListStatusReportsRequest{
 			Application: app.Ref(),
 			Workspace:   wsRef,
 		})
+
 		if status.Code(err) == codes.NotFound {
```
briancain, line 116: Might need to rebase to get Raes fix for when the API for status reports is unimplemented here.

internal/server/proto/server.proto
```diff
@@ -2387,6 +2387,9 @@ message StatusReport {
     // Resource id
     string id = 4;
   }
+
+  // the time when this report was generated
+  google.protobuf.Timestamp generated_time = 10;
```
briancain, line 2392: We should leave a note on line 2370 that these variables (health, resources_health, and generated_time) are all copied from the plugin generated raw SDK status report and are a convenience for the UI and CLI. Also move this line to under `resources_health` so that the Health message is the last thing here (that's mostly a style thing)

briancain (changes requested):
Sweeeet, thanks for the fix izaak 😄 A couple of minor comments from me and some +1s on the other feedback given 👍🏻

internal/cli/deployment_list.go
mitchellh: Yep, more background for @izaaklauer if you aren't aware of our style, we do:

```go
import(
  <stdlib stuff>

  <3rd party stuff>

  <internal stuff>
)
```

internal/core/app_status_report.go
mitchellh: Why is one `AB` and the other `BA` (A = "Generated" B = "Time")?

Renaming fields is backwards compatible in proto so should we rename fields?

briancain: Good catch...yeah these should be named the same.

internal/cli/deployment_list.go
```diff
@@ -112,22 +108,14 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
 			err = nil
-			statusReportResp = nil
-		}
-		if err != nil {
```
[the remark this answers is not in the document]
izaaklauer: You make a great point. I misunderstood what this was checking for.

internal/core/app_status_report.go
izaaklauer: Ha, I actually [just changed this](https://github.com/hashicorp/waypoint/pull/1594/commits/0132e30ed5828bf2b85f3c46eeafa23238123985) after @jgwhite caught that many similar time fields that the server returns are in the AB order, and it might be better to be consistent with other server fields than with the sdk field that we're mirroring: https://github.com/hashicorp/waypoint/pull/1594#discussion_r645457560

internal/cli/deployment_list.go
izaaklauer: Ah ok. My IDE has a mind of it's own when it comes to import ordering - i'll wrangle it into matching our convention.

internal/core/app_status_report.go
izaaklauer: If renaming fields is backwards compatible, fixing the SDK field may be the way to go.

briancain: `realMsg.GeneratedTime` was added in this PR so we should match the naming from the SDK.

https://github.com/hashicorp/waypoint/pull/1594/files#diff-08cc775443acf74b3ff82990db5a6a1dc18774b7701d3851dad9ff91e5947b37R2381

briancain: Ah, I noticed Jamies comment. We could flip them in the SDK if it doesn't break backwards compatibility. It's too bad this wasn't caught sooner 😄

briancain (commented):
@izaaklauer I think we should fix the `GeneratedTime = TimeGenerated` issue here. We should either flip it back to `TimeGenerated` on the copied convenience server proto var to match the option in the raw report ([and be a little inconsistent](https://github.com/hashicorp/waypoint/pull/1594#discussion_r645457560)), or try to fix the name on the SDK side if it doesn't break backwards compatibility so it is consistent with other parts of the API. If we do change it, it'd be good to get it in a 0.4.1 before anyone really builds anything on top of this to reduce major breaking changes.

izaaklauer:
PR to rename the time field in waypoint-plugin-sdk: https://github.com/hashicorp/waypoint-plugin-sdk/pull/35. I tested using that change in a new server with an old version of the CLI and vice-versa, and the backwards compatibility works! Once the sdk change is merged, I'll update this pr to include it.

briancain:
Nice @izaaklauer ! You'll probably also need to update all of the plugins to use the new variable too when they finish generating a report, right? Maybe that can be done in a new PR when we update the SDK version this project is using.

izaaklauer:
> Nice @izaaklauer ! You'll probably also need to update all of the plugins to use the new variable too when they finish generating a report, right? Maybe that can be done in a new PR when we update the SDK version this project is using.

The [other plugin changes](https://github.com/hashicorp/waypoint/pull/1594/commits/30647573bb24dd6f822ba8c4782cd12b05aeeae3) weren't too big and addresses this feedback, so I looped it in here. We are now consistently using `<thing>_time`!

izaaklauer:
Note that because with this change we're [recording](https://github.com/hashicorp/waypoint/pull/1594/files#diff-2bc7621eb360095ad390e9e13befe84f01090f95e81531a1c52f62f7d28c9c90R348) and [reading](https://github.com/hashicorp/waypoint/pull/1594/files#diff-a480951daf0e1c05f5cc0afd46ba4fb0240c0a189668f4b529610263d11cdfaaR191) GeneratedTime from the outer protobuf, and [reading HealthStatus from the outer protobuf](https://github.com/hashicorp/waypoint/pull/1594/files#diff-a480951daf0e1c05f5cc0afd46ba4fb0240c0a189668f4b529610263d11cdfaaR178). We used to read these from the [inner sdk pb.any protobuf](https://github.com/hashicorp/waypoint/pull/1594/files#diff-a480951daf0e1c05f5cc0afd46ba4fb0240c0a189668f4b529610263d11cdfaaL130). This means that `deployment list` commands with this release will not show a time or health for old status reports. Example:

Screen Shot 2021-06-09 at 2 48 10 PM

(deployment 3 was made with 0.4.0, and 4 with this new version. Note that 3 is missing health info).

The UI is also unaffected, because it doesn't inspect the inner sdk proto. 

With the current implementation of health checks, health is only determined immediately after a deployment, and is most valuable at that point, so losing that historical health check data in this CLI view may be acceptable. The next deployment that anyone runs after upgrading waypoint will show health on `deployment list`.

I would argue that it's not worth the complexity to check the SDK proto in the case where the outer proto is blank, but I'm not wedded to that opinion.

briancain (approved):
⏳

PR #1600, opened 2021-06-04
ui: Apply consistent copywriting to English translations

almonk:
This PR makes changes to our translations file to align our UI copy with that of our design system guidelines (https://docs.google.com/document/d/1ABNkYIp6aLwPYNpuYOFcDFBFCcnH_4ePRaLP-gD_tKE/edit).

I am also beginning to codify other guidelines (such as respecting brand name capitalisation e.g. `Git` not `git`) in a separate document.


jgwhite (approved):

ui/translations/en-us.yaml
```diff
@@ -37,34 +37,34 @@ page:
...
     git_source_ssh_password_label: 'Passphrase (optional)'
     git_source_ssh_key_label: 'SSH Private Key'
     git_source_basic_username_label: 'Username'
     git_source_basic_password_label: 'Password'
-    git_sync_label: 'Automated Sync'
+    git_sync_label: 'Automated sync'
     git_sync_description_label: 'Enable automated deployment on changes'
     hcl_title: 'Waypoint.hcl settings'
-    hcl_location: 'Waypoint.hcl config location'
+    hcl_location: 'waypoint.hcl config location…'
```
gregone, line 64: why the ellipsis here?

gregone (approved):
LGTM, just one minor comment.

ui/translations/en-us.yaml
almonk: @gregone In the UI the action buttons finish the sentence, so I think it makes sense to use the ellipsis to indicate that. I don't feel super strongly about it though.

CleanShot 2021-06-08 at 14 33 02

briancain: What about something like this instead

> waypoint.hcl config location:

gregone: @briancain 
> What about something like this instead

> >waypoint.hcl config location:

The main reason against it is consistency, no other labels use colons, so I'd avoid introducing a change in text patterns. 

@dizzyup I agree with that choice, removing the "sentence" style gets rid of the problem

gregone:
@almonk can we finish this and merge it?

PR #1603, opened 2021-06-04
website: document configuration files and internal values

mitchellh:
We forgot to write the docs for this amazing feature!

### DIFF
--- a/website/content/docs/app-config/internal.mdx
+++ b/website/content/docs/app-config/internal.mdx
@@ -2,16 +2,16 @@
 layout: docs
 page_title: Internal Values - Application Configuration
 description: |-
-  Internal values are static or dynamic configuration values that can be referenced by environment variable or files to compose values.
+  Internal values are static or dynamic configuration values that can be referenced by environment variables or files to compose values.
 ---
 
 # Internal Values
 
 Internal values are static or [dynamic](/docs/app-config/dynamic)
-configuration values that can be referenced by environment variables
-or [files](/docs/app-config/files) to compose values. An example common
-usecase is building a database connection string which requires a database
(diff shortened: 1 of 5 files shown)


website/content/docs/app-config/internal.mdx
```diff
@@ -0,0 +1,51 @@
...
+page_title: Internal Values - Application Configuration
+description: |-
+  Internal values are static or dynamic configuration values that can be referenced by environment variable or files to compose values.
+---
+
+# Internal Values
+
+Internal values are static or [dynamic](/docs/app-config/dynamic)
+configuration values that can be referenced by environment variables
+or [files](/docs/app-config/files) to compose values. An example common
```
evanphx, line 12: internal can reference internal too.

evanphx (changes requested):
Should we clarify that internal can reference internal?

website/content/docs/app-config/internal.mdx
mitchellh: yes definitely! will update that

website/content/docs/app-config/internal.mdx
```diff
@@ -0,0 +1,51 @@
+---
+layout: docs
+page_title: Internal Values - Application Configuration
+description: |-
+  Internal values are static or dynamic configuration values that can be referenced by environment variable or files to compose values.
```
izaaklauer, line 5: environment variables (plural)

izaaklauer (commented):
These are great features, and i'm glad I know about them now!

briancain (approved):
Other than the two review comments looks good to me 👍🏻

PR #1620, opened 2021-06-08
Enhancement: moving api-common-protos from vendor to 3rdparty

izaaklauer:
If there is a `vendor` folder at the project root, the Goland IDE ignores go modules and downloads all dependencies into that vendor folder. Until now, I've been working around this by deleting the vendor folder until I need to regenerate protobufs. Renaming `vendor` to `3rdparty` also makes us more consistent with waypoint-plugin-sdk.

I also added tests to the `make gen` commands to verify that the submodule has been initialized, and instructs you to do so if not.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "3rdparty/proto/api-common-protos"]
-	path = 3rdparty/proto/api-common-protos
+[submodule "thirdparty/proto/api-common-protos"]
+	path = thirdparty/proto/api-common-protos
 	url = https://github.com/googleapis/api-common-protos
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
-[submodule "vendor/proto/api-common-protos"]
-	path = vendor/proto/api-common-protos
+[submodule "3rdparty/proto/api-common-protos"]
+	path = 3rdparty/proto/api-common-protos
 	url = https://github.com/googleapis/api-common-protos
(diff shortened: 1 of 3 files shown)


denysvitali:
Oh my god, thank you! You can't even imagine how much I struggled with this think. Goland as soon as it sees the `vendor` directory it completely ignores the GOPATH and syncs again every dependency in there. So frustrating!

Thanks for solving the issue, I owe you a beer! 🍺

mitchellh:
The only thing I don't like about `3rdparty` is that it sorts to the top, which feels a bit weird on browsing experiences like GitHub. Though, I can't think of any better name.

mitchellh (approved):

izaaklauer:
> The only thing I don't like about `3rdparty` is that it sorts to the top, which feels a bit weird on browsing experiences like GitHub. Though, I can't think of any better name.

Leading with a numeral is kind of weird too. We could do `thirdparty` instead.

briancain:
@izaaklauer `thirdparty` makes sense to me!

izaaklauer:
Changed to `thirdparty` here and in [waypoint-plugin-sdk](https://github.com/hashicorp/waypoint-plugin-sdk/pull/34/commits/d54d56410e3213622e68b1c251ed83de07a5c17b).

briancain (approved):
3️⃣ 🎉

mitchellh (approved):

Issue #1537, opened 2021-05-25
Upgrade from bolt to bbolt


izaaklauer:
[Boltdb](https://github.com/boltdb/bolt) is no longer maintained, and causes intermittent test failures. Bolt also causes a runtime crash when attempting to run `go test` with the `-race` flag.

Etcd maintains a fork, [bbolt](https://github.com/etcd-io/bbolt), that is api-compatible, actively maintained, and has addressed at least [some](https://github.com/etcd-io/bbolt/pull/201) of our issues.

It remains to be seen if bbolt's database file is structurally compatible with bolt's. If it is, it should be straightforward to swap out libraries. If not, we'll need to migrate users to the new database file on startup.

Prior work converting raft-boltdb for reference: https://github.com/hashicorp/raft-boltdb/pull/22/commits/22de16c8e211a122225ee5868ff64b01bf837c3f

PR #1631, opened 2021-06-09
Upgrade bolt to bbolt

izaaklauer:
[Boltdb](https://github.com/boltdb/bolt) is no longer maintained, and causes intermittent test failures. Bolt also causes a runtime crash when attempting to run `go test` with the `-race` flag.

Etcd's bbolt is api-compatible with boltdb, and it's database file appears to be structurally compatible too. I've tested reading builds, config, deployments, jobs, projects, status reports, and releases created by `bolt`, and writing and reading new values with `bbolt`, and it all works as expected. 

There are no documented structural incompatibilities or related issues that I've found, and there are examples of [other](https://github.com/influxdata/kapacitor/commit/91b7bd7d29db8ece58bc557a2a8c49ad16338ae0) [projects](https://github.com/wakatime/wakatime-cli/commit/175ec848d36d3d29fd4aa1c144edddd5fdca82ce) doing this successfully.

Note that [raft-boltdb](https://github.com/hashicorp/raft-boltdb/pull/23) chose to perform this upgrade by migrating the database file, but moreso out of an abundance of caution than to work around any specific issues.

Closes https://github.com/hashicorp/waypoint/issues/1537

### DIFF
--- a/.changelog/1631.txt
+++ b/.changelog/1631.txt
@@ -1,3 +1,0 @@
-```release-note:improvement
-core: Upgrade from unmaintained `boltdb/bolt` to `etcd-io/bbolt`
-```
--- a/.changelog/1631.txt
+++ b/.changelog/1631.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: Upgrade from unmaintained `boltdb/bolt` to `etcd-io/bbolt`
+```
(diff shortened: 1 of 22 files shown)


mitchellh (approved):
As long as tests pass I approve.

.changelog/1631.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core: Upgrade from unmaintained `boltdb/bolt` to `etcd-io/bbolt`
```
mitchellh, line 2: I almost feel like a changelog entry isn't even required for this change. I usually avoid non-user-impacting changes from the CL.

Issue #1614, opened 2021-06-07
internal/state: Add Peek and Complete funcs to appOperation


briancain:
**Is your feature request related to a problem? Please describe.**

To support running a queued poller for any kind of app related thing in Waypoint, `appOperation` state items should include a way to Peek at the _next_ thing to be polled, as well as a way to mark that thing as Completed.

**Describe the solution you'd like**

Update the `state/app_operation` implementation to include a generic way to `Peek` at the next item to be polled, as well as a way to mark that poll as `Complete`, similar to how this is handled in the `state/project` package. Then any app operation that has state storage that needs polling can simply use the app_operation interface to have a way to Peek and Complete.

In support of https://github.com/hashicorp/waypoint/issues/1612

mitchellh:
There's a ton of ways to do this but this makes sense to make an operation pollable, but the interface should be separate of course because things like `Project` are not operations :) And things like `Project` may actually have multiple things they want to poll on (right now its just Git stuff but we could add another field to project).

I think you're thinking about this right but just noting. :)

briancain:
I'm closing this. As mentioned in the linked PR, there isn't a clear use-case for this yet. Given we implemented application status polling already, I think we're safe to close this one out.

PR #1649, opened 2021-06-11
internal/server/state: Add Poll options with Peek and Complete funcs to App Operations

briancain:
This pull request introduces the ground work to poll on any app operation. It also includes the supporting functions required for a queued poller such as Peek and Complete.

Fixes: https://github.com/hashicorp/waypoint/issues/1614


internal/server/proto/server.proto
```diff
@@ -2399,6 +2410,18 @@ message StatusReport {
...
   string id = 6;
 
+  // data_source_poll starts at number 30 to reserve any space for copied status
+  // report options on this message. Hopefully with RFC WP-072, any further
+  // status report details needed out of the raw report can be obtained through
+  // the json structure rather than manually copied and added here.
+
+  // Each polling event is tracked as a separate job. You can query the
+  // poll operations and their success/failure by using the ListJobs API.
+  Poll data_source_poll = 30;
```
mitchellh, line 2420: This feels a bit weird. Why is the polling settings _per_ status report? Shouldn't the polling settings for status reports be on the application or project? That way you're saying "For this project/app I want a status report every X seconds" ?

briancain: Hmm, right, that makes sense. In that case, I'd probably place polling on the application, and I don't think this PR is really needed then for polling status reports per app. I'm pretty sure then I can essentially use the built-in project polling state funcs since application is built off project. So just adding an ApplicationPollPeek/ApplicationPollComplete similar to how its other funcs already use project.

App Operation polling still might come in handy in the future so it's probably still useful to go forward with this minus the status report poll message changes 🤔 If not I could close this and redo it based on the plan ☝🏻

mitchellh: Yeah app operation polling might still come in handy one day. Agreed you probably don't need the status report stuff. 

And yeah I'd create an implementation of `applicationPollHandler` similar to project poll handler. I don't think we need to expose any user facing settings, just run each application that has a deployment/release like every 30 seconds or something (we can tune that later). 

Might be worth talking to Evan to hear his thoughts while I'm out.

briancain:
I think this _could_ come in handy some day, but since there isn't anything directly going to be using this I'm going to close it. We can always come back and open it again later if we ever decide to start polling on various app operations.

PR #1652, opened 2021-06-14
plugin/aws/alb bug: adding generated time to ALB releaser status report

izaaklauer:
This was missing in the initial implementation of aws ALB status reports.

Also note: it looks like `ptypes.TimestampNow` has been [deprecated](https://github.com/golang/protobuf/blob/v1.5.2/ptypes/timestamp.go#L52) in favor of `timestamppb.Now`, so I'm using that here instead.


briancain (approved):
Ahh, sorry we missed this 👍🏻

briancain:
I think this should probably have a changelog, it seems like it's probably a bug?

izaaklauer:
My thinking behind not including a changelog is that it's not user facing because we don't expose the release status reports in the UI or CLI yet.

briancain:
@izaaklauer - true! fair point, well that makes sense 👍🏻

PR #1658, opened 2021-06-15
UI: project input variables

### LINKED ISSUE #1391: Input Variables UI
Addition to https://github.com/hashicorp/waypoint/issues/1228 

Adds a Settings screen on Projects to be able to override Input variables from the browser UI.

gregone:
Closes #1391

This PR adds the ability for users to add/edit serve-side input variables

https://user-images.githubusercontent.com/1416421/122088944-8f0c3680-ce06-11eb-9243-cfbef4f56eb1.mov

We could follow-up on this with some improvements to the form layout (esp. for HCL input), but i'd like to discuss that with @krantzinator


ui/app/components/project-input-variables/list-item.ts
```diff
@@ -0,0 +1,85 @@
...
+  cancelEdit() {
+    this.isCreating = false;
+    this.isEditing = false;
+    this.variable = this.initialVariable;
+  }
+
+  @action
+  toggleHcl(variable) {
+    if (this.isHcl) {
+      this.isHcl = false;
```
sabrinako, line 76: Could we after the if/else statement do `this.isHcl = !this.isHcl` to reduce some of the similar logic here?

ui/tests/integration/components/project-input-variables-list-test.ts
```diff
@@ -0,0 +1,61 @@
...
+  cancelButton: clickable('[data-test-input-variables-edit-cancel]'),
+  saveButton: clickable('[data-test-input-variables-edit-save]'),
+  varName: fillable('[data-test-input-variables-var-name]'),
+  varStr: fillable('[data-test-input-variables-var-str]'),
+});
+
+module('Integration | Component | project-input-variables-list', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
```
sabrinako, line 23: Is there a reason to keep all this on one test? It might be clearer to have multiple tests for the different asserts at the bottom, so we could see clearly if a test fails what specific logic has broken. Also helpful in case someone needs to write new tests, they can easily read the test name to tell if the logic they are trying to test is already covered or not

sabrinako (changes requested):
Looks great so far! Just a few adjustments and I think it'd be ready to merge 🤠 👍

ui/tests/integration/components/project-input-variables-list-test.ts
jgwhite: Sabrina makes some really good points here. We should definitely clarify the name of this test, at least.

I’d say in tests for this sort of component (with significant behavior and state) we want two kinds of tests:

1. **Pure-rendering** — given these inputs, I except this DOM
2. **Flows** — given these inputs and this sequence of events, I expect this DOM and these side-effects

It’s also useful to write what I’d call “**integrity**” tests — given these inputs and this sequence of events, I expect state not to have leaked or fallen out of sync. With these, we have to figure out ways to *observe* if internal state has became out-of-sync or leaked, as we did on the regression tests for `<ProjectRepositorySettings>`.

ui/tests/integration/components/project-input-variables-list-test.ts
```diff
@@ -0,0 +1,61 @@
...
+    assert.equal(
+      page.variablesList.length,
+      2,
+      'Attempt to create then cancel: the list still has the normal count of variables'
+    );
+    await page.createButton();
+    await page.varName('var_name');
+    await page.varStr('foozbarz');
+    await page.saveButton();
+    assert.equal(page.variablesList.length, 3, 'Create Variable: the list has the new variable');
```
jgwhite, line 56: Note for the future: I don’t think we have a Mirage handler for `UpsertProject` yet but, we when do, it’d be good to add some assertions here about the state of Mirage’s DB.

ui/tests/integration/components/project-input-variables-list-test.ts
gregone: Yeah,  this is more of a user-flow test. I used the default test title and never changed it. 🙈 Maybe I'll add a simple rendering test for the empty state and an example initial variable set, to separate the concerns.

almonk:
Added a more explanatory empty state...

Safari 2021-06-21 at 14 28 16

ui/app/components/project-input-variables/list-item.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  @action
+  async deleteVariable(variable) {
+    await this.args.deleteVariable(variable);
+  }
+
+  @action
+  editVariable(variable) {
```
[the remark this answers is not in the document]
gregone: yup, that was a leftover from previous implementation

ui/app/components/project-input-variables/list-item.ts
```diff
@@ -0,0 +1,85 @@
...
+    super(owner, args);
+    this.isCreating = false;
+    let { variable, isEditing, isCreating } = args;
+    this.variable = variable;
+    this.isEditing = isEditing;
+    this.isCreating = isCreating;
+    this.isHcl = false;
+    if (variable.hcl) {
+      this.isHcl = true;
+    }
```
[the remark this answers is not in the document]
gregone: Oooh I think i remembered why i didn't do that. Somehow, that causes a state issue

ui/app/components/project-input-variables/list.ts
```diff
@@ -0,0 +1,88 @@
...
+  async saveVariableSettings(variable?: Variable.AsObject) {
+    let project = this.project;
+    let ref = new Project();
+    ref.setName(project.name);
+    if (variable) {
+      let existingVar = this.variablesList.find((v) => v.name === variable.name);
+      if (existingVar) {
+        existingVar = variable;
+      }
+    }
```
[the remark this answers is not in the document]
gregone: Yup, I didn't clean this up. I think I was initially reinserting the modified variable here, but changed it later

ui/app/components/project-input-variables/list-item.ts
gregone: I reworked this whole statement, so it's addressed

ui/tests/integration/components/project-input-variables-list-test.ts
gregone: Added a base rendering test separated from the user flow

ui/app/components/project-input-variables/list.hbs
```diff
@@ -0,0 +1,29 @@
...
+  {{#each this.variablesList as |variable|}}
+    <ProjectInputVariables::ListItem
+      @variable={{variable}}
+      @isEditing={{eq variable.name this.activeVariable.name}}
+      @isCreating={{this.isCreating}}
+      @saveVariableSettings={{this.saveVariableSettings}}
+      @deleteVariable={{this.deleteVariable}}/>
+
+  {{else}}
+    <Pds::EmptyState as |P|>
```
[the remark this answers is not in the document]
gregone: @almonk moved the empty state outside of the `<ul>` so it's fine now

ui/tests/integration/components/project-input-variables-list-test.ts
```diff
@@ -0,0 +1,79 @@
...
+        {
+          name: 'Varname2',
+          str: 'foo2',
+        },
+      ],
+    };
+    this.set('project', project);
+    await render(hbs`<ProjectInputVariables::List @project={{this.project}}/>`);
+    assert.dom('.project-input-variables-list').exists('The list renders');
+    assert.equal(page.variablesList.length, 2, 'the list contains all variables');
```
sabrinako, line 61: Cool, didn't realize there was a 3rd optional argument for `assert.equal()`!

sabrinako (approved):
👍  Looks good to me! Thanks for adding the error messages, helps clarify things a lot

ui/tests/integration/components/project-input-variables-list-test.ts
gregone: all assertions have that optional last argument, which makes the test output much more descriptive

ui/app/components/project-input-variables/list.ts
```diff
@@ -0,0 +1,89 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { Ref, Project, UpsertProjectRequest, Variable } from 'waypoint-pb';
+import { inject as service } from '@ember/service';
+import ApiService from 'waypoint/services/api';
+
+interface ProjectSettingsArgs {
+  project;
```
jgwhite, line 9: Could we give this a type? (Is it `Project.AsObject`?)

jgwhite (changes requested):
This is looking good.

I spotted one bug which I think might be related to `isCreating` state. Repro steps:

1. Visit `/:workspace/:project/settings/variables`
2. Click “Add variable”
3. Fill in variable name and value
4. Click “Save variable”
5. Click “Actions” for the newly-created variable
6. Click “Edit”
7. Change the name or value
8. Click “Save variable”
9. Observe the toast appears but the form does not disappear
10. Click “Save variable” again
11. Observe the form disappears this time

Another, similar variant:

1. Visit `/:workspace/:project/settings/variables`
2. Click “Add variable”
3. Fill in variable name and value
4. Click “Save variable”
5. Click “Actions” for the newly-created variable
6. Click “Edit”
7. Change the name or value
8. Click “Save variable”
9. Observe the toast appears but the form does not disappear
10. Click “Cancel”
11. Observe the variable has disappeared from the list

Here’s a video demonstrating the issue:

https://user-images.githubusercontent.com/34030/122925901-946a0380-d367-11eb-9ace-f48e7fa419d9.mp4

almonk:
@jgwhite watching your video, it also struck me that when I click 'add variable' or 'edit variable' it should focus/select the first form field for best usability

gregone:
@jgwhite Both variations of the bug are fixed. 
While doing so, I added mocks for the `UpsertProject` api call (it was painful). And while adding much more comprehensive tests, I also found a fixed a bug when editing variable names. With the mocks and tests in place, this feels pretty solid now.

ui/mirage/factories/project.ts
```diff
@@ -14,6 +14,18 @@ export default Factory.extend({
...
 
+  'with-input-variables': trait({
+    name: 'with-input-variables',
+
+    afterCreate(project, server) {
+      let variables = [
+        server.create('variable', 'random-str', { project }),
+        server.create('variable', 'random-str', { project }),
+        server.create('variable', 'random-hcl', { project }),
+      ];
```
jgwhite, line 25: Since we’re not doing anything further with this variables in this function, we can drop the referencing array.

```suggestion
      server.createList('variable', 2, 'random-str', { project });
      server.create('variable', 'random-hcl', { project });
```

ui/mirage/models/project.ts
```diff
@@ -16,7 +16,24 @@ export default Model.extend({
     result.setRemoteEnabled(this.remoteEnabled);
     result.setWaypointHcl(this.waypointHcl);
     result.setWaypointHclFormat(Project.Format.HCL);
-
+    // Somehow adding the toProtoBuf method to the variable wasn't working
+    // (probably because of the embedded relationship), so we're converting here for now.
```
jgwhite, line 20: Do you mean the following didn’t work?

```ts
let varProtosList = this.variables.models.map((v) => v.toProtobuf());
```

Do you recall what the error was?

ui/mirage/models/variable.ts
```diff
@@ -0,0 +1,20 @@
...
+  toProtobuf(): Variable {
+    let result = new Variable();
+
+    result.setServer();
+    result.setName(this.name);
+    if (this.hcl) {
+      result.setHcl(this.hcl);
+    } else {
+      result.setStr(this.str);
+    }
```
jgwhite, line 16: Does this need to same “unset the other field” behavior we see above?

```suggestion
    if (this.hcl) {
      result.setHcl(this.hcl);
      result.setStr('');
    } else {
      result.setHcl('');
      result.setStr(this.str);
    }
```

(aside: this aspect of protobuf `oneof` fields is a little inelegant isn’t it?)

ui/mirage/services/project.ts
```diff
@@ -23,3 +23,26 @@ export function get(schema: any, { requestBody }: Request): Response {
...
+    .getVariablesList()
+    .map((v) => v.toObject());
+  let model = schema.projects.findBy({ name });
+
+  let newVars = [];
+  variablesList.forEach((v) => {
+    newVars.push(schema.variables.create(v));
+  });
+
+  model.variables.models = newVars;
```
jgwhite, line 41: Mirage has a nice convenience for this:

```suggestion
  variablesList.forEach((v) => model.createVariable(v));
```

[See the mirage guides for documentation](https://miragejs.com/docs/main-concepts/relationships/#hasmany)

ui/tests/integration/components/project-input-variables-list-test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    let api = this.owner.lookup('service:api');
+    let ref = new Ref.Project();
+    ref.setProject(dbproj.name);
+    let req = new GetProjectRequest();
+    req.setProject(ref);
+
+    let resp = await api.client.getProject(req, api.WithMeta());
+    let project = resp.getProject();
+    this.set('project', project.toObject());
```
jgwhite, line 42: While this probably is the most-accurate way to create the same shape of object that the real app will deal with, I think for the purposes of this test we can skip the API layer:

```suggestion
    let dbproj = await this.server.create('project', 'with-input-variables', { name: 'Proj1' });
    let project = dbproj.toProtobuf();
    this.set('project', project.toObject());

```

ui/tests/integration/components/project-input-variables-list-test.ts
```diff
@@ -0,0 +1,123 @@
...
```
jgwhite, line 61: Similar to above:

```suggestion
    let dbproj = await this.server.create('project', { name: 'Proj3' });
    let project = dbproj.toProtobuf();
    this.set('project', project.toObject());

```

ui/tests/integration/components/project-input-variables-list-test.ts
```diff
@@ -0,0 +1,123 @@
...
```
jgwhite, line 94: Similar to above:

```suggestion
    let dbproj = await this.server.create('project', { name: 'Proj3' });
    let project = dbproj.toProtobuf();
    this.set('project', project.toObject());

```

jgwhite (changes requested):
Looking good. I’ve left a few superficial suggestions, but I think the bones of this are ready to go 👍

ui/mirage/models/project.ts
gregone: Yeah, that didn't work. I think I remember somehow getting `toProtoBuf is not a function` (whereas when I tried to add `toProtoBufRef`, that would work)

jgwhite: Could it have been a capitalization issue? I think we use `toProtobuf` rather than `toProtoBuf` in all models so far.

krantzinator:
This looks pretty slick 🎉 
I have a few thoughts:
* The term we use for the variable name/key should be consistent. ("Variable name" is used when adding a var value; "Key" is used in the list of var values) Preference is to use `Key` for both.
* Clicking on "Add Variable" and then having the box appear further up the page feels a bit disconnected. Could we move the "Add Variable" button to the top (does that make UI/UX sense, in terms of best practices? I don't know)? I actually don't want to copy the TFC method of hiding the list of variables and only displaying the new variable box, because as a user I've wished I could see my list of variables while also adding a new one. So I do like this presentation overall! Just the button position-to-resulting-action-location felt odd.
* I have a comment about the below. The wording is inaccurate. Variables can only be *defined* in the waypoint.hcl. **Values** can be set here. The reason for this is related to how/when we load config and can evaluate values.
  * This is, admittedly, a sticky part of this implementation. I know we discussed early on. We can talk it over more next week!

> Added a more explanatory empty state...
> 
> Safari 2021-06-21 at 14 28 16

ui/tests/integration/components/project-input-variables-list-test.ts
gregone: I'm personally on the fence on this one (I like the accuracy of using the API), but I'll go with your suggestion.

gregone:
@krantzinator Thanks for the feedback I'll rename the key/value labels for consistency. As for the help text in the empty state, could we come up with something clear so that there is no confusion for the end user? 

As for the button placement: it could be better to move it to the upper-right corner, as common for a lot of primary actions. @almonk would you be ok with that?

gregone:
@krantzinator updated copy and labels:

Screen Shot 2021-06-28 at 17 48 36


Screen Shot 2021-06-28 at 17 54 57

ui/mirage/services/project.ts
```diff
@@ -23,3 +23,21 @@ export function get(schema: any, { requestBody }: Request): Response {
...
+export function update(schema: any, { requestBody }: Request): Response {
+  let requestMsg = decode(UpsertProjectRequest, requestBody);
+  let name = requestMsg.getProject().getName();
+  let variablesList = requestMsg
+    .getProject()
+    .getVariablesList()
+    .map((v) => v.toObject());
+  let model = schema.projects.findBy({ name });
+
+  variablesList.forEach((v) => model.createVariable(v));
```
jgwhite, line 36: I realized I was mistaken in my previous suggestion here. I forgot this was an upsert operation — sorry about that!

Maybe we can try something like this:

```suggestion
  model.variables = variablesList.map((v) => model.newVariable(v));
  model.save();
```

ui/mirage/models/project.ts
```diff
@@ -1,9 +1,9 @@
 import { Model, hasMany } from 'ember-cli-mirage';
-import { Project, Ref } from 'waypoint-pb';
+import { Project, Ref, Variable } from 'waypoint-pb';
```
jgwhite, line 2: Can we remove this unused import?

```suggestion
import { Project, Ref } from 'waypoint-pb';
```

ui/mirage/models/variable.ts
```diff
@@ -0,0 +1,24 @@
+import { Model, belongsTo, hasMany } from 'ember-cli-mirage';
+import { Variable, Ref } from 'waypoint-pb';
```
jgwhite, line 2: Can we remove these unused imports?

```suggestion
import { Model, belongsTo } from 'ember-cli-mirage';
import { Variable } from 'waypoint-pb';
```

ui/app/components/project-input-variables/list.ts
```diff
@@ -0,0 +1,90 @@
...
+      return variable;
+    });
+    ref.setVariablesList(varProtosList);
+    // Build and trigger request
+    let req = new UpsertProjectRequest();
+    req.setProject(ref);
+    try {
+      let resp = await this.api.client.upsertProject(req, this.api.WithMeta());
+      let respProject = resp.toObject().project;
+      this.project = respProject;
```
jgwhite, line 81: It would be handy if we could somehow give this updated project data back to the route at this point. As things currently stand, the route’s copy of `project` is unaware of the changes so we get this effect:

https://user-images.githubusercontent.com/34030/123796379-5e87ca80-d8e5-11eb-8884-f98bf263d0c8.mp4

I guess the simpler option would be to get the route to trigger a reload.

ui/app/components/project-input-variables/list-item.hbs
```diff
@@ -0,0 +1,116 @@
...
+              {{t "form.project_variables_settings.variable_name"}}
+            </label>
+            <Pds::Input
+              @type="text"
+              data-test-input-variables-var-name
+              id="input-variable-name"
+              placeholder={{t "form.project_variables_settings.variable_name_placeholder"}}
+              value={{this.variable.name}}
+              {{on "input" (pick "target.value" (set this "variable.name"))}}
+            />
```
jgwhite, line 20: Echoing @almonk’s suggestion, it’d be cool if this field auto-focused when it appears.

jgwhite (changes requested):
Looking great. Left a few more thoughts, both user-facing and mirage-y.

ui/app/components/project-input-variables/list-item.hbs
gregone: so I have been pondering this one: 
- when creating, yes
- when editing, variable value changes would be more frequent than key changes, so maybe the second input should be focused? (but then we have inconsistency and that's weird)

jgwhite: Ah yeah, I think you’re right. Focusing the value is the more likely case. Maybe we focus the `key` field if it’s empty? Otherwise the value?

gregone: @almonk we should talk about this one. we can solve it in a separate PR.

jgwhite: 👍 to solving in a separate PR

ui/app/components/project-input-variables/list.ts
```diff
@@ -0,0 +1,93 @@
...
+      this.flashMessages.success('Settings saved');
+      this.activeVariable = null;
+      this.isCreating = false;
+      this.router.refresh('workspace.projects.project');
```
jgwhite, line 87: Oh neat! I had no idea this API existed.

jgwhite (approved):
Lovely stuff 👍

ui/app/components/project-input-variables/list.ts
gregone: I also did not, but it's currently polyfilled: https://github.com/Windvis/ember-router-service-refresh-polyfill

krantzinator:
Quality of life suggestions, as I play with this more:
* I was able to save an empty key/value set as a variable. We should show an error instead of either the key or the value is empty.
* I was able to click "+Add variable" multiple times at once and have multiple entry boxes open. We should limit this to one at a time I think? Or do we not care? It doesn't seem to cause collisions with saving multiple variables right in a row.
* Should we show a confirmation dialog when someone Deletes a variable? I could see some fuzzy-clicks from people meaning to click Edit.

Bug:
* The value for `Variable_Source` is not set on the message. Is this something I should infer on my side, or can you add it to the variable protobuf on your side?
 ```
message Variable {
  string name = 1;

  oneof value {
...
  }

  oneof source {
...
    google.protobuf.Empty server = 8;
}
```

**Comment by gregone** [comment]
@krantzinator I'll solve the `Server` issue and add the validation for empty variables.
For the two other items:
- We can add confirmation for deletion later
- I don't think the multiple edition is an issue, so we can leave it, but I'll indeed prevent creating multiple variables at once

**Review by krantzinator — APPROVED** [review]
Woohoo it's working! Getting the source `server` now for error messages:
```
! <nil>: Invalid value for variable; The value set for variable "port" from source
  "server" is of value type "string" and is not compatible with the variable's
  type constraint: a number is required.
```

🚀 🎉 🎊 💃🏻

PR #1663, opened 2021-06-16
Pinning alpine versions when docker building.

izaaklauer:
This is solving a current build crash when compiling shadow.


jgwhite (approved):
Nice catch!

PR #1664, opened 2021-06-16
Dockerfile: Remove SSH key injection from circle-ci and dockerfile build

briancain:
This commit removes the injected ssh keys from the build process as well
as the various circle-ci jobs. It was originally added due to Waypoint
being a private repo and needing access to push the built container to
the alpha releases on github. Waypoint is no longer private, and now circle-ci
should still be able to push packages to github using the docker login step
instead.

We should probably backport this, but I'd like to wait until after this has merged and gone through the CI process. Just in case we have to revert. Our backport system can be applied most-merge now so waiting on applying the label isn't a huge deal.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,8 +218,6 @@ jobs:
           name: "Docker: Build image"
           command: |
             docker build \
-              --ssh default \
-              --secret id=ssh.config,src="${HOME}/project/.circleci/ssh_config" \
               --progress=plain \
               -t docker.pkg.github.com/hashicorp/waypoint/alpha:latest \
               -t docker.pkg.github.com/hashicorp/waypoint/alpha:$GIT_COMMIT \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,9 +32,6 @@ jobs:
     environment:
       GOTAGS: ""
     steps:
-      - add_ssh_keys:
-          fingerprints:
-            - "76:94:15:c8:a0:41:f0:8b:9b:f1:f7:36:4c:d5:7f:57"
       - checkout
       - run:
           name: Install golangci-lint
@@ -60,9 +57,6 @@ jobs:
     environment:
       <<: *ENVIRONMENT
     steps:
-      - add_ssh_keys:
-          fingerprints:
-            - "76:94:15:c8:a0:41:f0:8b:9b:f1:f7:36:4c:d5:7f:57"
       - checkout
       - run:
           command: go mod tidy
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -235,12 +217,9 @@ jobs:
       - run:
           name: "Docker: Build image"
           command: |
-            # Note that the "id_rsa_FOO" key name is based on the fingerprint of the SSH key
-            # added in add_ssh_keys, but with colons removed.
             docker build \
               --ssh default \
               --secret id=ssh.config,src="${HOME}/project/.circleci/ssh_config" \
```
izaaklauer, line 222: I think you can pull lines 241-242 too

Dockerfile
```diff
@@ -17,9 +17,7 @@ WORKDIR /tmp/wp-prime
 RUN mkdir -p -m 0600 ~/.ssh \
```
izaaklauer, line 17: I bet lines 17-19 can go, and you can pull openssh out of the `apk add`.

izaaklauer (changes requested):

Dockerfile
```diff
@@ -6,20 +6,16 @@
...
 
 RUN mkdir -p /tmp/wp-prime
 COPY go.sum /tmp/wp-prime
 COPY go.mod /tmp/wp-prime
 
 WORKDIR /tmp/wp-prime
 
-RUN mkdir -p -m 0600 ~/.ssh \
-    && ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
 RUN git config --global url.ssh://git@github.com/.insteadOf https://github.com/
```
izaaklauer, line 17: I think this can go - we can clone over https now.

izaaklauer (changes requested):

Dockerfile
briancain: Oops, yeah missed that one! 👍🏻

izaaklauer (approved):

briancain:
Requesting review from @evanphx and @pearkes this week since they might have more historical context for these keys, just before I merge 🙏🏻

evanphx (approved):
Looks good! Ship it!

.circleci/config.yml
```diff
@@ -342,9 +316,6 @@ jobs:
       <<: *ENVIRONMENT
     steps:
       - checkout
-      - add_ssh_keys:
-          fingerprints:
-            - "76:94:15:c8:a0:41:f0:8b:9b:f1:f7:36:4c:d5:7f:57"
```
jgwhite, line 347: It looks like this is actually required as `publish-static-assets` wants to push a commit back into the repo: https://app.circleci.com/pipelines/github/hashicorp/waypoint/6439/workflows/a20dc52f-c62a-44aa-ab5d-5c47e4b20059/jobs/50414

PR #1671, opened 2021-06-17
Docs: Remove `buildargs` required parameter

briancain:
This commit removes a mistakenly added doc option `buildargs`. The actual option is `build_args`, and it's not required.

### DIFF
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -125,9 +125,12 @@ build {
 	)
 
 	doc.SetField(
-		"buildargs",
-		"An array of strings of build-time variables passed as build-arg to docker"+
-			"or img for the build step.",
+		"build_args",
+		"build args to pass to docker or img for the build step",
+		docs.Summary(
+			"An array of strings of build-time variables passed as build-arg to docker",
+			" or img for the build step.",
+		),
 	)
 
 	doc.SetField(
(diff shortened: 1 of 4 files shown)


briancain:
@evanphx I can't seem to get the docs to auto-gen properly. Even though I've fixed the option, it keeps wanting to put back this entry with the old string. Any idea what I'm missing here?

```diff
diff --git a/website/content/partials/components/builder-docker.mdx b/website/content/partials/components/builder-docker.mdx
index 4a0fb052..488fa98a 100644
--- a/website/content/partials/components/builder-docker.mdx
+++ b/website/content/partials/components/builder-docker.mdx
@@ -37,6 +37,14 @@ build {
 }

+### Required Parameters
+
+These parameters are used in the [`use` stanza](/docs/waypoint-hcl/use) for this plugin.
+
+#### buildargs
+
+An array of strings of build-time variables passed as build-arg to dockeror img for the build step.
+
 ```

evanphx:
Oh, it's the mdx lint thing. So when generating it, we need to escape an _. The linter is looking for _ as markdown italics.

briancain:
@evanphx Ahh, well that makes sense. Is there a supported way to escape `_` characters? My vim-go is upset because it doesn't recognize `\_` as a valid escape character 🤔 and it doesn't look like that fixed the auto-gen so I'm missing something.

evanphx:
@briancain ah, you need to do `"\\_"` basically escape the backslash in the replacement value.

evanphx (approved):

PR #1676, opened 2021-06-19
website: upgrade code-block

zchsh:
👀 [Preview](https://waypoint-git-fork-zchsh-zsupgrade-code-block-hashicorp.vercel.app/docs)

This PR upgrades all MDX code snippets to use [our latest `code-block` component](https://github.com/hashicorp/react-components/pull/205), which contains a number of new features detailed in [[MKTG-035] Improvements to our Code Blocks](https://docs.google.com/document/d/1b2ROeAoy708dSqpJZjZRvuhJm9FB86skKEeQC3ZkLMY/edit?ts=5fadaaee).

We've also converted our `code-block` component to use CSS modules, which removes the need for global code-block related style imports.

### DIFF
--- a/website/content/partials/components/platform-exec.mdx
+++ b/website/content/partials/components/platform-exec.mdx
@@ -29,7 +29,7 @@ template. The top of the documentation there has details on the format.
 
 The following template values are always available:
 
-- ".Env" (map<string>string) - These are environment variables that should
+- ".Env" (map<string\>string) - These are environment variables that should
   be set on the deployed workload. These enable the entrypoint to work so
   you should set these if able.
 
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -1282,222 +1282,193 @@
         "@babel/highlight": "^7.10.4"
       }
     },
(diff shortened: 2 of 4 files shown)


website/content/partials/components/platform-exec.mdx
```diff
@@ -29,7 +29,7 @@ template. The top of the documentation there has details on the format.
 
 The following template values are always available:
 
-- ".Env" (map<string>string) - These are environment variables that should
+- ".Env" (map<string\>string) - These are environment variables that should
```
zchsh, line 32: This already needed to be escaped, but was silently failing (see [live](https://www.waypointproject.io/plugins/exec#common-values), the text `<string>` does not appear because it is interpreted as an opening HTML tag).

Now that we're processing partials with MDX, this issue no longer fails silently. Note that all other uses of similar tag-like text I could find (eg `<string>`) are rendered as inline code, or in code blocks, which is why this is typically not an issue.

jescalan (approved):

jescalan (approved):

PR #1683, opened 2021-06-22
ui: upgrade to ember-source@3.24.4 (latest LTS)

jgwhite:
This includes a load of transitive dependency noise (because of Yarn’s upgrade algorithm I guess?) which can be safely ignored. The operative change is https://github.com/hashicorp/waypoint/pull/1683/files#diff-3a968206d6de2fecfc5dacd7d94bab7744c9f5d5c999a816164d95cbc135c316R8840

### DIFF
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -2,12 +2,19 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4":
+"@babel/code-frame@^7.0.0":
   version "7.10.4"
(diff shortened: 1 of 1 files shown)


gregone (approved):
LGTM. Would it make sense to use `~3.24.4` in `package.json` to explicitly reflect the change?

jgwhite:
> Would it make sense to use ~3.24.4 in package.json to explicitly reflect the change?

That’s an interesting question. My two cents: In this case, I think the thing we want to communicate is “we want the latest in the 3.24.x series”. Whereas “~3.24.4” to me implies “there’s some specific bugfix in 3.24.4 that we depend upon”. What do you think?

gregone:
@jgwhite I'm fine with both. I usually look for upgrade history in `package.json`, hence my comment. But I agree in this case it's just a minor upgrade and probably not worth it.

PR #1690, opened 2021-06-22
cli: set default to 30s for runner polling

krantzinator:
The poll interval is not required and should default to 30s if not specified. This is handled on the `project apply` side by setting a default for the flag value. But if a user sets their runner config directly in the waypoint.hcl, then currently no default is set. This results in an error from `time`: `! time: invalid duration ""`.
This fix adds the default time duration in lieu of an empty string if the interval is not specified.

### DIFF
--- a/.changelog/1690.txt
+++ b/.changelog/1690.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: set runner poll interval default for runner defined in waypoint.hcl
+```
--- a/internal/cli/init.go
+++ b/internal/cli/init.go
@@ -309,9 +309,6 @@ func (c *InitCommand) validateProject() bool {
 				Enabled:  v.Enabled,
 				Interval: v.Interval,
 			}
-			if poll.Interval == "" {
-				poll.Interval = "30s"
-			}
 		}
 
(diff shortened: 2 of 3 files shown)


briancain (approved):
This makes sense!

If it's possible, it would be good to also set a default on the server too when a new poll handler is created with an empty interval. In the case where there are more CLI commands (or anything else) that initiate a project poll, our default interval is set in a single place for project polling. Otherwise they will end up with the same issue.

mitchellh:
Concur with @briancain: should we just do this on the server side instead? That way it'll work everywhere. I don't have the full context though!

briancain:
Assuming `state/project` is the right place for defining the default, I'd probably place the default here:

https://github.com/hashicorp/waypoint/blob/2900d227b7ef38f7c8233ee325f1b25f489c0f80/internal/server/singleprocess/state/project.go#L361-L363

I'm guessing this is where the original error came from when it tried to parse the interval duration as `""`. Then maybe we have a var block in the project state outside of this project record set that defines the default poll interval that gets inserted for each new project poll.

briancain (approved):
Thanks! Looks great 😄

PR #1699, opened 2021-06-23
Inspecting boolean env var values.

izaaklauer:
Previously, we were taking any value for boolean env vars to mean "true". If someone specified `WAYPOINT_SERVER_TLS=false`, we would take that to mean true. This fixes that.

Note on the implementation - I hate to be the person to introduce a generic "util" package. If anyone has any better ideas for where to put this little function, or feels like we should just strconv in-line every time, I am open to those ideas.

### DIFF
--- a/.changelog/1699.txt
+++ b/.changelog/1699.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+core: Inspect values of enviromment variables meant to be booleans
+```
+
--- a/.changelog/1699.txt
+++ b/.changelog/1699.txt
@@ -1,4 +1,4 @@
 ```release-note:improvement
-core: Inspect values of enviromment variables meant to be booleans
+core: Correct parsing of boolean environment variables
 ```
 
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -16,7 +16,7 @@ func GetEnvBool(key string, defaultValue bool) (bool, error) {
 	}
 	value, err := strconv.ParseBool(strings.ToLower(envVal))
 	if err != nil {
-		return defaultValue, fmt.Errorf("failed to parse a boolean from environment variable %s=%s",key,envVal)
+		return defaultValue, fmt.Errorf("failed to parse a boolean from environment variable %s=%s", key, envVal)
 	}
 	return value, nil
 }
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -1,20 +1,22 @@
 package env
 
 import (
+	"fmt"
 	"os"
 	"strconv"
+	"strings"
 )
 
 // GetEnvBool Extracts a boolean from an env var. Falls back to the default
 // if the key is unset or not a valid boolean.
-func GetEnvBool(key string, defaultValue bool) bool {
+func GetEnvBool(key string, defaultValue bool) (bool, error) {
 	envVal := os.Getenv(key)
 	if envVal == "" {
-		return defaultValue
+		return defaultValue, nil
 	}
-	value, err := strconv.ParseBool(envVal)
+	value, err := strconv.ParseBool(strings.ToLower(envVal)))
 	if err != nil {
-		return defaultValue
+		return defaultValue, fmt.Errorf("failed to parse a boolean from environment variable %s=%s",key,envVal)
 	}
-	return value
+	return value, nil
 }
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -1,13 +1,13 @@
-package util
+package env
 
 import (
 	"os"
 	"strconv"
 )
 
-// Extracts a boolean from an env var. Falls back to the default if
-// The key is unset or not a valid boolean.
-func GetEnvBool(key string, defaultValue bool) (value bool) {
+// GetEnvBool Extracts a boolean from an env var. Falls back to the default
+// if the key is unset or not a valid boolean.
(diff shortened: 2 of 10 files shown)


catsby (changes requested):
There are some other examples in the documentation where `WAYPOINT_SERVER_TLS_SKIP_VERIFY` is used and needs to be updated I think, should we encourage users to use `true` now instead of `1`?

.changelog/1699.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
+core: Inspect values of enviromment variables meant to be booleans
```
catsby, line 2:
```suggestion
core: Correct parsing of boolean environment variables
```

krantzinator (approved):
I'm good with the util package!
I'd update the docs (see comments) to only be as specific as our code actually requires.

internal/util/env.go
```diff
@@ -0,0 +1,20 @@
...
+func GetEnvBool(key string, defaultValue bool) (value bool) {
+	envVal := os.Getenv(key)
+	if envVal == "" {
+		return defaultValue
+	}
+	value, err := strconv.ParseBool(envVal)
+	if err != nil {
+		return defaultValue
+	}
+	return value
```
briancain, line 19: This func could use tests! Probably the three return cases covered here 👍🏻

internal/util/env.go
```diff
@@ -0,0 +1,20 @@
+package util
```
briancain, line 1: What if instead of `util` we make this `env`? Then in the future if we ever need to do more env manipulation we have a package already defined and I think it will end up reading a bit nicer when the function is invoked like `env.GetEnvBool(...)`.

briancain (changes requested):
Got a request for tests 😄 And a suggestion for not introducing util if we don't have to

website/content/docs/automating-execution/index.mdx
```diff
@@ -43,9 +43,9 @@ configure the Waypoint CLI for communication with the Waypoint server.
 - `WAYPOINT_SERVER_TOKEN`. Must be set to a Waypoint token, created with [`waypoint token new`](/commands/token-new)
 - `WAYPOINT_SERVER_ADDR`. The address to the Waypoint server gRPC address. This must
   be accessible from the network that the client is running on.
-- `WAYPOINT_SERVER_TLS`. Should be set to `1` to configure the client to use TLS when
+- `WAYPOINT_SERVER_TLS`. Should be set to a truthy value (e.g. "1") to configure the client to use TLS when
   connecting to the server.
-- `WAYPOINT_SERVER_TLS_SKIP_VERIFY`. Current must be set to `1` to disable TLS verification
+- `WAYPOINT_SERVER_TLS_SKIP_VERIFY`. Current must be set to truthy value (e.g. "1") to disable TLS verification
```
krantzinator, line 48:
```suggestion
- `WAYPOINT_SERVER_TLS_SKIP_VERIFY`. Current must be set to a truthy value (e.g. "1") to disable TLS verification
```

krantzinator (approved):

internal/env/env.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+// GetEnvBool Extracts a boolean from an env var. Falls back to the default
+// if the key is unset or not a valid boolean.
+func GetEnvBool(key string, defaultValue bool) bool {
+	envVal := os.Getenv(key)
+	if envVal == "" {
+		return defaultValue
+	}
+	value, err := strconv.ParseBool(envVal)
```
evanphx, line 15: Let's convert envVal to lower before passing it in just to support 🧽 👖 case. Dealers choice if we also support `yes`  / `no`.

evanphx (changes requested):

internal/env/env.go
```diff
@@ -0,0 +1,20 @@
+package env
+
+import (
+	"os"
+	"strconv"
+)
+
+// GetEnvBool Extracts a boolean from an env var. Falls back to the default
+// if the key is unset or not a valid boolean.
+func GetEnvBool(key string, defaultValue bool) bool {
```
mitchellh, line 10: I’d name this `GetBool` since its already in the `env` package. This is covered in the “zen of go” or w/e that Go document is (on golang.org). Try not to repeat the package name in the function names.

internal/env/env_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/stretchr/testify/require"
+	"os"
+	"testing"
+)
+
+func TestGetEnvBool(t *testing.T) {
+	envVarTestKey := "WAYPOINT_GET_ENV_BOOL_TEST"
+	require := require.New(t)
+
+	t.Run("Unset env var returns default", func(t *testing.T) {
```
mitchellh, line 13: This is all fine but a table test would probably be more appropriate here.

mitchellh (changes requested):

internal/env/env_test.go
izaaklauer: I started with that, but stopped when I realized I needed to set an env var for each table entry. That didn't need to stop me though.

mitchellh (approved):

mitchellh:
Btw, this can probably be backported if you want.

briancain (approved):

evanphx (approved):

catsby (approved):

PR #1709, opened 2021-06-24
add app name variable

psihachina:
This commit adds the ${app.name} variable, so that can have
use for generation strings for anything, for example

```
project = "nutcorp"

app "hamster" {
  config {
    env = {
      APP_NAME = "${app.name}"
      ENV = "dev"
    }
  }

  build {
    use "docker" {
      dockerfile = "../../../projects/${app.name}/Dockerfile"
      build_args = {
        PROJECT_PATH="./projects/${app.name}"
      }
      context    = "../../../"
    }

    registry {
      use "aws-ecr" {
        region     = "us-east-2"
        repository = "nutcorp-${app.name}"
        tag        = "latest"
      }
    }
  }
}

app "vole" {
  config {
    env = {
      APP_NAME = "${app.name}"
      ENV = "dev"
    }
  }

  build {
    use "docker" {
      dockerfile = "../../../projects/${app.name}/Dockerfile"
      build_args = {
        PROJECT_PATH="./projects/${app.name}"
      }
      context    = "../../../"
    }

    registry {
      use "aws-ecr" {
        region     = "us-east-2"
        repository = "nutcorp-${app.name}"
        tag        = "latest"
      }
    }
  }
}

```

### Test
[asciicast](https://asciinema.org/a/h5yIdbKEfjSafpZs1RwRJADIh)

### DIFF
--- a/.changelog/1709.txt
+++ b/.changelog/1709.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+internal/config: add `${app.name}` variable
+```
--- a/.changelog/1709.txt
+++ b/.changelog/1709.txt
@@ -1,3 +1,3 @@
 ```release-note:feature
-internal/config: add `${app.name}` variable
+config: add `${app.name}` variable
 ```
--- a/internal/config/app.go
+++ b/internal/config/app.go
@@ -86,7 +86,9 @@ func (c *Config) App(n string, ctx *hcl.EvalContext) (*App, error) {
 	// Build a new context with our app-scoped values
 	ctx = ctx.NewChild()
 	addPathValue(ctx, pathData)
-	addAppValue(ctx, rawApp.Name)
+	addMapVariable(ctx, "app", map[string]string{
+		"name": rawApp.Name,
+	})
 
 	// Full decode
 	var app App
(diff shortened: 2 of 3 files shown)


.changelog/1709.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+internal/config: add `${app.name}` variable
```
mitchellh, line 2: This should just be `config:` 

We're working on documenting all of our preferences here. :)

internal/config/app.go
```diff
@@ -86,6 +86,7 @@ func (c *Config) App(n string, ctx *hcl.EvalContext) (*App, error) {
 	// Build a new context with our app-scoped values
 	ctx = ctx.NewChild()
 	addPathValue(ctx, pathData)
+	addAppValue(ctx, rawApp.Name)
```
mitchellh, line 89: I'd just call `addMapVariable` directly since `addAppValue` isn't reused (like path value is).

mitchellh (commented):
This looks great and makes sense to me. I figured something like this would be added sooner or later. 😄

mitchellh:
This looks good! Backporting it as well.

Issue #1704, opened 2021-06-23
Bug: `waypoint runner agent -remote` panics


izaaklauer:
**Describe the bug**
Running `waypoint runner agent` with the `-remote` flag causes a panic for me under all circumstances.

Trace:
```
$ waypoint runner agent -remote

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x319bb61]

goroutine 1 [running]:
github.com/hashicorp/waypoint/internal/cli.(*baseCommand).Init(0xc000c00000, 0xc000d3fcb0, 0x4, 0x4, 0xc000caa3f0, 0xc000d3fb28)
	/home/circleci/project/waypoint/internal/cli/base.go:253 +0x821
github.com/hashicorp/waypoint/internal/cli.(*RunnerAgentCommand).Run(0xc000792c40, 0xc00040dab0, 0x1, 0x1, 0x0)
	/home/circleci/project/waypoint/internal/cli/runner_agent.go:40 +0x218
github.com/mitchellh/cli.(*CLI).Run(0xc000ca4000, 0xc000d3fee8, 0x3e2bf18, 0xc0003e84d0)
	/home/circleci/go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x41a
github.com/hashicorp/waypoint/internal/cli.Main(0xc00040da80, 0x4, 0x4, 0x0)
	/home/circleci/project/waypoint/internal/cli/main.go:119 +0x57a
main.main()
	/home/circleci/project/waypoint/cmd/waypoint/main.go:14 +0xa5
```

**Expected behavior**

I am honestly not sure what this flag is supposed to be doing. I ran across it accidentally while trying to set up a standalone runner.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: 0.4.0
* Waypoint Server Platform and Version: 0.4.0
* Waypoint Plugin: any

**Additional context**

It's crashing [here](https://github.com/hashicorp/waypoint/blob/321eb68d56e918968b0145392158826a9b77e5e8/internal/cli/base.go#L253), specifically when inspecting `c.cfg.Runner`.  The crazy thing is that c.cfg is, in fact, not nil _until_ you ask for `c.cfg.Runner`

[...]

PR #1710, opened 2021-06-24
Not embedding hclConfig as a pointer.

izaaklauer:
## What problem is this solving?

This addresses the panic within https://github.com/hashicorp/waypoint/issues/1704.

The problem here is that, if we embed a pointer to a private struct, and the private struct is nil, the caller has no way to determine if the embedded struct's values are safe to access.

This manifests here: https://github.com/hashicorp/waypoint/blob/124f4555ae306fb528adfaa65d39dfc72d41c075/internal/cli/base.go#L253

There is no way to determine of `c.cfg.Runner` is safe to access. `c.cfg` can be non-nil on its own, but it will _become_ nil when inspecting `c.cfg.Runner` if the embedded `*hclConfig` is nil.

A simplified example of the problem:

https://play.golang.org/p/sbAPT2HGJqW

If hclConfig is not a pointer, we won't get a nil pointer deference when looking at its fields.

## Is this change safe?

I believe so - we don't have any methods on hclConfig that require it to be a pointer. We also never statefully maniuplate it outside of the context of `cfg`, as far as I can see. I've tested this locally and it doesn't seem to break anything.

### DIFF
--- a/.changelog/1710.txt
+++ b/.changelog/1710.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: fix crash that could occur when running commands outside the context of a project with an hcl config file.
+```
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -13,7 +13,7 @@ import (
 )
 
 type Config struct {
-	*hclConfig
+	hclConfig
 
 	ctx      *hcl.EvalContext
 	path     string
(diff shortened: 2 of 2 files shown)


evanphx (approved):
Because the expectation is that users outside this package are going to be accessing things like Runner on Config, but via hclConfig, it makes sense to have hclConfig be a value and not a pointer because those outside users can't even nil check hclConfig. Ergo it makes sense to just have hclConfig be a value.

Issue #1513, opened 2021-05-19
UI and CLI should agree on which representation of ID or Seq Number is displayed


evanphx:
Presently, the CLI shows mostly sequence numbers by default but the UI shows ID. They should agree on what they show by default.

mitchellh:
My opinion is we should standardize on the seq number everywhere. The downside of the seq number is it requires the `(project, app)` to have meaning. The ID is globally unique, the seq number is only unique within the full tuple of `(project, app, seq)`. However, I think from a UX perspective, you're using operating in that context anyways, so it makes sense.

Internally, our APIs should probably continue accepting both cause its easy enough.

evanphx:
Could/should we establish a format to capture that tuple in the UI? For instance a common format of `#{app_name}:#{seq_no}` or something?

mitchellh:
> Could/should we establish a format to capture that tuple in the UI? For instance a common format of `#{app_name}:#{seq_no}` or something?

Yeah, the CLI currently accepts the `project/app` format for THAT, we could do like `project/app#seq` or something similar to GitHub's reference style...

evanphx:
👍🏻

izaaklauer:
We also use ULIDs as names for aws resources we create (i.e. my target group is called `example-nginx-4C4THAKBKPN61X25HA`). Sequence number would be much friendlier there anyway, and we should remember to change that too so platform owners can still tie resources back to deployments.

jgwhite:
Just to ensure we’re all on the same page about the scope of these changes:

![image](https://user-images.githubusercontent.com/34030/122240719-474ae500-cec2-11eb-9630-630eb122a110.png)

izaaklauer:
I like the idea of having some advanced way to get the internal ID out of the UI, both for us and for advanced users trying to reference debug logs.

Other than that, I agree with the scope of the UI changes.

evanphx:
Yes, you're correct @jgwhite. We have the sequence number there, it's the #4 or v4 in other views. View wise, we can just not show the ulid if the sequence number is right there too, giving more visual area for other things.

jgwhite:
@evanphx perfect. That fits in well with @almonk’s latest designs.

PR #1668, opened 2021-06-17
ui: Refine status indicators

jgwhite:
## Why the change?

0.4.0 introduced “status reports” (AKA “health checks”) to Waypoint. Status reports join the existing concepts of “state” and “status” on Waypoint entities. With the arrival of status reports, entities such as deployments have three complementary notions of “status”:

1. **Physical Status** status of underlying resources
2. **Operation Status** outcome of a waypoint operation (build, deploy, release)
3. **Health Status** outcome of most recent status report (AKA health check)

This PR aims to present all these statuses in an unambiguous manner, based on this proposed design:

image

Addresses #1581 and other issues with status indicators.

## What’s the plan?

- [x] Create `<StatusReportIndicator>` (to take over from `<StatusBadge>`)
- [x] Create `<OperationStatusIndicator>` (or similar)
- [x] Only show one “latest” card in each column
- [x] Update `<AppCard::Build>`
- [x] Update `<AppCard::Deployment>`
- [x] Update `<AppCard::Release>`
- [x] Update `<AppItem::Build>`
- [x] Update `<AppItem::Deployment>`
- [x] Update `<AppItem::Release>`
- [x] Update deployment detail page
- [x] Update release detail page
- [x] Retire `<StatusBadge>`
- [x] Await #1713 

## What’s out of scope?

- Artifact info

### Before

https://user-images.githubusercontent.com/34030/124609083-98217e00-de6f-11eb-8be2-cd6a4d77f264.mp4

### After

https://user-images.githubusercontent.com/34030/124751917-771c6400-df27-11eb-94b9-333cae1f13fa.mp4

https://user-images.githubusercontent.com/34030/124751922-77b4fa80-df27-11eb-9db6-276b710fa9fe.mp4

### Using Mirage

1. Check out the branch:
   ```sh
   git checkout ui/refine-status-indicators
   ```
2. Boot the dev server
   ```sh
   cd ui && ember serve
   ```
3. [Visit the app](http://localhost:4200/default/marketing-public/app/wp-matrix)
4. Verify you see the new status indicators
5. Verify the tooltips appear correctly
6. Navigate to builds, deployments, releases
7. Verify the new indicators look good wherever you see them

### For realsies

1. Check out the branch:
   ```sh
   git checkout ui/refine-status-indicators
   ```
2. Build the ember app
   ```sh
   (cd ui && make)
   ```
3. Bundle the static assets
   ```sh
   make static-assets
   ```
4. Build the dev server
   ```sh
   make docker/server
   ```
[...]

PR #1713, opened 2021-06-24
Removed references to id in favor of seq #

sabrinako:
Addresses #1513 
## What's Included
- Removed the instances of the build/deployment/release ID hash in the UI
- Updated the CLI hints to utilize sequence numbers instead of ID (see last screenshot)

## What's Not Included
- Updating the URL of individual build/deployment/release to use sequence number
- Creating a tooltip/UI for displaying the ID and allowing users to copy to clipboard

### **_List of builds without ID_**
builds list without id
### **_List of deployments without ID_**
deployments list without id
### **_List of releases without ID_**
releases list without id
### **_Individual build's page, note the cli hint `waypoint deploy 4` using the sequence # not ID_**
localhost_4200_default_marketing-public_app_wp-matrix_builds

### DIFF
--- a/.changelog/1713.txt
+++ b/.changelog/1713.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Removed references to IDs in favor of sequence # to align with CLI
+```
--- a/ui/app/components/actions/deploy.hbs
+++ b/ui/app/components/actions/deploy.hbs
@@ -7,7 +7,7 @@
   <div class="cli-hint">
     <p>To deploy this build you can run the following command in the CLI</p>
     <CopyableCode @ref="deploy-cli-hint">
-      <pre><code id="deploy-cli-hint">waypoint deploy {{@id}}</code></pre>
+      <pre><code id="deploy-cli-hint">waypoint deploy {{@sequence}}</code></pre>
     </CopyableCode>
     <small><ExternalLink href="https://waypointproject.io/commands/deploy">Read more in our documentation</ExternalLink></small>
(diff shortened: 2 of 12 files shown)


jgwhite (approved):
The code looks perfect to me. I’m going to rebase my stuff in #1668 on top of this right away.

I think it’d be worth getting @almonk’s review, and possibly @evanphx and/or @mitchellh’s too.

Reminder: this is a step towards the designs presented in [WP-071](https://go.hashi.co/wp-071).

gregone (approved):
Looking good! 

Related: Should we update the docs to reflect the use of `id` as the primary reference to artifacts?

jgwhite:
> Is this a state that will be user-facing or a non-user-facing step toward the WP-071 design?
> 
> If the former, I think we'll need to play with the layout to accommodate the reduction in detail. If the latter then this seems fine to me.

That’s somewhat dependent on how quickly I can finish #1668 (which is now based on this work).

We’re planning to cut a minor release (0.4.1) this week. So perhaps we could wait for 0.4.1 to happen, then merge this change as-is, then proceed to land the #1668 (WP-071) changes on top.

A reminder of what we’re eventually aiming for:

122463146-87d75b00-cfb5-11eb-9643-017a1294e685

@sabrinako does that fit with how you were thinking about shipping this change?

sabrinako:
Okay since the minor release is happening soon, I think we can wait a little to merge this in until that occurs. I'll ping you @jgwhite when this is eventually merged in

PR #1716, opened 2021-06-25
Feature: plugin debugging

izaaklauer:
It is currently difficult to debug the builtin plugins, as waypoint itself launches plugin processes as it requires them. Debugging is easiest when you can use a debugger to launch the process of interest. 

This experimental change allows Waypoint to launch builtin plugins in debug mode, and connect to and use an existing plugin processes. It is dependent on these waypoint-plugin-sdk changes: https://github.com/hashicorp/waypoint-plugin-sdk/pull/39

These changes are inspired by (and in many cases copied from) [Terraform's plugin debug mode](https://www.terraform.io/docs/extend/debugging.html#debugger-based-debugging). I've tried to keep the user facing side (flag/env var names, workflow, etc) as similar to Terraform as possible for consistency.

### Debugging workflow

Step 1: Launch a plugin process of interest (via delve or goland debug) with the new debug flag (i.e. `waypoint plugin -debug pack`). The plugin will launch, print reconnection info to the console (example below), then wait for core to attach.

```
Plugin started, to attach Waypoint set the WP_REATTACH_PLUGINS env var:

        WP_REATTACH_PLUGINS='{"pack":{"Protocol":"grpc","ProtocolVersion":1,"Pid":38972,"Test":true,"Addr":{"Network":"unix","String":"/var/folders/ns/grk8kk196_106v37w9hk8hxm0000gq/T/plugin864517971"}}}'

```

Step 2: Export that environment variable, then run a core command that will require that plugin (i.e. `waypoint build`). When initializing plugins, core will unmarshall and inspect the contents of the `WP_REATTACH_PLUGINS` env var. If it finds that a requested plugin is in that map, it will attempt to reconnect to it, rather than launching it itself. 

The flow is a little clunky for me - if you're iterating fast and relaunching the plugin process frequently, you have to do a lot of env var copying and pasting. It would be nice if the plugin could use a consistent socket path on every launch. I think this is good enough for now though, and consistency with Terraform is nice.

### Documentation

My intent is to use this feature myself for this cycle, and document it next cycle once it's been hardened.


go.mod
```diff
@@ -49,7 +49,7 @@ require (
 	github.com/hashicorp/go-hclog v0.14.1
 	github.com/hashicorp/go-memdb v1.3.2
 	github.com/hashicorp/go-multierror v1.1.0
-	github.com/hashicorp/go-plugin v1.3.0
+	github.com/hashicorp/go-plugin v1.4.2
```
izaaklauer, line 52: Upgrading to get the ProtocolVersion field on ReattachConfig, which we need to know which VersionedPlugins to use.

internal/plugin/factory.go
```diff
@@ -95,6 +96,65 @@ func BuiltinFactory(name string, typ component.Type) interface{} {
 	return Factory(cmd, typ)
 }
 
+// UnmanagedPluginFactory produces a provider factory that uses the passed
+// reattach information to connect to go-plugin processes that are already
+// running, and implements Instance against it.
+func UnmanagedPluginFactory(reattach *plugin.ReattachConfig, typ component.Type) interface{} {
```
mitchellh, line 102: `unmanaged` is a bit confusing since go-plugin actually does support a concept of managed vs. unmanaged plugins (defined as whether it auto cleans them up).

My recommendation would be changing this to `ReattachPluginFactory` cause its reattaching.

internal/runner/operation.go
```diff
@@ -267,6 +294,46 @@ func (r *Runner) pluginFactories(
...
+			unmanangedPlugins[p] = &goplugin.ReattachConfig{
+				Protocol:        goplugin.Protocol(c.Protocol),
+				ProtocolVersion: c.ProtocolVersion,
+				Pid:             c.Pid,
+				Test:            c.Test,
+				Addr:            addr,
+			}
+		}
+	}
+	return unmanangedPlugins, nil
```
mitchellh, line 334: Same here, you actually call this `parseReattachPlugins` but then use `unmanagedPlugins` as the result name, which is a smell to me that the whole thign should be one or the other (and I prefer `reattach`).

internal/client/server.go
```diff
@@ -93,7 +94,7 @@ func (c *Project) initLocalServer(ctx context.Context) (*grpc.ClientConn, error)
 		Timeout: 1 * time.Second,
 	})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed opening boltdb path %s. Is another server already running against this db?: %w", path, err)
```
izaaklauer, line 97: This is mostly unrelated, but I spent a good long while trying to get to the bottom of why this change was breaking waypoint's local server mode. It turns out I wasn't initializing the plugin command with the `WithNoAutoServer()` option, so the plugin was trying to spool up another server, and crashing when it couldn't open the bolt's `data.db` file. The only thing I had to go on from the logs was this:

```
! Unrecognized remote plugin message: ! failed to create client: timeout
```

So i'm enhancing that log message here.

mitchellh (approved):

PR #1723, opened 2021-06-28
builtin/nomad: fix dropped error

alrs:
This fixes a dropped `err` variable in `builtin/nomad`.

### DIFF
--- a/.changelog/1723.txt
+++ b/.changelog/1723.txt
@@ -1,0 +1,3 @@
+se-note:bug
+plugin/nomad: Fix case where nomad error would be ignored during a status check
+```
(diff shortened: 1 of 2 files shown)


izaaklauer:
Thanks for the bug fix! Could you add a [changelog entry](https://github.com/hashicorp/waypoint/blob/main/.github/CHANGELOG_GUIDE.md#changelog-entry-examples) for it? Creating a file at `.changelog/1723.txt` with contents something like this would do it:
~~~
```release-note:bug
plugin/nomad: Fix case where nomad error would be ignored during a status check
```
~~~

izaaklauer (approved):

PR #1724, opened 2021-06-28
ui: add `lib` to `.eslintignore`

jgwhite:
## Why the change?

The contents of `ui/lib` are either generated or vendored code that we oughtn’t change, therefore it’s safe for the listing tooling to ignore this directory.


gregone (approved):
Good idea!

PR #1715, opened 2021-06-25
ui: Fix template lint violations

jgwhite:
## Why the change?

These changes all fix violations ember-template-lint rules.

## What does it look like?

Hopefully no user-facing differences.

## How do I test it?

1. `git checkout ui/fix-template-lint-violations`
2. `yarn lint:hbs`
3. Verify you see no errors

To really fine-tooth-comb this change, you may want to test the various pages where changes occur. I’ve been fairly diligent, but you never know.

## Any follow-up work?

After this is merged we can make `yarn lint:hbs` a part of our CI workflow.

PR #1729, opened 2021-06-28
add default auth to mirage scenario for mocks mode

Dhaulagiri:
When trying to run the web ui with [mocks as outlined in the readme](https://github.com/hashicorp/waypoint/blob/fdff0ef772cd40293dd8f3ae440f89c7d493ff58/ui/README.md#running-with-mocks) I found I could not make it past the authentication screen. To help make this easier I copied over the fake authentication we're doing in the tests to run by default when mirage boots up.

Certainly possible I'm missing something obvious here though so open to feedback or a 👎

### DIFF
--- a/ui/mirage/helpers/login.ts
+++ b/ui/mirage/helpers/login.ts
@@ -1,0 +1,3 @@
+export default function login(token?: string) {
+  window.localStorage.waypointAuthToken = token || 'default-test-token-value';
+}
--- a/ui/mirage/scenarios/default.ts
+++ b/ui/mirage/scenarios/default.ts
@@ -1,6 +1,8 @@
 import { Server } from 'ember-cli-mirage';
+import login from '../helpers/login';
 
 export default function (server: Server): void {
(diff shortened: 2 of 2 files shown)


jgwhite (approved):
I don’t see any obvious downside to this.

The missing line from the README was probably “enter any non-empty token value on the login screen” but this helps skip right past that confusion.

Dhaulagiri:
> The missing line from the README was probably “enter any non-empty token value on the login screen” but this helps skip right past that confusion.

FWIW i did try this and when you attempt to submit the form nothing happens

jgwhite:
@Dhaulagiri ah! This is a bug introduced by yours truly in #1715. Thanks for pointing it out!

Dhaulagiri:
Ahh, that makes more sense. If that gets resolved then this PR is of perhaps less value, but it doesn't save a couple steps folks are unlikely to care much about if they are solely trying to look at fake data.

jgwhite:
@Dhaulagiri I agree 👍

PR #1737, opened 2021-06-30
Declared resources wiring for deployments + server resource awareness protos

izaaklauer:
Includes both the new protobuf changes for server resource awareness, and the changes that were required for me to get the DeclaredResourcesResp outparameter working.

The idea here is that a plugin can optionally accept a *DeclaredResourcesResp arg in its deploy function, populate it with their resources, and then those resources make it back to the server an go on the final Deployment protobuf.

The wiring feels pretty clunky to me in a number of places - see notes in comments.

Corresponding plugin sdk changes: https://github.com/hashicorp/waypoint-plugin-sdk/pull/40


builtin/docker/platform.go
```diff
@@ -511,6 +512,15 @@ func (p *Platform) Deploy(
 	s.Done()
 
 	result.Container = crState.Id
+
+	//Add our experimental declared resource
+	declaredResource := sdk.DeclaredResource{
```
izaaklauer, line 517: Just filling in some dummy data here.

internal/server/proto/server.proto
```diff
@@ -2096,6 +2136,16 @@ message Deployment {
   }
 }
 
+// A deployment with additional related messages pre-fetched.
+message DeploymentExpanded {
```
briancain, line 2140: Should we include artifact too like how pre-load does?

internal/core/app_deploy.go
```diff
@@ -313,14 +315,48 @@ func (op *deployOperation) Do(ctx context.Context, log hclog.Logger, app *App, m
...
 		op.component,
 		op.component.Value.(component.Platform).DeployFunc(),
-		op.args()...,
+		args...,
 	)
 
+	// NOTE(izaak): Unfortunately, the returned type here is an waypoint-plugin-sdk/internal/plugioncomponent/Deployment,
+	// and because it's internal we can't cast to it. This is similar to how we're getting TemplateData
+	// (https://github.com/hashicorp/waypoint-plugin-sdk/blob/eda7ae600c2d69ba042ecac4d6d9333d37f6fd9d/docs/func.go#L106),
+	// but it still doesn't feel great.
```
briancain, line 338: Hrm, we can reference release, so we should be able to reference deploy:

https://github.com/hashicorp/waypoint/blob/6306d37c290f71dc0f5013845732b7afe534fd9c/internal/core/app_release.go#L217-L228

I haven't messed with this myself so there might be something I'm missing.

briancain (changes requested):

internal/server/proto/server.proto
briancain: @jgwhite Anything else we should include here for the UI?

internal/core/app_deploy.go
```diff
@@ -313,14 +315,48 @@ func (op *deployOperation) Do(ctx context.Context, log hclog.Logger, app *App, m
...
+	var pluginDeclaredResources []*sdkproto.DeclaredResource
+	pluginDeclaredResourcesField := reflect.ValueOf(val).Elem().FieldByName("DeclaredResources")
```
evanphx, line 340: You mention doing this like TemplateData, we should probably mask the digging around in a reflect value the same as we do with TemplateData.

evanphx (changes requested):
Coming along fine, with using reflect being the only place that gives me the willies (but that's how we've done it before so 🤷🏻 ).

internal/core/app_deploy.go
izaaklauer: I think the difference is that for release, we explicitly cast to the super broad component.Release on the plugin side:

https://github.com/hashicorp/waypoint-plugin-sdk/blob/main/internal/plugin/releaser.go#L288

We may be able to do something similar here though to avoid the reflection - i'll give it a shot.

internal/core/app_deploy.go
```diff
@@ -313,14 +315,48 @@ func (op *deployOperation) Do(ctx context.Context, log hclog.Logger, app *App, m
...
+	args = append(args, argmapper.Typed(declaredResourcesResp))
+	args = append(args, op.args()...)
+
 	val, err := app.callDynamicFunc(ctx,
 		log,
 		(*component.Deployment)(nil),
 		op.component,
 		op.component.Value.(component.Platform).DeployFunc(),
-		op.args()...,
+		args...,
```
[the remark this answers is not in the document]
izaaklauer: 👍 good call. I had this up here initially because I was hoping I could get that declaredResourcesResp actually populated after calling the plugin func (like a real outparameter), but I had no such luck.

internal/server/proto/server.proto
jgwhite: I think this is the right place to start, for sure.

Am I right in thinking nothing is returning these types yet?

izaaklauer: Right you are. That isn't implemented yet - this wip was intended just to demonstrate internal wiring of declared resources.

jgwhite: I just thought of one other preload-y thing we definitely will want right away: `deploy_url`

izaaklauer:
More context and detail around what this code is doing, and why:

In thinking about the best way for plugins to communicate their declared resources back to the server, we decided to give an "[outparameter](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier)" pattern a try. The idea is that plugins can optionally add a `*component.DeclaredResourcesResp` argument to their DeployFunc, which they fill in with their declared resources. It's pretty clean and backward compatable from the plugin side - plugins don't need to ask for the new arg. This will also let us make a new ResourceManager function that can take the DeclaredResourcesResp and fill it in on behalf of the plugin.

Here's the flow of invoking the DeployFunc and working with the outparameter.

1\. Core asks the docker plugin for its DeployFunc Spec [here](https://github.com/hashicorp/waypoint/blob/server-resources/internal/core/app_deploy.go#L327).

2\. The plugin SDK inspects the docker plugin's DeployFunc, and generates a spec that only takes proto type messages. Because the DeployFunc now needs a DeclaredResourcesResp component, we need a [proto -> component mapper](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal-shared/protomappers/mappers.go#L355), or we'll fail to generate the spec [here](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal/funcspec/spec.go#L50).

3\. In core, we're ready to call the plugin's DeployFunc. We make an empty DeclaredResourcesResp [here](https://github.com/hashicorp/waypoint/blob/server-resources/internal/core/app_deploy.go#L395), which gets [mapped](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal-shared/protomappers/mappers.go#L361) to an empty DeclaredResourcesResp proto. Without this, argmapper would fail [here](https://github.com/hashicorp/waypoint/blob/server-resources/internal/core/app.go#L254) because it would be missing a required argument.

4\. In core, argmapper turns the empty DeclaredResourcesResp in to a proto, and invokes the plugin sdk's Deploy function via GRPC.

5\. Inside the plugin framework, right before we invoke the Docker plugin's DeployFunc, we inject a new DeclaredResourcesResp argument ([here](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal/plugin/platform.go#L451)). If we didn't do this, argmapper would turn the DeclaredResourcesResp proto into a legit DeclaredResourcesResp component, and the plugin would use it, but we would be left without a reference to it after the plugin's DeployFunc exited. In practice, argmapper seems to take the shortest path and use our last-minute injected type. I included [a snippet](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal/plugin/platform.go#L438-L445) that would remove the bad proto, but we seem to not need it..

6\. Still inside the plugin framework, we call the DeployFunc with argmapper ([here](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal/plugin/platform.go#L447)). After it completes, we take the DeclaredResources out of the DeclaredResourceResp, and add them to the plugincomponent.Deployment response ([here](https://github.com/hashicorp/waypoint-plugin-sdk/blob/server-resources/internal/plugin/platform.go#L460)).

7\. The plugincomponent.Deployment gets serialized to/from proto and lands back on the core side as the response to invoking the plugin's Deploy func. That struct definition is private to the plugin SDK though, we need to use some funky reflection to dig the DeclaredResources out if it ([here](https://github.com/hashicorp/waypoint/blob/server-resources/internal/core/app_deploy.go#L340)).

8\. Huzzah, we have the plugin's DeclaredResources back in core. Mission accomplished.

It's a complex flow, and includes a fair amount of misdirection in core waypoint. Personally, I'm not convinced that this is the correction direction to take, but it's technically interesting and is nice for plugin developers.

internal/core/app_deploy.go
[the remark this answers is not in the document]
izaaklauer: I added this back - we need to hand onto a reference to the declaredResourceResp here in Do().

builtin/docker/platform.go
izaaklauer: Removed - this will be it's own PR

internal/server/proto/server.proto
izaaklauer: I like the idea of eventually deprecating all the fields in the Preload sections and moving them out to the Expanded apis. We can address that in another PR.

internal/core/app_deploy.go
izaaklauer: Solved this in a different way. I'm now passing an outparameter into the plugin framework with argmapper, and extracting the goodies from the private struct in the plugin code (running as part of the core process) and passing it back into the outparameter:

https://github.com/hashicorp/waypoint-plugin-sdk/pull/40/files#diff-1131f9061dccc1444fff98f5fc696b66ab7180c942cd0b66d768f13822614fb1R302

So we're using argmapper outparameters in two different ways as part of this change!

internal/core/app_deploy.go
izaaklauer: No longer needed! It gave me the willies too.

izaaklauer:
R/E the above very long comment:

After thinking and discussions, I have a new flow working! Instead of using reflection to dig values out of a private struct, I'm passing in _another_ outparameter into the plugin side of the core process, and the plugin is populating it ([here](https://github.com/hashicorp/waypoint-plugin-sdk/pull/40/files#diff-1131f9061dccc1444fff98f5fc696b66ab7180c942cd0b66d768f13822614fb1R283)). 

I've also removed the mappers that do nothing but satisfy argmapper, and replaced them with an argmapper filter function that prevents outparameters from being declared ([here](https://github.com/hashicorp/waypoint-plugin-sdk/pull/40/files#diff-6e083c58b681306ffed3d828652a8e0e068bcc3e4708ba40f1db6e355906a6dfR28))

evanphx (approved):
So clean now!

internal/server/proto/server.proto
izaaklauer: I'll leave this in here for now, but I think we will pretty shortly replace DeploymentExpanded with some different methods aimed directly at serving the UI.

briancain (approved):
🎉

PR #1738, opened 2021-06-30
ui: auto-fix linting issues in ui/app/components

jgwhite:
## Why the change?

One more step toward us being able to run the UI linters as part of CI.

## How do I verify it?

This should be entirely stylistic changes thanks to Prettier, so we can be fairly confident it won’t change behavior. If CI passes, we should be good. The one exception is the removal of the `<AppFooter>` backing class. To test this, boot the dev server and check that footer still renders (it appears fine in my testing).

### DIFF
--- a/ui/app/components/actions/context.ts
+++ b/ui/app/components/actions/context.ts
@@ -8,9 +8,9 @@ export default class ActionsRelease extends Component {
   @action
   toggleHint() {
     if (this.hintIsVisible === true) {
-      return this.hintIsVisible = false;
+      return (this.hintIsVisible = false);
     } else {
-      return this.hintIsVisible = true;
-    };
+      return (this.hintIsVisible = true);
+    }
   }
 }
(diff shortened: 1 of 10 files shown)


ui/app/components/actions/context.ts
```diff
@@ -8,9 +8,9 @@ export default class ActionsRelease extends Component {
   @action
   toggleHint() {
     if (this.hintIsVisible === true) {
-      return this.hintIsVisible = false;
+      return (this.hintIsVisible = false);
     } else {
-      return this.hintIsVisible = true;
-    };
+      return (this.hintIsVisible = true);
```
gregone, line 13: The fact that we're returning assignments here is weird, and wrapping those in expressions makes it look even weirder. Maybe we should be using `mut` or the `set/pick` helpers instead ? (not in this PR though)

jgwhite: Yeah, I agree, it is odd. As you say, something we can fix in a different PR.

jgwhite: That aside, are you happy to proceed with these changes?

gregone (approved):

PR #1740, opened 2021-06-30
builtin/aws: add ability to pass additional security groups

psihachina:
This commit adds the ability to pass additional security groups, for example
```
app "vole" {
 
  ...
  ...
  ...

  deploy {
    use "aws-ecs" {
      region = "us-east-2"
      memory = "512"
      cluster = "dev-nutcorp"
      security_groups = ["sg-0a98e407efe1aa683", "sg-078fc39329784c141"]
    }
  }
}
```
image

### Test
[asciicast](https://asciinema.org/a/xloZv22H6EkC44aFdefx3UzJ4)

### DIFF
--- a/.changelog/1740.txt
+++ b/.changelog/1740.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/aws: add ability to pass additional security groups for ecs
+```
--- a/.changelog/1740.txt
+++ b/.changelog/1740.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-plugin/aws: add ability to pass additional security groups for ecs
+plugin/aws: add ability to define existing security groups for ecs
 ```
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1247,9 +1247,14 @@ func (p *Platform) Launch(
 	// Create the service
 
 	L.Debug("creating service", "arn", *taskOut.TaskDefinition.TaskDefinitionArn)
-	sgecsport, err := createSG(ctx, s, sess, fmt.Sprintf("%s-inbound-internal", app.App), vpcId, int(p.config.ServicePort))
-	if err != nil {
-		return nil, err
+
+	if p.config.SecurityGroups == nil {
+		sgecsport, err := createSG(ctx, s, sess, fmt.Sprintf("%s-inbound-internal", app.App), vpcId, int(p.config.ServicePort))
+		if err != nil {
+			return nil, err
+		}
+
+		p.config.SecurityGroups = append(p.config.SecurityGroups, sgecsport)
 	}
 
 	count := int64(p.config.Count)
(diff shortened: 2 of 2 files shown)


krantzinator:
What do you think about providing a replacement set of security groups, rather than adding on to the default that Waypoint sets? So instead of `append` you'd just make a new slice.
Would this solve the root problem?

(Please remove the `waypoint-examples` submodule again 🙂 )

.changelog/1740.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+builtin/aws: add ability to pass additional security groups for ecs
```
briancain, line 2:
```suggestion
```release-note:improvement
plugin/aws: add ability to pass additional security groups for ecs
```

briancain (changes requested):
+1 to what @krantzinator mentioned.

We should fix up the changelog here too. We keep new functionalities as `feature`, where as new additions or improvements to existing plugins should be listed as `improvement`. We also opt for naming the changelog entry with `plugin` over `builtin` since its more obvious what that means to a user reading the CHANGELOG not familiar with the waypoint repo structure. Thanks!

.changelog/1740.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+plugin/aws: add ability to pass additional security groups for ecs
```
evanphx, line 2:
```suggestion
plugin/aws: add ability to define existing security groups for ecs
```

builtin/aws/ecs/platform.go
```diff
@@ -1567,6 +1572,9 @@ type Config struct {
 	// Subnets to place the service into. Defaults to the subnets in the default VPC.
 	Subnets []string `hcl:"subnets,optional"`
 
+	// SecurityGroups to pass security groups to deployment.
```
evanphx, line 1575: This should specify that it's the security group ids. Also, you need to add security_groups to the Documentation function below here.

evanphx (changes requested):

builtin/aws/ecs/platform.go
```diff
@@ -1567,6 +1572,9 @@ type Config struct {
 	// Subnets to place the service into. Defaults to the subnets in the default VPC.
 	Subnets []string `hcl:"subnets,optional"`
 
+	// Security Groups ARN to define existing security groups for ecs.
```
krantzinator, line 1575: Does the ARN work here? When I tested this I had to use the Security Group ID.

psihachina: Sorry, I'm confused about the concepts. Fixed

krantzinator:
Thanks for the quick response! We need to do the below items as well to get this fully ready. I can do it if you like, let me know.
* Update the `doc.SetField` to:
```
"Security Group IDs to define existing security groups for ecs",
		docs.Summary(
			"list of existing groups to be specified for the ecs",
```

* Rebase this branch onto `main`
* run `make gen/website-mdx`

That should do it!

psihachina:
@krantzinator It will probably be better if you do this because I have a `make get/website-mdx` error. 

image

PR #1745, opened 2021-06-30
Adjusted font size of view all button in card headers

sabrinako:
## Explanation
This PR fixes a small unexpected change in CSS with the refactor. The font size increased of the buttons in the card headers (ie "View all deployments/builds/releases")

### Before fix
localhost_4200_default_marketing-public_app_wp-matrix_logs (1)

### After fix
localhost_4200_default_marketing-public_app_wp-matrix_logs


gregone (approved):
✨ :shipit:

PR #1748, opened 2021-07-01
ui: reduce the impact of our app’s CSS on QUnit’s runner UI

jgwhite:
## Why the change?

At the moment, our app’s CSS affects QUnit’s UI in way that makes the module-selector unusable and also makes other things look broken and untrustworthy.

This is not a complete solution, but is the quickest stopgap I could think of makes the QUnit UI usable again without requiring too many other changes.

### Before

before

### After

after

## How do I test it?

1. `git checkout ui/qunit-inputs`
2. `cd ui && yarn start`
3. Visit [http://localhost:4200/tests](http://localhost:4200/tests)
4. Verify that the module selector is usable

### DIFF
--- a/ui/app/styles/components/input.scss
+++ b/ui/app/styles/components/input.scss
@@ -67,4 +67,12 @@ input {
     border: 1px solid;
     border-color: rgb(var(--input-disabled-border));
   }
+
+  #qunit & {
+    width: unset;
+    padding: unset;
+    font-family: unset;
+    font-size: unset;
+    line-height: unset;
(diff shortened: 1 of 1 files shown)


gregone (approved):
That works!

the global `input` style's `width: 100%` should probably be scoped to text-like inputs too (someday)

jgwhite:
@gregone yeah, I think ultimately we’ll want to use `<Pds::Input>` components explicitly and avoid this global style leakage altogether.

PR #1750, opened 2021-07-01
WIP: Encoding and validating token protocol version

izaaklauer:
Encoding server version info into waypoint tokens, and validating the protocol version before using the token.

Note that adding all this version info increases the token length by 25%, from 121 to 152 characters. I think the uint32 protocol version is the most useful thing here, and we could encode just that to save space.


mitchellh:
Going to close this since we don't want to do the whole protocol version :) i'll make a sep PR to see what this might look like w/ only the number.

PR #1752, opened 2021-07-01
User account system

mitchellh:
This completes most of WP-068. 

This introduces:
* The user account system models
* Associates tokens with users. 
* Enables signup tokens to create new users
* The APIs for users are all added.
* "Current user" available to all API calls

This does NOT do (and still needs to do in the future, as a separate PR):
* New CLI for tokens to setup new users
* Associate users with operations and other data model things

**Backwards compatibility:** old tokens from prior to 0.4 will continue to use the Waypoint superuser and will work _until_ the initial waypoint user (ID 1) is deleted.


internal/server/singleprocess/auth.go
```diff
@@ -57,44 +61,93 @@ var (
...
+			Id: &pb.Ref_UserId{Id: login.Login.UserId},
+		},
+	})
+	if status.Code(err) == codes.NotFound {
+		// If we are a legacy token logging into a WP 0.5+ server for the
+		// first time, we need to create the initial Waypoint user. This is
+		// purely a backwards compatibility case that we should drop at some
+		// point.
+		if !body.UnusedLogin || login.Login.UserId != DefaultUserId {
+			return nil, status.Errorf(codes.Unauthenticated, "Expired or invalid authentication token")
```
izaaklauer, line 130: When would an expired token make it here? I would think it would get caught on [on line 101](https://github.com/hashicorp/waypoint/pull/1752/files#diff-2155cc372482197f0003ac88874cc7fc653780a65ecea2305ab1a4b3b48d7d1eR101). Are you thinking about invite (non-login) tokens generated pre-update that haven't been redeemed?

izaaklauer: Also, if we merge this now, all tokens generated since #1644 will fail here because UnusedLogin is no longer set and defaults to false.

internal/server/singleprocess/auth.go
```diff
@@ -57,44 +61,93 @@ var (
...
+		}
+
+		// Bootstrap our user
+		if _, err := s.bootstrapUser(); err != nil {
+			// If we're already bootstrapped, give a slightly better error.
+			if status.Code(err) == codes.PermissionDenied {
+				return nil, status.Errorf(codes.Unauthenticated,
+					"Pre-Waypoint 0.5 token no longer accepted once the bootstrap user is deleted")
```
izaaklauer, line 138: It seems like `s.bootstrapUser` gives a PermissionDenied error if there are any users defined, not if a particular bootstrap user is deleted.

internal/server/singleprocess/auth.go
```diff
@@ -293,27 +349,27 @@ func (s *service) newToken(
...
 	if req.Login == nil {
 		req.Login = &pb.Token_Login{
-			// TODO(mitchellh): when we have a login system, set this
-			// to the currently logged in user.
-			UserId: DefaultUser,
+			UserId: currentUser.Id,
 		}
 	}
 
 	// We don't currently have any other users.
```
izaaklauer, line 371: Can probably remove this comment

internal/server/singleprocess/auth.go
mitchellh: Yes, but its impossible to create a user without first creating the bootstrap user (because the only way is a signup token, and that creates the bootstrap user first)! This might change with OIDC but should be valid here.

internal/server/singleprocess/auth.go
mitchellh: You actually found a missing feature. I have to implement this.

internal/server/singleprocess/auth.go
mitchellh: @izaaklauer `UnusedLogin` will be set to true for old tokens (tokens issued before 0.4). 

And the expired point: it never will be expired, let me fix the error message.

internal/server/singleprocess/auth.go
mitchellh: Fixed: https://github.com/hashicorp/waypoint/pull/1752/commits/0aebac367ca51b8f7c4c377167011487df517325

briancain (changes requested):
Requesting changes for the doc comment, but if the other comment doesn't make sense to change then no need to block on it :)

evanphx (changes requested):
A couple of things, mostly a race condition to protect from.

internal/server/proto/server.proto
```diff
@@ -493,6 +506,43 @@ message Workspace {
...
+  repeated Link links = 5;
+
+  // Link is a connection between an authentication provider and
+  // the user identity. This is used to lookup a unique user account
+  // within Waypoint from multiple auth sources (i.e. GitHub auth,
+  // user/pass, etc.).
+  message Link {
+    oneof method {
+      // TODO: one day!
+      string todo = 1;
```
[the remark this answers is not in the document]
mitchellh: It's probably okay cause its going to be replaced very shortly with the OIDC work.

internal/server/singleprocess/auth.go
```diff
@@ -57,44 +61,95 @@ var (
...
+		// point.
+		if !body.UnusedLogin || login.Login.UserId != DefaultUserId {
+			return nil, status.Errorf(codes.Unauthenticated,
+				"Pre-Waypoint 0.5 token must be for the default user. This should always "+
+					"be the case so the token is likely corrupt.")
+		}
+
+		// Bootstrap our user
+		if _, err := s.bootstrapUser(); err != nil {
+			// If we're already bootstrapped, give a slightly better error.
```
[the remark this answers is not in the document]
mitchellh: Yes, I thought of this: Two distinct users will never be created because they're created with the same ID. So the race will simply overwrite one with the other. I also felt that given bootstrap users PLUS legacy tokens are going to be so rare, its unlikely to ever cause real issues in any scenario.

internal/server/singleprocess/auth.go
```diff
@@ -57,44 +61,95 @@ var (
...
+			if status.Code(err) == codes.PermissionDenied {
+				return nil, status.Errorf(codes.Unauthenticated,
+					"Pre-Waypoint 0.5 token no longer accepted once the bootstrap user is deleted")
+			}
+
+			return nil, err
+		}
+
+		goto LOOKUP_USER
```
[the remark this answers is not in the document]
mitchellh: Good call, fixed!

briancain (approved):

internal/server/singleprocess/auth.go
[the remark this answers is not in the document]
evanphx: Ah ah ok. Gotcha sure.

evanphx (approved):

internal/server/singleprocess/auth.go
izaaklauer: r/e `UnusedLogin`: I hit this error when I presented a token created by main, because main no longer sets `UnusedLogin`. I don't think it's a huge issue, because we didn't backport #1644, but I think it'll mean everyone running off main will have to abandon their contexts, or cut a new token using 0.4.1.

mitchellh: Ah yes this is true, this is backwards incompatible with `main`, but we haven't ever really tried to maintain compatibility there, only with released versions. :)

mitchellh:
I was waiting on @izaaklauer's follow up and I think he gave it, so merging!

izaaklauer:
Haha you merged just as I was looking for the approve button 😄

PR #1761, opened 2021-07-01
Managing grpc and http server agents manually

### LINKED ISSUE #1476: Server panic on sigint
**Describe the bug**
The waypoint server panic-crashed after sigint via ctrl-c, with the following trace:

```
2021-05-17T15:53:54.024Z [INFO]  waypoint.server.http: HTTP request: POST /hashicorp.waypoint.Waypoint/ListBuilds: date=2021-05-17T15:53:54.023985604Z http.host=34.201.38.155:9702 http.method=POST http.request_path=/hashicorp.waypoint.Waypoint/ListBuilds http.remote_addr=47.6.36.181 http.response_size=2504 http.scheme= http.status_code=200 http.useragent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36" http.version=HTTP/1.1
2021/05/17 15:53:54 http: TLS handshake error from 47.6.36.181:55081: remote error: tls: unknown certificate
2021-05-17T15:53:54.132Z [INFO]  waypoint.server.grpc: /hashicorp.waypoint.Waypoint/ListDeployments response: error=<nil> duration=169.069908ms
2021-05-17T15:53:54.132Z [INFO]  waypoint.server.http: HTTP request: POST /hashicorp.waypoint.Waypoint/ListDeployments: date=2021-05-17T15:53:53.962851867Z http.host=34.201.38.155:9702 http.method=POST http.request_path=/hashicorp.waypoint.Waypoint/ListDeployments http.remote_addr=47.6.36.181 http.response_size=1380 http.scheme= http.status_code=200 http.useragent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36" http.version=HTTP/1.1
^C2021-05-17T15:54:01.869Z [WARN]  waypoint: interrupt received, cancelling context
2021-05-17T15:54:01.870Z [INFO]  waypoint.server.grpc: /hashicorp.waypoint.Waypoint/GetLogStream response: error="context canceled" duration=3m6.962165008s
2021-05-17T15:54:01.870Z [INFO]  waypoint.server.http: HTTP request: POST /hashicorp.waypoint.Waypoint/GetLogStream: date=2021-05-17T15:50:54.907719383Z http.host=34.201.38.155:9702 http.method=POST http.request_path=/hashicorp.waypoint.Waypoint/GetLogStream http.remote_addr=47.6.36.181 http.response_size=0 http.scheme= http.status_code=200 http.useragent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36" http.version=HTTP/1.1
2021-05-17T15:54:01.870Z [INFO]  waypoint.server.grpc: shutting down gRPC server
panic: Drain() is not implemented

goroutine 613 [running]:
google.golang.org/grpc/internal/transport.(*serverHandlerTransport).Drain(0xc000f950e0)
	/home/circleci/go/pkg/mod/google.golang.org/grpc@v1.33.1/internal/transport/handler_server.go:440 +0x39
google.golang.org/grpc.(*Server).GracefulStop(0xc000818000)
	/home/circleci/go/pkg/mod/google.golang.org/grpc@v1.33.1/server.go:1683 +0x385
github.com/hashicorp/waypoint/internal/server.grpcInit.func2.1(0xc000e6cae0, 0x3317c78, 0xc0001383f0, 0xc000818000)
	/home/circleci/project/waypoint/internal/server/grpc.go:73 +0x91
created by github.com/hashicorp/waypoint/internal/server.grpcInit.func2
	/home/circleci/project/waypoint/internal/server/grpc.go:70 +0xa5
```

There was no ongoing build, deploy, or release at the time of the panic. 

**Steps to Reproduce**
Unfortunately, I haven't been able to reproduce this. I was running a server manually on ec2, with this command:

```
waypoint server run -accept-tos -advertise-addr=<addr> -listen-grpc=0.0.0.0:9701 -listen-http=0.0.0.0:9702 -db=/opt/waypoint/data.db -advertise-tls-skip-verify -vvv
```

**Expected behavior**
Graceful shutdown

applicable:

* Waypoint CLI Version: v0.3.2
* Waypoint Server Platform and Version: v0.3.2

izaaklauer:
This addresses the panic described in https://github.com/hashicorp/waypoint/issues/1476.

We need to create the grpc server first, but close the http server first, which run groups do not allow for. This removes the run groups and manages grpc and http server struct lifecycles directly.

Closes #1476

### DIFF
--- a/internal/server/http.go
+++ b/internal/server/http.go
@@ -90,20 +90,21 @@ func (s *httpServer) close() {
 	gracefulCh := make(chan struct{})
 	go func() {
 		defer close(gracefulCh)
-		log.Info("shutting down HTTP server")
+		log.Info("stopping")
 		if err := s.server.Shutdown(ctx); err != nil {
-			log.Error("failed gracefully shutting down http server: %s", err)
+			log.Error("failed graceful shutdown: %s", err)
 		}
 	}()
 
 	select {
 	case <-gracefulCh:
+		log.Debug("exited gracefully")
 
 	// After a timeout we just forcibly exit. Our HTTP endpoints should
 	// be fairly quick and their operations are atomic so we just kill
 	// the connections after a few seconds.
 	case <-time.After(2 * time.Second):
-		log.Debug("stopping http server after waiting unsuccessfully for graceful shutdown")
+		log.Debug("stopping forcefully after waiting unsuccessfully for graceful stop")
 		cancelFunc()
 	}
 }
--- a/internal/server/http.go
+++ b/internal/server/http.go
@@ -2,6 +2,8 @@ package server
 
 import (
 	"context"
+	"github.com/hashicorp/go-hclog"
+	"google.golang.org/grpc"
 	"net"
 	"net/http"
 	"strings"
@@ -10,11 +12,17 @@ import (
 	assetfs "github.com/elazarl/go-bindata-assetfs"
 	"github.com/hashicorp/waypoint/internal/server/gen"
 	"github.com/improbable-eng/grpc-web/go/grpcweb"
-	"github.com/oklog/run"
 )
 
-// httpInit initializes the HTTP server and adds it to the run group.
-func httpInit(group *run.Group, opts *options) error {
+type httpServer struct {
+	opts *options
+	log hclog.Logger
+	server *http.Server
+}
+
+// newHttpServer initializes a new http server.
+// Uses grpc-web to wrap an existing grpc server.
+func newHttpServer(grpcServer *grpc.Server, opts *options) *httpServer{
 	log := opts.Logger.Named("http")
 	if opts.HTTPListener == nil {
 		log.Info("HTTP listener not specified, HTTP API is disabled")
@@ -22,7 +30,7 @@ func httpInit(group *run.Group, opts *options) error {
 	}
 
 	// Wrap the grpc server so that it is grpc-web compatible
-	grpcWrapped := grpcweb.WrapServer(opts.grpcServer,
+	grpcWrapped := grpcweb.WrapServer(grpcServer,
 		grpcweb.WithCorsForRegisteredEndpointsOnly(false),
 		grpcweb.WithOriginFunc(func(string) bool { return true }),
 		grpcweb.WithAllowNonRootResource(true),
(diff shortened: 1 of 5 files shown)


internal/server/server.go
```diff
@@ -32,32 +29,43 @@ func Run(opts ...Option) error {
...
+		}()
 	}
 
-	// Run!
-	return group.Run()
+	ctx, cancel := context.WithCancel(cfg.Context)
+	defer cancel()
+
+	select {
+	case err := <-errch:
```
mitchellh, line 59: Do we need to close here as well? i.e. if there is an error we should probably clean up the running goroutines?

catsby: We also don't know which goroutine this error came from, so we need to make sure calling `close()` on an already closed thing is OK. I looked briefly and it seems so, but depending on the nature of the error maybe the http/grpc server isn't even around anymore?

izaaklauer: 👍 good points. I'm now tracking which server exited, and shutting down the _other_ server. From my reading, it looks like when the server run processes exit they close their listeners, so you don't have to close them.

I'm also no waiting for each server routine to finish before exiting this function. This makes sure we have time to log the final error message before we quit.

internal/server/server.go
```diff
@@ -32,32 +31,62 @@ func Run(opts ...Option) error {
...
-	// We first add an actor that just returns when the context ends. This
-	// will trigger the rest of the group to end since a group will not exit
-	// until any of its actors exit.
-	ctx, cancelCtx := context.WithCancel(cfg.Context)
-	cfg.Context = ctx
-	group.Add(func() error {
-		<-ctx.Done()
-		return ctx.Err()
-	}, func(error) { cancelCtx() })
+	log := cfg.Logger.Named("server")
```
mitchellh, line 34: Another thing to consider is making another sub-logger `log.named("grpc")` (and HTTP) since I noticed your log messages in each start with "HTTP Server" or "GRPC Server" a lot which is a smell to me you want a named logger.

internal/server/http.go
```diff
@@ -47,43 +55,56 @@ func httpInit(group *run.Group, opts *options) error {
...
+	s.log.Info("starting HTTP server", "addr", ln.Addr().String())
+	return s.server.Serve(ln)
+}
 
-		select {
-		case <-gracefulCh:
+// close stops the grpc server, gracefully if possible.
+// Warning: before closing the GRPC server, this HTTP server must first be closed.
+// Attempting to gracefully stop the GRPC server first will cause it to drain HTTP connections,
+// which will panic.
```
mitchellh, line 83: I'd note this is expected to be called exactly once. Calling it multiple times seems... unknown.

internal/server/server.go
```diff
@@ -32,32 +31,62 @@ func Run(opts ...Option) error {
...
 	}
 
-	// Run!
-	return group.Run()
+	ctx, cancel := context.WithCancel(cfg.Context)
+	defer cancel()
+
+	defer wg.Wait() // Wait for server run processes to exit before returning
+	select {
+	case err := <-gprcErrs:
```
mitchellh, line 71: This is all fine, another option I usually do (esp when there are a lot more goroutines to juggle) is to make `Close` idempotent for each one, then whenever there is an error, I don't care where it came from, I just call close again on all of them. If they're already closed its fine.

mitchellh (approved):
Left a few comments. You don't need to address them. 

I didn't carefully test this logic either I'm trusting you ran it and it worked. :)

internal/server/http.go
```diff
@@ -47,43 +55,56 @@ func httpInit(group *run.Group, opts *options) error {
...
-		// After a timeout we just forcibly exit. Our HTTP endpoints should
-		// be fairly quick and their operations are atomic so we just kill
-		// the connections after a few seconds.
-		case <-time.After(2 * time.Second):
-			cancelFunc()
+	// Graceful in a goroutine so we can timeout
+	gracefulCh := make(chan struct{})
+	go func() {
+		defer close(gracefulCh)
+		log.Info("stopping")
```
briancain, line 93:
```suggestion
		log.Debug("stopping")
```

Should this be `debug` like the grpc server and other logs here?

internal/server/server.go
briancain: I had the same suggestion. I added a bunch of suggestions for logging messages to be more explicit about which server is doing what. But if the grpc and http server is formatted like ☝🏻 then I think most of those suggestions won't be needed.

briancain: I removed those review comments, we should just structure each servers logger to be named.

internal/server/server.go
```diff
@@ -32,32 +31,62 @@ func Run(opts ...Option) error {
...
+		// If the GRPC server errored, we can assume it's closed and shut down the http server if necessary
+		if httpEnabled {
+			httpServer.close()
+		}
+		return fmt.Errorf("failed running the grpc server: %w", err)
```
briancain, line 76: We shouldn't re-wrap the error returned from starting the server here, especially if the original error contains any `status` `codes` from grpc, We lose them if we do this I think and can't check the kind of error (i.e. we can't check later that `status.Code(err) == codes.gRPC_Status_Code`). Instead we can log above the return that we failed to run the grpc server and return the error as given?

Same comment below for failing to run the http server.

briancain (changes requested):
I still need to test this too before merge, but overall looks great!

internal/server/server.go
izaaklauer: 👍 Logging the error too seems like a good idea here, because it looks like the error is discarded upstream.

internal/server/server.go
izaaklauer: Hmm, I am creating and using named [http](https://github.com/hashicorp/waypoint/pull/1761/files#diff-5f2a899e8c4b83b44a53b5bea4e1a3129c4c9cadeab0331c3d3e6e6affac42a2R26) and [grpc](https://github.com/hashicorp/waypoint/pull/1761/files#diff-85d6db9d4d33f7064a02bbe45907ffdaca1cd1382bbfdb5ab8c4a7e77841cea5R23) loggers in their respective files. Are you suggesting that I create those loggers earlier, here in server.go, and use them [here](https://github.com/hashicorp/waypoint/pull/1761/files#diff-d0706a4ae4ecf91c5224ca388f31a087a36af06c40dc81321da7958b25c70b23R59) and [here](https://github.com/hashicorp/waypoint/pull/1761/files#diff-d0706a4ae4ecf91c5224ca388f31a087a36af06c40dc81321da7958b25c70b23R79)?

briancain: I would keep this logger, and instead of creating a new one in each function, pass this `log` through to the functions that need it and you can modify it using the `With` to give it the detail for which kind of server it is:

```golang
func newHttpServer(grpcServer *grpc.Server, log hclog.Logger, opts *options) *httpServer {
  log = log.With("http")
  // stuff
}
```

Then it should show up namespaced like `waypoint.server.http` or `waypoint.server.grpc` for example.

izaaklauer: Hmm, from my reading it looks like log.With is for adding key/value pairs. The logs do come through named the way I would expect in the current implementation - here's a snippet from shutdown:
```
2021-07-07T09:56:09.441-0500 [DEBUG] waypoint.server.http: stopping
2021-07-07T09:56:09.441-0500 [DEBUG] waypoint.server: HTTP server has exited: err="http: Server closed"
2021-07-07T09:56:09.442-0500 [DEBUG] waypoint.server.http: exited gracefully
2021-07-07T09:56:09.442-0500 [DEBUG] waypoint.server.grpc: stopping
2021-07-07T09:56:11.443-0500 [DEBUG] waypoint.server.grpc: stopping forcefully after waiting unsuccessfully for graceful stop
2021-07-07T09:56:11.443-0500 [DEBUG] waypoint.server: gRPC server has exited: err=<nil>
```

The parent logger is passed into the sub-functions as part of opts instead of as a standalone argument, which was the existing behavior.

briancain: Ahh, alright! I wasn't sure if that was the case or if the namespace was lost when you create a new one off that options config. That makes sense to me then 👍🏻

briancain (approved):
Gave running a server manually a shot and it works for me when I shut it down! 👍🏻

PR #1764, opened 2021-07-03
internal/serverinstall: Add support for installing the ECS Waypoint server with internal ELB

Licht-T:
We can choose if the LB is Internet-facing or not when installing the Waypoint server on k8s, but we cannot do so when installing on ECS. (I am not sure about Nomad.)

This PR adds the support for installing the ECS Waypoint server with internal ELB.

### DIFF
--- a/internal/serverinstall/ecs.go
+++ b/internal/serverinstall/ecs.go
@@ -74,6 +74,9 @@ type ecsConfig struct {
 	CPU string `hcl:"cpu,optional"`
 	// Memory configures the default amount of memory for the task
 	Memory string `hcl:"memory,optional"`
+
+	// InternalScheme configures the ELB Internet-facing or not
+	InternalScheme bool `hcl:"internal,optional"`
 }
 
 // Install is a method of ECSInstaller and implements the Installer interface to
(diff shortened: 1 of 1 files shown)


internal/serverinstall/ecs.go
```diff
@@ -1063,6 +1073,13 @@ func (i *ECSInstaller) InstallFlags(set *flag.Set) {
 		Usage:   "Configures the requested memory amount for the Waypoint server task in ECS.",
 		Default: "1024",
 	})
+
+	set.BoolVar(&flag.BoolVar{
+		Name:    "ecs-alb-internal",
```
catsby, line 1078: Nit: At this time, it's technically a [Network Load Balancer](https://aws.amazon.com/elasticloadbalancing/features/#Product_comparisons) . Ideally it would be an ALB but there's issues with using gRPC and ALBs at the moment that we have not solved, but I'm hopeful in the future we could. 

For now I think we should rename this to `ecs-lb-internal` and just refer to it as "load balancer", dropping the specifics of Application or Network. Thoughts?

catsby (commented):
I was unable to successfully setup Waypoint Server in ECS with this branch, because at this time we rely on the load balancer's public DNS for connection information:

https://github.com/hashicorp/waypoint/blob/67fda7a5574749506d9d48d948bf048d5953caaf/internal/serverinstall/ecs.go#L371

When installing waypoint server with an internal scheme like this, I assume a user would do so from _inside_ the VPC (for example from an EC2 server?). Otherwise we have no access to the private IP address that's used. 

I'm not sure how to get around this, any ideas? How do you imagine this being used?

PR #1791, opened 2021-07-07
builtin/aws: add custom name prefix for AWS resources

psihachina:
This commit adds the ability to specify a prefix for AWS resources, for example
```
app "vole" {
 
  ...
  ...
  ...

  deploy {
    use "aws-ecs" {
      region = "us-east-2"
      memory = "512"
      cluster = "dev-nutcorp"
      name_prefix = "dev-nutcorp"
    }
  }
}
```
image
image

### Test
[asciicast](https://asciinema.org/a/pNW6mJtIFOuZcCCcPYNq86YVB)

### DIFF
--- a/.changelog/1791.txt
+++ b/.changelog/1791.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/aws: add custom name prefix for AWS resources
+```
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -35,6 +35,8 @@ func (p *Platform) Config() (interface{}, error) {
 	return &p.config, nil
 }
 
+var prefix string
+
 // ConfigSet is called after a configuration has been decoded
 // we can use this to validate the config
 func (p *Platform) ConfigSet(config interface{}) error {
@@ -266,6 +268,11 @@ func (p *Platform) Deploy(
 		err error
 	)
 
+	prefix = "waypoint"
+	if p.config.NamePrefix != nil {
+		prefix = *p.config.NamePrefix
+	}
+
 	if p.config.ALB != nil {
 		if p.config.ALB.ListenerARN != "" {
 			if p.config.ALB.ZoneId != "" || p.config.ALB.FQDN != "" {
@@ -460,7 +467,7 @@ func (p *Platform) SetupExecutionRole(ctx context.Context, s LifecycleStatus, L 
 	roleName := p.config.ExecutionRoleName
 
 	if roleName == "" {
-		roleName = "ecr-" + app.App
+		roleName = fmt.Sprintf("%s-%s-%s", prefix, "ecr", app.App)
 	}
 
 	// role names have to be 64 characters or less, and the client side doesn't validate this.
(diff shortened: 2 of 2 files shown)


krantzinator:
Hello! I have a question around the purpose here. It looks like you are wanting the name prefix to give you more fine tuned organization of your AWS resources. Does it make more sense to support [AWS tags](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html) for resources that Waypoint creates?
If tags does not fit your intended purpose here, please provide more context as to what this change is helping you to achieve. Thanks!

(In the meantime, please remove the `waypoint-examples` submodule)

AutomationD:
Hey, @krantzinator

While I agree with you, that tags should solve most of the organization requirements in the proper modern way, this feature would help us to align with our standardized (read legacy) resource naming.

In the enterprise environments we use specific naming patterns, sometimes it's `<ENV>-<NAMESPACE>-<NAME>` sometimes it's just `<ENV>-<NAME>`. It helps to visually distinguish the resources. Also, as far as I understand, tags my not supported in all of the AWS resources (but that might've changed), as well as work same way across all cloud platforms.

This is basically a feature that would allow us to use Waypoint as a part of some legacy system (let's say Cloudformation or Terraform-managed), where a service name is usually prefixed with some namespace/container.

My 5 cents here.

krantzinator:
@AutomationD that makes sense.
@psihachina For the execution of this PR, I have a few suggestions:
1) we should remove the `waypoint-examples` submodule
2) add a changelog entry (you can see examples in `./changelog`; there is additional info in our [changelog guide](https://github.com/hashicorp/waypoint/blob/e7173e8179e1f01768c9f468e49b9070028979ae/.github/CHANGELOG_GUIDE.md) as well)
3) We should set a top-level variable assigned to a prefix default, and then overwrite that default if there is a value set in the config. We should also use the prefix for all resources created in AWS, rather than just a couple, to be consistent.

I think we should include `waypoint` in the prefix still, in addition to whatever prefix is specified, so it's clear what is creating these resources. What do you think?

psihachina:
@krantzinator 
 Include waypoint in the prefix still, in addition to whatever prefix is specified is a good idea, but as result there will be a very long name and ID that did not fit, for example, prefix `dev-nutcorp` - `
waypoint-dev-nutcorp-hamster-01F`

krantzinator:
@psihachina That makes sense. I think this gives greater importance to the work around resource awareness so we clean up everything under a given application, even if a user updates their prefix.

evanphx:
Thanks for the PR @psihachina. After discussion, we're going to reject this PR because we like the idea so much that we actually want to make this prefix available as a workspace setting that is adhered to by all the plugins. Because we want to do it more holistically, we don't want to have `name_prefix` be a one-off in ECS.

Additionally, there are aspects of your PR that will create issues with users after merge (mostly around accidentally creating new load balancers and other resources). Waypoint isn't tracking the resources well enough yet, and we need to do that better before subjecting folks to name changes.

evanphx:
@psihachina We've made an issue to track adding this to the workspace:  #1862

Issue #1612, opened 2021-06-07
Feature: Continuously generate status reports on a queued poll interval or on demand for deployments and releases


briancain:
**Is your feature request related to a problem? Please describe.**

Waypoint 0.4.0 introduced generating a status report which forwarded the current platform reported health of a deployment or release. This report was only generated after a deploy or releaser finished executing, rather than continuously throughout an applications lifetime. Continuously generating a report on an interval or on demand will increase the confidence that the application Waypoint deployed or released is up and ready for requests _now_ rather than the few seconds after it finished with the app. Checking on the status of an app is also important when applications are slow to become healthy. Additionally, these reports will give the Waypoint UI a more accurate picture as to the current state of an application.

**Describe the solution you'd like**

Waypoint server should poll for and report on the status of its deployment or releases continuously during the lifecycle of a deployment or release on a poll interval that is respectful enough to not overload a system or max out a platforms request limit.

**Steps to Implementation**

- [x] (PR: https://github.com/hashicorp/waypoint/pull/1604) Refactor the current polling system to be more generic and support multiple kinds of items to queue a poll job on 
- [x] (Issue: https://github.com/hashicorp/waypoint/issues/1614, PR: https://github.com/hashicorp/waypoint/pull/1649) Add Peek and Complete funcs to appOperation then update status reports state package to include `PollPeek` and `PollComplete` funcs
- [x] (This closes this issue) Add a `poll_status_report` poll queuer to Waypoint servers service

PR #1801, opened 2021-07-07
feature: Queue status reports on continuous interval for applications

briancain:
This pull request introduces a queued poll handler for applications which queues a status report job every 5 minutes by default after the initial deployment or release of an application. It includes a peek and complete function for applications that behaves similar to ProjectPeek and ProjectComplete. It includes its own index table to keep track of the applications poll times and intervals. Initially, each projects applications poll handler is turned off until a user does a deployment or release. This saves us cycles in the server where there are no applications to generate a status for. The status report operation is in charge of updating the apps poll status once a status report has been initially generated.

This feature also allows users to turn off the application poll queuer or adjust the interval, given that it's likely with enough applications users could run into rate limits on their accounts for platforms such as AWS.

Fixes https://github.com/hashicorp/waypoint/issues/1612

Note: Currently this will only poll if the project is backed by a remote data source with git.


internal/server/singleprocess/poll_test.go
```diff
@@ -75,3 +75,195 @@ func TestServicePollQueue(t *testing.T) {
...
+	// ensure the next poll is after the initial poll before waiting
+	// next poll time gets set when a project poll is marked complete
+	p, nextPollTime, err := state.ProjectPollPeek(nil)
+	require.NoError(err)
+	require.NotNil(p)
+	require.NotNil(nextPollTime)
+	require.True(nextPollTime.After(pollTime))
+}
+
+func TestApplicationPollHandler(t *testing.T) {
```
mitchellh, line 158: Big picture comment 💭 : these tests are _fine_ and good! But now that we extracted poll handling to an interface, we can probably test the interface methods do what we expect.

For example: these big picture tests should test that the poller handles the interface properly (and probably use a mock/test interface to test all the edges). Separately, we should test that our interface implementation _returns_ the proper data for each method call.

The actionable feedback here would be to make those lines clear. I would say to probably just skip that in this PR and noodle on it for another PR.

internal/server/singleprocess/service_job.go
```diff
@@ -81,28 +81,14 @@ func (s *service) QueueJob(
...
 	}
 
-	if job.DataSource == nil {
-		if project.DataSource == nil {
-			return nil, status.Errorf(codes.FailedPrecondition,
-				"Project %s does not have a data source configured. Remote jobs "+
-					"require a data source such as Git to be configured with the project. "+
-					"Data sources can be configured via the CLI or UI. For help, see : "+
-					"https://www.waypointproject.io/docs/projects/git#configuring-the-project",
-				job.Application.Project,
```
mitchellh, line 99: I understand why you got rid of this error. 

Did this error go anywhere else (in the runner perhaps)? I'm wondering what error you get if you try to queue an `up` for a project that doesn't have a remote data source configured, for example. That is what this was protecting.

internal/server/singleprocess/state/application_test.go
```diff
@@ -2,13 +2,15 @@ package state
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 	/*
 		"google.golang.org/grpc/codes"
 		"google.golang.org/grpc/status"
 	*/
```
mitchellh, line 11: This isn't you, but this is weird and we should prob delete it.

internal/server/singleprocess/state/project.go
```diff
@@ -350,8 +350,9 @@ func (s *State) projectDelete(
 // projectIndexSet writes an index record for a single project.
 func (s *State) projectIndexSet(txn *memdb.Txn, id []byte, value *pb.Project) error {
 	record := &projectIndexRecord{
-		Id:   string(id),
-		Poll: false, // being explicit that we want to default poll to false
+		Id:       string(id),
+		Poll:     false, // being explicit that we want to default poll to false
+		ApplPoll: true,  // application polling off by default until a deployment or release happens
```
mitchellh, line 355: Probably want to use `AppPoll` (not `Appl`). `Appl` is a fine shortening of application but we don't use it anywhere else while we use App prodigiously.

mitchellh (changes requested):

internal/server/singleprocess/service_job.go
briancain: It did! It got moved directly into the `internal/server/singleprocess/poll_project.go` file. The change is in this PR, there's just a lot in the diff so easy to miss 😅

internal/server/singleprocess/state/application_test.go
briancain: Good call! Very strange, I'll remove 👍🏻

internal/server/singleprocess/state/project.go
briancain: 👍🏻 makes sense! I used `Appl` in a bunch of places in this func so I'll fix them all up

internal/server/singleprocess/poll_application.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	return jobList, nil
+}
+
+// PollJob will generate a job to queue a project on.
+// When determining what to generate a report on, either a Deployment or Release,
+// this job assumes that the _Release_ was the last operation taken on the application.
+// If there's a release, this job will queue a status report genreation on that.
+// Otherwise if there's just a deployment, it return a job to generate a
```
krantzinator, line 85:
```suggestion
// Otherwise if there's just a deployment, it returns a job to generate a
```

internal/server/singleprocess/state/project.go
```diff
@@ -350,8 +351,9 @@ func (s *State) projectDelete(
 // projectIndexSet writes an index record for a single project.
 func (s *State) projectIndexSet(txn *memdb.Txn, id []byte, value *pb.Project) error {
 	record := &projectIndexRecord{
-		Id:   string(id),
-		Poll: false, // being explicit that we want to default poll to false
+		Id:      string(id),
+		Poll:    false, // being explicit that we want to default poll to false
+		AppPoll: true,  // application polling off by default until a deployment or release happens
```
krantzinator, line 356: Should this be false?

internal/server/singleprocess/state/project.go
```diff
@@ -466,27 +536,51 @@ const (
 	projectIndexTableName         = "project-index"
 	projectIndexIdIndexName       = "id"
 	projectIndexNextPollIndexName = "next-poll"
+	appIndexNextPollIndexName     = "app-next-poll"
 
 	projectWaypointHclMaxSize = 5 * 1024 // 5 MB
 
-	defaultPollInterval = "30s"
+	// defaultProjectPollInterval is used the project poll handler
```
krantzinator, line 543:
```suggestion
	// defaultProjectPollInterval is used by the project poll handler
```

internal/server/singleprocess/state/project.go
```diff
@@ -466,27 +536,51 @@ const (
...
+	// for setting up a default interval time
+	defaultProjectPollInterval = "30s"
+
+	// defaultAppPollInterval is used for polling a status report job for each
+	// application defined in a project. It is initially set to a long interval
+	// so that Waypoint doesn't overrun and rate limit user accounts like on AWS.
```
krantzinator, line 549: 👏🏻

krantzinator (approved):
This was a joy to review, excellent documentation 🎉

internal/server/singleprocess/poll.go
```diff
@@ -158,16 +166,31 @@ func (s *service) runPollQueuer(
...
+				time.Sleep(1 * time.Second)
+				break
+			}
+			log.Debug("queued polling job", "job_id", resp.JobId)
+		}
+
+		if failedQueueJob {
+			log.Warn("A job failed to be queued. Any further downstream jobs were " +
+				"not queued. This should not happen repeatedly. If you see this in your " +
+				"log frequently, report a bug.")
```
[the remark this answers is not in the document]
briancain: Hm, this is a good point. Honestly any more than once I think is bad and should be reported. I'll rework this message!

internal/server/singleprocess/state/project.go
briancain: Yep.... This was originally set to true when I was thinking we'd have app poll settings for every app. Essentially the apps wouldn't poll until the `AppPoll{}` struct existed inside a specific app record, but app polling for a project as a whole could be switched on and off.

Anyway TL;DR we decided to make app polling a single setting across a project so yes! This should be fixed up

internal/cli/project_apply.go
```diff
@@ -454,6 +467,21 @@ func (c *ProjectApplyCommand) Flags() *flag.Sets {
 			Default: "30s",
 			Usage:   "Interval between polling if polling is enabled.",
 		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:    "app-poll",
```
mitchellh, line 472: Should this be `status-poll`? What else are you expecting to poll in the future at the app level?

internal/server/proto/server.proto
```diff
@@ -493,6 +501,16 @@ message Project {
     // than the minimum will be ignored.
     string interval = 2;
   }
+
+  message AppPoll {
```
mitchellh, line 505: Still a bit confused why some things are status polling and some are app polling and what other things might be polled in an app.

internal/server/singleprocess/poll.go
```diff
@@ -33,6 +33,12 @@ type pollHandler interface {
 	// is used to define the job returned.
 	PollJob(hclog.Logger, interface{}) (*pb.QueueJobRequest, error)
 
+	// GeneratePollJobs will build a list of QueueJobRequests through PollJob to
+	// schedule for execution. If a pollItem has multiple elements that need to
+	// be polled for, this is required for scheduling a job for each poll item.
+	// If there is only one poll item, the list should only contain one queue job request.
+	GeneratePollJobs(hclog.Logger, interface{}) ([]*pb.QueueJobRequest, error)
```
mitchellh, line 40: Should we just have `PollJob` return a slice? I don't thikn we need two methods for this: existing things can return just one queue job request.

internal/server/singleprocess/poll.go
```diff
@@ -158,16 +166,31 @@ func (s *service) runPollQueuer(
...
 		}
 
-		resp, err := s.QueueJob(ctx, queueJobRequest)
-		if err != nil {
-			log.Warn("error queueing a poll job", "err", err)
-			time.Sleep(1 * time.Second)
+		totalRequests := len(queueJobRequests)
+		log.Trace("queueing jobs for poller", "job_total", totalRequests)
+
+		// If one job fails to queue, we break out of the entire queue loop
```
mitchellh, line 172: I think important to note: but we _keep_ the ones that did succeed queueing, so its possible by design to have partial queueing here.

If we don't want that, its certainly possible to make the state store treat the entire thing like a transaction.

internal/server/singleprocess/poll_project.go
```diff
@@ -59,6 +78,9 @@ func (pp *projectPoll) PollJob(
 			// any time queued per project.
 			SingletonId: fmt.Sprintf("poll/%s", p.Name),
 
+			// Project polling requires a data source to be configured for the project
+			DataSource: p.DataSource,
```
mitchellh, line 82: I thought that jobs default to their project-configured data source at some point?

This probably does no harm, but was curious if you saw this break for some reason? They should default to it because the CLI generates jobs without data sources attached.

internal/runner/operation_status_report.go
```diff
@@ -40,9 +47,71 @@ func (r *Runner) executeStatusReportOp(
...
 	}, nil
 }
+
+// enableApplicationPoll will switch on its queued poll handler. This means
+// on the defined interval, Waypoint will generate a status report for this
+// application. When an application is initially inserted, like on a `waypoint init`,
+// it won't enable the poller to generate a status report. This method switches
+// it on after the first status report is generated. Each time after it should
+// do nothinng when its polling has been enabled.
+func (r *Runner) enableApplicationPoll(
```
mitchellh, line 71: This feels a little weird to me. My immediate question is: why can't polling be enabled by default, and the poll query just ignores any applications that never had any status reports? This should be solvable in the data layer, and not through a state machine. 

i.e. if somehow all the deployments get deleted, won't the polling still be enabled? That feels somewhat broken. If we solve it in the data layer: only poll things that (1) have a deployment and (2) have at least one status report then it will be resilient to any state changes, right?

mitchellh (changes requested):

internal/cli/project_apply.go
briancain: Nothing at the moment! I think I'll go with `app-status-poll` since it's a flag on project, I think that's a bit clearer.

internal/server/singleprocess/poll_project.go
briancain: I believe this was left over from when I had mistakenly moved the datasource job validation. But I think this shouldn't be harmful since it gets attached again in service_job later when we queue this job struct and validate a data source was configured with the project.

I can remove it back to how it was though if that's preferred 🤔

internal/server/proto/server.proto
briancain: Renamed everything from `App*` to `AppStatus*` to be extra clear what the settings are for

internal/runner/operation_status_report.go
briancain: Ok, this makes more sense to me! I've removed this func in favor of turning on app status polling by default for projects that enable polling and have a remote data source! If the app doesn't have a status report or a deployment, a job for generating a status report won't be queued already by the poll handler.

mitchellh (approved):

internal/server/singleprocess/poll_application.go
```diff
@@ -0,0 +1,210 @@
...
+			// SingletonId so that we only have one poll operation at
+			// any time queued per application.
+			SingletonId: fmt.Sprintf("appl-poll/%s", app.Name),
+
+			Application: &pb.Ref_Application{
+				Application: app.Name,
+				Project:     app.Project.Project,
+			},
+
+			// Applicatioon polling requires a data source to be configured for the project
```
izaaklauer, line 167:
```suggestion
			// Application polling requires a data source to be configured for the project
```

internal/server/singleprocess/service_job.go
```diff
@@ -119,7 +119,7 @@ func (s *service) queueJobReqToJob(
 	project, err := s.state.ProjectGet(&pb.Ref_Project{Project: job.Application.Project})
 	if status.Code(err) == codes.NotFound {
 		return nil, status.Errorf(codes.NotFound,
-			"Project %s was not found! Please ensure that 'waypoint init' was run with this project.",
+			"Project %q was not found! Please ensure that 'waypoint init' was run with this project.",
```
izaaklauer, line 122: TIL about %q! very nice.

izaaklauer (approved):

PR #1803, opened 2021-07-07
Fix the exec/platform and pack/builder docs

izaaklauer:
For the last few releases, we've been massaging these mdx files by hand at the last minute, because our automation produced broken mdx files. With these changes, the automation produces good mdx files.

Here's what the pack/builder complex docs section looks like after this change:

Screen Shot 2021-07-07 at 4 51 45 PM

index.mdx also had some minor whitespace changes that i'm not going to question too much.

### DIFF
--- a/builtin/exec/platform.go
+++ b/builtin/exec/platform.go
@@ -290,7 +290,7 @@ template. The top of the documentation there has details on the format.
 
 The following template values are always available:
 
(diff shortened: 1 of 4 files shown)


mitchellh (approved):
A hero!

briancain (approved):

PR #1816, opened 2021-07-08
Verify no mdx changes during circleCI runs

izaaklauer:
This PR introduces a new circleCI workflow which runs `make gen/website-mdx` and verifies there were no changes.

In practice, this means that after making a change to our docs strings, we will need to generate mdx files and validate that the website pages look acceptable before CI will pass. It will prevent us from having to perform this step at release-time.

**NOTE**: The circleCI check here is intentionally failing on the new gen/website-mdx check, so that you all can see what it looks like. If we approve this PR, I'll run `make gen/website-mdx` and commit to clear the check.

This also includes a few related changes:

### Consistent ordering of platform options
Our platform options were randomly ordered before, which would have made this check unreliable. Changes to `cli/install.go`, `cli/server-upgrade.go`, and `cli/server-uninstall.go` fix this.

### Fix "too many open files" panic when making website mdx files
We used to be instantiating and holding open every plugin at once to generate mdx files, which would sometimes cause a panic. Changes to `cli/app_docs.go` fix this.

### Do not display platform options with no flags

Mostly unrelated - it bothered me that we were showing headers for platforms with no options in the server uninstall command docs here: https://www.waypointproject.io/commands/server-uninstall

Screen Shot 2021-07-08 at 10 28 51 AM

Changes to `cli/docs.go` omit sections with no flags.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,52 @@ jobs:
           fi
       - notify_main_failure
 
+  generate-website-mdx:
+    docker:
+      - image: *GOLANG_IMAGE
+    environment:
+      <<: *ENVIRONMENT
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run:
+          name: generate plugin docs
+          command: go run ./cmd/waypoint docs -website-mdx
+      - run:
+          name: generate cli docs
+          command: go run ./tools/gendocs
+      - run:
+          command: rm waypoint-restore.db.lock; git status
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
+      - notify_main_failure
+
+  check-website-mdx:
+    docker:
+      - image: *EMBER_IMAGE
+    environment:
+      <<: *ENVIRONMENT
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          command: git status
+      - run:
+          command: cd ./website; npm install
+      - run:
+          command: cd ./website; npx --no-install next-hashicorp format
+      - run: |
+          if ! git diff --exit-code; then
+            echo "Git directory has website mdx changes. Run `make gen/website-mdx`"
+            echo "and validate that the corresponding website pages look acceptable."
+            git status
+            exit 1
+          fi
+      - notify_main_failure
+
   go-test:
     docker:
       - image: *GOLANG_IMAGE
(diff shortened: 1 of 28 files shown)


internal/cli/install.go
```diff
@@ -337,7 +338,17 @@ func (c *InstallCommand) Flags() *flag.Sets {
...
 
-		for name, platform := range serverinstall.Platforms {
+		// Add platforms in alphabetical order. A consistent order is important for repeatable doc generation.
+		i := 0
+		sortedPlatformNames := make([]string, len(serverinstall.Platforms))
+		for name := range serverinstall.Platforms {
+			sortedPlatformNames[i] = name
+			i++
+		}
+		sort.Strings(sortedPlatformNames)
```
briancain, line 348: We have this loop in three places now....is there a place where we can move it more directly into the docs generation so it always sorts alphabetically and these callers get the behavior automatically? Then future CLI packages get it "for free" too.

.circleci/config.yml
```diff
@@ -67,6 +67,52 @@ jobs:
...
+      - run:
+          command: git status
+      - run:
+          command: cd ./website; npm install
+      - run:
+          command: cd ./website; npx --no-install next-hashicorp format
+      - run: |
+          if ! git diff --exit-code > /dev/null; then
+            echo "Git directory has website mdx changes. Run make gen/website-mdx"
+            echo "and validate that the corresponding website pages look acceptable."
```
briancain, line 110: What should someone do after this to resolve the failure? It's not clear to me reading this error message. Maybe another sentence to say what to do to resolve the issue would be good

```suggestion
            echo "'website' directory has unstaged mdx changes. Run make gen/website-mdx"
            echo "and validate that the corresponding website pages look acceptable."
```

briancain (changes requested):
Nice @izaaklauer ! Happy to be removing some extra manual steps to our release process, looks good. Some minor comments from me 😄

internal/cli/install.go
izaaklauer: I thought about that, but we usually want to preserve the order that plugins insert flags, so that they can set "global options" before "command options", so we can't alphabetize flag order for everyone.

This is also a weird case related only to server commands, where we have our own little internal plugin system for the different kinds of platforms you can install waypoint on. I seem to remember us thinking we weren't going to double down on this pattern in the future if possible.

briancain (approved):

PR #1817, opened 2021-07-08
core: fix error handling to check for nil value

krantzinator:
Writing docs is a great way to find bugs!

Before, we were checking if the key existed for a variable (which it always will at this point; error handling for missing keys is earlier). Now, we are actually checking if the value is nil or not.

Before:
```
$ waypoint init
⠙ Validating configuration file...panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xe17198]

goroutine 67 [running]:
github.com/hashicorp/waypoint/internal/config/variables.Values.values(0xc000964030, 0xc000b13b78)
	/home/rae/waypoint-project/waypoint/internal/config/variables/variables.go:592 +0xb8
```

After _(note the links for [docs] is still WIP and is OK for now)_:
```
$ wdi
✓ Project "example-python" and all apps are registered with the server.
❌ Failed to load and validate plugins!

! This validation check ensures that you have all the required plugins available
  and the configuration for each plugin (if it exists) is valid. The error message
  below should tell you which plugin(s) failed.

! <nil>: Unset variable "port"; A variable must be set or have a default value;
  see [docs] for details.
```

### DIFF
--- a/internal/config/variables/variables.go
+++ b/internal/config/variables/variables.go
@@ -400,8 +400,8 @@ func EvaluateVariables(
 
 	// check that all variables have a set value, including default of null
 	for name := range vs {
-		value := iv[name]
-		if value == nil {
+		v, ok := iv[name]
+		if !ok || v == nil {
 			return nil, append(diags, &hcl.Diagnostic{
 				Severity: hcl.DiagError,
 				Summary:  fmt.Sprintf("Unset variable %q", name),
--- a/internal/config/variables/variables.go
+++ b/internal/config/variables/variables.go
@@ -290,7 +290,6 @@ func EvaluateVariables(
 ) (Values, hcl.Diagnostics) {
 	var diags hcl.Diagnostics
 	iv := Values{}
-
 	for v, def := range vs {
 		iv[v] = def.Default
 	}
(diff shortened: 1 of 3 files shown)


internal/config/variables/variables.go
```diff
@@ -400,8 +400,8 @@ func EvaluateVariables(
 
 	// check that all variables have a set value, including default of null
 	for name := range vs {
-		_, found := iv[name]
-		if !found {
+		value := iv[name]
+		if value == nil {
```
izaaklauer, line 404: Is it possible for a variable be found, but nil? It looks like we're not checking for that case anymore with this change.

mitchellh: If that's the case, a common pattern I use is:

```go
v, ok := m[k]
if !ok || v == nil {
```

mitchellh: If the value of a map is nullable, I usually do the above anyways out of caution.

krantzinator: @izaaklauer yes, the case we are checking for here is if a found variable is nil. At this point, we have checked for "non found" variables already. But, to your and Mitchell's points, caution is good! I'll incorporate Mitchell's suggestion.

briancain (approved):
🦖 🎉

izaaklauer (approved):

PR #1821, opened 2021-07-08
Git polling with a subpath specified should ignore git changes outside that path

### LINKED ISSUE #1796: Polling should consider sub-directory of project for changes
The way Waypoint works currently is well described by the following paragraph of the documentation:
> A project in Waypoint is a 1:1 mapping to a VCS repository (if one exists). A project can be comprised of one or more applications. Each application in a project represents a single thing that can be deployed. A project is the first thing you ever create after installing Waypoint.

Waypoint seems to be more focused on tackling the CI + CD part, but most of the time the CI part might be super complex and thus be handled by another tool such as Jenkins. This is why (I guess?) Waypoint lets you have a "Build" step that is not really a build per se (e.g: `docker-pull`).

We're currently using Waypoint as a multi-cloud / multi-target continuous delivery software, and as such we're trying to use Waypoint as we would use [Flux.CD](https://fluxcd.io/), but for generic `waypoint.hcl` files.

In Flux.CD you usually have a single repository (single source of truth) containing all your Kubernetes manifests, and an agent on your Kubernetes cluster(s) takes care of polling your repository (exactly as Waypoint does) and `kubectl apply`ing the manfiests.

Unfortunately, for the way Waypoint is built, the separation between code and the continuous delivery part is not that easy to achieve for the following problems:

- Per-workspace project settings cannot be applied (this means that you cannot use branch `develop` for the `develop` workspace, branch `staging` for the `staging` workspace and so on...) (#1665)
- A git commit on the set-up repository triggers a `waypoint up`, despite maybe the files in the _Git Source Path (optional)_ did not change

In our current set-up, we have a `waypoint` repository configured as follows:
- 3 branches (`develop`, `staging`, `prod`)
- `project/waypoint.hcl` structure

The folder looks like following:
```
.
├── README.md
└── key
    ├── prj-1
    │   ├── VERSION
    │   └── waypoint.hcl
    └── prj-2
        ├── VERSION
        └── waypoint.hcl
```

The `waypoint.hcl` file looks like as follows:

```hcl
project = "prj-1"
labels = { "product" = "project-1" }

app "app-1" {
    build {
        use "docker-pull" {
            image = "prj-1"
            tag   = trimspace(file("${path.app}/VERSION"))
            disable_entrypoint = true
        }
    }

    deploy {
        use "cloudfoundry" {
            organisation = "org"
            space = "prj-1-${workspace.name}"
            domain = "xyz.swisscom.ch"
            envFromFile = file("${path.app}/.env")
            quota {
                memory = "512Mi"
                disk = "2Gi"
            }
        }
    }

    release {
        use "cloudfoundry" {
            domain = "xyz.swisscom.ch"
            hostname = "prj-1-${workspace.name}"
        }
    }
}
```

Jenkins opens a PR to our single source of truth repo with the change (bump of the `VERSION` file), so that then Waypoint can deploy the new stuff:

image

Unfortunately, with this approach, having as a Data Source for `prj-1` and `prj-2` the same git repository causes `prj-2` to be deployed as well if you change something in `prj-1`: this is due to the fact that file changes are not taken into consideration but the poll algorithm is merely checking for a newer commit and then running `waypoint up` with the _Git Source Path_ (which is different from `prj-1` and `prj-2`) 

It would be awesome if the polling took into consideration paths as well (e.g: no files changed in `prj-1` directory, let's skip this commit). Bonus points if this polling is centralized so that we don't have to poll 10 times the same repository to check for changes when a single repository is specified for every project.

`1:1` project - repo mapping, although we're trying to move away from this approach because we would like to have a GitOps repository where all the CD related changes happen

N/A

**Additional context**

N/A

mitchellh:
Fixes #1796 

**Open questions at the end of this desc for the Waypoint team, please check it out.**

Currently, our git poller works by simply comparing commit hashes. If they aren't equal, then we trigger an `up`. 

In #1796, it was pointed out that for monorepos where many Waypoint projects exist in one repository, this triggers many superfluous `up` operations because they get triggered when there are changes in paths unrelated to the project. 

This PR changes the behavior of the Git change detector so that _if a subpath is specified for a project_, we only consider there to be changes if the differences between the two Git trees contain a change within that subpath. 

**This has a downside:** if users put their Waypoint HCL file in a subdirectory and WANT to trigger for all changes ANYWHERE (even outside the subpath), this will break their behavior. So... open questions: 

* do we want this PR behavior at all?
* do we want to add an override to NOT do this? (default true, default false?)
* we can also add a more complex option such as "additional trigger paths" that we can check


evanphx (changes requested):
So I think this should be an opt-in behavior because while it seems like only triggering on subpath modifications would be the correct behavior, it's actually trickier than that. Monorepos many times have shared components and an applications subpath likely won't include the shared component dir. Ergo it wouldn't deploy when shared components change, which some users will want.

internal/datasource/git.go
```diff
@@ -386,7 +417,141 @@ func (s *GitSource) Changes(
...
+	}
+
+	// Clone
+	var output bytes.Buffer
+	log.Trace("cloning repository", "url", source.Git.Url)
+	repo, err := git.PlainCloneContext(ctx, td, false, &git.CloneOptions{
+		URL:      source.Git.Url,
+		Auth:     auth,
+		Progress: &output,
+	})
```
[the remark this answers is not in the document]
mitchellh: Yeah, it is very big. The way runners are currently written, we can’t really because each job gets its own directory that is blown away at the end of the job. I can try to pull in some caching mechanism here in this PR though.

The other option if this is opt-in is to just accept this, because this is happening ANYWAYS right now (we’re triggering a superfluous `up` which is forcing the checkout) and so at worst this is just making the current behavior better.

There is a “part 2” to the issue this fixes where 10 projects tracking the same repo will poll the repo 10 times, and it’d be nicer to actually poll 1 repo 1 time and trigger changes in 10 projects at once. That is fairly significant so not this PR, but I can also use that as a way to do caching cause we can just cache the “poll repo” operation.

mitchellh:
> So I think this should be an opt-in behavior because while it seems like only triggering on subpath modifications would be the correct behavior, it's actually trickier than that. Monorepos many times have shared components and an applications subpath likely won't include the shared component dir. Ergo it wouldn't deploy when shared components change, which some users will want.

This was my thinking as well. I think the default behavior of “always trigger” is the safest behavior.

mitchellh:
I made this opt-in.

@hashicorp/waypoint-frontend Note there is a new setting here we probably want to expose in the UI. I'll bring it up in our meetings too.

.changelog/1821.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core: git poller ignores changes outside of the configured project path
```
briancain, line 2:
```suggestion
core: git poller optionally ignores changes outside of the configured project path
```

briancain (commented):
Nice! I think this should come with some docs too for where you set up git in a project? Maybe we could hold off and open a TODO issue since we'll wanna include how to expose the setting in the UI and CLI

denysvitali:
You should definitely include the new option here:
https://github.com/hashicorp/waypoint/blob/3e32c4e4070fa9d5cc6ee7df235b41bb6b94fd8b/internal/datasource/git.go#L610-L616

mitchellh:
> You should definitely include the new option here:

Good catch. Done!

.changelog/1821.txt
denysvitali: @mitchellh I guess you mistakenly reverted @briancain's suggestion :S

PR #1822, opened 2021-07-09
internal/server: harden git path validation

mitchellh:
This adds more validation to ensure that the path specified for a git
repository is what we expect. There are no bugs I've seen related to
this and users seem to be using subpaths fine, but working on #1821 made
me realize the need to be confident our paths are normalized.

### DIFF
--- a/.changelog/1822.txt
+++ b/.changelog/1822.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core: git datasource `path` field has more validation
+```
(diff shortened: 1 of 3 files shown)


krantzinator (approved):
Too minimal to write a test for?

mitchellh:
Nah, I'll write a test!

mitchellh:
The tests actually found a bug! It was in previously written code (so not this PR), but now this PR introduces new validation and fixes an old one. 😄

Issue #1827, opened 2021-07-12
ui: Invalid URL for Release


denysvitali:
**Describe the bug**

With the latest commit on `main` (currently 2b60ebb), the following behavior can be noticed:
![image](https://user-images.githubusercontent.com/4939519/125315786-c1db1900-e337-11eb-99ae-ac0300113e9b.png)

Clicking the link redirects the user to `https://http://some-url`, which is obviously wrong. This seems the only place where we're prepending the `https` protocol: every other URL (Release + Deployment) seems to be OK.

https://github.com/hashicorp/waypoint/blob/34c4d60eb6b53f6154c72fee76f331817167dcea/ui/app/components/app-card/release.hbs#L28-L29

https://github.com/hashicorp/waypoint/blob/ab573517e46184b1a21d25cf0196ea8027a35084/ui/app/components/latest-release-url/index.hbs#L3

https://github.com/hashicorp/waypoint/blob/34c4d60eb6b53f6154c72fee76f331817167dcea/ui/app/components/app-item/release.hbs#L24

**Steps to Reproduce**

1. Perform a release with `http://some-release-url`
2. Check the links in the Release tab and the Release buttons, you'll notice that the one mentioned here is the only one that's not working, the reason is that we're prepending `https://`

**Expected behavior**

Either `http://some-release-url` everywhere, or `https://http://some-release-url` everywhere (implementation error on Waypoint Server side?), but a mix of the two is definitely an UI issue

**Waypoint Platform Versions**

2b60ebb

**Additional context**
N/A

PR #1828, opened 2021-07-12
Fix release URL links

gregone:
Fixes https://github.com/hashicorp/waypoint/issues/1827

### DIFF
--- a/ui/app/components/app-card/release.hbs
+++ b/ui/app/components/app-card/release.hbs
@@ -25,8 +25,8 @@
   <:actions>
     {{#if @model.url}}
       <ExternalLink
-        href="https://{{@model.url}}"
-        title="Visit https://{{@model.url}}"
+        href={{@model.url}}
+        title="Visit {{@model.url}}"
         class="button button--primary app-card__launch-link"
       >
         <Pds::Icon @type="exit" class="icon" />
--- a/.changelog/1828.txt
+++ b/.changelog/1828.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: fix prepending release links with duplicate https:// protocol
+```


ui/app/components/app-card/release.hbs
```diff
@@ -25,7 +25,7 @@
   <:actions>
     {{#if @model.url}}
       <ExternalLink
-        href="https://{{@model.url}}"
+        href={{@model.url}}
         title="Visit https://{{@model.url}}"
```
koesbong, line 29: Should the title here be updated too without the `https://`?

jgwhite (approved):

jgwhite:
I wonder if `<AppCard::Deployment>` has the same issue.

denysvitali:
Uh! Yes it does :(

jgwhite:
I just tested this against waypoint+k8s and it seems like `deployment.preload.deploy_url` does *not* include the protocol, while `release.url` *does*. I found I kept forgetting which way around it was. Maybe it would be worth us writing an `ensure-protocol` helper, or making `ExternalLink` automatically add the protocol only if it’s missing?

```hbs
{{ensure-protocol "http://example.com"}} {{! => http://example.com }}
{{ensure-protocol "example.com"}}        {{! => https://example.com }}
```

denysvitali:
(I opened #1834 in the meantime)

IMHO, an object called `xyzUrl` should be an URL. That means that the protocol should be included.

I guess that we need to check and make sure that our plugins always return an URL rather than having an hackish way in the frontend to mitigate the issue.

cc/ @mitchellh, @krantzinator, @briancain  ?

jgwhite:
Now that `deployment.url` exists (thanks @denysvitali!), we should probably transition away from `deployment.preload.deploy_url` and use `deployment.url` directly instead.

denysvitali:
I'm not sure about that, I always though that the `deployment.preload.deploy_url` was for Horizon: but maybe we can remove this `preload` part and start using `deployment.url` directly as you mentioned, yes :)

jgwhite:
> I'm not sure about that, I always though that the `deployment.preload.deploy_url` was for Horizon: but maybe we can remove this `preload` part and start using `deployment.url` directly as you mentioned, yes :)

I guess this is another thing we should clarify. An explicit `horizon_url` would probably be clearer.

For the purposes of this PR, it’s worth nothing that the changes #1387 weren’t back-ported, so we’ll need to stick with `deployment.preload.deploy_url` for the purposes of this PR either way.

jgwhite:
> I guess that we need to check and make sure that our plugins always return an URL rather than having an hackish way in the frontend to mitigate the issue.

I agree but I also don’t think there’s any harm in adding the frontend mitigation too.

denysvitali:
This will solve the issues we're having in the short term, which seems a win-win situation. We can then start to migrate everything to use proper URLs and expect that past a certain Waypoint version, only URLs are accepted on that field - so that plugins have time to adapt.

gregone:
Are we still good to merge this one?

jgwhite:
@gregone yeah, let’s merge this.

PR #1830, opened 2021-07-12
Adds ability to launch one-off tasks with TaskLauncher

evanphx:
This is just the first bit of the work for TaskLaunch and ondemand runners. It builds on the functionality in https://github.com/hashicorp/waypoint-plugin-sdk/pull/42.

It's currently missing plumbing for StopTask, should will either be added to the PR or come in a second PR.


internal/plugin/factory.go
```diff
@@ -20,9 +21,17 @@ var exePath string
 
 func init() {
 	var err error
-	exePath, err = os.Executable()
-	if err != nil {
-		panic(err)
+	exp := os.Getenv("WAYPOINT_PLUGIN_PATH")
```
mitchellh, line 24: Is this something we expect people to generally use or was this for dev only? If its for dev only maybe we should rename this to not the `WAYPOINT_` prefix which makes it look official.

internal/plugin/invoke.go
```diff
@@ -0,0 +1,203 @@
...
+// Plugin is the state of a created plugin to be invoked, returned by OpenPlugin.
+type Plugin struct {
+	req      *PluginRequest
+	Instance *Instance
+}
+
+// OpenPlugin resolves the plugin information in req and returns a Plugin value
+// to have Invoke called upon it. OpenPlugin also returns the raw component
+// interface, which is used to get the function value that will be invoked
+func OpenPlugin(
```
mitchellh, line 107: Maybe just name this `Open`, since this is in the plugin package already: `plugin.Open` vs `plugin.OpenPlugin` the latter feels redundant.

internal/plugin/invoke.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if diags.HasErrors() {
+		return nil, nil, diags
+	}
+
+	if file.Body != nil {
+		component.Configure(pi.Component, file.Body, hclCtx.NewChild())
+	}
+
+	return &Plugin{req: req, Instance: pi}, pi.Component, nil
```
mitchellh, line 138: Feel like we should just return the `*Plugin, error` tuple and maybe introduce a helper `func (p *Plugin) Component()` to access `pi.Component` since that's common. That feels like it's more Go-ish.

internal/runner/operation_task.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	fn := c.(component.TaskLauncher).StartTaskFunc()
+
+	_, err = pi.Invoke(ctx, log, fn, tli)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pb.Job_Result{}, nil
```
mitchellh, line 58: This isn't passing state back yet, on purpose?

mitchellh (changes requested):
Some small changes but this looks very good.

Makefile
```diff
@@ -19,6 +19,10 @@ bin: # bin creates the binaries for Waypoint for the current platform
 	cd internal/assets && go-bindata -pkg assets -o prod.go -tags assetsembedded ./ceb
 	CGO_ENABLED=$(CGO_ENABLED) go build -ldflags $(GOLDFLAGS) -tags assetsembedded -o ./waypoint ./cmd/waypoint
 
+.PHONY: bin-dirty
+bin-dirty: # bin-dirty only recompiles waypoint
+	CGO_ENABLED=$(CGO_ENABLED) go build -ldflags $(GOLDFLAGS) -tags assetsembedded -o ./waypoint ./cmd/waypoint
```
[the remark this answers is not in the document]
evanphx: It's just to speed up testing, I can call it something else sure.

internal/plugin/factory.go
evanphx: I mean, it is useful to be able to control where the builtin plugins are executed from, in case you want to use a wrapper script or something.

internal/plugin/invoke.go
evanphx: Totes.

internal/plugin/invoke.go
evanphx: Yeah, I had that originally, but you have to dig it out as `p.Instance.Component` each time which during experimentation was really long given you HAVE to use that value to actually use the plugin interface. I also experiment with having `p.C()` ie, a C() function on Plugin to return the component. It felt cleaner to just return the value from Open.

internal/runner/operation_task.go
evanphx: Just incomplete. Finishing it now.

internal/plugin/factory.go
mitchellh: No problem, it was a genuine question, I didn't know.

In that case, I might recommend naming this `WAYPOINT_BUILTIN_PLUGIN_EXE` or something. I imagine at some point we'll drop the builtin plugins, but sitll have plugins, and `WAYPOINT_PLUGIN_PATH` sounds too general to me.

internal/plugin/invoke.go
mitchellh: Okay that's fine

go.mod
```diff
@@ -119,7 +119,7 @@ require (
 // development at the moment it is common to be working on these libs too.
 // replace github.com/hashicorp/go-argmapper => ../go-argmapper
 // replace github.com/hashicorp/horizon => ../horizon
-// replace github.com/hashicorp/waypoint-plugin-sdk => ../waypoint-plugin-sdk
+replace github.com/hashicorp/waypoint-plugin-sdk => ../waypoint-plugin-sdk
```
mitchellh, line 122: Don't forget to re-comment this prior to merge. I'm guessing CI tests will fail anyways to remind you.

internal/runner/operation_task.go
```diff
@@ -50,7 +52,60 @@ func (r *Runner) executeStartTaskOp(
...
+
+	stop := c.(component.TaskLauncher).StopTaskFunc()
+
+	msg, err := ptypes.AnyMessageName(op.StopTask.State)
+	if err != nil {
+		// This should never happen.
+		panic(err)
+	}
+
+	arg := argmapper.NamedSubtype("state", op.StopTask.State, msg)
```
mitchellh, line 112: Ah, there is a helper for this in `internal/core` I'm sure you found. I wonder if we should extract this. (internal/core/arg.go)

evanphx: Yup, that's where I pulled this code over from actually. Should we just export `argNamedAny` ?

internal/plugin/factory.go
evanphx: Sure, that sounds good.

internal/runner/operation_task.go
mitchellh: I think its weird for `core` to do this, makes a lot of sense imo to pull it into `internal/plugin` and export.

evanphx: I went ahead and exported it.

mitchellh: I think it still feels out of scope for the "core" package to do this and more in line with the goals of the "plugin" package.

(Thinking through someone reading through a `godoc` I'd think... why?)

evanphx: I can move it over to plugin, sure.

evanphx: Ok, moved it over to plugin.

internal/plugin/factory.go
```diff
@@ -21,9 +22,17 @@ var exePath string
 
 func init() {
 	var err error
-	exePath, err = os.Executable()
-	if err != nil {
-		panic(err)
+	exp := os.Getenv("WAYPOINT_BUILTIN_PLUGIN_EXE")
```
izaaklauer, line 25: Are you planning to document this?

evanphx: Yeah, we should.

mitchellh (approved):

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+		docs.Summary(
+			"A map of key/value pair(s), stored in docker as a string. Each key/value pair must",
+			"be unique. Validiation occurs at the docker layer, not in Waypoint. Label",
+			"keys are alphanumeric strings which may contain periods (.) and hyphens (-).",
+		),
+	)
+
+	doc.SetField(
+		"networks",
+		"An list of strings with network names to connect the container to.",
```
krantzinator, line 131:
```suggestion
		"A list of strings with network names to connect the container to.",
```

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+	)
+
+	doc.SetField(
+		"static_environment",
+		"environment variables to expose to the application",
+		docs.Summary(
+			"these environment variables should not be run of the mill",
+			"configuration variables, use waypoint config for that.",
+			"These variables are used to control over all container modes,",
+			"such as configuring it to start a web app vs a background worker",
```
krantzinator, line 155:
```suggestion
			"These variables are used to control all of a container's modes,",
			"such as configuring it to start a web app vs a background worker.",
			"These environment variables should not be common",
			"configuration variables normally set in `waypoint config`.",
```
Rearranged this a bit. I feel like the purpose should go first and the anti-purpose should be a clarifier. Also removed the slang term that, tbh, kind of confused me 😅

krantzinator: For ease of copy/pasting if you don't want to use `Commit suggestion` 😄 
```
			"These variables are used to control all of a container's modes,",
			"such as configuring it to start a web app vs a background worker.",
			"These environment variables should not be common",
			"configuration variables normally set in `waypoint config`.",
```

krantzinator (commented):
I didn't get all the way through this yet, but FYI some comment comments.

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+) error {
+	args := filters.NewArgs()
+	args.Add("reference", img)
+
+	// only pull if image is not in current registry so check to see if the image is present
+	// if force then skip this check
+	if !b.config.ForcePull {
+		sum, err := cli.ImageList(context.Background(), types.ImageListOptions{Filters: args})
+		if err != nil {
+			return fmt.Errorf("unable to list images in local Docker cache: %w", err)
```
briancain, line 181:
```suggestion
			return status.Errorf(codes.FailedPrecondition, "unable to list images in local Docker cache: %w", err)
```

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+			log.Info("reusing existing image for task", "image", img, "id", sum[0].ID)
+			return nil
+		}
+	}
+
+	img = makeImageCanonical(img)
+
+	out, err := cli.ImagePull(context.Background(), img, types.ImagePullOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to pull image: %w", err)
```
briancain, line 197:
```suggestion
		return status.Errorf(codes.FailedPrecondition, "unable to pull image: %w", err)
```

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+
+	// An array of strings with network names to connect the container to
+	Networks []string `hcl:"networks,optional"`
+
+	// Resources configures the resource constraings such as cpu and memory for the
+	// created containers.
+	Resources TaskResources `hcl:"resources,block"`
+
+	// Environment variables that are meant to configure the application in a static
+	// way. This might be control an image that has mulitple modes of operation,
```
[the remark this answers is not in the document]
evanphx:
```suggestion
	// way. This might be start an image in a specific mode,
```

briancain (changes requested):

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+			log.Info("reusing existing image for task", "image", img, "id", sum[0].ID)
+			return nil
+		}
+	}
+
+	img = makeImageCanonical(img)
+
+	out, err := cli.ImagePull(context.Background(), img, types.ImagePullOptions{})
+	if err != nil {
+		return status.Errorf(codes.FailedPrecondition, "unable to pull image: %w", err)
```
briancain, line 197:
```suggestion
		return status.Errorf(codes.FailedPrecondition, "unable to pull image: %s", err)
```

builtin/docker/task.go
```diff
@@ -0,0 +1,369 @@
...
+) error {
+	args := filters.NewArgs()
+	args.Add("reference", img)
+
+	// only pull if image is not in current registry so check to see if the image is present
+	// if force then skip this check
+	if !b.config.ForcePull {
+		sum, err := cli.ImageList(context.Background(), types.ImageListOptions{Filters: args})
+		if err != nil {
+			return status.Errorf(codes.FailedPrecondition, "unable to list images in local Docker cache: %w", err)
```
briancain, line 181:
```suggestion
			return status.Errorf(codes.FailedPrecondition, "unable to list images in local Docker cache: %s", err)
```

briancain (changes requested):

briancain (approved):
Nice! 🎉 🚀

Issue #1834, opened 2021-07-13
ui: Invalid URL for Deployment


denysvitali:
**Describe the bug**
Pretty identical to #1827, but for Deployments.
For some reason we are automatically prepending `https://` to some URLs, but we're not on some others:

https://github.com/hashicorp/waypoint/blob/b698cd270b295d09ad178ae3aae67d3de034d9af/ui/app/components/app-card/deployment.hbs#L26-L34

https://github.com/hashicorp/waypoint/blob/e1c8ae868e0ac8c48dfc71bf0fc152335938c50b/ui/app/templates/workspace/projects/project/app/deployment.hbs#L13-L16

We should make this consistent, either _with_ `https://` in front, or without. This also depends on the data model, but as the name suggests `deployUrl` and `@release.url` (for #1827) should already be URLs, hence `https://` in front shouldn't be needed.

**Steps to Reproduce**
Deploy an app that returns an URL like `https://some-url.com`: check the UI, you'll see `https://https://some-url` or `https://http://some-url` if it was an HTTP URL.

**Expected behavior**
Consistent behavior, either we treat the Deployment URL / Release URL as a 

**Waypoint Platform Versions**
5cc3a43

**Additional context**
Issue similar to #1827, so probably the twos should be tackled together

PR #1836, opened 2021-07-13
Add enforce-protocol helper

gregone:
Fixes https://github.com/hashicorp/waypoint/issues/1834 

Following https://github.com/hashicorp/waypoint/issues/1827 it seems that URL formatting is not consistent and correctly enforced on the backend. This PR adds a `enforce-protocol` helper to make sure protocols are always added to URLs before displaying them as links. 

This is an interim fix, and as discussed with @briancain yesterday, the URLs should be consistent on the server, and we *should not* be using this helper. 

here's the plan: 
- [ ] merge this and resume normal operations
- [ ] enforce URL consistency on the server in another PR
- [ ] remove the helper from the links
- [ ] profit

### DIFF
--- a/ui/app/helpers/enforce-protocol.ts
+++ b/ui/app/helpers/enforce-protocol.ts
@@ -1,0 +1,15 @@
+import { helper } from '@ember/component/helper';
+
+export function enforceProtocol(params/*, hash*/) {
+  let str = params[0];
+
+  let isHttps = str.indexOf('https://') !== -1;
+  let isHttp = str.indexOf('http://') !== -1;
+  if (isHttps || isHttp) {
+    return str;
+  } else {
+    return `https://${str}`;
+  }
+}
+
+export default helper(enforceProtocol);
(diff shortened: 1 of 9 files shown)


ui/app/helpers/enforce-protocol.ts
```diff
@@ -0,0 +1,15 @@
+import { helper } from '@ember/component/helper';
+
+export function enforceProtocol(params/*, hash*/) {
+  let str = params[0];
+
+  let isHttps = str.indexOf('https://') !== -1;
+  let isHttp = str.indexOf('http://') !== -1;
```
jgwhite, line 7: [sand] I’d suggest using [`String.prototype.startsWith`](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) for these.

```suggestion
  let isHttps = str.startsWith('https://');
  let isHttp = str.startsWith('http://');
```

jgwhite (approved):
Looking good! One [⏳sand](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/#sand-non-blocking-but-requires-future-consideration) suggestion but all good apart from that.

Issue #1287, opened 2021-04-07
ui: if a registry push fails, can't see that anywhere


mitchellh:
This is a weird one kind of. If you have a registry configuration in your build, then the push operation isn't shown anywhere in the UI and so you can't see the failure. In the video below, we're seeing a situation where the _build_ succeeded (Docker image built) but the _push_ failed (auth error). Internally, Waypoint tracks these as two separate operations, so I can see why this is happening, just not sure what we should do from a UX perspective.

https://user-images.githubusercontent.com/1299/113939341-bca68500-97b0-11eb-9a75-65cc45438293.mp4

Waypoint.hcl:

```
app "learn-waypoint-lambda" {
  build {
    use "docker" {}

    registry {
      use "aws-ecr" {
        region = "us-west-2"
        repository = "learn-waypoint-lambda"
        tag = "howdy-waypoint-lambda"
      }
    }
  }
}
```

mitchellh:
@gregone I think we can do this after 0.3.0, this feels a little more involved.

gregone:
@mitchellh sounds good. In fact, after working on the log output component yesterday, I realized we're filtering the output quite a bit, and I need to do a dive into the log format to clean up some things, so this would be a good issue to do so.

gregone:
https://github.com/hashicorp/waypoint/issues/1329

jgwhite:
I’m going to take a quick look at this in light of #1668, and the new way we approach status indicators. I don’t think that change will have brought any accidental fixes, but may give us more of a steer on where to go with this. Either way, I propose deferring this until a later milestone (given that we’re approaching 0.5 and should be clearing 0.4.x items).

jgwhite:
I’ve started to address this in #1840. It’s a significant change so I suggest targeting 0.5.0 or 0.5.x rather than 0.4.x, just to give us some more breathing room.

jgwhite:
#1840 is looking good so I think a fix for this will land in 0.4.x (probably 0.4.2)

jgwhite:
Closed by #1840

PR #1840, opened 2021-07-13
ui: Incorporate pushed artifacts into build display

jgwhite:
## Why the change?

Primarily to address #1287, though adding the foundations for artifact-awareness will likely be useful for other features.

## Could we load pushed artifacts in a more efficient manner?

“Yes, but not yet…”

We’re currently exploring the idea of adding dedicated aggregate API methods for aspects of the UI such as this. Watch this space. Until we have such a thing, this load-and-cross-reference approach is the only one available.

## What’s the plan?

- [x] Add changelog entry
- [x] Add pushed artifacts to Mirage
- [x] Refactor app route
- [x] Load pushed artifacts in app route
- [x] Incorporate push status into `<AppCard::Build>`
- [x] Incorporate push status into `<AppItem::Build>`
- [x] Fix capitalization of initialisms (i.e. AWS)
- [x] Add aws-ecr to icon-for-component
- [x] Write some test steps
- [x] Improve design

## What does it look like?

2-1-building
2-2-built
2-3-pushing
2-4-pushed
2-8-build-error
2-9-push-error

Obligatory video:

https://user-images.githubusercontent.com/34030/126233744-a8fdbf9b-8c62-4318-934e-d55e1ba432f4.mov

### Using Mirage

1. Check out the branch:
   ```sh
   git checkout ui/pushed-artifacts
   ```
2. Boot the dev server
   ```sh
   cd ui && ember serve
   ```
3. [Visit the app](http://localhost:4200/default/marketing-public/app/wp-matrix/builds)
4. Verify you see information about pushed artifacts in addition to builds
5. Verify that the UI makes sense

### For realsies

1. Check out the branch:
   ```sh
   git checkout ui/pushed-artifacts
   ```
2. Build the ember app
   ```sh
   (cd ui && make)
   ```
3. Bundle the static assets
   ```sh
   make static-assets
   ```
4. Build the dev server
   ```sh
   make docker/server
   ```
5. Build the CLI
   ```sh
   make bin
   ```
6. Install waypoint on your platform of choice, i.e.
   ```sh
   ./waypoint install -platform=kubernetes -accept-tos -k8s-server-image=waypoint:dev
   ```
7. Open the UI
   ```
   ./waypoint ui -authenticate
   ```
8. Try out an example app of your choice (i.e. [kubernetes/nodejs](https://github.com/hashicorp/waypoint-examples/tree/main/kubernetes/nodejs))
9. Verify that you see information about pushed artifacts in the UI
10. Trying triggering a failed push (see #1287 for an example)
11. Verify you see the overall state of the build is failed

### DIFF
--- a/ui/translations/en-us.yaml
+++ b/ui/translations/en-us.yaml
@@ -162,3 +162,18 @@ app_card_build:
         state-1: Pushing to # RUNNING
         state-2: Pushed to # SUCCESS
         state-3: Failed to push to # ERROR
+
+app_item_build:
+  label:
+    prefix:
+      type-1: # BUILDER
+        state-0: Building with # UNKNOWN
+        state-1: Building with # RUNNING
+        state-2: Built with # SUCCESS
+        state-3: Failed to build with # ERROR
+      type-2: # REGISTRY
+        state-0: Pushing to # UNKNOWN
+        state-1: Pushing to # RUNNING
+        state-2: Pushed to # SUCCESS
+        state-3: Failed to push to # ERROR
+  built_in: Built in {duration}
--- a/ui/translations/en-us.yaml
+++ b/ui/translations/en-us.yaml
@@ -148,3 +148,17 @@ status_report_indicator:
     checking_now: Checking now…
     last_checked: Last checked
     unknown: Status unknown
+
+app_card_build:
+  label:
+    prefix:
+      type-1: # BUILDER
+        state-0: Building with # UNKNOWN
+        state-1: Building with # RUNNING
+        state-2: Built with # SUCCESS
+        state-3: Failed to build with # ERROR
+      type-2: # REGISTRY
+        state-0: Pushing to # UNKNOWN
+        state-1: Pushing to # RUNNING
+        state-2: Pushed to # SUCCESS
+        state-3: Failed to push to # ERROR
--- a/ui/translations/en-us.yaml
+++ b/ui/translations/en-us.yaml
@@ -149,31 +149,17 @@ status_report_indicator:
     last_checked: Last checked
     unknown: Status unknown
 
-app_card_build:
-  label:
-    prefix:
-      type-1: # BUILDER
-        state-0: Building with # UNKNOWN
-        state-1: Building with # RUNNING
-        state-2: Built with # SUCCESS
-        state-3: Failed to build with # ERROR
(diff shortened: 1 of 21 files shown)


jgwhite:
@almonk there’s a bit of a design challenge with this change.

CleanShot 2021-07-13 at 16 20 29@2x

In this screenshot, we display both operations (build and push) in the build card and list items, and display the status of whichever is later (should be the push). The result is potentially a lot of icon duplication.

image

My gut says this is not terrible but rather inelegant and I suspect the layout probably falls down in some edge cases. Would love your take when you get a change to look at it.

jgwhite:
Idea from @almonk:

125468599-158ad068-b627-4f3c-b24e-650f76815440

jgwhite:
Reflecting on this, I think @almonk’s design suggestion make it realistic for this to land in `0.5.0`. Have updated milestones and project tracking accordingly.

ui/translations/en-us.yaml
```diff
@@ -148,3 +148,32 @@ status_report_indicator:
     checking_now: Checking now…
     last_checked: Last checked
     unknown: Status unknown
+
+app_card_build:
```
sabrinako, line 152: [sand] Is there a reason why we need to have the same text separated for the `app_card_build` and `app_item_build` (besides the additional `built_in` text)? Could we make it more generic like `app_build`?

jgwhite: We absolutely could. My thought here was that we might want to leave ourselves the option of having much terser text in `<AppCard::Build>` because there’s so much less space to play with.

jgwhite:
Just noticed that we could do with applying the same logic to the build detail page. Going to add that in here.

ui/translations/en-us.yaml
jgwhite: Thought about it some more and decided you were right. Applied your suggestion in f0aa30c08ce4caac23402f068cf00320b9530d1e

jgwhite:
> Just noticed that we could do with applying the same logic to the build detail page. Going to add that in here.

Addressed in 000f4940ec29ca859ae86782287a1f2bd7e357e1

gregone (approved):
Looking great!

sabrinako (approved):
Looks good! Saw the demo video

PR #1842, opened 2021-07-13
Backport of Feature: plugin debugging into release/0.4.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #1716 to be assessed for backporting due to the inclusion of the label backport/0.4.x.

---

It is currently difficult to debug the builtin plugins, as waypoint itself launches plugin processes as it requires them. Debugging is easiest when you can use a debugger to launch the process of interest. 

This experimental change allows Waypoint to launch builtin plugins in debug mode, and connect to and use an existing plugin processes. It is dependent on these waypoint-plugin-sdk changes: https://github.com/hashicorp/waypoint-plugin-sdk/pull/39

These changes are inspired by (and in many cases copied from) [Terraform's plugin debug mode](https://www.terraform.io/docs/extend/debugging.html#debugger-based-debugging). I've tried to keep the user facing side (flag/env var names, workflow, etc) as similar to Terraform as possible for consistency.

### Debugging workflow

Step 1: Launch a plugin process of interest (via delve or goland debug) with the new debug flag (i.e. `waypoint plugin -debug pack`). The plugin will launch, print reconnection info to the console (example below), then wait for core to attach.

```
Plugin started, to attach Waypoint set the WP_REATTACH_PLUGINS env var:

        WP_REATTACH_PLUGINS='{"pack":{"Protocol":"grpc","ProtocolVersion":1,"Pid":38972,"Test":true,"Addr":{"Network":"unix","String":"/var/folders/ns/grk8kk196_106v37w9hk8hxm0000gq/T/plugin864517971"}}}'

```

Step 2: Export that environment variable, then run a core command that will require that plugin (i.e. `waypoint build`). When initializing plugins, core will unmarshall and inspect the contents of the `WP_REATTACH_PLUGINS` env var. If it finds that a requested plugin is in that map, it will attempt to reconnect to it, rather than launching it itself. 

The flow is a little clunky for me - if you're iterating fast and relaunching the plugin process frequently, you have to do a lot of env var copying and pasting. It would be nice if the plugin could use a consistent socket path on every launch. I think this is good enough for now though, and consistency with Terraform is nice.

### Documentation

My intent is to use this feature myself for this cycle, and document it next cycle once it's been hardened.


briancain:
@izaaklauer Looks like the backport failed.

PR #1846, opened 2021-07-13
website: docs for input variables

krantzinator:
One of the links in here is dependent on a PR on the Learn side (https://github.com/hashicorp/learn/pull/3653) that will be merged when we release 0.5.0.

Since these are docs I'll let them speak for themselves. Let me know if something is unclear or you find a bug.


website/content/docs/waypoint-hcl/variables/input.mdx
```diff
@@ -0,0 +1,353 @@
...
+  define, reference, and use input variables.
+---
+
+# Input Variables
+
+Input variables let you customize aspects of Waypoint configuration with values
+that may change, or that need to be set per Waypoint run.
+
+You can parameterize your configuration by declaring variables in the
+waypoint.hcl, and setting their values using CLI options, environment variables,
```
mitchellh, line 16:
```suggestion
waypoint.hcl file, and setting their values using CLI options, environment variables,
```

website/content/docs/waypoint-hcl/variables/input.mdx
```diff
@@ -0,0 +1,353 @@
...
+`auto` files, and the Waypoint server UI.
+
+> **Hands-on:** Try the [Input Variables]() tutorial on HashiCorp Learn.
+
+## Declaring an Input Variable
+
+Each input variable referenced in the waypoint.hcl must be declared using a
+`variable` block:
```
mitchellh, line 24: I would link to the stanza reference documentation for variable here. I don't see that in this PR so I recommend adding that, example: https://www.waypointproject.io/docs/waypoint-hcl/app That should be the exhaustive reference of the syntax and available args. You can keep everything here, but that should exist too!

The idea there is anyone should be able to look at that sidebar and see every valid stanza in a waypoint file.

website/content/docs/waypoint-hcl/variables/input.mdx
```diff
@@ -0,0 +1,353 @@
...
+```
+
+The label after the `variable` keyword is a name for the variable, which must
+be unique among all variables in the same waypoint.hcl. This name is used to
+assign a value to the variable and to reference the variable's value when
+evaluating the configuration.
+
+The name of a variable can be any valid [identifier](/docs/waypoint-hcl/syntax#identifiers).
+
+## Arguments
```
mitchellh, line 56: You may not need this with the stanza reference docs... up to you. There'll be some duplication either way.

mitchellh (commented):
Looks good!

I noticed this is missing "setting variables in the UI". Do you want to include that too?

krantzinator:
@mitchellh yes I absolutely did want to include setting in the UI 😅 I included it in the Learn guide but forgot here, will do!

website/content/docs/waypoint-hcl/variables/input.mdx
```diff
@@ -0,0 +1,353 @@
...
+```
+
+If your intended value includes literal double quotes then you'll need to
+escape those with a backslash:
+
+```
+waypoint up -var "name=va\"lue"
+```
+
+PowerShell on Windows cannot correctly pass literal quotes to external programs,
```
briancain, line 237: Hm, not sure if this paragraph was taken from some where but you can totally quote within quotes on powershell: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.1 You should be able to escape a single quote with a backtick for powershell to not interpret it as "quotes".

briancain (changes requested):
Looks legit! 📝

website/content/docs/waypoint-hcl/variables/input.mdx
```diff
@@ -0,0 +1,353 @@
...
+# Input Variables
+
+Input variables let you customize aspects of Waypoint configuration with values
+that may change, or that need to be set per Waypoint run.
+
+You can parameterize your configuration by declaring variables in the
+waypoint.hcl file, and setting their values using CLI options, environment variables,
+`auto` files, and the Waypoint server UI.
+
+> **Hands-on:** Try the [Input Variables]() tutorial on HashiCorp Learn.
```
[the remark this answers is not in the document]
krantzinator: Yep, forgot! Thanks!

website/content/docs/waypoint-hcl/variables/input.mdx
krantzinator: I did lift this from other docs of ours. I think I'll just remove this sentence for now.

briancain (approved):

krantzinator:
@mitchellh you good with the updates?

mitchellh:
> @mitchellh you good with the updates?

Yep!

PR #1847, opened 2021-07-13
feat(internal): add project delete (hashicorp/waypoint#1418)

### LINKED ISSUE #1418: feature-request: Add `waypoint project destroy`
Currently, there seems to be no way to delete a project or a remote app (= easily without creating an hcl file locally).

It would be great to be able to delete projects from the UI / CLI and maybe being able to do so _per workspace_.

Destroying the Waypoint DB and restarting from scratch - although it doesn't seem like a good alternative for production 😄 

- Being able to delete certain apps from certain workspaces
- Being able to delete certain projects from certain workspaces
- Being able to delete all projects from a certain workspace

**Additional context**
None

denysvitali:
Fixes #1418

```
waypoint project -h
Usage: waypoint project SUBCOMMAND

  Project management.

  Projects are comprised of one or more applications. A project maps
  to a VCS repository (if one exists).

Subcommands:
    apply     Create or update a project.
    delete    Delete the specified project.
    detail    Show details for the specified project.
    list      List all registered projects.
```

```
waypoint project delete -h
Usage: waypoint project delete PROJECT-NAME

  This command deletes a project entirely
```

### Project exists

```
$ waypoint project delete -force waypoint-1234
  project waypoint-1234 deleted successfully
```

### Project doesn't exist

```
$ waypoint project delete -force waypoint-1234
! unable to delete project: the project doesn't exist
```

**Note:** for now `-force` is always needed, the reason is that waypoint-plugin-sdk still doesn't have a valid terminal that accepts user input (or I wasn't able to find one)


internal/cli/project_delete.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (c *ProjectDeleteCommand) Flags() *flag.Sets {
+	return c.flagSet(0, func(sets *flag.Sets) {
+		f := sets.NewSet("Command Options")
+		f.BoolVar(&flag.BoolVar{
+			Name:    "force",
+			Target:  &c.flagForce,
+			Default: false,
+			Usage:   "skip the confirmation prompt",
+		})
```
briancain, line 85: Instead of `force`, we are using `auto-approve` for these kinds of behaviors. You can see one example in Waypoints server upgrade:

https://github.com/hashicorp/waypoint/blob/6e6dbd95405adfdf6ee6b75b1f9ce8d014c08ced/internal/cli/server_upgrade.go#L385-L390

internal/cli/project_delete.go
```diff
@@ -0,0 +1,108 @@
...
+	return c.Flags().Completions()
+}
+
+func (c *ProjectDeleteCommand) Synopsis() string {
+	return "Delete the specified project."
+}
+
+func (c *ProjectDeleteCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint project delete PROJECT-NAME
```
briancain, line 103:
```suggestion
Usage: waypoint project delete PROJECT-NAME [options]
```

internal/cli/project_delete.go
```diff
@@ -0,0 +1,108 @@
...
+
+  This command deletes a project entirely.
+
```
briancain, line 106: We should print the flags here too. If not here, where we print help in the `Run()`

internal/server/singleprocess/service_project.go
```diff
@@ -48,6 +47,25 @@ func (s *service) GetProject(
 	}, nil
 }
 
+// DeleteProject processes a DeleteProjectRequest and deletes the requested project
+// TODO: test
```
briancain, line 51: We could use a test for this please before merge! 🙏🏻

internal/cli/project_delete.go
```diff
@@ -0,0 +1,108 @@
...
+	if len(args) != 1 {
+		c.ui.Output(c.Help(), terminal.WithErrorStyle())
+		return 1
+	}
+	name := args[0]
+
+	// TODO: Use the Interactive terminal, when it's implemented
+	if !c.ui.Interactive() && !c.flagForce {
+		c.ui.Output(
+			"operation interrupted, terminal is not interactive and force not provided",
```
briancain, line 38: I don't think it's worth mentioning an interactive terminal yet until we actually support it. So I'd say let's split out these two error message separately. Then the auto-approve error message could read similar to how we've handled this before:

https://github.com/hashicorp/waypoint/blob/6e6dbd95405adfdf6ee6b75b1f9ce8d014c08ced/internal/cli/server_upgrade.go#L456-L457

```
Deleting a project requires confirmation. 
Rerun the command with '-auto-approve' to continue with deleting the project.
```

internal/cli/project_delete.go
```diff
@@ -0,0 +1,108 @@
...
+		c.ui.Output(fmt.Sprintf("unable to delete project: %v", err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	if !resp.Successful {
+		c.ui.Output("unable to delete project: the project doesn't exist", terminal.WithErrorStyle())
+		return 1
+	}
+
+	c.ui.Output(fmt.Sprintf("project %s deleted successfully", name), terminal.WithInfoStyle())
```
briancain, line 73:
```suggestion
	c.ui.Output(fmt.Sprintf("project %s deleted successfully", name), terminal.WithSuccessStyle())
```

briancain (changes requested):
Thanks @denysvitali! Haven't tried this yet, but I have a few comments on the code itself 😄

mitchellh (commented):
I don't think we can accept this as-is.

The reason we don't support delete today is because it leaves a bunch of dangling data in the database. We don't currently support a way to do a cascading delete. And that data will never be reclaimed in any way. It could be argued you want to keep that data in some cases, but in its current form its effectively just a memory leak.

The roadmap for delete has some ideas for how we can implement cascading deletes on top of BoltDB that we currently use.

internal/cli/project_delete.go
denysvitali: Oh yes, like an inverse dry-run. I like it!

denysvitali:
I agree with you @mitchellh , maybe I should try to tackle the cascade deletion too - what are your ideas there?

Should I maybe convert the PR to a draft again until we sort out the business logic part of the ProjectDelete method?

mitchellh:
> I agree with you @mitchellh , maybe I should try to tackle the cascade deletion too - what are your ideas there?

The only VERY rough idea I have is to use reflection with the `Ref.` types (in proto) to build up cascades automatically. I dunno if thats possible, but that was the line I was going to investigate.

I think it'll be tough and time consuming (my guess) and its something I was going to tackle at some point eventually, if you wanted to wait. If you want to give it a shot, that's fine too!

> 
> Should I maybe convert the PR to a draft again until we sort out the business logic part of the ProjectDelete method?

I'd rather close this for now (its always available in history to reopen if we wanted).

denysvitali:
Closed, thanks!

PR #1848, opened 2021-07-13
internal/cli: waypoint login supports tokens

mitchellh:
This builds on `waypoint login` introduced in #1831 for OIDC to provide a much nicer token-based login experience.

Prior to this, users had to use `waypoint context create` and specifically had to use a _login_ token with that. If they had an _invite_ token, they first had to convert it using `waypoint token exchange`. This is all friction and not obvious to users (though the underlying machinery is important).

Users can now use `waypoint login` with just the server address and ANY token and the CLI automatically does the right thing. It checks if the token is an invite token and automatically exchanges it. It also infers all the TLS and server connection settings based on the hostname (same as OIDC).

## Demo

https://user-images.githubusercontent.com/1299/125501910-bd673d5f-7221-43a3-a732-057e2210f2bf.mp4

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -78,6 +78,12 @@ func (c *LoginCommand) Run(args []string) int {
 	// which is already configured with the basic server connection stuff
 	// from this command.
 	newContext := c.flagConnection
+	if c.clientContext != nil {
+		// clientContext is always set to our actual context we used to
+		// create our client. So this will accurately grab non-flag based
+		// access i.e. loading our default context.
+		newContext = *c.clientContext
+	}
 	newContext.Server.AuthToken = token
 	newContext.Server.RequireAuth = true
 
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -285,7 +285,7 @@ Usage: waypoint login [server address]
 const (
 	errNoAuthMethods = `
 Only token-based authentication is allowed by this server. To login using
-a token, use the "waypoint context create" command.
+a token, use the "waypoint login" command with the "-token" flag.
 `
 
 	errManyAuthMethods = `
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -24,6 +24,7 @@ type LoginCommand struct {
 	*baseCommand
 
 	flagAuthMethod string
+	flagToken      string
 }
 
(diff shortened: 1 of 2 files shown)


internal/cli/login.go
```diff
@@ -53,6 +64,12 @@ func (c *LoginCommand) Run(args []string) int {
 	// which is already configured with the basic server connection stuff
 	// from this command.
 	newContext := c.flagConnection
+	if c.clientContext != nil {
+		// clientContext is always set to our actual context we used to
+		// create our client. So this will accurately grab non-flag based
+		// access i.e. loading our default context.
+		newContext = *c.clientContext
```
mitchellh, line 71: This also fixes a bug with OIDC, too.

internal/cli/login.go
```diff
@@ -73,6 +90,41 @@ func (c *LoginCommand) Run(args []string) int {
...
+	})
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return "", 1
+	}
+	token := decodeResp.Token
+
+	// If we have a login token, then we're just done cause that can be stored directly.
+	if _, ok := token.Kind.(*pb.Token_Login_); ok {
+		return c.flagToken, 0
```
krantzinator, line 107: Super yak shave-y question, but wouldn't `token` here and on line 188 make more semantic sense?

krantzinator (approved):
Massive QOL improvement 🚀

internal/cli/login.go
mitchellh: `token` is the full `pb.Token` structure, not the string-encoded token 😄

krantzinator: Ah ok, so is the `convertResp.Token` on line 125 a different type then than the `decodeResp.Token`?

krantzinator: I'm not familiar with the larger OIDC picture yet 🙈 Tomorrow! hah

mitchellh: Yeah haha. This is just the token system APIs so no need to know OIDC. `decodeResp` decodes a string token into a `*pb.Token` and Convert just returns the string token (converted from an invite token).

mitchellh:
Merging this since it fixes a bug with OIDC!

Issue #1510, opened 2021-05-19
It is difficult to determine where to put the `-workspace` flag during `waypoint deployment destroy`


izaaklauer:
**Describe the bug**
It is not obvious how to specify a workspace when destroying a deployment on the CLI.

`waypoint deployment destroy <number> -workspace=<workspace>` treats the workspace flag as a deployment ID, and deletes nothing:
```
! value with given id not found: -workspace=test
```

`waypoint -workspace=<workspace> destroy <number>` gives this error, which _does_ hint at the correct answer:

```
Invalid flags before the subcommand. If these flags are for
the subcommand, please put them after the subcommand.
```

The correct answer is `waypoint destroy -workspace=<workspace> <number>`, which does make some sense upon reflection but is also the last combination I would have considered.

**Expected behavior**
It would be great if the flag could work anywhere. 

Barring that, it would be nice if it could work before _or_ after the destroy args. It seems like we could handle this case in [the command itself](https://github.com/hashicorp/waypoint/blob/main/internal/cli/deployment_destroy.go#L43), but it feels to me like something that the flag parser should be doing for us.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: 0.3.2
* Waypoint Server Platform and Version: n/a
* Waypoint Plugin: n/a

**Additional context**
I don't see any other commands where we take any number of arguments _and_ allow subcommand flags, but it's bound to happen somewhere else someday.

krantzinator:
We do have an order specified under the `help`, do you think we could enhance this a bit to be more obvious?
```
$ waypoint deployment destroy -h
Usage: waypoint deployment destroy [options] [id...]
```

mitchellh:
Note that underlying Waypoint (and all HC products) we use the Go `flag` stdlib. This forces the order of `[flags] [args]`. So the only valid place is in where the help noted by @krantzinator is.

You CANNOT do: `waypoint -workspace foo deployment destroy` or `waypoint deployment -workspace foo destroy` or `waypoint deployment destroy foo -workspace foo`. You HAVE to do `waypoint deployment destroy -workspace foo foo`

evanphx:
The PR to close this should just he to enhance the help text to guide the user to the proper syntax. Perhaps the `Invalid flags before subcommand` text needs to be highlight more and/or provide an example output.

evanphx:
A larger, organizational discussion point would be "should HC products support more flexible option positions."

[...]

PR #1849, opened 2021-07-13
internal/cli: detect flags after positional args and error

mitchellh:
This introduces a heuristic for detecting flag usage after positional
arguments. This is an error for the stdlib flag package that we use.
This has come up many times before (see #1510) and I believe this will
make the experience smoother.

**Note:** I plan on writing tests for this check function, but want to gauge interest.

Example:

CleanShot 2021-07-13 at 11 45 28@2x

### DIFF
--- a/.changelog/1849.txt
+++ b/.changelog/1849.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: detect flags after arguments and show user-friendly error
+```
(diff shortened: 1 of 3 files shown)


krantzinator:
I like it.

izaaklauer (approved):
Actually closes https://github.com/hashicorp/waypoint/issues/1510. Thank you!

mitchellh:
Added tests, will merge this.

PR #1863, opened 2021-07-14
UI: Default URLS for details pages use sequence #

### LINKED ISSUE #1513: UI and CLI should agree on which representation of ID or Seq Number is displayed

sabrinako:
Additional tweaks for #1513

### Notes
- The current URL format for artifact detail pages (`.../[deployment OR build OR release]/[ULID]`) still works (to support users who may have bookmarked these pages and also users who are more interested in using ULIDs than sequence #s)
- The new format is `.../[deployment OR build OR release]/seq/[SEQUENCE]` to differentiate what identifier is being used

### DIFF
--- a/ui/app/routes/workspace/projects/project/app/build.ts
+++ b/ui/app/routes/workspace/projects/project/app/build.ts
@@ -29,11 +29,13 @@ export default class BuildDetail extends Route {
   async model(params: BuildModelParams) {
     // Setup the build request
     let { builds } = this.modelFor('workspace.projects.project.app');
-    let { id: build_id } = builds.find((obj) => obj.sequence == params.sequence);
+    let { id: build_id } = builds.find((obj) => obj.sequence === Number(params.sequence));
+
     let ref = new Ref.Operation();
     ref.setId(build_id);
     let req = new GetBuildRequest();
     req.setRef(ref);
+
     let build = await this.api.client.getBuild(req, this.api.WithMeta());
     return build.toObject();
   }
--- a/ui/app/routes/workspace/projects/project/app/build.ts
+++ b/ui/app/routes/workspace/projects/project/app/build.ts
@@ -5,7 +5,7 @@ import { Ref, GetBuildRequest } from 'waypoint-pb';
 import { AppRouteModel } from '../app';
 
 interface BuildModelParams {
-  build_id: string;
+  sequence: number;
 }
 export default class BuildDetail extends Route {
   @service api!: ApiService;
@@ -28,11 +28,12 @@ export default class BuildDetail extends Route {
 
   async model(params: BuildModelParams) {
     // Setup the build request
+    let { builds } = this.modelFor('workspace.projects.project.app');
+    let { id: build_id } = builds.find((obj) => obj.sequence == params.sequence);
     let ref = new Ref.Operation();
-    ref.setId(params.build_id);
+    ref.setId(build_id);
     let req = new GetBuildRequest();
     req.setRef(ref);
-
     let build = await this.api.client.getBuild(req, this.api.WithMeta());
     return build.toObject();
   }
--- a/ui/app/routes/workspace/projects/project/app/deployment.ts
+++ b/ui/app/routes/workspace/projects/project/app/deployment.ts
@@ -5,7 +5,7 @@ import { GetDeploymentRequest, Deployment, Ref, StatusReport } from 'waypoint-pb
 import { AppRouteModel, ResolvedModel as ResolvedAppRouteModel } from '../app';
 
 interface DeploymentModelParams {
-  deployment_id: string;
+  sequence: number;
 }
 
 interface Breadcrumb {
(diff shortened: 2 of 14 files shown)


ui/app/routes/workspace/projects/project/app/build.ts
```diff
@@ -28,11 +28,12 @@ export default class BuildDetail extends Route {
 
   async model(params: BuildModelParams) {
     // Setup the build request
+    let { builds } = this.modelFor('workspace.projects.project.app');
+    let { id: build_id } = builds.find((obj) => obj.sequence == params.sequence);
```
jgwhite, line 32: [sand] Let’s use threequals here just to avoid any javascript typecasting funny business.

```suggestion
    let { id: build_id } = builds.find((obj) => obj.sequence === params.sequence);
```

ui/app/routes/workspace/projects/project/app/build.ts
```diff
@@ -28,11 +28,12 @@ export default class BuildDetail extends Route {
...
     let ref = new Ref.Operation();
-    ref.setId(params.build_id);
+    ref.setId(build_id);
     let req = new GetBuildRequest();
     req.setRef(ref);
-
     let build = await this.api.client.getBuild(req, this.api.WithMeta());
     return build.toObject();
```
jgwhite, line 38: This approach is totally great and very pragmatic 👍 

[dust] For completeness (and apologies if you’re already aware of this) it’s also possible to use `GetBuild` with a sequence number. It’d look something like this:

```ts
async model(params: BuildModelParams): Promise<Build.AsObject> {
  let { sequence } = params;
  let { project_id: projectName } = this.paramsFor('workspace.projects.project') as ProjectRouteParams;
  let { app_id: appName } = this.paramsFor('workspace.projects.project.app') as AppRouteParams;

  let req = new GetBuildRequest();
  let ref = new Ref.Operation();
  let appRef = new Ref.Application();
  let seqRef = new Ref.OperationSeq();

  appRef.setProject(projectName);
  appRef.setApplication(appName);

  seqRef.setApplication(appRef);
  seqRef.setNumber(sequence);

  ref.setSequence(seqRef);

  req.setRef(ref);

  let build = await this.api.client.getBuild(req, this.api.WithMeta());

  return build.toObject();
}
```

This is true of `GetDeployment` and `GetRelease` too. Let’s not use this now, but it’s useful to know it’s available should we need it in future.

ui/app/routes/workspace/projects/project/app/deployment.ts
```diff
@@ -38,8 +38,10 @@ export default class DeploymentDetail extends Route {
   }
 
   async model(params: DeploymentModelParams): Promise<Deployment.AsObject> {
+    let { deployments } = this.modelFor('workspace.projects.project.app');
+    let { id: deployment_id } = deployments.find((obj) => obj.sequence == params.sequence);
```
jgwhite, line 42: [sand] Same threequals comment:

```suggestion
    let { id: deployment_id } = deployments.find((obj) => obj.sequence === params.sequence);
```

jgwhite (approved):
I love this approach. So elegant and pragmatic, and works a charm.

I’ve left a couple of [sand] and [dust] comments (see the [Netlify feedback ladder](https://www.netlify.com/blog/2020/03/05/feedback-ladders-how-we-encode-code-reviews-at-netlify/)). Feel free to take-or-leave them.

ui/app/routes/workspace/projects/project/app/build.ts
sabrinako: Used the `==` because the threequals was not passing. Looked into the reason why that happened a bit more, and for some reason the parameter's `sequence` value gets converted to a string at some point in the process of coming down to the build router (even though I have the interface saying it should be a number), any ideas why that would happen?

ui/app/routes/workspace/projects/project/app/build.ts
sabrinako: Didn't realize this, or maybe I found that out before I could grab values from parent components and then forgot when implementing here. Good to know in the future though!

ui/app/routes/workspace/projects/project/app/build.ts
jgwhite: Ahhh, very interesting! I actually think this is a great example of where the implicit type-coercion of `==` can lead to trouble. Without `===`, we wouldn’t have realized that we were dealing with mixed types, and might have wound up with some weird coercion bugs.

As to why the parameter comes through as a string rather than a number — I think that’s the only permitted behavior of Ember’s router. I _think_ it’s because that parameter originates in the URL, so therefore it starts off as a string and should really stay a string unless we do some explicit conversion.

So, TL;DR I think having an explicit type conversion here would be the ideal. Something like:

```ts
obj.sequence === Number(params.sequence)
```

PR #1867, opened 2021-07-15
internal/clicontext: parse 127.0.0.1:1234 correctly for server addr

mitchellh:
We use net/url.parse which requires a valid URL. We do some sniffing to
try to turn common values into valid URLs. This is backed up by tests.


izaaklauer (commented):
👍

PR #1868, opened 2021-07-15
Add logging for log streaming, potentially fix infinite loop case.

mitchellh:
We were creating a new "ws" variable meaning that we'd enter an infinite
loop since the last watchset would constantly refire. This is super
subtle not sure how to make it safer.

I added a bunch more logging as well so that if this doesn't fix it we can
better trace what's going on.

### DIFF
--- a/internal/server/singleprocess/service_logs.go
+++ b/internal/server/singleprocess/service_logs.go
@@ -168,6 +168,7 @@ func (s *service) GetLogStream(
 		// This flag is set when we create the Deployment value by detecting if the plugin
 		// had a LogsFunc defined.
 		if deployment.HasLogsPlugin {
+			log.Debug("deployment supports log plugin. spawning log plugin")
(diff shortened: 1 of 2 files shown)


internal/server/singleprocess/service_logs.go
```diff
@@ -468,12 +478,12 @@ func (s *service) sendInstanceLogs(
 		}
 
 		// Get all current records
-		ws := memdb.NewWatchSet()
+		ws = memdb.NewWatchSet()
```
mitchellh, line 481: This is the key change.

evanphx: AAAAH

evanphx (approved):

briancain (approved):
Gave this a spin 3 times with 3 different projects, didn't seem to reproduce it! I think given the fix and the behavior described this should fix it.

izaaklauer (approved):

PR #1872, opened 2021-07-15
AWS ECS: validate memory + cpu pairs

catsby:
AWS ECS / Fargate require the memory and cpu values for containers to be agreeable, and will error when trying to create a container or task definition if they do not match the prescribed values. See this article for more information:

- https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html

The Waypoint ECS deployment already validated these to a point, but the ECS Server install did not. This PR refactors that validation into a shared utility and uses it for both the platform and server install.

UX for using bad memory value in an ECS deploy:

Screen Shot 2021-07-15 at 16 19 17

UX for using bad memory / cpu values for server install:

Screen Shot 2021-07-15 at 16 49 04

### DIFF
--- a/.changelog/1872.txt
+++ b/.changelog/1872.txt
@@ -1,0 +1,7 @@
+```release-note:bug
+serverinstall/ecs: validate memory and cpu values
+```
+
+```release-note:bug
+plugin/aws/ecs: validate memory and cpu values
+```
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1065,9 +1065,6 @@ func (p *Platform) Launch(
 		runtime = aws.String("EC2")
 		cpuShares = p.config.CPU
 	} else {
-		// if p.config.Memory == 0 {
-		// 	return nil, fmt.Errorf("memory value required for fargate")
-		// }
 		if err := utils.ValidateEcsMemCPUPair(p.config.Memory, p.config.CPU); err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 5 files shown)


mitchellh (approved):
I can't believe this is the right way to do this but I've seen this done in Terraform and so many other places that it just must be. LOL.

builtin/aws/ecs/platform.go
```diff
@@ -1063,58 +1056,29 @@ func (p *Platform) Launch(
 	L.Debug("registering task definition", "id", id)
 
 	var cpuShares int
+	family := "waypoint-" + app.App
+
+	s.Status("Registering Task definition: %s", family)
```
catsby, line 1061: Note: I moved this because it's the first status update in this method, and without moving it (or adding another one) when the validation fails it looks like the previous step failed, and not this one

.changelog/1872.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
+serverinstall/ecs: validate memory and cpu values
+```
+
+```release-note:bug
+plugin/aws/ecs: validate memory and cpu values
+```
```
evanphx, line 7: I think these should be in one block, mostly because I think go-changelog only supports one block per file?

evanphx (approved):
Looks good, just a question about the changelog file.

.changelog/1872.txt
briancain: You can have multiple entries in a single go-changelog txt file! https://github.com/hashicorp/waypoint/blob/main/.github/CHANGELOG_GUIDE.md#multiple-entries

evanphx: bitmoji

PR #1873, opened 2021-07-15
internal/core: Add unimplemented field for a release 

### LINKED ISSUE #1478: Release Operation always returns proto message even if platform has no releaser configured
**Describe the bug**

I'm not 100% sure if the decision to return a Release proto was intentional, but I believe this might be a cause of confusion and potentially a bug at the moment for the UI and release CLI command. There might be some logic we aren't fully checking which causes some confusion behavior. I'll try my best to explain what I think is going on and why the behavior might be occuring!

The [operation_release job](https://github.com/hashicorp/waypoint/blob/main/internal/runner/operation_release.go#L139-L143) always returns a Release proto message, even [if the requested platform doesn't implement a release](https://github.com/hashicorp/waypoint/blob/main/internal/core/app_release.go#L60-L64). This makes sense since the operation is returning the top level Release proto message, but it looks like the UI still shows that a "release" happens with a SHA. It's also possible there are other parts of the code that make this assumption that a release occurs, like with the `waypoint release` CLI for example:

```
# release an old release (docker platform, with no releaser implemented)

$ waypoint release 01F5XKRRCRRCZ2RWA8W820X3D8
» Releasing...
» Pruning old deployments...
  Deployment: 01F5XKRRCRRCZ2RWA8W820X3D8
Deployment 01F5XMMCEDGE8S3ZMGC9R42P60 marked as released.
No release manager was configured and the configured platform doesn't
natively support releases. This means that releasing doesn't generate any
public URL. Waypoint marked the deployment aboved as "released" for server
history and to prevent commands such as "waypoint destroy" from deleting
the deployment by default.
```

Then try to release a different release after a few `waypoint deploy`s:

```
# 'foo' could also be another recent release after the initial successful release

$ waypoint release foo
The deployment "01F5XN9V4318GS5ZFSWXYWAB6X" is already the released deployment. Nothing to be
done.
```

Additionally, as a Waypoint dev, determining if a release actually happened on the result returned by the release operation requres this check:

```golang
if releaseResult != nil && releaseResult.Release != nil && releaseResult.Release.Release != nil {}
``` 

Which makes me wonder if there are other places that assume if `releaseResult` is not nil, a release occured.

The ultimate question here is:

Should the release operation be returning a Release proto message if no releaser was configured? Should it include a higher level option in the proto message that marks the release as skipped or unimplemented?

**Steps to Reproduce**

Look at release result on a platform with no releaser, like docker.

**Expected behavior**

Unclear, I can see returning a Release proto message as being intentional, even if no releaser was configured. Maybe we need to be more explicit when checking for releases beyond if the release result was nil or empty.

n/a, I assume this has been this way for a while but it's at least on `main` and the latest release 0.3.2.

**Additional context**

briancain:
Because Waypoint always stores a release message even if the release
was unimplemented, certain actions will behave poorly if it assumes
that getting the latest release means a release was performed. This
commit adds a new field to the release, Unimplemented, which will be
marked if the app_release determines if the plugin does not support
releasers.

Fixes #1478


internal/runner/operation_up.go
```diff
@@ -93,6 +93,10 @@ func (r *Runner) executeUpOp(
 	}
 	releaseResult := result.Release
 
+	if releaseResult.Release.Unimplemented {
+		app.UI.Output("Release not supported for platform, skipping...")
```
evanphx, line 97: I think this will happen even if the platform does support releasing but just doesn't have a default releaser AND the user didn't define one.

As such, I'd probably make the message `No release phase specified, skipping...`

evanphx (changes requested):
Fine, just want to adjust the messaging the UI outputs.

evanphx (approved):

internal/server/proto/server.proto
```diff
@@ -2680,6 +2680,11 @@ message Release {
   // ID of the job that created this. This may be empty.
   string job_id = 13;
 
+  // A boolean to mark this release message as unimplemented by the plugin
+  // that generated the message. If true, that means a release was never
+  // executed and Waypoint failed to build a valid releaser.
+  bool unimplemented = 15;
```
mitchellh, line 2686: I also will nitpick this comment to be a little more clear on how Waypoint works: a `Release` is _always_ made, but this means that there was no actual logic run, this was just automatically created.

briancain: Ah, good point! Important distinction to make.

mitchellh (approved):

internal/core/app_release.go
```diff
@@ -58,10 +58,12 @@ func (a *App) Release(ctx context.Context, target *pb.Deployment) (
 			"err", err)
 	}
 
+	unimplemeneted := false
 	c, err := a.createReleaser(ctx, &evalCtx)
 	if status.Code(err) == codes.Unimplemented {
 		c = nil
 		err = nil
+		unimplemeneted = true
```
krantzinator, line 66: typo 😱 
😁

briancain: Good catch!! Will be fixed via https://github.com/hashicorp/waypoint/pull/1801/commits/a9ef3f27783b971fc08e08f00dd0b64347e212be in the app poll feature PR 🙏🏻

PR #1876, opened 2021-07-15
Backport of internal/core: Add unimplemented field for a release  into release/0.4.x

### LINKED ISSUE #1478: Release Operation always returns proto message even if platform has no releaser configured

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #1873 to be assessed for backporting due to the inclusion of the label backport/0.4.x.

---

Because Waypoint always stores a release message even if the release
was unimplemented, certain actions will behave poorly if it assumes
that getting the latest release means a release was performed. This
commit adds a new field to the release, Unimplemented, which will be
marked if the app_release determines if the plugin does not support
releasers.

Fixes #1478


briancain:
Hey @izaaklauer -

I think there was an issue with your previous backport. I'm getting these errors when I try to generate the server protos:

```
brian@localghost:waypoint λ make gen/server                                                                                                                                                                              ±[●●][backport/release/0.4.x/unimpl-releasse]
go generate ./internal/server
google/rpc/status.proto: File not found.
internal/server/proto/server.proto:10:1: Import "google/rpc/status.proto" was not found or had errors.
internal/server/proto/server.proto:575:3: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:709:3: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:1029:7: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:1036:7: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:824:3: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:1417:5: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:1424:5: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:1522:5: "google.rpc.Status" is not defined.
internal/server/proto/server.proto:2259:12: "DeclaredResource" is not defined.
internal/server/proto/server.proto:2789:5: "google.rpc.Status" is not defined.
internal/server/server.go:12: running "sh": exit status 1
Makefile:81: recipe for target 'gen/server' failed
make: *** [gen/server] Error 1
```

Notice this one:

```
internal/server/proto/server.proto:2259:12: "DeclaredResource" is not defined.
```

There isn't a DeclaredResource field defined now.

Any idea what's wrong?

briancain:
Fixed in https://github.com/hashicorp/waypoint/pull/1877

PR #1878, opened 2021-07-15
OIDC Claim Mapping and Access Control

mitchellh:
This allows operators to configure a test for allowing access via an OIDC auth method using OIDC claims. For example, we can choose to allow only GitLab users that are part of the "hashicorp" group/organization like this: `hashicorp in list.groups`.

This fixes an issue where certain public OIDC providers -- such as GitLab -- by default allow _anyone_ with an account to authenticate. For something like Waypoint that's certainly not what we want!

Custom claims usually come in the form of user info claims and not ID token claims, so this also adds to the Waypoint server the ability to request the user info claims. This happens automatically/behind the scenes as part of the auth.

## Demo

(Secrets are safe to share, I already revoked them)

https://user-images.githubusercontent.com/1299/125871885-48bb0556-5304-4380-b617-6ac9e1eca00f.mp4


izaaklauer (changes requested):
Looks good! 

You'll need to add the new user commands to `waypoint/website/data/commands-nav-data.json`, or the docs won't render. This blob worked for me:

```
  {
    "title": "user inspect",
    "path": "user-inspect"
  },
  {
    "title": "user invite",
    "path": "user-invite"
  },
  {
    "title": "user modify",
    "path": "user-modify"
  },
  {
    "title": "user token",
    "path": "user-token"
  },
```

mitchellh:
Done. I also moved login and the auth method ones cause we alphabetize there.

izaaklauer (approved):

PR #1879, opened 2021-07-16
builtin/aws: add ability to add AWS tags to resources

psihachina:
This commit adds ability to add AWS tags to resources, for example
```
app "vole" {
 
  ...
  ...
  ...

  deploy {
    use "aws-ecs" {
      region = "us-east-2"
      memory = "512"
      cluster = "dev-nutcorp"
      tags = {
        test_tag="test-tag"
        test_tag2="test-tag2"
      }
    }
  }
}
```
image

### DIFF
--- a/.changelog/1879.txt
+++ b/.changelog/1879.txt
@@ -1,0 +1,2 @@
+```release-note:improvement
+plugin/aws: add ability to add AWS tags to resources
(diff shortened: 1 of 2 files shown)


izaaklauer:
Hi there,

This is a great idea! We're thinking through a general solution to this problem that involves mapping waypoint labels as resource tags across all platforms.

We've opened this issue to track that work. Thank you for the concrete use-case!

https://github.com/hashicorp/waypoint/issues/1909

izaaklauer:
We really appreciate your use of Waypoint and the thought put into these PRs! In the future, if you open issues for your ideas beforehand, we can discuss them there first and collaborate on the implementation.

Thanks!

PR #1885, opened 2021-07-19
web: migrate to web platform- packages

BRKalow:
Migrate away from @hashicorp/nextjs-scripts to the new suite of @hashicorp/platform-* packages

[_Created by Sourcegraph campaign `bkalow/migrate-nextjs-scripts-packages`._](https://sourcegraph.hashi-mktg.com/users/bkalow/campaigns/migrate-nextjs-scripts-packages)


izaaklauer:
Hey one questions about this:

We're using next-js explicitly in our make file to format our mdx files in CI (https://github.com/hashicorp/waypoint/blob/main/Makefile#L137). Do you know offhand what the change would be there to migrate to the new packages?

BRKalow:
Good catch! `website-mdx` is fixed up now after my latest commit, so no changes should be necessary for that workflow 😄

website/pages/downloads/index.jsx
```diff
@@ -1,14 +1,15 @@
...
+import ProductDownloadsPage from '@hashicorp/react-product-downloads-page'
+import { generateStaticProps } from '@hashicorp/react-product-downloads-page/server'
 import styles from './style.module.css'
 
 export default function DownloadsPage({ releases }) {
   return (
     <span className={styles.downloads}>
       <HashiHead title={`Downloads | ${productName} by HashiCorp`} />
-      <ProductDownloader
+      <ProductDownloadsPage
```
jescalan, line 12: wrapping span and head can be dropped here since this is now a full page component

jescalan (changes requested):
just one small change with the downloads page

website/pages/downloads/index.jsx
BRKalow: Aha I knew I migrated another usage somewhere...thank you 🙏

jescalan (approved):

izaaklauer (approved):

izaaklauer:
Oh actually, when I re-run `npm install` from the website directory after this change, I get a large change to `package-lock.json`. Is this expected?

BRKalow:
@izaaklauer What's your `npm -v`? If it's `7`, it introduces a new lockfile format which causes quite a few changes. More info if you're curious: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

It should be fine all things considered, we've just gotta start enforcing npm v7.

BRKalow:
@zchsh Let me know if I can help with the updated global styles rollout here. I think it should fix the terminal background color issue 👀

izaaklauer:
> @izaaklauer What's your `npm -v`

@BRKalow Ah yup, it's `7.19.1`, and I can see that the lockFileVersion bumped from 1 to 2. Might make sense for us to commit those changes to embrace the new npm version.

PR #1886, opened 2021-07-19
serverinstall/k8s: update log msg and add terminal warn for RollingUpdate

### LINKED ISSUE #1795: `waypoint server upgrade` on kubernetes does not upgrade if `latest` tag image is cached.
**Describe the bug**
If you don't explicitly set a `k8s-server-image`, running `waypoint server upgrade` will default to using `hashicorp/waypoint:latest`. If that's already cached on the node waypoint is running on, the new deployment will use the cached image instead of re-pulling to get the actual latest version.

**Steps to Reproduce**
- Install waypoint on a single-node kubernetes cluster, without specifying a `k8s-server-image`
- Wait for the waypoint team to publish a new version
- Run `waypoint server upgrade -auto-approve`, then check the version of the waypoint server. It will be the previous version.

**Expected behavior**
Running `waypoint server upgrade` when a new version is available should always result in an upgrade to the latest version.

Options to fix this: 
- Waypoint should detect that a newer version is available, and use that instead of the `latest` tag.
- We could default to an image-pull-policy of `always`

applicable:

* Waypoint CLI Version: 0.4.1
* Waypoint Server Platform and Version: 0.4.0/0.4.1
* Waypoint Plugin: kubernetes, but this may also be a problem for docker/ecs/nomad.

krantzinator:
Fixes #1795

If a user does not specify a `-k8s-server-image` or `-k8s-pull-policy` on the initial install, then we default to K8S defaults, which is to use an imagePullPolicy of `Always` for images tagged `latest`. Rather than force a restart of the pod during a `waypoint upgrade`, we are adding a note in the CLI output to alert users to these settings so they can take appropriate action per their cluster settings and needs.

```
$ waypoint server upgrade -platform=kubernetes -auto-approve
✓ Context "install-1626724718" validated and connected successfully.
✓ Snapshot of server written to: 'waypoint-server-snapshot-1626728262'

» Upgrading...
  Waypoint server will now upgrade from version "v0.4.0-663-g4bc86276"
✓ Gathering information about the Kubernetes cluster...
⚠️ Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image
✓ Image set to update!
✓ Upgrade complete!
...
```

### DIFF
--- a/internal/serverinstall/k8s.go
+++ b/internal/serverinstall/k8s.go
@@ -443,10 +443,6 @@ func (i *K8sInstaller) Upgrade(
 		log.Info("Pod(s) deleted, k8s will now restart waypoint server ", serverName)
 	} else if waypointStatefulSet.Spec.UpdateStrategy.Type == "RollingUpdate" {
 		log.Info("Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image")
-		s.Update("Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image")
-		s.Status(terminal.StatusWarn)
-		s.Done()
-		s = sg.Add("")
 	} else {
 		log.Warn("Update Strategy is not recognized, so no action is taken", "UpdateStrategy",
 			waypointStatefulSet.Spec.UpdateStrategy.Type)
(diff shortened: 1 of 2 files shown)


internal/serverinstall/k8s.go
```diff
@@ -442,7 +442,11 @@ func (i *K8sInstaller) Upgrade(
 
 		log.Info("Pod(s) deleted, k8s will now restart waypoint server ", serverName)
 	} else if waypointStatefulSet.Spec.UpdateStrategy.Type == "RollingUpdate" {
-		log.Info("Update Strategy is 'RollingUpdate', no further action required")
+		log.Info("Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image")
+		s.Update("Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image")
+		s.Status(terminal.StatusWarn)
+		s.Done()
+		s = sg.Add("")
```
briancain, line 449: I think we can move the `s.Update` checks to further below on `567`: https://github.com/hashicorp/waypoint/blob/4bc862762b6f56bcde7343c596c25c96a370710e/internal/serverinstall/k8s.go#L563-L566

But good call on the Warning, we should update ☝🏻 to be Warn

briancain (changes requested):
Looks good to me! Just a quick note on moving where the output happens

internal/serverinstall/k8s.go
krantzinator: new output
```
✓ Image set to update!
⚠️ Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image
✓ Upgrade complete!

Kubernetes is now set to upgrade waypoint server image with its
'RollingUpdate' strategy. This means the pod might not be updated immediately.
```
image

briancain (approved):

Issue #1880, opened 2021-07-18
"lookup waypoint.local: no such host" when deploying to Nomad


davidmytton:
**Describe the bug**
Following the steps in the [deploy to Nomad guide](https://learn.hashicorp.com/tutorials/waypoint/get-started-nomad), Waypoint shows the build, deploy, release as successful but it fails on Nomad with the error:

```
 [ERROR] client.driver_mgr.docker: failed pulling container: driver=docker image_ref=waypoint.local/web:latest error="API error (500): Get "https://waypoint.local/v2/": dial tcp: lookup waypoint.local: no such host"
```

**Steps to Reproduce**
```
# david @ dm-mba-m1 in ~/Code/shincha-web on git:main o [7:29:33] C:130
$ export NOMAD_ADDR='http://192.168.1.244:4646'

# david @ dm-mba-m1 in ~/Code/shincha-web on git:main o [7:29:48] 
$ waypoint install -platform=nomad -nomad-dc=dc1 -accept-tos
✓ Waypoint server ready
✓ Server installed and configured!
✓ Installing runner...
✓ Waypoint runner installed
Waypoint server successfully installed and configured!

The CLI has been configured to connect to the server automatically. This
connection information is saved in the CLI context named "install-1626589812".
Use the "waypoint context" CLI to manage CLI contexts.

The server has been configured to advertise the following address for
entrypoint communications. This must be a reachable address for all your
deployments. If this is incorrect, manually set it using the CLI command
"waypoint server config-set".

To launch and authenticate into the Web UI, run:
waypoint ui -authenticate

Advertise Address: 192.168.1.244:31917
Web UI Address: https://192.168.1.244:9702

# david @ dm-mba-m1 in ~/Code/shincha-web on git:main x [7:31:15] 
$ nomad status
ID               Type     Priority  Status   Submit Date
waypoint-runner  service  50        running  2021-07-18T07:31:11+01:00
waypoint-server  service  50        running  2021-07-18T07:30:13+01:00

# david @ dm-mba-m1 in ~/Code/shincha-web on git:main x [7:31:22] 
$ waypoint init
✓ Configuration file appears valid
✓ Connection to Waypoint server was successful
✓ Project "shincha-web" and all apps are registered with the server.
✓ Plugins loaded and configured successfully

Project initialized!

You may now call 'waypoint up' to deploy your project or
commands such as 'waypoint build' to perform steps individually.

[...]

PR #1892, opened 2021-07-20
website: note about using local Docker registry

davidmytton:
Clarifies the use of the `local` option of the Docker `registry` stanza to make locally built images available to Nomad, as discussed in https://github.com/hashicorp/waypoint/issues/1880


briancain (changes requested):
Hey there @davidmytton - thanks for tackling this issue!

This kind of change should be covered here instead: https://tip.waypointproject.io/docs/lifecycle/deploy#private-registries

Given the note about Nomad, it looks like we'd actually have some work to support setting up some additional options for authenticating to a private registry though. I think we could at least include a note about how to do it locally when using Nomad in dev mode.

I'm happy to pick this up since it's a bit more work, if so please feel free to close this PR and I can open an issue to track the work. Thanks!

davidmytton:
No problem. Probably better to close this then!

Issue #1532, opened 2021-05-24
Upgrade ember-template-lint to 3.x series


jgwhite:
We’re currently on ember-template-lint 2.18.1. Let’s upgrade to the 3.x series (3.4.2 at the time of writing).

Three motivations to upgrade:

1. The 3.x series includes a very useful [TODO workflow](https://github.com/ember-template-lint/ember-template-lint/blob/v3.4.2/docs/todos.md#todos) that will be a great help in gradually and sustainably improving the health of our template code. 
2. Newer versions of ember-template-lint include more comprehensive accessibility rules.
3. Newer versions make it even easier to write custom plugins, tailored exactly to the needs of the project.

PR #1893, opened 2021-07-20
ui: refactor breadcrumbs

jgwhite:
## Why the change?

A few reasons:

1. We were using private API (`@params`) which might disappear in future.
2. This clears the way to #1532 
3. Makes the interface to breadcrumbs clearer

## Any other notes?

We might want to reintroduce the ability to specify explicit `models` at some point:

```ts
breadcrumbs: Breadcrumb[] = [
  {
    label: 'Example',
    route: 'example',
    models: ['some-id']
  }
];
```

But for now I don’t think we need it.

### Using Mirage

1. Check out the branch:
   ```sh
   git checkout ui/refactor-breadcrumbs
   ```
2. Boot the dev server
   ```sh
   cd ui && ember serve
   ```
3. [Visit the app](http://localhost:4200)
4. Browse around
5. Verify the breadcrumbs work as production

### DIFF
--- a/ui/app/components/app-breadcrumbs/index.hbs
+++ b/ui/app/components/app-breadcrumbs/index.hbs
@@ -3,7 +3,7 @@
     {{#if breadcrumb.isPending}}
       <a href="#" aria-label="loading" data-test-breadcrumb="loading">&hellip;</a>
     {{else}}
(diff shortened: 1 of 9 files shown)


sabrinako (approved):
Looks good to me! Tested it out and works as expected

PR #1739, opened 2021-06-30
feat(ui): show deployment URL if available

denysvitali:
UI Changes related to #1387

This PR only contains the UI changes to ease the frontend review. This change basically makes the deployment URL visible and removes the "Visit Deployment" title to be consistent with the Release URL button:

image

In addition to this, a consistency change was performed (Deployments were displayed in the form `#1` instead of `v1` when you visited the deployment page)

PR #1894, opened 2021-07-20
Generate API bindings for OIDC protos

gregone:
Updates the Javascript client to include proto changes from https://github.com/hashicorp/waypoint/pull/1878

### DIFF
--- a/ui/lib/waypoint-client/ServerServiceClientPb.ts
+++ b/ui/lib/waypoint-client/ServerServiceClientPb.ts
@@ -76,6 +76,126 @@ export class WaypointClient {
     this.methodInfoGetVersionInfo);
   }
 
+  methodInfoListOIDCAuthMethods = new grpcWeb.AbstractClientBase.MethodInfo(
+    internal_server_proto_server_pb.ListOIDCAuthMethodsResponse,
+    (request: google_protobuf_empty_pb.Empty) => {
+      return request.serializeBinary();
+    },
+    internal_server_proto_server_pb.ListOIDCAuthMethodsResponse.deserializeBinary
+  );
+
+  listOIDCAuthMethods(
+    request: google_protobuf_empty_pb.Empty,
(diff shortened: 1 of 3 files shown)


ui/lib/api-common-protos/google/api/annotations_pb.js
```diff
@@ -2,7 +2,6 @@
 /**
  * @fileoverview
  * @enhanceable
- * @suppress {missingRequire} reports error on implicit type usages.
```
jgwhite, line 5: [boulder]

I’m ready to approve this, but I’m a bit confused by the removal of these `@suppress` lines.

Can we compare tool versions? Here’s mine:

| Tool | Version | Package Manager |
| --- | --- | --- |
| `protoc` | `3.17.3` | Homebrew |
| `ts-protoc-gen` | `0.15.0` | Volta |
| `protoc-gen-grpc-web` | `1.2.1` | Homebrew |

gregone: |`protoc` | `3.14.0`| Binary (manual install) |
| `ts-protoc-gen` | `0.15.0`| yarn |
|`protoc-gen-grpc-web ` |`1.2.1`| Binary (manual install) |

FWIW I've started seeing the removals in https://github.com/hashicorp/waypoint/pull/1739 , but was not expecting my setup to do the same. I wonder what has changed.

jgwhite: Interestingly, when I run `make gen/ts` with protoc `3.17.3` I don’t see any change to those `@suppress {missingRequire}` lines. So I suspect the `@suppress {missingRequire}` stuff was introduced to the core protoc javascript/closure-compiler generator some time *after* `3.14.0`. Could you try upgrading to `3.17.3` and see if the behavior changes?

gregone: @jgwhite just did so, the `missingRequire` lines are missing no more.

jgwhite (approved):

PR #1895, opened 2021-07-20
Bug: allow nomad runner and server memory to be configurable

izaaklauer:
We had the wrong target for the nomad memory flags - it was overwriting CPU config.


krantzinator (approved):

briancain (approved):
Ahh, whoops 🙈 👍🏻

Issue #1508, opened 2021-05-19
Expose Git Source information when available


gregone:
**Is your feature request related to a problem? Please describe.**
Projects that use git and automated syncing have information about git changes that triggered a new deploy, we should surface that information in the UI for better visibility to the user

**Describe the solution you'd like**
Display commit & author information along with operations in the browser UI

PR #1920, opened 2021-07-22
[WIP] Git ref data on operations

gregone:
## Very WIP

prerequisite work for https://github.com/hashicorp/waypoint/issues/1508

Adds `DataSourceRef` information on `Operation` objects returned by the server, starting with `Build`.


mitchellh (commented):
I think we should just populate this in the API rather than string it all the way through. We already have the data attached to the `Job` so if we want/need this for the UI/CLI we should just query it in the read-path and populate it rather than store it directly on the object.

If we do it that way I think you just have to edit the protos and `internal/server/singleprocess.`

gregone:
Closing this, as @mitchellh solved it in https://github.com/hashicorp/waypoint/pull/1980

PR #1922, opened 2021-07-22
internal/plugin: fix muliple imports

alrs:
`ssm` and `vault` were being imported twice each in `internal/plugin`. This fixes that.

Could someone add a "no-changelog" label?


mitchellh:
Thanks for catchign this.

PR #1932, opened 2021-07-26
Internal function to queue multiple jobs transactionally

mitchellh:
This exposes `queueJobMulti` in `internal/server/singleprocess` and modifies the state store to support queueing multiple jobs transactionally: either all queue, or all do not queue. This is useful in @briancain's queueing work and probably useful in the future as well.

I chose to not expose this publicly as an RPC API since we don't need it right now, but all the primitives are there for when we do.


catsby (approved):
👍

briancain (approved):
Looks good to me! ✨

PR #1935, opened 2021-07-27
Feature/aws ecs/add ability attaching policy for task role

psihachina:
This commit adds the ability to pass IAM Policy JSON for attaching to new task role, for example
```
app "hamster" {
 
  config {
    env = {
      test= configdynamic("aws-ssm", {
        path = "test"
      })
    }
  }
  ...
  ...

    deploy {
    use "aws-ecs" {
      region = "us-east-2"
      memory = "512"
      task_role_name = "newTaskRoleForTest"
      task_role_policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "ssm:*"
      ],
      "Resource": [
        "arn:aws:ssm:us-east-2:xxxxxxxx:*"
      ]
     }
  ]
}
    EOF
    }
  }
}
```
image

image

### Test
[asciicast](https://asciinema.org/a/crPmJbbZZcwrKfOKoRiATVDAD)


izaaklauer:
Hi @psihachina,

Does your use-case require that policies change from deployment to deployment, or are you looking for a way to use the same long-lived policy for every deployment for a given app?

psihachina:
Hi @izaaklauer,

The most important thing in our use is the function of automatic policy creation itself so that the user can describe the policy he needs and the waypoint created it himself. 
I may have understood your point. I did not foresee the possibility of changing the policy. 
I will try to fix this so that the change in the description of the policy is tracked and after the role is changed or recreated.

izaaklauer:
Hey @psihachina,

I asked because right now waypoint has a great system for creating and managing operation scoped resources (i.e. resources that need to be created and destroyed for every deployment or release), but not a great system for managing project-scoped resources (i.e. resources that need to be created once for each app and can be re-used for each deployment). We'd like to avoid waypoint managing more project-scoped resources until we can build better support for them.

The fact that it's impossible to use a custom task-role policy is definitely a problem though, and thank you for bringing it to light! Our preferred solution to this for now would be to add a config field called `task_role_policy_arn` field to the waypoint ECS config. Users would create this policy outside of waypoint one time, but would be able to use it for every deployment and release.

Would that work for your use-case?

psihachina:
Hey @izaaklauer, 

I understand your point of view. You already have such a field as `task_role_name` in the config. Since it is not yet possible to create custom policies in the waypoint, I have another idea. It is necessary to somehow transfer the role name to the waypoint, for example from terraform. For example, using AWS SSM or some other secret store. 
It seems like terraform has so-called `data` with this logic.
It would be great to have a similar waypoint, since in addition to the `role name`, it was planned to generate and transfer `subnets` in a similar way.

AutomationD:
Hey, @izaaklauer, I see your point with project-scoped resources and a solution with passing task role `task_role_policy_arn` would work for me as well. Maybe there should be a documentation gist on how to create one via AWS CLI. I know it's a simple step, but a oneliner maybe super helpful for most users.

Hey, @psihachina. Thanks for the PR! Regarding the `data` thing, it seems like a potential solution for external data communication, but maybe we should create another issue for that to discuss it (and link it here).

psihachina:
Hi @izaaklauer , 

I added a field `task_role_policy_arn`  and updated the PR description

Issue #1881, opened 2021-07-18
Docker build step does not support buildx


davidmytton:
**Describe the bug**
[Docker buildx](https://docs.docker.com/desktop/multi-arch/) is needed to build for multiple CPU architectures e.g. different ARM architectures. Currently, the Waypoint Docker build step plugin executes `docker build` and cannot be configured to support `docker buildx`.

This may be a feature request but could also be a bug because executing `docker install buildx` [should alias](https://docs.docker.com/buildx/working-with-buildx/#set-buildx-as-the-default-builder) `docker build` to `docker buildx`. This should allow Waypoint to automatically support `buildx` with the additional `build_args` setting. However, even when this alias has been configured, Waypoint continues to execute `docker build`.

**Steps to Reproduce**
```shell
docker install buildx
waypoint build
```

**Expected behavior**
Waypoint executes using the `buildx` alias.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI: v0.4.1 (4d5c838e)
* Waypoint Plugin: `docker`

briancain:
Hey there @davidmytton - thanks for the feature request!

It's not as easy as aliasing a shell command for Waypoint, because we don't `exec` the docker CLI, but rather we use their [Go client to build images](https://pkg.go.dev/github.com/ph/moby/client#Client.ImageBuild) with our [Waypoint docker build plugin](https://github.com/hashicorp/waypoint/blob/86f5d261c3bbd221d6802c9b2d8330dcbc31078f/builtin/docker/builder.go#L299).

This _might_ work if we expose a `Platform` option in our docker builder plugin, which would get passed as an [image build option to the docker engine](https://pkg.go.dev/github.com/docker/docker/api/types#ImageBuildOptions) when building the image. However I don't know if `Platform` is a buildx option or something separate since there's not a codedoc there.

Anyway thanks for opening the issue!

PR #1937, opened 2021-07-27
plugin/docker: Adding a session when building with buildkit

### LINKED ISSUE #1056: dockerfile.v0: failed to create LLB definition: no active sessions
**Describe the bug**

`waypoint build` fails with:

```
$ waypoint build
⠦ Building image...
! unable to stream build logs to the terminal: failed to solve with frontend
  dockerfile.v0: failed to create LLB definition: no active sessions
```

**Steps to Reproduce**

Run `waypoint build` with the configuration file below

**Expected behavior**

Waypoint should build the app

**Additional context**

Waypoint 0.2.1.

waypoint.hcl:

```
project = "myproject"

app "myproject-myapp" {

  labels = {
    "service" = "myapp"
    "env" = "dev"
  }

  build {
    use "docker" {
      buildkit = true
    }

    registry {
      use "docker" {
        image = "gcr.io/myproject/myapp"
        tag = "test-0"
      }
    }
  }

  deploy {
    use "google-cloud-run" {
      project  = "myproject"
      location = "us-central1"
      port = 8080
      unauthenticated = true

      capacity {
        memory = 2048
        cpu_count = 1
        max_requests_per_container = 80
        request_timeout = 900
      }
    }
  }

  release {
    use "google-cloud-run" {}
  }
}
```

Dockerfile:

```
FROM minimalcompact/thumbor
ENV THUMBOR_PORT=8080
```

izaaklauer:
Fixes #1056

### Why this change?

We observed that we could replicate #1056 on most of the example repos by enabling buildkit. The failure seems to be governed by the FROM image, but isn't consistent - on my machine, `FROM nginx:stable` would be fine, but wouldn't work for @xiaolin-ninja. In any case, the error mentions not having a session, and this change fixes it.

### Future considerations

The docker CLI's usage of moby sessions is _much_ more sophisticated than this. If you're curious, [here](https://github.com/docker/cli/blob/13e4a097ea4cd6df652c5f57f68771fe526fcb1d/cli/command/image/build_buildkit.go#L50)'s where to start reading. Things they are doing that we are not include:

- Generating a session name based on the build context dir, not statically: https://github.com/docker/cli/blob/13e4a097ea4cd6df652c5f57f68771fe526fcb1d/cli/command/image/build_session.go#L34

- Generating a sharedKey that includes randomness but is saved to disk and re-used: https://github.com/docker/cli/blob/13e4a097ea4cd6df652c5f57f68771fe526fcb1d/cli/command/image/build_session.go#L33

- Adding session "Allow" methods, which it looks like turn into GRPC services, which seems to enable non-standard context paths, dockerfile paths, and docker daemon auth methods: (e.x. [here](https://github.com/docker/cli/blob/master/cli/command/image/build_buildkit.go#L167-L177) and [here](https://github.com/docker/cli/blob/master/cli/command/image/build_buildkit.go#L194))

I haven't found any documentation on what any of this session configuration is for, and would rather not dig deep into the moby code to figure it out at this moment. I'm comforted that [img also uses a static session name and blank sharedKey](https://github.com/genuinetools/img/blob/master/client/session.go#L33), and because we only ever connect to a local docker daemon, I don't believe this hurts our security model.

I wouldn't be shocked if some advanced combination of buildkit parameters requires the fancy session configuration. At worst though, this change brings building with buildkit from mostly broken to mostly working, so I think it's worth merging.

Dependency for #1881

### DIFF
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -32,6 +32,8 @@ import (
 	"github.com/hashicorp/waypoint/internal/pkg/epinject"
 )
 
+const minBuildkitDockerVersion = "1.39"
+
 // Builder uses `docker build` to build a Docker image.
 type Builder struct {
 	config BuilderConfig
@@ -251,7 +253,7 @@ func (b *Builder) Build(
 		step.Done()
 		step = nil
 		if err := b.buildWithDocker(
-			ctx, ui, sg, cli, contextDir, relDockerfile, result.Name(), b.config.BuildArgs,
+			ctx, ui, sg, cli, contextDir, relDockerfile, result.Name(), b.config.BuildArgs, log,
 		); err != nil {
 			return nil, err
 		}
@@ -308,6 +310,7 @@ func (b *Builder) buildWithDocker(
 	relDockerfile string,
 	tag string,
 	buildArgs map[string]*string,
+	log hclog.Logger,
 ) error {
 	excludes, err := build.ReadDockerignore(contextDir)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


krantzinator (approved):
I can't speak to the security model but this fix is 🎉

builtin/docker/builder.go
```diff
@@ -341,13 +344,30 @@ func (b *Builder) buildWithDocker(
...
 		Dockerfile: relDockerfile,
 		Tags:       []string{tag},
 		Remove:     true,
 		BuildArgs:  buildArgs,
-	})
+	}
+
+	// Buildkit builds need a session under most circumstances, but sessions are only supported in >1.39
+	if ver == types.BuilderBuildKit && versions.GreaterThanOrEqualTo(cli.ClientVersion(), "1.39") {
+
```
mitchellh, line 357: This empty line is a bit weird.

mitchellh (approved):
One stupid little comment but this looks good.

builtin/docker/builder.go
```diff
@@ -341,13 +344,30 @@ func (b *Builder) buildWithDocker(
...
+		dialSession := func(ctx context.Context, proto string, meta map[string][]string) (net.Conn, error) {
+			return cli.DialHijack(ctx, "/session", proto, meta)
+		}
+
+		go s.Run(ctx, dialSession)
+		defer s.Close()
+
+		buildOpts.SessionID = s.ID()
+	}
```
briancain, line 368: We should probably log something here in the case where the CLI version isn't supported with sessions but buildkit was requested. It could be inside the StepGroup as a warning so it's visible, or maybe we could pass the `log` through to this func and just log a warning instead.

briancain (changes requested):
Nice job getting this working! I've just got a request for more logging in the unsupported case 👍🏻

briancain (approved):

PR #1911, opened 2021-07-21
UI: Add on-demand status reports (health checks) to app

sabrinako:
### Related PRs
#1943 + #1982

### Overview
This is a PR for implementing a button in the UI for refreshing health checks (aka status reports). This is a MVP implementation, will further quality-of-life improvements to be added during 0.5.x updates.

### Testing
1. Pull down this branch to your machine
2. From `waypoint` directory cd into `contrib/waypoint-local`
3. `sh setup-waypoint-local.sh`
4. `waypoint ui -authenticate`
    1. NOTE: I've had issues with authenticating when directed to `127.0.0.1:9702` so also visit `localhost:9702` if you cannot successfully log in with the ember test environment
4. Navigate to the `waypoint/ui` directory
5. `ember serve local` + `waypoint user token`
6. Create new project with git repo settings as in picture: Screen Shot 2021-08-10 at 11 20 05 AM  
7. Wait for Waypoint to do an auto `waypoint up`
8. Check the individual page for the deployment that was created from this process
9. Click `re-run health check` button
    1. NOTE: as this is an MVP, the status does not update in real time, refresh the page to see new status once re-run button is no longer disabled (means job is completed)
10. Stop the docker container generated on your machine by the waypoint server
11. Rerun the health check again, see that it is reporting as down now
12. Check project page to ensure status shows correctly there as well

Issue #1511, opened 2021-05-19
Add the ability to trigger a status check from the UI


gregone:
**Is your feature request related to a problem? Please describe.**
As a follow-up to https://github.com/hashicorp/waypoint/issues/1237, once we have the API to support it, we should enable the user to trigger a status check from the UI

**Describe the solution you'd like**
As defined in the UI component design, add a button to trigger a status check (if the target supports it)

sabrinako:
Linked the wrong issue with the PR above, reopening!

PR #1946, opened 2021-07-28
internal/server: Add endpoint to queue a status report job on demand

### LINKED ISSUE #1943: core/server: Add a service func to support one-off status report requests

Users want to know the status of their deployment or release on demand in the UI. The work detailed in issue https://github.com/hashicorp/waypoint/issues/1511 and pull request https://github.com/hashicorp/waypoint/pull/1911 by @sabrinako will add a new button that gives users on-demand requests to refresh the latest report.

The server should implement a new service function `ExpediteStatusReportRequest` that accepts a one-of target that is either a Deployment or Release, generate a QueuedJobRequest to generate a status report, queue that job and return the id back to the front end that made the request.

briancain:
Users want to know the status of their deployment or release on demand in the UI. The work detailed in issue #1511 and pull request #1911 by @sabrinako will add a new button that gives users on-demand requests to refresh the latest report.

This pull request implements the service function responsible for generating and queuing a status report job on demand from the UI.

Fixes https://github.com/hashicorp/waypoint/issues/1943

Not adding a changelog for now, I think it'll be more a more useful entry tied with the UI pull request.


internal/server/singleprocess/service_status_report.go
```diff
@@ -77,3 +78,75 @@ func (s *service) GetStatusReport(
...
+			SingletonId: fmt.Sprintf("status-report-ondemand/%s", applicationRef.Application),
+
+			Application: applicationRef,
+
+			// Application polling requires a data source to be configured for the project
+			// Otherwise a status report can't properly eval the projects hcl context
+			// needed to query the deploy or release
+			DataSource: project.DataSource,
+
+			Workspace: &pb.Ref_Workspace{Workspace: "default"},
```
briancain, line 129: Note: if the UI has the workspace available we should hook up `workspace` as a parameter to `ExpediteStatusReportRequest`

jgwhite (commented):
One minor suggestion about field naming but otherwise looking great 👍

internal/server/proto/server.proto
```diff
@@ -2828,6 +2831,24 @@ message GetStatusReportRequest {
...
+    Deployment deployment = 2;
+
+    // the release id that this status report was generated on
+    Release release = 3;
+  }
+}
+
+// id of job expedited by the request
+message ExpediteStatusReportResponse {
+    string id = 1;
```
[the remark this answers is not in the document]
briancain: Yes totally! Job ID sounds better

internal/server/singleprocess/service_status_report.go
jgwhite: The UI technically does have the workspace available (it’s the first chunk of the path i.e. `/default/projects/...`) but it’s not really wired in so this needn’t be a blocker.

briancain: Ok cool! I think it might be more reliable right now for the UI to request the workspace. I think the only way the server can know is on the Project proto, which is an array of Workspaces. So if the UI already knows what workspace the project is in, that might be the best approach at the moment.

If that makes I'll update the PR to expect a workspace param

internal/server/proto/server.proto
```diff
@@ -2828,6 +2831,24 @@ message GetStatusReportRequest {
   // a report by deployment or release operation id
 }
 
+message ExpediteStatusReportRequest {
+  Ref.Operation ref = 1;
```
jgwhite, line 2835: It doesn’t look like we actually use this in the service. Maybe we can do without it?

internal/server/proto/server.proto
```diff
@@ -2828,6 +2831,27 @@ message GetStatusReportRequest {
...
+
+  // Target is used to determine what a status report should be run against
+  oneof target {
+    // the deployment id that this status report was generated on
+    Deployment deployment = 2;
+
+    // the release id that this status report was generated on
+    Release release = 3;
```
jgwhite, line 2843: I was expecting these to be `Ref.Deployment` and `Ref.Release` (rather than embedding whole `Deployment` and `Release` messages) but of course we don’t actually have `Ref.Deployment` and `Ref.Release` in the schema, just `Ref.Operation` which doesn’t tell you what type of operation is being referenced.

What do you think is the best course of action? Add `Ref.Deployment` and `Ref.Release`?

briancain: We could add those, that isn't so bad. Then the UI passes along the name inside one of the Refs rather than the entire message. Will push that commit now.

jgwhite: Or it could potentially be something like this:

```proto
message ExpediteStatusReportRequest {
  // Target is used to determine what a status report should be run against
  oneof target {
    Ref.Operation deployment = 1;
    Ref.Operation release = 2;
  }
}
```

briancain: Haha I was about to suggest that! We can get the deployment or release from the Ref_Operation, so yep, that's what we need 👍🏻

jgwhite (approved):
Niiiice

image

internal/server/proto/server.proto
```diff
@@ -2828,6 +2831,25 @@ message GetStatusReportRequest {
   // a report by deployment or release operation id
 }
 
+message ExpediteStatusReportRequest {
```
mitchellh, line 2834: Should this be one of them newfangled `UI_` endpoints instead?

It feels weird to have a special API for queueing a status report when we have the generic `QueueJob` that supports this (and internally we're just building that up anyways).

briancain: That's a good point.... @jgwhite should the UI just be using the `QueueJob` API endpoint? This service func isn't doing anything that the frontend could handle if that's the case.

jgwhite: That’s a good question. Originally we were planning to call `QueueJob` directly from the UI but as we discussed it we realized that we wanted something that would queue the job whilst respecting the new app polling logic (rather than sidestepping it). To me this doesn’t feel like it’s explicitly tailored to the UI but I’m certainly not opposed to adding the `UI_` prefix.

briancain: Ah right, I think we wouldn't have to worry about the app status polling logic, since this sidesteps that as well and queues a job directly rather than trying to update an apps "next poll" to now. So maybe instead of this the front end queues the job?

briancain: On the other hand, it could be worth having a dedicated endpoint that does all of the construction to refresh a status report 🤔

jgwhite: > So maybe instead of this the front end queues the job?

Personally I’d like the UI to use `QueueJob` directly. The one argument I can see against it is the special logic around constructing the singleton ID:

https://github.com/hashicorp/waypoint/blob/d0fdb750fc8b79d9fdd39c73e54750a607d06f09/internal/server/singleprocess/service_status_report.go#L127-L129

We could certainly recreate this in the UI but it might be a little brittle? Maybe I’m being over-cautious?

briancain: You should be able to use `SingletonId` when constructing a job from the UI and sending it through the `QueueJob` endpoint. I think it's nice in this case too prevent people from accidentally clicking refresh and queuing a bunch of jobs?

sabrinako: Chiming in here!
@jgwhite and I just tried walking through a `QueueJob` implementation and there's ...a lot... of code to send a request. An issue is that `StatusReportOp` needs a `Deployment` or a `Release`, not a `Ref.Deployment` or `Ref.Release`. We have all the information about the deployment or release available, but in object format. The JS protobuf library does _**not**_ have the ability to convert an object into a specified class.  Here's just some of the code:
```
    let app = new Ref.Application();
    app.setApplication(...);
    app.setProject(...);

    let workspace = new Ref.Workspace();
    workspace.setWorkspace(...)

    let runner = new Ref.Runner();
    runner.setAny(new Ref.RunnerAny());

    let statusReportOp = new Job.StatusReportOp();
    // NOTE: This is where we would have to create a new Deployment()
    // which adds even more lines to this than what is already here
    statusReportOp.setDeployment();

    let job = new Job();
    job.setApplication(app);
    job.setWorkspace(workspace);
    job.setTargetRunner(runner);
    job.setStatusReport(statusReportOp);

    let req = new QueueJobRequest();
    req.setJob(job);
    let resp = await this.api.client.queueJob(req, this.api.WithMeta());
```
This might be an argument to have a separate `ExpediteStatusReport` function, which we can add the `UI_` prefix to

mitchellh: One other option, which we do for HCL formatting for example, is have a `UI_` endpoint that just returns the pre-populated `QueueJobRequest` that it would’ve sent. That way its more flexible in the future if the UI ever wants to tweak it. So could do like `UI_StatusReportQueueJobRequest` or something (prob too long) and just get the right structure back.

I’m not against the original approach either if that makes sense, just trying to give you the most flexible tools.

briancain: Just to circle back to this thread....we discussed this in slack and we are thinking for now this is OK as is. In the future if the UI ever needs a more tweakable status report refresh job, we should introduce an endpoint that returns a `QueuedJobRequest` proto message instead and let the UI manipulate the request before queuing.

sabrinako (approved):
Still shaky on my Go knowledge, but from what I can decipher + the `server.proto` file everything looks good to me!

PR #1948, opened 2021-07-28
internal/server: invite tokens for other users are allowed

mitchellh:
For now, everyone is a superuser (we have no policy system yet) so we
allow anyone to do this. However, we do want to ensure that the user
exists. This implements and tests that.


izaaklauer (approved):

izaaklauer:
^ I merged this now because I've tested this, @mitchellh is OOO, and I'd like to have the fix 😄

Issue #1942, opened 2021-07-28
GCR Deployment just fails without any info why


bnrosa:
I'm trying to deploy my project to GCR using the [tutorial](https://learn.hashicorp.com/tutorials/waypoint/google-cloud-run?in=waypoint/deploy-google-cloud#overview) but it simply fails on the deployment step without comprehensible errors, is there a way I can debug it further? The logs aren't of any help either. Also, I'm have enabled my GC account, the API as well, made it billable, so it shouldn't be related to that.
```
project = "konf-auth-dev"

app "konf-auth-dev" {

  labels = {
    "service" = "konf-auth-dev",
    "env"     = "dev"
  }

  build {
    use "pack" {
    }

    registry {
      use "docker" {
        image = "gcr.io/konf-auth-dev/dev"
        tag   = "latest"
      }
    }
  }

  deploy {
    use "google-cloud-run" {
      project  = "konf-auth-dev"
      location = "us-east1"

      port = 5000

      capacity {
        memory                     = 128
        cpu_count                  = 1
        max_requests_per_container = 10
        request_timeout            = 300
      }

      auto_scaling {
        max = 2
      }


      vpc_access {
        connector = "custom-vpc-connector"
        egress = "all"
      }
    }
  }

  release {
    use "google-cloud-run" {}
  }
}
```
![image](https://user-images.githubusercontent.com/14277020/127369408-7f63b123-e45f-4a86-84fa-3c1c82cd3b85.png)

krantzinator:
@bnrosa Are you able to share what the solution was? If it's something that would help our tutorial our we can add it!

krantzinator:
Or our error messaging :)

bnrosa:
Well, first there's some inconsistency between the tutorial and docs, if you look into the [docs](https://www.waypointproject.io/plugins/google-cloud-run) it says that the **vpc_access** parameter is required, but it doesn't appears in the tutorial setup.
I'm still trying to fix it, but I can identify what the issue is, so the problem is simply the status of the GCR service, it is not ready. The issue is caused by the connector not being found.
![image](https://user-images.githubusercontent.com/14277020/127387031-f891593c-fedd-4e12-9122-dbe74e626269.png)

Hope it's helpful info 😄

bnrosa:
Also, this error gaves the same message: 
![image](https://user-images.githubusercontent.com/14277020/127387834-53aa15cc-2b0a-4163-a98b-a7545954a5a1.png)
[...]

PR #1958, opened 2021-07-29
plugin/gcp: a bit more helpful error msg

krantzinator:
Updating the error message per feedback in #1942

### DIFF
--- a/builtin/google/cloudrun/deployment.go
+++ b/builtin/google/cloudrun/deployment.go
@@ -171,7 +171,7 @@ func (d *Deployment) pollServiceReady(
 				return service, nil
 
 			case "False":
-				return nil, fmt.Errorf("service failed to get ready")
+				return nil, fmt.Errorf("Service failed to get ready; check GCR logs for troubleshooting info.")
 			}
 		}
 
--- a/builtin/google/cloudrun/deployment.go
+++ b/builtin/google/cloudrun/deployment.go
@@ -171,7 +171,7 @@ func (d *Deployment) pollServiceReady(
 				return service, nil
 
 			case "False":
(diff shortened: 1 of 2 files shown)


builtin/google/cloudrun/deployment.go
```diff
@@ -171,7 +171,7 @@ func (d *Deployment) pollServiceReady(
 				return service, nil
 
 			case "False":
-				return nil, fmt.Errorf("service failed to get ready")
+				return nil, fmt.Errorf("Service failed to get ready; check GCR logs for troubleshooting info.")
```
izaaklauer, line 174: GCR is Google Container Registry - I think "Cloud Run" might be better here.

I also hesitate a bit to punt them to the underlying platform here - it seems like the problem could just be that their app is crashing, and their app logs in waypoint could be the most useful thing.

izaaklauer (changes requested):

builtin/google/cloudrun/deployment.go
krantzinator: In this case there is no app in Waypoint to check, because the service didn't start. The deploy never succeeded.

krantzinator: How about scrap that line and just specify Cloud Run at the start, to clarify which service exactly failed to get ready: `Cloud Run service failed to get ready`

izaaklauer: Oh, got it! Maybe now that we have status checks, it would make sense not to block the deploy on this check, so they can see those logs in waypoint? Not related to this change though 👍

izaaklauer (approved):

PR #1963, opened 2021-07-29
Backport of internal/server: invite tokens for other users are allowed into release/0.4.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #1948 to be assessed for backporting due to the inclusion of the label backport/0.4.x.

---

For now, everyone is a superuser (we have no policy system yet) so we
allow anyone to do this. However, we do want to ensure that the user
exists. This implements and tests that.


izaaklauer:
This shouldn't be backported.

Issue #1501, opened 2021-05-18
feature: `waypoint status` CLI command


briancain:
**Is your feature request related to a problem? Please describe.**

With the addition of a [resource manager](https://github.com/hashicorp/waypoint-plugin-sdk/pull/29), and [status](https://github.com/hashicorp/waypoint-plugin-sdk/pull/23) [reporting](https://github.com/hashicorp/waypoint/pull/1488), we should include some diagnostic tooling in the Waypoint CLI to expose this information in a useful way for deployments and releases. Much of that content is relevant for a given lifecycle and exposing that information in a way that's easily digestible should give Waypoint users a better understanding of what the tool is doing for them, what resources are being created and managed, and the overall health associated with their environment.

**Describe the solution you'd like**

Introduce a `waypoint status` CLI command for giving a more detailed structure of the resources on a given deployment/release as well as the current reported health associated with it.

Not to be confused with a `waypoint deployment list` or `waypoint releases list`, this command should give an overall view of the current _state_ of the project in a way that's useful for "now".

PR #1966, opened 2021-07-29
feature: waypoint status CLI

briancain:
This pull request introduces a new command `waypoint status`.

- [ ] Future PR: `waypoint status check`

Fixes https://github.com/hashicorp/waypoint/issues/1501

### DIFF
--- a/internal/cli/filter_status.go
+++ b/internal/cli/filter_status.go
@@ -119,11 +119,11 @@ func (ff *filterFlags) orderOp() *pb.OperationOrder {
 func initFilterFlags(set *flag.Sets, ff *filterFlags, opts filterOption) {
 	f := set.NewSet("Filter Options")
 
-	if opts == filterOptionAll || opts&filterOptionState != 0 {
+	if opts == filterOptionAll || opts == filterOptionState {
 		f.EnumVar(stateFlagVar(&ff.flagStatusFilter))
 	}
 
-	if opts == filterOptionAll || opts&filterOptionPhyState != 0 {
+	if opts == filterOptionAll || opts == filterOptionPhyState {
 		f.EnumSingleVar(phyStateFlagVar(&ff.flagPhysState))
 	}
 
--- a/internal/cli/filter_status.go
+++ b/internal/cli/filter_status.go
@@ -127,7 +127,7 @@ func initFilterFlags(set *flag.Sets, ff *filterFlags, opts filterOption) {
 		f.EnumSingleVar(phyStateFlagVar(&ff.flagPhysState))
 	}
 
-	if opts == filterOptionAll || opts&filterOptionOrder != 0 {
+	if opts == filterOptionAll || opts == filterOptionOrder {
 		f.EnumSingleVar(&flag.EnumSingleVar{
 			Name:   "order-by",
 			Target: new(string),
--- a/internal/cli/filter_status.go
+++ b/internal/cli/filter_status.go
@@ -10,7 +10,7 @@ import (
 type filterOption int
 
 const (
-	fillterOptionAll filterOption = iota
+	filterOptionAll filterOption = iota
 	filterOptionState
 	filterOptionPhyState
 	filterOptionOrder
@@ -119,15 +119,15 @@ func (ff *filterFlags) orderOp() *pb.OperationOrder {
 func initFilterFlags(set *flag.Sets, ff *filterFlags, opts filterOption) {
 	f := set.NewSet("Filter Options")
 
-	if opts == fillterOptionAll || opts&filterOptionState != 0 {
+	if opts == filterOptionAll || opts&filterOptionState != 0 {
 		f.EnumVar(stateFlagVar(&ff.flagStatusFilter))
 	}
 
-	if opts == fillterOptionAll || opts&filterOptionPhyState != 0 {
+	if opts == filterOptionAll || opts&filterOptionPhyState != 0 {
 		f.EnumSingleVar(phyStateFlagVar(&ff.flagPhysState))
 	}
 
-	if opts == fillterOptionAll || opts&filterOptionOrder != 0 {
+	if opts == filterOptionAll || opts&filterOptionOrder != 0 {
 		f.EnumSingleVar(&flag.EnumSingleVar{
 			Name:   "order-by",
 			Target: new(string),
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -158,7 +158,7 @@ func (c *StatusCommand) FormatProjectAppStatus(projectTarget string) error {
 	//   App list
 
 	appHeaders := []string{
-		"App", "Workspace", "Latest Status", "Last Check",
+		"App", "Workspace", "Latest Report", "Latest Status", "Last Check",
 	}
 
 	appTbl := terminal.NewTable(appHeaders...)
@@ -187,10 +187,21 @@ func (c *StatusCommand) FormatProjectAppStatus(projectTarget string) error {
 			return err
 		}
 
+		var reportType string
+		switch appStatusResp.TargetId.(type) {
+		case *pb.StatusReport_DeploymentId:
+			reportType = "Deployment"
+		case *pb.StatusReport_ReleaseId:
+			reportType = "Release"
+		default:
+			reportType = "None"
+		}
+
 		statusColor := ""
 		columns := []string{
 			app.Name,
 			workspace,
+			reportType,
 			statusReportComplete,
 			statusReportCheckTime,
 		}
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/clicontext"
 	"github.com/hashicorp/waypoint/internal/clierrors"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
@@ -27,6 +28,8 @@ type StatusCommand struct {
 	flagJson        bool
 	flagAllProjects bool
 	filterFlags     filterFlags
+
+	serverCtx *clicontext.Config
 }
 
 func (c *StatusCommand) Run(args []string) int {
@@ -58,6 +61,7 @@ func (c *StatusCommand) Run(args []string) int {
 		c.ui.Output("Error loading context %q: %s", ctxName, err.Error(), terminal.WithErrorStyle())
 		return 1
 	}
+	c.serverCtx = ctxConfig
 
 	cmdArgs := flagSet.Args()
 
@@ -92,7 +96,9 @@ func (c *StatusCommand) Run(args []string) int {
 	// Generate a status view
 	if projectTarget == "" || c.flagAllProjects {
 		// Show high-level status of all projects
-		c.ui.Output(wpStatusMsg, ctxConfig.Server.Address)
+		if !c.flagJson {
+			c.ui.Output(wpStatusMsg, ctxConfig.Server.Address)
+		}
 
 		err = c.FormatProjectStatus()
 		if err != nil {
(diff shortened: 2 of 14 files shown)


internal/cli/status.go
```diff
@@ -0,0 +1,845 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	project := resp.Project
+
+	var workspace string
+	if len(resp.Workspaces) == 0 {
+		// this happens if you just wapyoint init
+		workspace = "???"
```
denysvitali, line 159: I would suggesting using `nil`. Althogh rare, I can see that somebody could eventually use `???` as a workspace and break this part 😅

briancain: Definitely! Good call.

krantzinator: Hm, this seems like odd behavior, in that after an `init` we should know the workspace? And in fact we know it if we specify the app.

```
$ waypoint status
Current status for project "example-go" in server context "localhost:9701".

APP       	WORKSPACE	LATEST STATUS	LAST CHECK 
example-go	default  	N/A          	          	

...
$ waypoint status -app=example-go
Current status for application "example-go" in project "example-go" in server context "localhost:9701".

Application Summary
APP       	WORKSPACE	LATEST STATUS	LAST CHECK 
example-go	???      	N/A          	          	
```

internal/cli/status.go
```diff
@@ -0,0 +1,845 @@
...
+		}
+	} else if projectTarget != "" && appTarget != "" {
+		// Advanced view of a single app status
+		err = c.FormatAppStatus(projectTarget, appTarget)
+		if err != nil {
+			if status.Code(err) == codes.NotFound {
+				c.ui.Output(wpAppNotFound, appTarget, projectTarget, c.serverCtx.Server.Address, terminal.WithErrorStyle())
+			} else {
+				c.ui.Output("CLI failed to format app status", terminal.WithErrorStyle())
+				c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
```
krantzinator, line 128: UX nit -- seeing both of these marked as an error makes me as a user think two errors happened, so I was confused for half a second. Can we combine them like `c.ui.Output("CLI failed to format app status: "+err.Error(), terminal.WithErrorStyle())`

internal/cli/status.go
```diff
@@ -0,0 +1,845 @@
...
+	if len(cmdArgs) > 1 {
+		c.ui.Output("No more than 1 argument required.\n\n"+c.Help(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	// Determine which view to show based on user input
+	var projectTarget, appTarget string
+	if len(cmdArgs) >= 1 {
+		s := cmdArgs[0]
+		target := strings.Split(s, "/") // This breaks if we allow projects with "/" in the name
```
krantzinator, line 76: We do allow projects with `/` in the name. I would support removing this ability and requiring the user to specify the app via the `-app` flag, rather than having two ways of doing that.

internal/cli/status.go
```diff
@@ -0,0 +1,845 @@
...
+		}
+	} else if projectTarget != "" && appTarget == "" {
+		// Show status of apps inside project
+		err = c.FormatProjectAppStatus(projectTarget)
+		if err != nil {
+			if status.Code(err) == codes.NotFound {
+				c.ui.Output(wpProjectNotFound, projectTarget, c.serverCtx.Server.Address, terminal.WithErrorStyle())
+			} else {
+				c.ui.Output("CLI failed to format project app statuses:", terminal.WithErrorStyle())
+				c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
```
krantzinator, line 116: Another combo place!

internal/cli/status.go
```diff
@@ -0,0 +1,845 @@
...
+		if err != nil {
+			return err
+		}
+
+		var workspace string
+		if len(resp.Workspaces) == 0 {
+			// this happens if you just wapyoint init
+			workspace = "???"
+		} else {
+			workspace = resp.Workspaces[0].Workspace.Workspace // TODO: assume the first workspace is correct??
```
krantzinator, line 456: I think this flow needs a bit more work. The workspace we specify via the global `workspace` flag is not reflected.
```
$ waypoint status -workspace=two
Current status for project "example-python" in server context "localhost:9701".

APP           	WORKSPACE	LATEST STATUS	LAST CHECK     
example-python	default  	✔ READY      	25 minutes ago	
```

internal/cli/filter_status.go
```diff
@@ -119,15 +119,15 @@ func (ff *filterFlags) orderOp() *pb.OperationOrder {
...
 		f.EnumVar(stateFlagVar(&ff.flagStatusFilter))
 	}
 
-	if opts == fillterOptionAll || opts&filterOptionPhyState != 0 {
+	if opts == filterOptionAll || opts == filterOptionPhyState {
 		f.EnumSingleVar(phyStateFlagVar(&ff.flagPhysState))
 	}
 
-	if opts == fillterOptionAll || opts&filterOptionOrder != 0 {
+	if opts == filterOptionAll || opts&filterOptionOrder != 0 {
```
mitchellh, line 130: should this be `opts ==` too? Or did we mean for this to actually be a bitmask?

If its a bitmask we need to change our constant `iota` above to `1 << iota` I think.

internal/cli/status.go
briancain: Good catch @krantzinator ! That behavior was from me trying to be helpful if worksace was unset and just "try" default. But I think we can rely on the API here to do the right thing if we pass it as unset so I'll remove it.

internal/cli/filter_status.go
briancain: Oops, this needs to be the same change I made with the other if-statements and I had just missed it! I think the original intention was to use a bitmask but checking against `opts` like the change I made above works as intended too.

internal/cli/status.go
```diff
@@ -0,0 +1,830 @@
...
+	for _, projectRef := range projNameList {
+		resp, err := client.GetProject(c.Ctx, &pb.GetProjectRequest{
+			Project: projectRef,
+		})
+		if err != nil {
+			return err
+		}
+
+		var workspace string
+		if len(resp.Workspaces) != 0 {
```
denysvitali, line 442: This part is repeated a couple of times. I guess it's a good candidate for a function

internal/cli/status.go
```diff
@@ -0,0 +1,850 @@
...
+		if a.Name == appTarget {
+			app = a
+			break
+		}
+	}
+	if app == nil {
+		return fmt.Errorf(fmt.Sprintf("Did not find aplication %q in project %q", appTarget, projectTarget))
+	}
+
+	appStatusResp, err := client.GetLatestStatusReport(c.Ctx, &pb.GetLatestStatusReportRequest{
```
izaaklauer, line 262: Same problem here - I think you need more than just the most recent deploy or release status report to determine overall app status.

One piece of WP-075 is building APIs that get you a deployment/release with it's corresponding latest status report, which I think would be useful here. I've been holding off building that until WP-080 stabilizes, but I think it's just what you need here.

izaaklauer (changes requested):
Nice! I'm really looking forward to this feature 🤩. 

It might be worth adding screenshots of the CLI output to the description.

internal/cli/status.go
```diff
@@ -0,0 +1,850 @@
...
+
+	appHeaders := []string{
+		"App", "Workspace", "Latest Status", "Last Check",
+	}
+
+	appTbl := terminal.NewTable(appHeaders...)
+
+	appFailures := false
+	for _, app := range project.Applications {
+		appStatusResp, err := client.GetLatestStatusReport(c.Ctx, &pb.GetLatestStatusReportRequest{
```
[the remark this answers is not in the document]
briancain: That's expected in this case I think. The column we're displaying for this app is the _latest status_ check recorded rather than an aggregate of its status history.

internal/cli/status.go
```diff
@@ -0,0 +1,850 @@
...
+	if len(respDeployList.Deployments) > 0 {
+		deploy := respDeployList.Deployments[0]
+		statusColor := ""
+
+		columns := []string{
+			deploy.Application.Application,
+			fmt.Sprintf("v%d", deploy.Sequence),
+			deploy.Status.State.String(),
+			deploy.Id,
+			deploy.ArtifactId,
```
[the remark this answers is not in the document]
briancain: Yep good call, I removed these ids, I don't think they were very useful! I'm just showing sequence id in the way `deployment list` shows it now (i.e. `v<sequence_id>`)

internal/cli/status.go
```diff
@@ -0,0 +1,850 @@
...
+			Workspace: workspace,
+		},
+		LoadDetails: pb.Deployment_BUILD,
+	})
+	if err != nil {
+		return err
+	}
+
+	deployHeaders := []string{
+		"App Name", "Version", "Physical State", "Id", "Artifact Id", "Exec", "Logs",
```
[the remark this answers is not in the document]
briancain: I might as well update it to be lifecycle! Probably matches closer to our docs on what an operation is too here.

internal/cli/status.go
```diff
@@ -0,0 +1,850 @@
...
+	}
+
+	deployTbl := terminal.NewTable(deployHeaders...)
+
+	resourcesHeaders := []string{
+		"Type", "Platform", "Category",
```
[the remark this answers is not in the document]
briancain: I felt like for certain headers passing in the string slice directly ran off the screen and wasn't too readable so I kept them separate for now.

internal/cli/status.go
briancain: Since this is the most detailed view, I could basically grab the status report associated with the latest deployment and or release (if a release happened) and show them both here in separate columns.

internal/cli/status.go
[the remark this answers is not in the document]
izaaklauer: Hmm, this deviates from the logic in the RFC:

> For the top level status command, we want to let users know the health of a project given all of its applications health. When we determine the value of this column, we’ll need to look through each application's overall health status and show the total statuses for each app. We will omit any status state that isn’t covered. For example, if we have 3 apps, 2 of which are READY and 1 is error, we should show “2 READY 1 ERROR” for the overall aggregate app status.

I think I much prefer the RFC logic. As an app developer, I don't think this field is useful to me if it's flapping between new and old deployments and releases. I would expect this to be some kind of view into the health of my application.

With the bundle APIs introduced in https://github.com/hashicorp/waypoint/pull/1994, it should be pretty easy to get the latest status report for the latest deploy, the latest status report for the latest release, and then either display them both, or one combined status that can be `PARTIAL`.

internal/cli/status.go
```diff
@@ -0,0 +1,992 @@
...
+		// TODO(briancain): Add resource health when it exists
+		if statusReport != nil {
+			if statusReport.Health.HealthStatus == "ERROR" ||
+				statusReport.Health.HealthStatus == "DOWN" {
+				appFailures = true
+			}
+
+			for _, dr := range statusReport.Resources {
+				columns := []string{
+					dr.Name,
```
izaaklauer, line 384: You're putting `Name` in the `Type` column here.

izaaklauer (changes requested):

internal/cli/status.go
```diff
@@ -0,0 +1,992 @@
...
+		return err
+	}
+
+	deployHeaders := []string{
+		"App Name", "Version", "Workspace", "Platform", "Details", "Lifecycle State",
+	}
+
+	deployTbl := terminal.NewTable(deployHeaders...)
+
+	resourcesHeaders := []string{
```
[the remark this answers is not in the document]
briancain: Fixed that, I think platform might still be useful and I don't mind duplicating the info too much. Especially in the json view.

internal/cli/status.go
briancain: Fixed!

internal/cli/status.go
```diff
@@ -0,0 +1,992 @@
...
+			Workspace: workspace,
+		},
+		LoadDetails: pb.Deployment_BUILD,
+	})
+	if err != nil {
+		return err
+	}
+
+	// Pregrab status report list
+	statusReportsResp, err := client.ListStatusReports(c.Ctx, &pb.ListStatusReportsRequest{
```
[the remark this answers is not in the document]
briancain: Agreed. I will resolve these for now, because they should all be fixed once the UI methods is merged.

internal/cli/status.go
```diff
@@ -0,0 +1,992 @@
...
+	return workspace, nil
+}
+
+func (c *StatusCommand) getLatestStatusReportByDeployId(
+	statusReportsResp *pb.ListStatusReportsResponse,
+	deployId string,
+) (*pb.StatusReport, error) {
+	for _, statusReport := range statusReportsResp.StatusReports {
+		if deploymentTargetId, ok := statusReport.TargetId.(*pb.StatusReport_DeploymentId); ok {
+			if deploymentTargetId.DeploymentId == deployId {
```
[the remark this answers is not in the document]
briancain: This whole method will be deleted once the UI methods are available so I'll go ahead and resolve this.

internal/cli/status.go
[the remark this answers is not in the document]
briancain: We can just display them both. I'll wait to update it once the UI bundle methods are available.

internal/cli/status.go
```diff
@@ -0,0 +1,992 @@
...
+		workspace, err := c.getWorkspaceFromProject(resp)
+		if err != nil {
+			return err
+		}
+
+		// Get App Statuses
+		var appStatusReports []*pb.StatusReport
+		var ready, alive, down, unknown int
+		for _, app := range resp.Project.Applications {
+			appStatusResp, err := client.GetLatestStatusReport(c.Ctx, &pb.GetLatestStatusReportRequest{
```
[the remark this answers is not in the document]
briancain: I think I'd rather just show both deployment and release statuses as two separate columns in the project view. Then we don't have to guess and people can interpret them separately rather than trying to combine deployment and releases statuses as 1 health report. This will also match the rest of the views too.

krantzinator (approved):
Looks great! Minor nit on formatting of the `waypoint status -app=APP-NAME` result: the `VERSION` and `WORKSPACE` words a bit close together and look like they are the same column. But that might be the responsibility of our cli formatter?
image

briancain:
> But that might be the responsibility of our cli formatter?

Correct...we use [tablewriter](https://github.com/olekukonko/tablewriter) to format this, which is the same package that's used for `waypoint deployment list`. Agreed that the spacing it sometimes chooses is odd 😞

izaaklauer (approved):

PR #1967, opened 2021-07-30
ui: handle incoming base64 strings from grpc-web

### LINKED ISSUE #1923: Waypoint HCL file becomes garbled in the UI if saved on the server via UI
**Description:**

When copy/pasting HCL configuration into the server via the UI, if you go back to edit it later it gets garbled as seen in the screenshot:

image

I assume this is because the config is compressed when stored on the instance to save space, but doesn't get decompressed when viewing it. I know the environment settings are a bit behind, but I checked for issues and changelog stuff and couldn't find anything related to this issue - will update version and check again soon but in the middle of a sprint and it's not a pressing issue on our end so low priority and just wanted to bring awareness.

**Steps to Reproduce**

1. Initialize a new project with a context configured with a remote server and runner via the CLI (`waypoint init`) and then re-run the command to register it with the server so the project shows up in the UI.
2. Copy the `waypoint.hcl` configuration file that was generated (in our case it was still just the boilerplate) and paste it into the box in the UI when managing project settings.
3. Apply the configuration, then go back to project settings and it will be garbled

**Expected behavior**

I would expect to see the original HCL configuration I pasted within the text box so that I could edit it.

Waypoint:
CLI: v0.4.1 (4d5c838e)
Server: v0.4.0
Chrome: Version 91.0.4472.114 (Official Build) (64-bit)

Server and Runner are deployed on an EC2 Ubuntu 20.04 instance. 

In addition, our Server/Runner are running on a custom Docker image that we built since one of our plugins has some dependencies we needed to bake in - all other functionality works as expected so I don't expect our image to be the culprit.

Happy to look into this myself as well as I've become quite familiar with the codebase, but if anyone knows of where the culprit could be and point me in the right direction it would be helpful!

grahamhub:
fixes #1923 

Quick synopsis of the issue: data being received for the fields `waypointHcl` and `privateKeyPem` is being transmitted as base64 encoded strings, which was unaccounted for when rendering that data in the UI, leading it to display the base64 encoded strings instead of the readable data.

The approach for this solution allows the data to continue "living" as base64, and is simply decoded when rendered in the UI (via computed properties `decodedWaypointHcl` and `decodedPrivateKey`). A caveat here is the `on "input"` event listeners. If a user _doesn't_ trigger that event (i.e. they don't change the value), there's no need to re-encode that field back to base64 to be sent to the server. This is why I chose to let the data stay as a base64 string - encoding will happen on input (in the actions `setWaypointHcl` and `setSshAuth`), thus removing the need to encode the data on form submission.


gregone:
@grahamhub Thanks for investigating the bug and for taking the time to dive into this with @jgwhite ! The approach is good, and after trying it out, everything works as expected. 🎉 

I just had to push a changelog entry in https://github.com/hashicorp/waypoint/pull/1967/commits/320d29cdb166044a032fcee2bd526880a3fe0458 to make the checks pass on our side. 

Thanks for contributing 💯 !

gregone (approved):

grahamhub:
thanks for the quick review @gregone - glad to have contributed to something I've been using extensively!

PR #1973, opened 2021-08-02
Allow disabling exec only for the entrypoint

### LINKED ISSUE #1720: Ability to disable `waypoint exec` to mitigate security risk.
The existence of `waypoint exec` creates a security vulnerability in my environment, but the other capabilities of the entrypoint are nice.

I'd like to be able to disable support for `waypoint exec` without disabling the Waypoint entrypoint entirely.

Disabling the Waypoint entrypoint is the only alternative, but it seems like overkill considering that `waypoint logs` is pretty useful.  On the other hand, there are plenty of service mesh sidecar proxy solutions that might, overall, be a better fit for monitoring instances.

-

**Additional context**
-

mitchellh:
Fixes #1720 

This allows the `WAYPOINT_CEB_DISABLE_EXEC` env var to be set to a truthy value to disable `waypoint exec`. This request is threaded through to the server so that the server should not ever assign an exec session to that instance, but also the entrypoint itself prevents any exec session if the server is poorly behaved or outdated.

Docs also updated.

**This also removes the `FindExecInstance` API.** @evanphx and I agreed to remove it since it was never used. And I had to remove it for this PR since it would've had to take into account whether exec was enabled or not and rather than enhancing that API, I'd rather remove it if we were going to remove it anyways.


catsby (approved):
👍

website/content/docs/entrypoint/disable.mdx
```diff
@@ -55,3 +55,28 @@ app "my-app" {
...
 alter your build scripts to prevent entrypoint installation.
+
+## Disabling Specific Features
+
+### Exec
+
+You can disable only the [`waypoint exec`](/docs/exec) functionality by
+setting the `WAYPOINT_CEB_DISABLE_EXEC` environment variable to a truthy
+value (e.g. "1"). Other entrypoint features such as logs, application config,
+etc. will continue to work.
```
izaaklauer, line 66: Might be worth calling out explicitly where this needs to be set. It looks like it needs to be done in some plugin `env` block in waypoint.hcl, and `waypoint config set WAYPOINT_CEB_DISABLE_EXEC=true` doesn't because it runs before the CEB can pull down env vars from the server? I can also imagine someone setting this at build time or on the server though and expecting it to work.

izaaklauer (changes requested):

website/content/docs/entrypoint/disable.mdx
mitchellh: I do already in the last paragraph. Should I move that up?

mitchellh: I moved it up and added an example for k8s

krantzinator (approved):
Works great!

izaaklauer (approved):

briancain (approved):

PR #1980, opened 2021-08-04
internal/server: add data source to operations

mitchellh:
Replaces #1920

This preloads the data source ref to operations. The data source ref is always available using an N+1 query to get the job via the `job_id`, but it is a frequent piece of data so we load it on every operation read.

I reused the `Preload` field semantics for this because (1) we already had it and (2) the app operation internals have some automation around preload already to ensure it is never persisted to disk, and that it is non-nil on read. We can reuse these semantics since we want that same behavior with the ref field.

Note there are a variety of cases the job ref might be nil, as documented in the proto docs.


gregone (approved):
Looks great @mitchellh! Thank you for taking to time to add this!

briancain (approved):
Awesome!

PR #1984, opened 2021-08-04
Improve ECS Server uninstall

### LINKED ISSUE #1974: waypoint uninstall in ECS leaves tags of deleted resources
**Describe the bug**
Repeatedly installing/uninstalling a waypoint server accumulates resources in ECS that were deleted, but are listed in the search by tag and are only visible in the [AWS CLI](https://gist.github.com/psihachina/13262db33102e29d321c979c9812822d). This results in an error on the following waypoint server uninstallation. Deleted resources are ELB listeners and EFS Access points. These resources are not found as [supported](https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html) in the resources group in the official documentation, but these resources can tag in AWS GO SDK. Maybe is a bug in AWS.

**Steps to Reproduce**

To see the removed resources in the AWS CLI, you need to run waypoint server `install -platform=ecs` / `waypoint server uninstall -auto-approve -platform=ecs` several times

If you repeat (a lot) of these actions, there will be an [error](https://gist.github.com/psihachina/4ec4844deb230eaa681cf95e59f12122).

**Expected behavior**
Installing and uninstalling waypoint in ECS without error

- Waypoint CLI Version: `main`
- Waypoint Server Platform and Version:  `ecs`

**Additional context**
One solution to the problem would be to remove the tag during installation for ELB listener and EFS Access point.

Also the [difference](https://gist.github.com/psihachina/0e86c2fef48a45b9596f21b7d9714362) between serving resources by tag in the GO AWS SDK in a tag-cluttered region and almost clean

catsby:
Improve ECS Server uninstall by getting the full list of tagged resources, and handling "not found" type errors when destroying resources. This PR also removes the code that was used to tag ELB Listeners and EFS Access Points. The purpose of the tags is for use in AWS Resource Manager queries to find all the tagged resources, and at this time those 2 resource types [are not currently supported][awsrm]. This is deemed safe because the resource identifiers that would be associated with those tags are not used themselves during server uninstall; they are cleaned up during the process of cleaning up their parent resources (Network load balancer and EFS file system, respectively). 

    - Check full paginated list of tag based search results
    - Bump the time allowed for the security group to be free of it's
      dependency

This should mostly resolve #1974 in that it handles "not found" errors cited in that issue. 

[awsrm]: https://docs.aws.amazon.com/ARG/latest/userguide/supported-resources.html

### DIFF
--- a/.changelog/1984.txt
+++ b/.changelog/1984.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+serverinstall/ecs: handle errors when resources are already destroyed
+```
--- a/internal/serverinstall/ecs.go
+++ b/internal/serverinstall/ecs.go
@@ -1393,12 +1393,6 @@ EFSLOOP:
 			},
 			Path: aws.String("/waypointserverdata"),
 		},
-		Tags: []*efs.Tag{
-			{
-				Key:   aws.String(defaultServerTagName),
-				Value: aws.String(defaultServerTagValue),
-			},
-		},
 	})
 	if err != nil {
 		return nil, fmt.Errorf("error creating access point: %w", err)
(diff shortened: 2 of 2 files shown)


briancain (approved):
👍🏻

PR #1994, opened 2021-08-05
UI bundle apis for deployments and releases

izaaklauer:
Lists deployments and releases, and bundles with latest status report. First UI bundle method following the pattern laid out in WP-080. 

Implementing this now to avoid having to write very similar logic inside the status report CLI method.

Testing this is a little tricky as we'd need to set up deployments, releases, and status reports, so leaving it as a TODO for now. Will tackle later.


internal/server/singleprocess/service_ui_deploy.go
```diff
@@ -0,0 +1,64 @@
...
+		state.ListWithOrder(req.Order),
+		state.ListWithWorkspace(req.Workspace),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: make this more efficient. We should be able to just
+	// grab the relevant status report in one go, not have to iterate
+	// over all of them.
```
briancain, line 26: You could enhance this to say we need to redo how `GetLatestStatusReport` is implemented. Right now it just calls its inherited func from app operation to get the latest item in the database. For us to target a deployment or release status report we'll have to not use that abstraction and instead write our own query for grabbing a status report if a target is requested.

internal/server/proto/server.proto
```diff
@@ -319,8 +319,78 @@ service Waypoint {
...
+
+  message ListDeploymentsRequest {
+    // Application that the deployments belong to.
+    Ref.Application application = 1;
+
+    // Workspace that the deployments belong to.
+    // If undefined, deployments from all workspaces will be listed.
+    Ref.Workspace workspace = 2;
+
+    OperationOrder order = 3;
```
briancain, line 350: Looks like the RFC had pagination settings, we should include that I think.

internal/server/proto/server.proto
```diff
@@ -319,8 +319,78 @@ service Waypoint {
 
   // UpsertStatusReport updates or inserts a statusreport.
   rpc UpsertStatusReport(UpsertStatusReportRequest) returns (UpsertStatusReportResponse);
+
+  //----------------------------------------------------------------------
+  // UI endpoints. The endpoints below are convenience methods for waypoint's
```
briancain, line 324:
```suggestion
  // UI endpoints. The endpoints below are convenience methods for Waypoint's
```

briancain (changes requested):
I left most of my comments on the Deployment side but they also apply to the Release side too!

internal/server/proto/server.proto
izaaklauer: I left it out for now because I don't plan to implement it right now, and I thought it would be dishonest to have the field that doesn't do anything.

internal/server/singleprocess/service_ui_deploy.go
```diff
@@ -0,0 +1,64 @@
...
+
+	for _, deploy := range deployList {
+		var matchingStatusReport *pb.StatusReport
+	MATCH_STATUS_REPORT_LOOP:
+		for _, statusReport := range statusReports {
+			switch target := statusReport.TargetId.(type) {
+			case *pb.StatusReport_DeploymentId:
+				if target.DeploymentId == deploy.Id {
+					matchingStatusReport = statusReport
+					break MATCH_STATUS_REPORT_LOOP
```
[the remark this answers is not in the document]
izaaklauer: Break actually breaks the switch case, believe it or not! It took me a while to figure out why my break statement wasn't working 😆.

internal/server/singleprocess/service_ui_deploy.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/waypoint/internal/server/singleprocess/state"
+
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+)
+
+// TODO: test
```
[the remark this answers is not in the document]
izaaklauer: Done

internal/server/singleprocess/service_ui_release.go
```diff
@@ -0,0 +1,61 @@
...
+
+	var releaseBundles []*pb.UI_ReleaseBundle
+
+	for _, release := range releaseList {
+		var matchingStatusReport *pb.StatusReport
+		for _, statusReport := range statusReports {
+			switch target := statusReport.TargetId.(type) {
+			case *pb.StatusReport_ReleaseId:
+				if target.ReleaseId == release.Id {
+					if matchingStatusReport == nil || statusReport.GeneratedTime.GetSeconds() > matchingStatusReport.GeneratedTime.Seconds {
```
briancain, line 46: I think this could use a comment here. If I'm understanding, we're doing this check to ensure we get the most recent status report for a release (or deployment in the other file).

briancain (approved):
Sweet, looks good! I think the matching status report if-statement could use a comment for why just so it's clear, but otherwise looks great!

internal/server/proto/server.proto
jgwhite: Good thing you didn't touch the pagination stuff because [further input on the RFC](https://docs.google.com/document/d/1RHuMDPJlWPWoZYx9AkRTw8ppHkg9j7RZwBjA4uUcbcQ/edit?disco=AAAANZQwK30) from @dadgar suggests we should revisit the pagination schema.

I think I’ll write a short follow-up RFC, but it’ll probably be a little simpler/flatter:

```diff
  message ListDeploymentsRequest {
    // Application that the deployments belong to.
    Ref.Application application = 1;

    // Workspace that the deployments belong to.
    // If undefined, deployments from all workspaces will be listed.
    Ref.Workspace workspace = 2;

-   OperationOrder order = 3;
+   OperationOrder deprecated_order = 3;

    // The filters to apply to this request. These are ORed, so you should
    // specify multiple filters in the StatusFilter for AND behavior.
    repeated StatusFilter status = 4;

    // The physical state to filter for. If this is zero or unset then no
    // filtering on physical state will be done.
    Operation.PhysicalState physical_state = 5;

+   repeated string order_by = 6;
+   uint32 page_size = 7;
+   string next_page_token = 8;
+   string previous_page_token = 9;
  }
```

PR #1998, opened 2021-08-06
internal/server: Ensure ExpediteStatusReport sets a default workspace

briancain:
Prior to this commit, if a caller requests a status report job but
doesn't specify a workspace, the server would panic on the Workspace ref
being unset when it attempted to build the job. This commit fixes it by
defaulting to the default workspace if the caller does not specify which
workspace the report should run in.


gregone (approved):
Thanks!

briancain:
I'll go ahead and merge this to help unblock the UI work.

PR #2001, opened 2021-08-06
Fix for auto-backporting: max out pagination limits

krantzinator:
We automatically apply `backport/website` PRs to the most recently created `backport/0.X.x` label as well, so that when we release off of a `release/0.X.x` branch, we can force-push it on top of `stable-website` and include both release-specific website updates and any updates that have been applied selectively to `stable-website` between releases.

The GitHub API defaults to 30 items per page in its pagination of responses, which we recently exceeded in our total labels, so the backports were being applied to `release/0.1.x`. The max items per page is 100, so this fix will ensure this functionality is good until we go over 100 labels 😁 .

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -22,7 +22,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.ELEVATED_GITHUB_TOKEN }}
       - name: Run Backport Assistant backport website changes to latest release branch
         run: |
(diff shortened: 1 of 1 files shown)


izaaklauer (approved):
Should buy us another year probably!

gregone (approved):
~this will break again once we have 100 releases, right? 😅~

Yes it will, I didn't read the description

Issue #2002, opened 2021-08-06
Deployment disappears from list even if a `destroy` call fails


krantzinator:
**Describe the bug**
If you run `waypoint deployment destroy` and the destroy fails, it still removes the deployment from Waypoint's list of deployments, therefore leaving whatever resources were attached to the deployment stuck.

**Steps to Reproduce**
Due to a bug with the GCP plugin in destroying deployments, you can test using that one currently.

* Spin up a GCP Waypoint setup (can use [our Learn guide](https://learn.hashicorp.com/tutorials/waypoint/google-cloud-run?in=waypoint/deploy-google-cloud))
* run `waypoint up`
* verify you see one deployment in `waypoint deployment list`
* run `waypoint deployment destroy -all`
* see an error message in your terminal that says the revision cannot be stopped
* verify you see 0 deployments in `waypoint deployment list`
* verify in your GCP console that the revision for the waypoint app's service was not destroyed

You can also try `waypoint destroy -auto-approve` to observe the same behavior. The first destroy fails, then subsequent destroys appear successful because it no longer has deployments it is attempting to destroy.

**Expected behavior**
Deployment destroys that fail should not remove that deployment from Waypoint's consciousness.

**Waypoint Platform Versions**

All

PR #2005, opened 2021-08-06
plugin/google: several improvements related to resource management

### LINKED ISSUE #1945: Update GCR plugin to use Resource Manager

### LINKED ISSUE #1985: google/cloudrun: CloudRun plugin doesn't destroy active revisions or services
**Describe the bug**
The Google CloudRun plugin only ever destroys inactive revisions created underneath a service. This means Waypoint doesn't clean up everything it creates when `waypoint destroy` is run, because active revisions and the parent service are not destroyed.

**Steps to Reproduce**
Run `waypoint up` a couple of times using [the Learn guide for the GCP plugin](https://learn.hashicorp.com/tutorials/waypoint/google-cloud-run?in=waypoint/deploy-google-cloud).

Navigate to Cloud Run in your GCP Console, then click on the service that corresponds to your Waypoint deployment. You should see two revisions listed.

Run `waypoint deployment destroy` to see a successful destroy of both a deployment on the Waypoint side, and the revision in your GC console for the service that was created.

Now run `waypoint destroy`. You should receive an error message like:
```
! Error destroying the deployment: rpc error: code = Unknown desc = googleapi: Error 400: Revision "example-python-js5z8" cannot be directly deleted because it
is actively serving., failedPrecondition
```

This is because we are attempting to delete the current revision. We should be destroying the entire service at this point since a destroy of all Waypoint-created resources was called. (Note: you cannot delete an active/latest revision; you must delete the entire service; [see docs](https://cloud.google.com/run/docs/managing/revisions#delete).)

If you run `waypoint destroy` again, it appears successful because Waypoint did successfully destroy its record of the deployment. But the service and revision are still active in GCP.

A similar pattern is observed if you run `waypoint deployment destroy -all`.

**Expected behavior**
I expect `waypoint destroy` to destroy the service that was created. We should implement  `DestroyWorkspace` for this purpose.

I am not sure if I expect `waypoint deployment destroy -all` to destroy the service or not. Since we can't destroy an active revision, I guess I'd expect it to at least provide platform-specific messaging. So instead of showing the platform's error of `Revision "example-python-js5z8" cannot be directly deleted because it
is actively serving.,`, maybe a warning-style message like, ```You cannot delete the current active revision for this service. To destroy the service, run `waypoint destroy`.```

applicable:

* Waypoint CLI Version: current
* Waypoint Server Platform and Version: all
* Waypoint Plugin: google/cloudrun

krantzinator:
This is kind of a lot but they were all interconnected so I figured it was still small enough to include in one PR, as change requests for one part may affect other parts here anyway.

Things done:
* Fixes #1945
  * Add Resource Manager to the plugin
  * Note: this plugin already handled its own resources quite well, so I wasn't sure how the current setup should relate to a Resource Manager setup. I opted to integrate rather than consolidate the two approaches. Feedback welcome.
* Fixes #1985 
  * Stopped trying to delete active revisions, and replace error with a more helpful warning message
  * Implemented DestroyWorkspace for `waypoint destroy` calls to specifically target the service rather than just revisions

_(note: there is still a bug with deployment destroys, see #2002; because of that, the first example below requires running `waypoint up` between destroy attempts in order to accurately see the errors.)_
Before changes:
```
$ waypoint deployment destroy -all

» 1 deployments will be destroyed.
  Destroying deployment: 01FCEJ15WCCFCTV19G1WXAS42N
! Error destroying the deployment: rpc error: code = Unknown desc = googleapi: Error 400: Revision "example-python-dw8lp" cannot be directly deleted because it is actively serving.,
failedPrecondition

$ waypoint destroy -auto-approve

» Destroying releases for application 'example-python'...

» Destroying deployments for application 'example-python'...
! Error destroying: rpc error: code = Unknown desc = googleapi: Error 400: Revision "example-python-nlmzx" cannot be directly deleted because it is actively serving.,
failedPrecondition
```

After changes:
```
$ waypoint deployment destroy -all

» 2 deployments will be destroyed.
  Destroying deployment: 01FCEHET53H4SEE7AA080SD3MV
namespaces/rae-waypoint-test-003/revisions/example-python-r7tmw
⚠️ Cannot destroy deployment with revision "namespaces/rae-waypoint-test-003/revisions/example-python-r7tmw", as revision is actively receiving traffic.
  Destroying deployment: 01FCEH7B80HPB4RTVP2RZEQDMW
namespaces/rae-waypoint-test-003/revisions/example-python-fr2tr

$ waypoint destroy -auto-approve

» Destroying releases for application 'example-python'...

» Destroying shared deploy resources for application 'example-python'...
namespaces/rae-waypoint-test-003/services/example-python
```


builtin/google/cloudrun/platform.go
```diff
@@ -383,7 +387,7 @@ func (p *Platform) Deploy(
 
 		service, err = deployment.createService(ctx, service)
 		if err != nil {
-			return nil, status.Errorf(codes.Aborted, "Unable to create Cloud Run service: %s", err.Error())
+			return status.Errorf(codes.Aborted, "Unable to create Cloud Run service: %s", err.Error())
```
catsby, line 390: Nit: In this method sometimes we error out with `fmt.Errorf` and others like this we use `status.Errorf`. I realize you may not have been the author of this but I'm curious if you think we should use this chance to standardize on one?

catsby (approved):
One minor comment/suggestion, otherwise 👍

builtin/google/cloudrun/platform.go
krantzinator: Makes sense to me! Done.

briancain (approved):
Looks good! ✨

builtin/google/cloudrun/platform.go
```diff
@@ -392,27 +396,140 @@ func (p *Platform) Deploy(
...
+
+	// Store our resource state
+	deployment.ResourceState = rm.State()
+
+	// Get our service state
+	servState := rm.Resource(rmResourceServiceName).State().(*Resource_Service)
+	if servState == nil {
+		return nil, status.Errorf(codes.Internal,
+			"service state is nil, this should never happen")
+	}
```
izaaklauer, line 532: What is the purpose of this chunk?

krantzinator: It was there previously; I assume it's one of those "this should ever happen but just in case it does, let's exit early with an explicit error"

krantzinator: Oh wait. I lied. that's my fault!

krantzinator: Yeah OK double-checked, and it's basically a check on "did we set the state on the resource manager, is it there." It's arguably not needed, but in the weird case that it _is_ needed, we don't get into a situation where the user expects state management to be there and it isn't.

izaaklauer (approved):

PR #2008, opened 2021-08-09
`deployment list` command uses new UI API

izaaklauer:
Prior to this, `deployment list` called ListStatusReports, which in this release could return too much data and result in an error.

This converts it to use the new UI_DeploymentList method. It also updates that method to contain all of the old Preload fields, as the DeploymentBundle is meant to supersede preload. 

Also added much more nil-checking based on panics I encountered while testing. All the fields should be non-nill now, but better to fill in less data than to panic.

Note that this breaks backward compat with the pre-0.5 server - we've decided that's acceptable, and i'm emitting a helpful error asking people to upgrade.


internal/cli/deployment_list.go
```diff
@@ -211,27 +201,33 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
-				details = append(details, "build-user:"+user)
-			}
+			} else if deployBundle.Build != nil {
+				build := deployBundle.Build
+				// preload labels have been set, safe to use them
 
-			if bp, ok := b.Preload.Build.Labels["common/languages"]; ok {
-				details = append(details, niceLanguages(bp))
-			}
+				if user, ok := build.Labels["common/user"]; ok {
```
briancain, line 208: I think we end up appending this twice? We're also looking at this label on 202.

briancain: Although one is user and one is build-user 🤔 not sure the difference.

internal/cli/deployment_list.go
```diff
@@ -95,34 +95,26 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
 			PhysicalState: phyState,
 			Status:        c.filterFlags.statusFilters(),
-			Order:         c.filterFlags.orderOp(),
-			LoadDetails:   pb.Deployment_BUILD,
 		})
 		if err != nil {
+			if s, ok := status.FromError(err); ok {
+				if s.Code() == codes.Unimplemented {
+					c.project.UI.Output(
+						"This CLI version is incompatible with the current server - missing UI_ListDeployments method. Upgrade your server or downgrade your CLI.",
```
mitchellh, line 108: I would mention the version of the server that you need. You can also use `c.project.ServerVersion()` I think to get the current server version (its cached, the client gets it on initialiization) if you want to show that to the user.

internal/cli/deployment_list.go
```diff
@@ -211,27 +215,33 @@ func (c *DeploymentListCommand) Run(args []string) int {
...
+				if len(details) == 0 {
+					details = append(details, artDetails)
+				} else if c.flagVerbose && deployBundle.Build != nil {
+					details = append(details,
+						artDetails,
+						fmt.Sprintf("build:%s", c.flagId.FormatId(deployBundle.Build.Sequence, deployBundle.Build.Id)))
+				}
 			}
 
 			if c.flagVerbose {
```
briancain, line 247: I can't reach this comment on the PR, but there's a panic on line 246. It should use the deployment bundle instead of `b.Preload.Artifact`. I think the whole `-verbose` output needs this fix up.

briancain (changes requested):
Gave the updates a shot and found a panic! ❗I also had a comment about the `common/user` label showing up twice. Otherwise I think these updates look good!

internal/cli/deployment_list.go
izaaklauer: I was just replicating the old logic, but with more nil checking 🙃. It looks to me like we'll either set that `user` field from the top-level label's `common/user`, or from the build labels `common/user` if the former isn't present. Either way, it should be the same as before.

briancain: I see! Should the details here just be `user:` then if it's the same?

internal/cli/deployment_list.go
izaaklauer: Oh thank you! I thought I found all the old uses of `Preload`, but I missed those _and_ neglected to test `verbose`. Fixing now.

internal/cli/deployment_list.go
izaaklauer: I don't think it'll necessarily be the same - if we don't find user on top level labels and resort to pulling it off the build, I think it makes sense to say we're showing the `build-user`.

internal/cli/deployment_list.go
briancain: Hm, there's more. Still getting a panic on 260, and once that's resolved there will be another one on 270.

izaaklauer: Yup, just pushed fixes for the last of them.

I also updated the integration test to do a few list calls, in leiu of actual tests for the CLI commands, which should help catch future panics.

briancain (approved):

briancain (approved):

PR #2010, opened 2021-08-09
cli/docs: Generate usage text from commands Help() func

briancain:
Prior to this pull request, our CLI docs auto-gen would manually format the
Usage string based on the command name. This would leave out various
command line arguments or other custom formatting. This pull request fixes
this behavior by looking at the output from the commands `Help()` function
and attempt to parse and format its Usage string in the way it would show
up on a terminal with `-help`.

### DIFF
--- a/internal/cli/artifact_build.go
+++ b/internal/cli/artifact_build.go
@@ -73,7 +73,7 @@ func (c *ArtifactBuildCommand) Synopsis() string {
 func (c *ArtifactBuildCommand) Help() string {
 	return formatHelp(`
 Usage: waypoint artifact build [options]
-Alias: waypoint build
+Alias: waypoint build [options]
 
   Build a new versioned artifact from source.
 
--- a/website/content/commands/artifact-list-builds.mdx
+++ b/website/content/commands/artifact-list-builds.mdx
@@ -17,6 +17,8 @@ List builds.
 
 Usage: `waypoint artifact list-builds [options]`
 
+List builds.
+
 #### Global Options
 
 - `-plain` - Plain output: no colors, no animation.
--- a/website/content/commands/artifact-list-builds.mdx
+++ b/website/content/commands/artifact-list-builds.mdx
@@ -17,7 +17,9 @@ List builds.
 
 Usage: `waypoint artifact list-builds [options]`
 
-List builds.
+List artifacts created from a build. An artifact is the result of a build or
+registry. This is the metadata only. The binary contents of an artifact are
+expected to be stored in a registry.
 
 #### Global Options
 
--- a/website/content/commands/docs.mdx
+++ b/website/content/commands/docs.mdx
@@ -20,7 +20,7 @@ Usage: `waypoint docs [options]`
 Output documentation about the plugins. By default, it lists the documentation
 for the plugins configured by this project.
 
-    The flags can change which plugins are listed and in which format.
+The flags can change which plugins are listed and in which format.
 
 #### Global Options
 
--- a/website/content/commands/docs.mdx
+++ b/website/content/commands/docs.mdx
@@ -17,6 +17,11 @@ Show documentation for components
 
 Usage: `waypoint docs [options]`
 
+Output documentation about the plugins. By default, it lists the documentation
+for the plugins configured by this project.
+
+    The flags can change which plugins are listed and in which format.
+
 #### Global Options
 
 - `-plain` - Plain output: no colors, no animation.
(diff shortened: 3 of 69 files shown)


website/content/commands/server-restore.mdx
```diff
@@ -15,7 +15,20 @@ Stage a snapshot on the current server for data restoration
...
+
+    If -exit is passed, the server process will exit after staging the data. This allows a process
+    monitor to restart the server, where it will see the staged snapshot and restore the data.
+
+    If -exit is not passed, an operator must restart the server manually to finish the restoration
+    process.
+
+    The argument should be to a file written previously by 'waypoint server snapshot'.
+    If no name is specified and standard input is not a terminal, the backup will read from
+    standard input. Using a name of '-' will force reading from standard input.
```
izaaklauer, line 31: Same issue here - this whole thing is in a code block:

Screen Shot 2021-08-10 at 9 48 54 AM

website/content/commands/docs.mdx
```diff
@@ -17,6 +17,11 @@ Show documentation for components
 
 Usage: `waypoint docs [options]`
 
+Output documentation about the plugins. By default, it lists the documentation
+for the plugins configured by this project.
+
+    The flags can change which plugins are listed and in which format.
+
```
izaaklauer, line 24: This indented line shows up in a code block:
Screen Shot 2021-08-10 at 9 47 29 AM

internal/cli/artifact_build.go
```diff
@@ -73,7 +73,7 @@ func (c *ArtifactBuildCommand) Synopsis() string {
 func (c *ArtifactBuildCommand) Help() string {
 	return formatHelp(`
 Usage: waypoint artifact build [options]
-Alias: waypoint build
+Alias: waypoint build [options]
```
izaaklauer, line 76: As it stands, Usage and Alias show up on the same line:

Screen Shot 2021-08-10 at 9 41 26 AM

I think we probably want those on separate lines, like the CLI does. We'd need a little more regex then though - just adding a newline causes the alias command to lose code block formatting.

website/content/commands/server-snapshot.mdx
```diff
@@ -15,7 +15,12 @@ Write a backup of the server data
 
 ## Usage
 
-Usage: `waypoint server snapshot [options]`
+Usage: `waypoint server snapshot [<filename>]`
+
+    Generate a snapshot from the current server and write it to a file specified
+    by the given name. If no name is specified and standard out is not a terminal,
+    the backup will be written to standard out. Using a name of '-' will force writing
+    to standard out.
```
izaaklauer, line 23: Indentation causes codeblock 
Screen Shot 2021-08-10 at 9 49 50 AM

website/content/commands/artifact-list-builds.mdx
```diff
@@ -17,6 +17,8 @@ List builds.
 
 Usage: `waypoint artifact list-builds [options]`
 
+List builds.
```
izaaklauer, line 20: Just pointing out here that our `description` is often identical to our help text, which makes the page look a bit repetitive:

Screen Shot 2021-08-10 at 9 51 35 AM

Not a big problem though, and I don't have any great ideas for how to make it better.

izaaklauer (approved):
What a massive improvement! We have _so much_ useful info buried in the help text.

website/content/commands/artifact-list-builds.mdx
krantzinator: I agree we don't need to fix this in this PR. Maybe the best fix is updating some of our help texts to be more involved than our description.

internal/cli/artifact_build.go
krantzinator: Hm, the docs gen appears to be working for me with just adding a newline like so
```
	fmt.Fprintf(w, fmt.Sprintf("\nAlias: `waypoint %s`\n", aliasMatch[1]))
```

website/content/commands/server-snapshot.mdx
krantzinator: Hey! This is because the formatting in our actual help text is off.
image

If I fix the spacing in the command from a tab to a space, then it fixes both things.

internal/cli/artifact_build.go
izaaklauer: Oh nice! I only tried adding the newline in the help string.

website/content/commands/server-snapshot.mdx
briancain: nice! this is revealing a lot of stuff that should be fixed with our docs it looks like 😄 I can fix this all up

website/content/commands/artifact-list-builds.mdx
briancain: Yeah I think I'll add a better more descriptive sentence for one of those.

PR #2015, opened 2021-08-10
cli/status: Handle case where CLI has no default server context loaded

briancain:
This commit adds some extra handling for when the local CLI has no
default CLI context set. If not set, it will display a warning and
display the server address as an empty string rather than panic on
not having a CLI context to load.

### DIFF
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -54,12 +54,16 @@ func (c *StatusCommand) Run(args []string) int {
 	}
 	ctxName = defaultName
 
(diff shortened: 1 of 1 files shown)


evanphx (approved):

krantzinator:
Wait! I have a suggestion!
Can we include a note in that warning message like `If Waypoint is running in local mode, this is expected.`.

briancain:
Oops @krantzinator - yes totally, I'll make another PR 😄

PR #2016, opened 2021-08-10
Add note about local mode

briancain:
Responding to PR feedback from @krantzinator https://github.com/hashicorp/waypoint/pull/2015#issuecomment-896150686


krantzinator (approved):
🌮

PR #2020, opened 2021-08-10
ui: Add note about compiling static assets for UI changes

briancain:
A quick note about trying out UI changes from pull requests.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -19,6 +19,24 @@ You will need the following things properly installed on your computer.
 - `cd waypoint/ui`
 - `yarn install`
 
+## Testing UI Changes
+
+If you are trying out UI changes on a pull request and don't want to run Ember
+locally, you can build the static assets and compile it into Waypoint server.
+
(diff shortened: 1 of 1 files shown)


izaaklauer:
Unless you're editing files directly in `ui/dist`, I think you still need to run `ember build` to update the dist files that `make static-assets` compiles into bindata.

briancain:
Unless something has changed in the last month or two, this was the method suggested by @jgwhite and worked for me when I was testing out the initial status report updates in the UI 🤔

izaaklauer:
> Unless something has changed in the last month or two, this was the method suggested by @jgwhite and worked for me when I was testing out the initial status report updates in the UI 🤔

Ha sorry yes, I deleted my comment because I realized it was wrong, but I wasn't fast enough! For posterity, I suggested that you'd still need to run `ember build`.

You're running `cd ui; make`, which will run `yarn build`, which will do basically the equivalent of `ember build`, so this is fine 👍

izaaklauer (approved):

gregone (approved):

PR #2031, opened 2021-08-11
resolve TODO for docs link

krantzinator:
I used the `tip.waypointproject.io` link and replaced `tip` with `www` so this should be a good link when all is published.


catsby (approved):
Checks out 👍

briancain (approved):

PR #2032, opened 2021-08-11
Update gRPC

mitchellh:
We need to bring in #4561 because gRPC made a breaking change in the
intermediary version that we're forced to upgrade to due to the
buildpack upgrade.


briancain (approved):

mitchellh:
Fixed em!

PR #2036, opened 2021-08-12
UI: Change layout of status-row

almonk:
Some minor changes to fix alignment and type rhythm of the `status-row`. This PR makes the `status-indicator` text the same size as the supporting text on this screen and also moves the status text next to the indicator to avoid uneven whitespace.

Before:
Safari 2021-08-12 at 11 18 42

After:
Safari 2021-08-12 at 11 14 51

Safari 2021-08-12 at 11 14 47

### DIFF
--- a/.changelog/2036.txt
+++ b/.changelog/2036.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Improve design of status row on Build/Deployment/Release detail pages
+```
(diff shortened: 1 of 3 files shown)


gregone (approved):

almonk:
@gregone fwiw I can't merge so I'll leave it to you

PR #2039, opened 2021-08-12
Build poll jobs for the latest deployment and release

izaaklauer:
Previously, if a release was present, we would stop polling on the latest deployment. This means we would stop looking at the statuses of instances, which are managed (and polled) by the deployments, and are most likely to have status that changes over time.

### How to verify this

- Make a k8s project with a releaser (e.x.: https://github.com/hashicorp/waypoint-examples/tree/main/kubernetes/java)

- Init the project

- Set app polling on the project (e.x. `waypoint project apply -app-status-poll-interval=10s -data-source=git -git-url=https://github.com/hashicorp/waypoint-examples -git-path=kubernetes/java -app-status-poll example-java`)

- Run a `waypoint up`

- Watch the status reports for 30 seconds or so. You should notice the latest deployment _and_ the latest release reports update every 10 seconds:

Screen Shot 2021-08-12 at 12 55 20 PM

### DIFF
--- a/.changelog/2039.txt
+++ b/.changelog/2039.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+core: Always generate status reports for latest deployment and release if present
+```
--- a/.changelog/2039.txt
+++ b/.changelog/2039.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-core: Always generate status reports for latest deployment and release if present
+core: App status polling will always queue status reports refresh jobs for latest deployment and release if present
 ```
--- a/internal/server/singleprocess/poll_application.go
+++ b/internal/server/singleprocess/poll_application.go
@@ -114,11 +114,22 @@ func (a *applicationPoll) buildPollJobs(
 	log.Trace("looking at latest deployment and release to generate status report on")
 	latestDeployment, err := a.state.DeploymentLatest(appRef, &pb.Ref_Workspace{Workspace: a.workspace})
 	if err != nil {
-		return nil, err
+		// If the deployment isn't found, it's ok
+		s, ok := status.FromError(err)
+		if !ok || s.Code() != codes.NotFound {
+			return nil, err
+		}
 	}
 	latestRelease, err := a.state.ReleaseLatest(appRef, &pb.Ref_Workspace{Workspace: a.workspace})
+	if err != nil {
+		// If the release isn't found, it's ok.
+		s, ok := status.FromError(err)
+		if !ok || s.Code() != codes.NotFound {
+			return nil, err
+		}
+	}
 	// Some platforms don't release, so we shouldn't error here if we at least got a deployment
-	if err != nil && latestDeployment == nil {
+	if latestRelease == nil && latestDeployment == nil {
 		log.Warn("no deployment or release found, cannot generate a poll job")
 		return nil, nil
 	}
--- a/internal/server/singleprocess/poll_application.go
+++ b/internal/server/singleprocess/poll_application.go
@@ -142,7 +142,7 @@ func (a *applicationPoll) buildPollJobs(
 			},
 
 			// Application polling requires a data source to be configured for the project
-			// Otherwise a status report can't properly eval the projects hcl context
+			// Otherwise a status report can't properly eval the project's hcl context
 			// needed to query the deploy or release
 			DataSource: project.DataSource,
 
@@ -184,7 +184,7 @@ func (a *applicationPoll) buildPollJobs(
 		}
 		// SingletonId so that we only have one poll operation at
 		// any time queued per app/operation.
-		deploymentJob.Job.SingletonId = fmt.Sprintf("app-status-poll/%s/deployment", app.Name)
+		deploymentJob.Job.SingletonId = appStatusPollSingletonId(app.Name, appStatusPollOperationTypeDeployment)
 
 		jobs = append(jobs, deploymentJob)
 	}
@@ -203,7 +203,7 @@ func (a *applicationPoll) buildPollJobs(
 		}
 		// SingletonId so that we only have one poll operation at
 		// any time queued per app/operation.
-		releaseJob.Job.SingletonId = fmt.Sprintf("app-status-poll/%s/release", app.Name)
+		releaseJob.Job.SingletonId = appStatusPollSingletonId(app.Name, appStatusPollOperationTypeRelease)
 
 		jobs = append(jobs, releaseJob)
 	}
@@ -235,4 +235,18 @@ func (a *applicationPoll) Complete(
 		return err
 	}
 	return nil
+}
+
+// The name of an operation type that status polling is possible for
+type appStatusPollOperationType string
+
+const (
+	appStatusPollOperationTypeDeployment appStatusPollOperationType = "deployment"
+	appStatusPollOperationTypeRelease    appStatusPollOperationType = "release"
+)
+
+// appStatusPollSingletonId generates an application status polling job singleton ID
+// for the given app and operation type.
+func appStatusPollSingletonId(appName string, operationType appStatusPollOperationType) string {
+	return fmt.Sprintf("app-status-poll/%s/%s", appName, operationType)
 }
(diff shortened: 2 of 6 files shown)


internal/server/singleprocess/poll_application.go
```diff
@@ -184,8 +152,57 @@ func (a *applicationPoll) buildPollJob(
...
 		// TODO define timeout interval based on projects app poll interval
 	}
+	var jobs []*pb.QueueJobRequest
+
+	log.Trace("Determining which target to generate a status report on")
+
+	// Default to poll on the "latest" lifecycle operation, so if there's a
+	// release, queue up a status on release. If the latest is deploy, then queue that.
+	if latestRelease != nil && !latestRelease.Unimplemented {
+		releaseJobCopy, err := copystructure.Copy(baseJob)
```
izaaklauer, line 162: Copystructure is my new hammer and now everything looks like a nail 😆

internal/server/singleprocess/poll_application.go
```diff
@@ -184,8 +152,57 @@ func (a *applicationPoll) buildPollJob(
...
+		releaseJob.Job.Operation = &pb.Job_StatusReport{
+			StatusReport: &pb.Job_StatusReportOp{
+				Target: &pb.Job_StatusReportOp_Release{
+					Release: latestRelease,
+				},
+			},
+		}
+		// SingletonId so that we only have one poll operation at
+		// any time queued per app/operation.
+		releaseJob.Job.SingletonId = fmt.Sprintf("appl-poll/%s/deployment", app.Name)
```
izaaklauer, line 176: I was pretty sure I found some kind of weird bug in the polling subsystem until I got to here and realized what the deploy and release jobs needed different singleton IDs:

https://github.com/hashicorp/waypoint/blob/c2455052d7e9ad304ae30a155d04c9764caa17ec/internal/server/singleprocess/state/job.go#L1000

.changelog/2039.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core: Always generate status reports for latest deployment and release if present
```
briancain, line 2:
```suggestion
core: App status polling will always queue status reports refresh jobs for latest deployment and release if present
```

internal/server/singleprocess/poll_application.go
```diff
@@ -67,34 +68,43 @@ func (a *applicationPoll) PollJob(
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Application polling requires a remote data source, otherwise a status report
+	// cannot be generated without a project and its hcl context. This returns
+	// an error so we fail early instead of queueing an already broken job
+	if project.DataSource == nil {
+		log.Debug("cannot poll a job without a remote data source configured.")
+		return nil, status.Error(codes.FailedPrecondition, "application polling requires a remote data source")
```
briancain, line 105: I wrote this originally but this should say `application status polling` to be more specific.

```suggestion
		return nil, status.Error(codes.FailedPrecondition, "application status polling requires a remote data source")
```

internal/server/singleprocess/poll_application.go
```diff
@@ -67,34 +68,43 @@ func (a *applicationPoll) PollJob(
...
 	}
 	log = log.Named(app.Name)
 
+	// App polling needs the parent project to obtain its datasource
+	project, err := a.state.ProjectGet(&pb.Ref_Project{Project: app.Project.Project})
+	if err != nil {
+		return nil, err
+	}
+
+	// Application polling requires a remote data source, otherwise a status report
```
briancain, line 100:
```suggestion
	// Application status polling requires a remote data source, otherwise a status report
```

internal/server/singleprocess/poll_application.go
```diff
@@ -67,34 +68,43 @@ func (a *applicationPoll) PollJob(
...
+	// cannot be generated without a project and its hcl context. This returns
+	// an error so we fail early instead of queueing an already broken job
+	if project.DataSource == nil {
+		log.Debug("cannot poll a job without a remote data source configured.")
```
briancain, line 104:
```suggestion
		log.Debug("cannot build an application poll job without a remote data source configured.")
```

internal/server/singleprocess/poll_application.go
```diff
@@ -184,8 +152,57 @@ func (a *applicationPoll) buildPollJob(
...
+		deploymentJob.Job.Operation = &pb.Job_StatusReport{
+			StatusReport: &pb.Job_StatusReportOp{
+				Target: &pb.Job_StatusReportOp_Deployment{
+					Deployment: latestDeployment,
+				},
+			},
+		}
+		// SingletonId so that we only have one poll operation at
+		// any time queued per app/operation.
+		deploymentJob.Job.SingletonId = fmt.Sprintf("appl-poll/%s/deployment", app.Name)
```
briancain, line 176: This was another one of my things, but we should stay consistent and have this be `app-status-poll`
```suggestion
		deploymentJob.Job.SingletonId = fmt.Sprintf("app-status-poll/%s/deployment", app.Name)
```

internal/server/singleprocess/poll_application.go
```diff
@@ -184,8 +152,57 @@ func (a *applicationPoll) buildPollJob(
...
+		releaseJob.Job.Operation = &pb.Job_StatusReport{
+			StatusReport: &pb.Job_StatusReportOp{
+				Target: &pb.Job_StatusReportOp_Release{
+					Release: latestRelease,
+				},
+			},
+		}
+		// SingletonId so that we only have one poll operation at
+		// any time queued per app/operation.
+		releaseJob.Job.SingletonId = fmt.Sprintf("appl-poll/%s/release", app.Name)
```
briancain, line 195:
```suggestion
		releaseJob.Job.SingletonId = fmt.Sprintf("app-status-poll/%s/release", app.Name)
```

internal/server/singleprocess/poll_test.go
```diff
@@ -668,7 +668,7 @@ func TestApplicationPollHandler_turnoff(t *testing.T) {
 		raw, err := testServiceImpl(impl).state.JobList()
 		for _, j := range raw {
 			if j.State != pb.Job_ERROR &&
-				j.SingletonId == fmt.Sprintf("appl-poll/%s", appName) {
+				j.SingletonId == fmt.Sprintf("appl-poll/%s/deployment", appName) {
```
briancain, line 671: We should add a new test for `TestApplicationPollHandler` to ensure if we have a deployment _and_ release, we return and queue two jobs and then a test with _just_ a deployment and we queue 1 job that's a deployment.

briancain (changes requested):
Thanks for the fixup! This approach looks good I think 👍🏻 Mostly some minor comments from me 😄

internal/server/singleprocess/state/application.go
```diff
@@ -242,18 +242,14 @@ func (s *State) appPollPeek(
...
-	err = s.db.View(func(dbTxn *bolt.Tx) error {
-		var err error
-		result, err = s.projectGet(dbTxn, memTxn, &pb.Ref_Project{
-			Project: rec.Id,
-		})
 
-		return err
+	result, err = s.projectGet(dbTxn, memTxn, &pb.Ref_Project{
+		Project: rec.Id,
 	})
```
izaaklauer, line 253: I _think_ that this fixes an intermittent deadlock I was seeing in the test, but I'm not entirely sure why. In any case, this seems like the right thing to be doing.

internal/server/singleprocess/state/app_operation.go
```diff
@@ -358,7 +358,7 @@ func (op *appOperation) Latest(
 		}
 	}
 
-	return nil, status.Errorf(codes.NotFound, "No application named %q is available!", ref.Application)
```
izaaklauer, line 361: I ran into this case while testing.

mitchellh (approved):

internal/server/singleprocess/state/application.go
mitchellh: Oh yes. Lowercase functions should not be creating transactions at all, they assume they're in a transaction already. Only exported functions setup transactions.

briancain (approved):

PR #2047, opened 2021-08-13
cli/status: fix broken output when no generated time is set

catsby:
If a status report is returned without `GeneratedTime` set, the CLI will fail to parse it:

```
$ waypoint status -app=nginx-example
Current status for application "nginx-example" in project "example-nginx-ami" in server context "waypoint-server-nlb-774e79e21a01ca87.elb.us-west-2.amazonaws.com:9701".
! CLI failed to format app status:timestamp: nil Timestamp
```

In this PR we check for a valid time. If it's not valid, we insert `(unknown)` as a placeholder

### DIFF
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -947,12 +947,12 @@ func (c *StatusCommand) FormatStatusReportComplete(
 		statusReportComplete = "? UNKNOWN"
 	}
 
-	t, err := ptypes.Timestamp(statusReport.GeneratedTime)
-	if err != nil {
-		return statusReportComplete, "", err
+	reportTime := "(unknown)"
+	if statusReport.GeneratedTime.IsValid() {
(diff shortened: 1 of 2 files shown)


internal/cli/status.go
```diff
@@ -947,12 +947,12 @@ func (c *StatusCommand) FormatStatusReportComplete(
 		statusReportComplete = "? UNKNOWN"
 	}
 
-	t, err := ptypes.Timestamp(statusReport.GeneratedTime)
```
catsby, line 950: `ptypes.Timestamp` is deprecated https://pkg.go.dev/github.com/golang/protobuf/ptypes#Timestamp

briancain (approved):
Looks good to me! I'd backport this and I think it's worth a changelog entry

PR #2049, opened 2021-08-13
Update docker and k8s plugins to use the new StatusReport helper func.

izaaklauer:
Depends on https://github.com/hashicorp/waypoint-plugin-sdk/pull/55

The new StatusReport function pushes all of this work into the plugin sdk!

### DIFF
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -1098,7 +1098,7 @@ their current state, run ‘waypoint status -app=APP-NAME PROJECT-NAME’.
 `)
 
 	wpStatusAppSuccessMsg = strings.TrimSpace(`
-The application and its declared resources listed above represents its current state known
+The application and it's resources listed above represents it's current state known
 in the Waypoint server.
 `)
 
--- a/builtin/docker/platform.go
+++ b/builtin/docker/platform.go
@@ -317,7 +317,7 @@ func (p *Platform) Status(
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


internal/cli/status.go
```diff
@@ -1098,7 +1098,7 @@ their current state, run ‘waypoint status -app=APP-NAME PROJECT-NAME’.
 `)
 
 	wpStatusAppSuccessMsg = strings.TrimSpace(`
-The application and its declared resources listed above represents its current state known
+The application and it's resources listed above represents it's current state known
```
izaaklauer, line 1101: Totally incidental change -  we're currently displaying actual resources, not declared resources.

evanphx (approved):

briancain (approved):
> +8 −117

Nice! :chef_kiss: 👨🏻‍🍳 ❤️

internal/cli/status.go
krantzinator: FYI it still would be `its` -- the apostrophe in `it's` does not denote possession, only contraction.
```suggestion
The application and its resources listed above represents its current state known
```

krantzinator (approved):
Nice!

internal/cli/status.go
izaaklauer: 🤦‍♂️🙏

izaaklauer: Fixing in https://github.com/hashicorp/waypoint/pull/2062

PR #2052, opened 2021-08-13
cli/uninstall: Add more validation to requested platform for uninstall

### LINKED ISSUE #2046: `waypoint server uninstall` can delete the wrong context
**Describe the bug**
Running `waypoint server uninstall` deletes the current context, even if that context does not use the platform you're uninstalling.

**Steps to Reproduce**
- Install waypoint with `waypoint install -platform=docker -accept-tos`
- Run a waypoint server manually, generating a bootstrap context
- Run `waypoint server uninstall -platform=docker -auto-approve`
- Notice your waypoint bootstrap context is gone, not your docker context!

**Expected behavior**
Only the waypoint docker context should be deleted. I don't think it's currently possible to use `waypoint install` to get more than one context for any given platform, so scanning the contexts to find the one with the matching platform name should be safe.

* Waypoint CLI Version: 0.5.0
* Waypoint Server Platform and Version: 0.5.0
* Waypoint Plugin: any (tested docker)

briancain:
Prior to this commit, the CLI uninstall command assumed that the
platform requested matched the server platform. This could be confusing
if a user has both a docker platform server and a local server from
`waypoint server run`. This commit adds some extra validation with the
platform requested by loading the server platform and taking into
account what platform was requested.

This commit also changes the hard requirement that the uninstall
command needs a platform flag. If a server context has a platform set,
and no platform flag was requested, use that platform instead. Note
that the platform flag, if requested, always overrides what the server
context has. This is because some early versions of Waypoint do
not set the Platform context. If the flag and server platform conflict,
we'll print a note to say we're honoring the platform flag requested.

Fixes #2046

Screenshot of an example when no server platform was set or requested, and then the warning for if no server platform was set but a platform flag was requested:

Screenshot from 2021-08-13 16-33-30

### DIFF
--- a/.changelog/2052.txt
+++ b/.changelog/2052.txt
@@ -1,0 +1,5 @@
+```release-note:improvement
+cli/uninstall: Remove hard requirement on platform flag, attempt to read server
+platform from server context. Platform flag overrides anytihng set in a server
+platform context.
+```
--- a/internal/cli/uninstall.go
+++ b/internal/cli/uninstall.go
@@ -66,28 +66,43 @@ func (c *UninstallCommand) Run(args []string) int {
 		}
 	}
 
+	// Validate platform requested matches the server contexts platform
 	serverPlatform := c.platform
-	if ctxConfig.Server.Platform == "" && serverPlatform != "" {
-		c.ui.Output(
-			"No platform set on server context. Will attempt to uninstall requested "+
-				"platform %q",
-			serverPlatform,
-			terminal.WithWarningStyle(),
-		)
-	} else if serverPlatform == "" {
-		// attempt to set the server platform so the platform flag isn't required.
-		serverPlatform = ctxConfig.Server.Platform
+	if ctxConfig != nil {
+		if serverPlatform != "" {
+			if ctxConfig.Server.Platform == "" {
+				c.ui.Output(
+					"No platform set on server context. Will attempt to uninstall requested "+
+						"platform %q",
+					serverPlatform,
+					terminal.WithWarningStyle(),
+				)
+			} else if ctxConfig.Server.Platform != serverPlatform {
+				c.ui.Output(
+					"The current server platform is %q but the requested platform through "+
+						"the -platform flag was %q",
+					ctxConfig.Server.Platform,
+					serverPlatform,
+					terminal.WithErrorStyle(),
+				)
 
-		if serverPlatform == "" {
-			// It's still empty
-			c.ui.Output(
-				"Cannot determine what platform to uninstall Waypoint. "+
-					"The -platform flag is required since the server context did not include "+
-					"a server platform.",
-				terminal.WithErrorStyle(),
-			)
+				return 1
+			}
+		} else {
+			// attempt to set the server platform so the platform flag isn't required.
+			serverPlatform = ctxConfig.Server.Platform
 
-			return 1
+			if serverPlatform == "" {
+				// It's still empty
+				c.ui.Output(
+					"Cannot determine what platform to uninstall Waypoint. "+
+						"The -platform flag is required since the server context did not include "+
+						"a server platform.",
+					terminal.WithErrorStyle(),
+				)
+
+				return 1
+			}
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


.changelog/2052.txt
```diff
@@ -0,0 +1,5 @@
+```release-note:improvement
+cli/uninstall: Remove hard requirement on platform flag, attempt to read server
+platform from server context. Platform flag overrides anytihng set in a server
```
krantzinator, line 3:
```suggestion
platform from server context. Platform flag overrides anything set in a server
```

website/content/commands/server-uninstall.mdx
```diff
@@ -17,8 +17,15 @@ Uninstall the Waypoint server
 
 Usage: `waypoint server uninstall [options]`
 
-Uninstall the Waypoint server. The platform should be specified as kubernetes,
-nomad, or docker. '-auto-approve' is required.
+Uninstall the Waypoint server. This command is not intended to uninstall a
+server that was manually run with the 'waypoint server run' CLI, but with
+a Waypoint server that was installed via 'waypoint server install'.
+
+The platform can be specified as kubernetes, nomad, ecs, or docker. If not,
```
krantzinator, line 24:
```suggestion
The platform can be specified as kubernetes, nomad, ecs, or docker. If not
```

internal/cli/uninstall.go
```diff
@@ -55,24 +56,43 @@ func (c *UninstallCommand) Run(args []string) int {
...
+			"No platform set on server context. Will attempt to uninstall requested "+
+				"platform %q",
+			c.platform,
+			terminal.WithWarningStyle(),
+		)
+	} else if serverPlatform == "" {
+		// attempt to set the server platform so the platform flag isn't required.
+		serverPlatform = ctxConfig.Server.Platform
+
+		if serverPlatform == "" && c.platform == "" {
```
krantzinator, line 81: At this point, we've already checked that `c.platform` is empty via the `else if` line checking if `serverPlatform` is empty, since we set `serverPlatform` to `c.platform` on line 69 and don't reset it until after we check if it's empty. Maybe consolidate some of our `c.platform` and `serverPlatform` references so it's clear what we're checking all the way through?
```suggestion
		if serverPlatform == "" {
```

krantzinator (approved):
Didn't test it manually on account of me checking this real quick before I'm offline tomorrow, but the code looks 👍🏻 !

internal/cli/uninstall.go
```diff
@@ -55,24 +56,44 @@ func (c *UninstallCommand) Run(args []string) int {
...
+	if contextDefault != "" {
+		ctxConfig, err = c.contextStorage.Load(contextDefault)
+		if err != nil {
+			c.ui.Output("Error loading context %q: %s", contextDefault, err.Error(), terminal.WithErrorStyle())
+			return 1
+		}
+	}
+
+	serverPlatform := c.platform
+	if ctxConfig.Server.Platform == "" && serverPlatform != "" {
```
evanphx, line 70: `ctxConfig` can be nil here according to if there is a default context or not, so we'll need to guard that.

internal/cli/uninstall.go
```diff
@@ -55,24 +56,44 @@ func (c *UninstallCommand) Run(args []string) int {
...
-				"The -platform flag is required.",
+				"Cannot determine what platform to uninstall Waypoint. "+
+					"The -platform flag is required since the server context did not include "+
+					"a server platform.",
 				terminal.WithErrorStyle(),
 			)
 
 			return 1
 		}
+	}
```
evanphx, line 92: Should we reject if there is a context and a serverPlatform and they conflict about the Platform value?

evanphx (changes requested):
A nil check and a question about if we should check that the platform value agrees.

internal/cli/uninstall.go
briancain: Sounds like a good idea to me!

izaaklauer (approved):

evanphx (approved):

PR #2053, opened 2021-08-16
ui: Fix tab styles for dark mode

almonk:
Fixes the background color and border color of tabs in dark mode

Before
Safari 2021-08-16 at 14 51 00

After
Safari 2021-08-16 at 14 48 37

### DIFF
--- a/.changelog/2053.txt
+++ b/.changelog/2053.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Improve tab styles for dark mode
+```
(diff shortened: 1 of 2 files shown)


sabrinako (approved):
I wonder if it'd be helpful to submit an issue or PR to the structure library about this, but I'm approving for now!

almonk:
@sabrinako I think you're probably right – I actually don't know the extent to which dark mode is supported by structure currently? cc @Dhaulagiri

Dhaulagiri:
Not 100% sure but I believe the answer is minimal support if at all. Definitely on our roadmap to improve this in the design system.

PR #2055, opened 2021-08-16
cli/project: Add project inspect command

### LINKED ISSUE #1267: cli: 'waypoint project inspect <project>' command

When configuring a project for Git runners with polling for example, we allow users to enable the feature through the UI or the CLI with `waypoint project apply -poll`. We allow users to see if git polling is enabled through the UI, but there's no way through the `waypoint project` command to see if git polling (or other options) are configured because the command currently only has `apply` and `list` subcommands.

Provide a new subcommand `inspect` or something else to show project setting details on the CLI.

A different way to list project options on the CLI, maybe.

briancain:
This PR adds a new subcommand to `waypoint project` named `inspect`.
It lists the details of the requested project, or if in a project
directory, list that projects details.

Fixes #1267

### DIFF
--- a/internal/cli/project_inspect.go
+++ b/internal/cli/project_inspect.go
@@ -120,8 +120,8 @@ func (c *ProjectInspectCommand) FormatProject(projectTarget string) error {
 		statusColor := ""
 		columns := []string{
 			project.Name,
-			strings.Join(appNames, "\n"),
-			strings.Join(workspaceNames, "\n"),
+			strings.Join(appNames, ", "),
+			strings.Join(workspaceNames, ", "),
 			strconv.FormatBool(project.RemoteEnabled),
 			dataSource,
 			strconv.FormatBool(projectPollEnabled),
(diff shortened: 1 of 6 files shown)


internal/cli/project_inspect.go
```diff
@@ -0,0 +1,327 @@
...
+// Takes a terminal Table and formats it into a map of key values to be used
+// for formatting a JSON output response
+func (c *ProjectInspectCommand) formatJsonMap(t *terminal.Table) []map[string]interface{} {
+	result := []map[string]interface{}{}
+	for _, row := range t.Rows {
+		c := map[string]interface{}{}
+
+		for j, r := range row {
+			// Remove any whitespacess in key
+			header := strings.ReplaceAll(t.Headers[j], " ", "")
```
evanphx, line 273: We should probably lowercase the keys too.

internal/cli/project_inspect.go
```diff
@@ -0,0 +1,327 @@
...
+			},
+		)
+
+		c.ui.Table(projectTbl)
+	}
+
+	return nil
+}
+
+func (c *ProjectInspectCommand) outputProjectJson(t *terminal.Table) error {
```
evanphx, line 251: I sort of want to make this something on terminal so that we can use it from all the commands. Thoughts? Or maybe for now, just make it a toplevel function rather than something on projectInspectCommand?

evanphx (changes requested):
Couple of things, plus we'll need to modify this once the ondemand runner bits go in.

internal/cli/project_inspect.go
briancain: I want this too. I almost made it a top level function since this function is mostly identical from the one I wrote for the Status command. The only difference is for applications and workspaces I'm splitting on `\n` to make an array. But maybe we're safe to say if we detect any `\n` in a value, it's multiple values and should be an array? I could see this coming in handy a bunch though as we add more CLI helpers.

izaaklauer:
It would be nice if our `waypoint inspect` commands all had similar UX. Here's the output of `waypoint context inspect` (for my tiny bootstrap context):

image

I think that `terminal.WithNamedValues` could make more sense here than a two-column table.

briancain:
Done @izaaklauer

internal/cli/project_inspect.go
```diff
@@ -0,0 +1,261 @@
...
+	var appPollInterval string
+	if project.StatusReportPoll != nil {
+		appPollEnabled = project.StatusReportPoll.Enabled
+		appPollInterval = project.StatusReportPoll.Interval
+	}
+
+	if c.flagJson {
+		projectHeaders := []string{
+			"Project", "Applications", "Workspaces", "Remote Enabled", "Data Source", "Project Poll Enabled",
+			"Project Poll Interval", "App Status Poll Enabled", "App Status Poll Interval",
```
izaaklauer, line 115: Can we show everything that can be configured with `waypoint project apply`? Looks like we're missing `-git-url` and `-git-path`, among other things.

izaaklauer (changes requested):
Looks great! 
Screen Shot 2021-08-17 at 1 11 34 PM

internal/cli/project_inspect.go
briancain: Yes. If those values are unset they won't show up in a NamedValue output though.

izaaklauer: Gotcha. 

I think it's confusing that the `-poll-interval` flag in `project apply`, the 
[`data_source_poll { enabled }` proto](https://github.com/hashicorp/waypoint/blob/main/internal/server/proto/server.proto#L596), and `Project Poll Enabled` here all refer to the same thing. I feel like "Data source polling" is a good vernacular to standardize around, because that's what we're actually polling?

Looking at the project proto, it looks like we're still missing:
- `file_change_signal` (also missing from `waypoint project apply`)
- `waypoint_hcl_format`

briancain: I can show the file change signal, but I'd probably omit the waypoint hcl format, I'm not sure how useful that is or what you'd do with that information 🤔 Also got the headers replaced with _Data Source Poll_!

izaaklauer (approved):

internal/cli/project_inspect.go
briancain: Done!

briancain:
@evanphx - got any more feedback on this? The on-demand runner stuff is in, is there something I should be adding to this PR? Thanks!

mitchellh (approved):

internal/cli/project_inspect.go
```diff
@@ -0,0 +1,248 @@
...
+
+	fileChangeSignal := project.FileChangeSignal
+
+	if c.flagJson {
+		projectHeaders := []string{
+			"Project", "Applications", "Workspaces", "Remote Enabled", "Data Source",
+			"Git URL", "Git Ref", "Git Path", "Data Source Poll Enabled",
+			"Data Source Poll Interval", "App Status Poll Enabled", "App Status Poll Interval",
+			"File Change Signal",
+		}
```
mitchellh, line 125: This is fine if we want to do this (I know we do this in other commands), but for the JSON output I've been just proto-json-encoding w/ `jsonpb` so that it shows the full API structure so as we add new fields I can easily see the values without having to modify the CLI (i.e. if they're not particularly useful fields but I still want to see them for debugging). Not sure if we want to maintain a stable JSON output format here or do something like that.

catsby (approved):
👍

internal/cli/project_inspect.go
briancain: Hm, I did notice that on the ODR list command! I like the idea of having a stable json output at least, but showing the complete picture for a project or other proto would be nice too. Maybe we could adopt the pattern for when we have json output to also include a `-json-raw` that does this? Maybe that's over complicated 🤔

mitchellh: We will [eventually] have a pretty strict backwards compatibility requirement over our API and I believe it'll be easier to maintain compatibility dumping an API object than maintaining a CLI, just my opinion. APIs you _have_ to have backwards compat at a certain point of project maturity, but CLIs we've never covered in our entire portfolio.

briancain: That's a good point, I'll switch this over to just show the raw project output instead.

briancain:
> @evanphx - got any more feedback on this? The on-demand runner stuff is in, is there something I should be adding to this PR? Thanks!

Never mind, included the new field for the ODR config id. That should be it since I don't see any other changes in the project proto 👍🏻

PR #2056, opened 2021-08-16
Introduce On-Demand Runners

evanphx:
This is a big PR and so I suggest reading it a commit at a time rather than all at once. The commits are sequenced and cohesive as standalone changes.

This change corresponds with a plugin-sdk PR as well: https://github.com/hashicorp/waypoint-plugin-sdk/pull/56

This builds on the task plugin work to add the concept on an ondemand runner (or ODR). This commit covers 4 main areas:

1. Introduction of state to register plugin types and configuration, where those fields are used to launch tasks that contain the ondemand runners.
2. Modifications to the runner code (albiet minor) so that a runner knows it's in ODR mode AND advertises this fact to plugins, so they can operate specially if they wish.
3. A new docker builder ODR mode that uses kaniko to execute the Docker build in the context of the runner.
4. A new pack builder ODR mode that uses kaniko to execute the pack build in the context of the runner.


Makefile
```diff
@@ -52,16 +52,15 @@ format: # format go code
 	gofmt -s -w ./
 
 .PHONY: docker/server
-docker/server:
+docker/server: docker/odr
```
briancain, line 55: I think this should be reverted back, right?

Makefile
```diff
@@ -52,16 +52,15 @@ format: # format go code
...
 
-.PHONY: docker/evanphx
-docker/evanphx:
-	DOCKER_BUILDKIT=1 docker build -f hack/Dockerfile.evanphx \
-					--ssh default \
-					-t waypoint:latest \
+.PHONY: docker/odr
+docker/odr:
+	DOCKER_BUILDKIT=1 docker build -f Dockerfile.odr \
+					-t waypoint:latest-odr \
```
briancain, line 63:
```suggestion
					-t waypoint:dev-odr \
```

Maybe just `dev-odr` to match the tag naming that `docker/server` generates?

briancain (commented):
Reviewing the rest of this now...but figured I'd comment on these now since it's not entirely related to the rest of the PR reivew 😄

builtin/docker/kaniko.go
```diff
@@ -0,0 +1,148 @@
...
+
+	step.Done()
+	step = sg.Add("Testing registry and uploading entrypoint layer")
+
+	err = os.SetupEntrypointLayer(refPath, data)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error setting up entrypoint layer: %s", err)
+	}
+
+	li, err := net.Listen("tcp", ":5000")
```
briancain, line 100: This port should probably be configurable? Same with the `img` version?

Makefile
evanphx: Nope, we want to build them both.

Makefile
evanphx: Oh shoot, yeah.

internal/cli/install.go
```diff
@@ -459,6 +459,28 @@ func installRunner(
...
+			if err != nil {
+				s.Update("Error creating ondemand runner: %s", err)
+			} else {
+				s.Update("Registered ondemand runner!")
+			}
+		} else {
+			s.Update("Install type did not provide an ondemand runner config")
+		}
+
+		s.Done()
```
briancain, line 481: It'd be good to close the UX loop here with a stepgroup saying it finished registered the runner.

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+			Target: &c.flagEnvVars,
+			Usage: "Environment variable to expose to the ondemand runner. Typically used to " +
+				"introduce configuration for the plugins that the runner will execute.",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "plugin-type",
+			Target:  &c.flagPluginType,
+			Default: "",
+			Usage:   "The type of the plugin to launch for the ondemand runner, such as aws-ecs, kubernetes, etc.",
```
briancain, line 215: We should include valid plugin types here in the usage description like we do `waypoint install`.

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (c *OndemandRunnerApplyCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint ondemand-runner apply [OPTIONS]
+
+  Create or update an ondemand runner.
+
+  This will create a new ondemand runner with the given options. If
+	a ondemand runner with the same id already exists, this will update the
+	existing runner using the fields that are set.
```
briancain, line 257:
```suggestion
  This will create a new ondemand runner with the given options. If
  a ondemand runner with the same id already exists, this will update the
  existing runner using the fields that are set.
```

internal/cli/ondemand_runner_list.go
```diff
@@ -0,0 +1,72 @@
...
+func (c *OndemandRunnerListCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (c *OndemandRunnerListCommand) AutocompleteFlags() complete.Flags {
+	return c.Flags().Completions()
+}
+
+func (c *OndemandRunnerListCommand) Synopsis() string {
+	return "List all registered projects."
```
briancain, line 60:
```suggestion
	return "List all registered ondemand runners."
```

internal/cli/ondemand_runner_list.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (c *OndemandRunnerListCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint ondemand-runner list
+
+  List all registered ondemand runners.
+
+  Ondemand runners are used to dynamically start tasks to execute operations for
+	projects such as building, deploying, etc.
```
briancain, line 70:
```suggestion
  Ondemand runners are used to dynamically start tasks to execute operations for
  projects such as building, deploying, etc.
```

briancain (changes requested):
Some more code review! ✨ This looks cool 😎 excited to take it for a spin 🌀

builtin/docker/kaniko.go
evanphx: Nah, it's local to the container only. It's not a port anything sees or uses except for the kaniko.

internal/cli/ondemand_runner_apply.go
evanphx: It's not builtin like install, it's per plugins like the build or deploy plugins. So there isn't a definitive list to show.

builtin/docker/kaniko.go
briancain: ok cool! Then maybe a comment explaining that would be good? Maybe something similar to ☝🏻

evanphx: Sure!

internal/cli/install.go
evanphx: We do that in the updates before this.

briancain: Oops, I see that now! :roger:

Makefile
briancain: Hm, it'd be good to have a separate server make target at least for when you're doing server dev work and only want to rebuild that. Maybe we have a `docker/all` that's this versus `docker/server` which is just for the server? or do we need to always build the ondemand runner Dockerfile for any changes in the server?

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+	}()
+
+	var (
+		od      *pb.OndemandRunner
+		updated bool
+	)
+
+	if c.flagId != "" {
+		s = sg.Add("Checking for an existing ondemand runner: %s", c.flagId)
```
briancain, line 61:
```suggestion
		s = sg.Add("Checking for an existing ondemand runner config: %s", c.flagId)
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			c.ui.Output(
+				"Error checking for project: %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return 1
+		}
+
+		od = resp.OndemandRunner
+		s.Update("Updating ondemand runner %q...", od.Id)
```
briancain, line 83:
```suggestion
		s.Update("Updating ondemand runner config %q...", od.Id)
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+		updated = true
+	} else {
+		s = sg.Add("Creating new ondemand runner")
```
briancain, line 86:
```suggestion
		s = sg.Add("Creating new ondemand runner config")
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+
+	od.PluginType = c.flagPluginType
+
+	// Upsert
+	_, err := c.project.Client().UpsertOndemandRunner(ctx, &pb.UpsertOndemandRunnerRequest{
+		OndemandRunner: od,
+	})
+	if err != nil {
+		c.ui.Output(
+			"Error upserting ondemand runner: %s", clierrors.Humanize(err),
```
briancain, line 170:
```suggestion
			"Error upserting ondemand runner config: %s", clierrors.Humanize(err),
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	if updated {
+		s.Update("Ondemand Runner updated")
```
briancain, line 177:
```suggestion
		s.Update("Ondemand Runner configuration updated")
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+	} else {
+		s.Update("Ondemand Runner created")
```
briancain, line 179:
```suggestion
		s.Update("Ondemand Runner configuration created")
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+func (c *OndemandRunnerApplyCommand) AutocompleteArgs() complete.Predictor {
+	return complete.PredictNothing
+}
+
+func (c *OndemandRunnerApplyCommand) AutocompleteFlags() complete.Flags {
+	return c.Flags().Completions()
+}
+
+func (c *OndemandRunnerApplyCommand) Synopsis() string {
+	return "Create or update an ondemand runner."
```
briancain, line 246:
```suggestion
	return "Create or update an ondemand runner configuration."
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (c *OndemandRunnerApplyCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint ondemand-runner apply [OPTIONS]
+
+  Create or update an ondemand runner.
+
+  This will create a new ondemand runner with the given options. If
+  a ondemand runner with the same id already exists, this will update the
+  existing runner using the fields that are set.
```
briancain, line 257:
```suggestion
  Create or update an ondemand runner configuration.

  This will register a new ondemand runner config with the given options. If
  a ondemand runner config with the same id already exists, this will update the
  existing runner config using the fields that are set.
  
  Waypoint will use an ondemand runner configuration to spawn containers for
  various kinds of work as needed on the platform requested during any given
  lifecycle operation.
```

briancain (changes requested):

Makefile
evanphx: The issue is that you need both images to do local work since we create an ondemand runner by default.

briancain: Suggested in slack that we could have a `docker/server-only` target. We wanna make sure peoples workflows don't break if they're used to `docker/server` giving everything they need so that's the main reason for the change.

internal/cli/project_apply.go
```diff
@@ -514,6 +537,12 @@ func (c *ProjectApplyCommand) Flags() *flag.Sets {
 			Default: "5m",
 			Usage:   "Interval between polling to generate status reports if polling is enabled.",
 		})
+
+		f.StringVar(&flag.StringVar{
+			Name:   "ondemand-runner",
```
mitchellh, line 542: I continue to think for user facing stuff, we shouldn't separate "ondemand" from "runner" and consider ondemand is just an implementation detail of how Waypoint just has "runners." Understanding both feels like some cognitive overhead. In that case, this should probably just be `runner-config` or something. We should maybe come up with a noun to describe our runner configs, is it "config" "role" "class" etc.

mitchellh (commented):
I did a partial review, maybe 30% of the thing. I'll submit what I have so far and review more of this tomorrow. I think I'm just nitpicking the surface layer of the thing at the moment but I think that's important since that's the part users will scratch! This is looking great though!

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (c *OndemandRunnerApplyCommand) Flags() *flag.Sets {
+	return c.flagSet(0, func(sets *flag.Sets) {
+		f := sets.NewSet("Command Options")
+
+		f.StringVar(&flag.StringVar{
+			Name:    "id",
+			Target:  &c.flagId,
+			Default: "",
+			Usage:   "The id an existing ondemand runner to update.",
```
krantzinator, line 194:
```suggestion
			Usage:   "The id of an existing ondemand runner to update.",
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,264 @@
...
+			Target:  &c.flagPluginType,
+			Default: "",
+			Usage:   "The type of the plugin to launch for the ondemand runner, such as aws-ecs, kubernetes, etc.",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "project-config",
+			Target:  &c.flagPluginConfig,
+			Default: "",
+			Usage: "Path to a an hcl file that contains the configuration for the plugin." +
```
krantzinator, line 222: Ah yes, the time-old debate -- is it "a hcl" or "an hcl" 🧙🏻  (Also I think we need a space after the period for spacing reasons with the next line that isn't actually a newline)
```suggestion
			Usage: "Path to an hcl file that contains the configuration for the plugin. " +
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,264 @@
...
+				"This is only necessary when the plugin's defaults need to be adjusted for " +
+				"the environment the plugin will launch the ondemand runner.",
```
krantzinator, line 224: This or "in which", depending on how fancy/grammatically correct you want to be.
```suggestion
				"the environment the plugin will launch the ondemand runner in.",
```

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,264 @@
...
+		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:    "default",
+			Target:  &c.flagDefault,
+			Default: false,
+			Usage: "Indicates that this ondemand runner should be used by any project that doesn't " +
+				"otherwise specify it's own ondemand runner.",
```
krantzinator, line 232: Izaak _just_ fixed these 😂 
```suggestion
				"otherwise specify its own ondemand runner.",
```

internal/server/singleprocess/service_ondemand_runner.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+
+	"github.com/golang/protobuf/ptypes/empty"
+
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+	serverptypes "github.com/hashicorp/waypoint/internal/server/ptypes"
+)
+
+// TODO: test
```
krantzinator, line 12: Team process question: should we make issues out of these TODOs that are liable to get forgotten? We've got half a dozen or maybe even a full dozen scattered throughout our codebase. @hashicorp/waypoint-core @hashicorp/waypoint-frontend

briancain: These tests should def be written before merging this PR.

krantzinator (commented):
Handful of docs grammar things for now

internal/server/singleprocess/service_ondemand_runner.go
krantzinator: Maybe I was too quick to @-mention everyone 😅

internal/cli/ondemand_runner_list.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	tbl := terminal.NewTable("id", "type", "oci-url")
+
+	for _, p := range resp.OndemandRunners {
+		tbl.Rich([]string{
+			p.Id,
+			p.PluginType,
+			p.OciUrl,
+		}, nil)
```
briancain, line 39: We should show the "Default" runner config specified in this list too.

builtin/docker/kaniko.go
```diff
@@ -0,0 +1,148 @@
...
+		return nil, status.Errorf(codes.Internal, "unable to parse image name: %s", err)
+	}
+	host := reference.Domain(ref)
+	log.Trace("auth host", "host", host)
+
+	var os ociregistry.Server
+	os.DisableEntrypoint = b.config.DisableCEB
+	os.Auth = auth
+	os.Logger = log
+	os.Upstream = "http://" + host
```
evanphx, line 83: Fix this http to https unless host is localhost OR they set configuration for insecure.

builtin/pack/builder.go
```diff
@@ -57,7 +69,7 @@ type BuilderConfig struct {
 	ProcessType string `hcl:"process_type,optional" default:"web"`
 }
 
-const DefaultBuilder = "heroku/buildpacks:18"
+const DefaultBuilder = "heroku/buildpacks:20"
```
evanphx, line 72: Should I have done this? ..... 🤔

Dockerfile.odr
```diff
@@ -0,0 +1,56 @@
...
+# Notes on img and what is required to make it work, since there's a lot
+# of small details below that are absolutely required for everything to
+# come together:
+#
+#  - img, runc, newuidmap, newgidmap need to be installed
+#  - libseccomp-dev must be installed for runc
+#  - newuidmap/newgidmap need to have suid set (u+s)
+#  - /etc/subuid and /etc/subgid need to have an entry for the user
+#  - USER, HOME, and XDG_RUNTIME_DIR all need to be set
+#
```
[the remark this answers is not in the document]
evanphx: Yup!

Dockerfile.odr
```diff
@@ -0,0 +1,56 @@
...
+COPY --from=busybox /bin/busybox /kaniko/busybox
+COPY --from=busybox /tmp /kaniko/tmp
+
+RUN ["/kaniko/busybox", "mkdir", "/kaniko/bin"]
+RUN ["/kaniko/busybox", "--install", "-s", "/kaniko/bin"]
+
+ENV PATH $PATH:/kaniko/bin
+ENV TMPDIR /kaniko/tmp
+
+ENTRYPOINT ["/kaniko/waypoint"]
```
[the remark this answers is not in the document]
evanphx: We'd still have to run `docker build` twice and because of the way docker layer caching works, the layers in Dockerfile.odr can be seen from Dockerfile, ergo if you run them one after another, they'll use the same build cache.

builtin/docker/registry.go
```diff
@@ -44,10 +62,9 @@ func (r *Registry) Push(
 	// Depending on whethere the image is, we diverge at this point.
 	switch img.Location.(type) {
 	case *Image_Registry:
-		// We can't push an image that isn't pulled locally in some form.
-		return nil, status.Errorf(codes.FailedPrecondition,
-			"Input image is not pulled locally and therefore can't be pushed. "+
-				"Please pull the image or use a builder that pulls the image first.")
+		// This indicates that the builder used the AccessInfo and published the image
+		// directly. Ergo we don't need to do anyhting and can just return the image as is.
```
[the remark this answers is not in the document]
evanphx: Remember that we're always calling AccessInfo now, so we're not sure if the Builder needed it or not. Ergo we have to call both of them for now.

I'll change the logic to only throw that warning IF the image is in a registry AND the image in the registry doesn't match the name and tag that Push was expecting.

Dockerfile.odr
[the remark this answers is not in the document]
mitchellh: Yeah, but we wouldn't have to duplicate the build steps. If we ever changed how we build Waypoint, we'd have to remember to update two locations here, vs. sharing the same definition.

builtin/docker/registry.go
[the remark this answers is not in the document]
mitchellh: That makes sense. But a registry can also not implement push at all now right? Thats optional.

internal/cli/main.go
```diff
@@ -555,12 +555,28 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"user token": func() (cli.Command, error) {
 			return &UserTokenCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
+		"ondemand-runner": func() (cli.Command, error) {
```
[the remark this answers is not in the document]
evanphx: There is no `waypoint runner list` atm, so I'll just make `waypoint ondemand-runner lis` into `waypoint runner list` and make sure the header indicates it's for ondemand runner configuration.

builtin/docker/kaniko.go
evanphx: Went ahead and did the kernel assigned port trick.

Dockerfile.odr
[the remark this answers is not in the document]
evanphx: Fair, did the change.

builtin/docker/registry.go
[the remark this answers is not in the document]
evanphx: No, they still have to implement Push. We didn't make that optional.

internal/server/singleprocess/service_ondemand_runner.go
evanphx: That's a copypasta comment, i'll check.

evanphx: Yeah, we don't have tests for most of the service files today.

internal/serverinstall/docker.go
```diff
@@ -645,13 +646,17 @@ func (i *DockerInstaller) InstallRunner(
...
 		Env:          opts.AdvertiseClient.Env(),
-		Cmd:          []string{"runner", "agent", "-vvv"},
+		Cmd:          []string{"runner", "agent", "-vv"},
 		Labels: map[string]string{
 			"waypoint-type": "runner",
 		},
 	}, &container.HostConfig{
+		Privileged: true,
+		CapAdd:     []string{"CAP_DAC_OVERRIDE"},
+		Binds:      []string{"/var/run/docker.sock:/var/run/docker.sock"},
```
izaaklauer, line 659: I feel like introducing on-demand runners makes this less necessary. After 0.6, I feel like a normal installation should have a static runner that launches on-demand runners in a platform like k8s or ecs.

I think a side-effect of this change is that there will be a lot more mystery orphaned containers on their k8s workers from people accidentally using the docker plugin with remote runners. It will also make it harder for some security-focused companies to adopt waypoint.

website/content/commands/server-install.mdx
```diff
@@ -52,6 +52,7 @@ you do not need to accept any terms.
 #### docker Options
 
 - `-docker-server-image=<string>` - Docker image for the Waypoint server.
+- `-docker-odr-image=<string>` - Docker image for the Waypoint Ondemand Runners
```
izaaklauer, line 55: Once we have more ODR types, it will probably make sense to make this a `type=image` map, but not a blocker now.

builtin/pack/builder.go
```diff
@@ -68,6 +80,170 @@ var skipBuildPacks = map[string]struct{}{
...
+
+	// Start constructing our arg string for img
+	args := []string{
+		"/kaniko/executor",
+		"-f", "Dockerfile.kaniko",
+		"--no-push",
+		"--context=dir:///" + dir,
+	}
+
+	// NOTE(mitchellh): we can probably use the img Go pkg directly one day.
```
izaaklauer, line 224: Can remove this comment

builtin/pack/builder.go
```diff
@@ -68,6 +80,170 @@ var skipBuildPacks = map[string]struct{}{
...
+
+	step.Done()
+	step = sg.Add("Testing registry and uploading entrypoint layer")
+
+	err = ocis.SetupEntrypointLayer(refPath, data)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error setting up entrypoint layer: %s", err)
+	}
+
+	li, err := net.Listen("tcp", ":5000")
```
izaaklauer, line 167: You might want to set this to `localhost:5000`. Otherwise I think this could bind to a non-loopback interface, and potentially be exposed on the internet.

izaaklauer (changes requested):

internal/serverinstall/docker.go
evanphx: Not true because without this change, the static docker runner can't launch tasks, and thusly can't launch ondemand runners.

internal/cli/runner_agent.go
```diff
@@ -178,6 +208,9 @@ func (c *RunnerAgentCommand) Run(args []string) int {
 					log.Warn("server unavailable, sleeping before retry")
 					time.Sleep(2 * time.Second)
 				}
+			} else if c.flagODR {
```
mitchellh, line 211: I've stared at this `else if` for far too long. It sends tingles down my brain and I think I figured it out. If there is no error AND its an ODR, then we're done. But so long as there are errors, we continue to retry. Right?

I wonder if this would be clearer to just break up the `if` into two separate (though slightly redundant) basic blocks. Or I wonder if I'm just being sensitive. I'm not sure. But maybe:

```go
if err == nil && flagODR {
  // Job success, we are ODR so handle one job only.
}

if err != nil {
  // Existing logic
}
```

internal/core/app_build.go
```diff
@@ -129,12 +148,57 @@ func (op *buildOperation) Upsert(
...
 
 func (op *buildOperation) Do(ctx context.Context, log hclog.Logger, app *App, _ proto.Message) (interface{}, error) {
+	args := []argmapper.Arg{
+		argmapper.Named("HasRegistry", op.HasRegistry),
+	}
+
+	// If there is a registry defined and it implements RegistryAccess...
+	if op.Registry != nil {
+		if ra, ok := op.Registry.Value.(component.RegistryAccess); ok {
+			fn := ra.AccessInfoFunc()
```
mitchellh, line 158: Should we not be using `callDynamicFuncAny2` here or whatever that thing is called?

internal/plugin/factory.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/hashicorp/waypoint-plugin-sdk/internal-shared/pluginclient"
 )
 
+// Indicates that the process is running inside an ondemand runner. This information
+// is passed to the plugins so they can configure themselves specially for this
+// unique context.
+var InsideODR bool
```
mitchellh, line 22: Ah hah. I wonderrrrr if we want to introduce a `internal/odr` package or something that exports an `InODR() bool` that we set globally and just use that. Kind of like an `os.GOARCH` feeling to it. That way if we have other ODR-specific logic we can just depend on that environmental-feeling package rather than sending this around everywhere.

Not crtical to this PR, but wanted to note it.

internal/server/proto/server.proto
```diff
@@ -296,6 +296,18 @@ service Waypoint {
   // JSON formatting is not supported.
   rpc WaypointHclFmt(WaypointHclFmtRequest) returns (WaypointHclFmtResponse);
 
+  //----------------------------------------------------------------------
+  // Ondemand runner functions.
+  //----------------------------------------------------------------------
+  rpc UpsertOndemandRunner(UpsertOndemandRunnerRequest) returns (UpsertOndemandRunnerResponse);
```
mitchellh, line 302: I think this should be `UpsertOndemandRunnerConfig` (the `Config`) part.

internal/server/proto/server.proto
```diff
@@ -296,6 +296,18 @@ service Waypoint {
...
+
+  // GetOndemandRunner returns the ondemand runner.
+  rpc GetOndemandRunner(GetOndemandRunnerRequest) returns (GetOndemandRunnerResponse);
```
mitchellh, line 305: Same with the `Config` suffix on this and List. Its a bit noisy but I think it'll be more idiomatic to our other APIs and if we introduce more ODR APIs will be clearer.

internal/server/proto/server.proto
```diff
@@ -801,6 +815,12 @@ message Ref {
   message DeclaredResource {
     string name = 1;
   }
+
+  // OndemandRunner references an OndemandRunner value to be used for contextual
+  // operation.
+  message OndemandRunner {
```
mitchellh, line 821: Similarly, this is referencing the ODR _config_ right, and not a running runner itself? As a reader of these protos, a `Config` suffix would clarify that for me.

internal/server/gen/mocks/is_restore_snapshot_request__event.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v1.1.2. DO NOT EDIT.
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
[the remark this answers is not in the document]
evanphx: Ah poop you're right.

website/content/commands/server-install.mdx
evanphx: Not sure what ya mean.

internal/server/proto/server.proto
```diff
@@ -2393,6 +2413,60 @@ message Artifact {
   google.protobuf.Any artifact = 1;
 }
 
+/********************************************************************
+* Ondemand Runners
+********************************************************************/
+message OndemandRunner {
```
mitchellh, line 2419: `OndemandRunnerConfig` ?

internal/server/proto/server.proto
```diff
@@ -2393,6 +2413,60 @@ message Artifact {
...
+  // plugin type is used to launch the plugin to start the batch task.
+  string plugin_type = 4;
+
+  // plugin config is the configuration for the plugin that is created. It
+  // is usually HCL and is decoded like the other plugins, and is plugin
+  // specific.
+  bytes plugin_config = 5;
+
+  // config format specifies the format of plugin_config
+  Project.Format config_format = 6;
```
mitchellh, line 2440: Speaking out loud again, not critical to this PR: we're using this in enough places now that maybe we should make this a standalone `HCLFormat` message/enum.

internal/server/proto/server.proto
```diff
@@ -2393,6 +2413,60 @@ message Artifact {
...
+  // sent in the request.
+  OndemandRunner ondemand_runner = 1;
+}
+
+message GetOndemandRunnerRequest {
+  Ref.OndemandRunner ondemand_runner = 1;
+}
+
+message GetOndemandRunnerResponse {
+  OndemandRunner ondemand_runner = 1;
```
mitchellh, line 2463: If we rename all these to have `config` then I think this field can just be `config`

internal/serverinstall/docker.go
izaaklauer: Oh gotcha, if waypoint is installed on docker, then on-demand runners won't work there. Makes sense.

We could update this later to only add these capabilities if we're installed on docker. Maybe add `runner-caps` and `runner-binds` to the server install commands or something, which we set automatically when installing on docker..

website/content/commands/server-install.mdx
izaaklauer: I'm imagining a world where you have multiple kinds of on-demand runners, and wouldn't want to use the same docker image for each of them.

internal/server/singleprocess/service_job.go
```diff
@@ -160,6 +166,48 @@ func (s *service) queueJobReqToJob(
 		}
 	}
 
+	_, anyTarget := job.TargetRunner.Target.(*pb.Ref_Runner_Any)
+
+	var od *pb.Ref_OndemandRunner
+
+	if anyTarget {
```
mitchellh, line 173: Personally would compress all the whitespace above here and make this one line:

`if _, anyTarget := job.TargetRunner.Target.(*pb.Ref_Runner_Any); anyTarget {`

mitchellh, line 173: Would comment this if case on why we're doing this.

internal/server/singleprocess/service_job.go
```diff
@@ -180,6 +228,136 @@ func (s *service) QueueJob(
...
+		"WAYPOINT_RUNNER_ID":        runnerId,
+		serverclient.EnvServerAddr:  addr.Addr,
+		serverclient.EnvServerToken: token,
+	}
+
+	if addr.Tls {
+		envVars[serverclient.EnvServerTls] = "1"
+	}
+
+	if addr.TlsSkipVerify {
```
mitchellh, line 296: Not sure if you want to use it, but if you convert the AdvertiseAddr to a `serverconfig.Config`, that struct has an `Env()` function that does all this for you. Might be a good idea to keep all that centralized.

internal/server/singleprocess/service_job.go
```diff
@@ -180,6 +228,136 @@ func (s *service) QueueJob(
...
+	// Get the next id
+	id, err := server.Id()
+	if err != nil {
+		return "", status.Errorf(codes.Internal, "uuid generation failed: %s", err)
+	}
+	job.Id = id
+
+	// Validate expiry if we have one
+	job.ExpireTime = nil
+	dur, err := time.ParseDuration("60s")
```
mitchellh, line 334: This is hardcoded at 60s, is that what you expected? (The comment makes it sound like its some sort of option)

internal/server/singleprocess/service_job.go
```diff
@@ -180,6 +228,136 @@ func (s *service) QueueJob(
...
+
+	// Follow the same logic as RunnerGetDeploymentConfig, which includes this note:
+	// Our addr for now is just the first one since we don't support
+	// multiple addresses yet. In the future we will want to support more
+	// advanced choicing.
+
+	var addr *pb.ServerConfig_AdvertiseAddr
+
+	// This should only happen during tests
+	if len(cfg.AdvertiseAddrs) == 0 {
```
mitchellh, line 265: Our TestServer function should probably initialize the server with this so we don't have to do this check. We can do that another time...

website/content/commands/server-install.mdx
mitchellh: I think for `server install` this is just setting up the initial config. Users can already setup multiple ODR pools (afaik) after the install.

internal/serverinstall/serverinstall.go
```diff
@@ -100,6 +100,12 @@ type InstallRunnerOpts struct {
 	AdvertiseClient *serverconfig.Client
 }
 
+// An optional interface that the installer can implement to request
+// an ondemand runner be registered.
+type OndemandRunnerConfig interface {
+	OndemandRunner() *pb.OndemandRunner
+}
```
mitchellh, line 107: Am I missing something in the PR or is this not implemented yet?

mitchellh (changes requested):

mitchellh:
Bigger question that I may have missed reading the PR: what currently causes an ODR to shut down after some timeout if say... a job is never given t it? Or is that not currently solved?

evanphx:
Currently nothing an orphan runner doesn't time out, totally right. I'll make that change to do a timeout of, say, 1 minute to start? That timeout is the time between the runner starting and receiving the job, not the total time to execute the job.

mitchellh:
> Currently nothing an orphan runner doesn't time out, totally right. I'll make that change to do a timeout of, say, 1 minute to start? That timeout is the time between the runner starting and receiving the job, not the total time to execute the job

Yes that feels good to me!

internal/cli/main.go
```diff
@@ -897,6 +913,15 @@ about the currently logged in user, generate new access, and invite new
 users directly into the Waypoint server.
 
 If you are looking to log in to Waypoint, use "waypoint login".
+`,
+	},
+	"ondemand-runner": {
+		"Ondemand Runner configuration",
+		`
+List and edit Ondemand Runner configuration.
```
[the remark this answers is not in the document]
evanphx: This is a worthwhile question. There is no english word `ondemand` so it becomes organizational jargon. `on-demand` is accepted english so perhaps that is what we should use. I'd say the one issue is situations where `on-demand` is not valid syntax we'll end up writing `ondemand`.

internal/cli/ondemand_runner_apply.go
```diff
@@ -0,0 +1,264 @@
...
+
+	flagId           string
+	flagOCIUrl       string
+	flagEnvVars      []string
+	flagPluginType   string
+	flagPluginConfig string
+	flagDefault      bool
+}
+
+func (c *OndemandRunnerApplyCommand) Run(args []string) int {
```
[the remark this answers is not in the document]
evanphx: Should we be changing project too, that uses `apply`? If we don't change project, should we use `apply` for the CLI to be consistent?

internal/server/singleprocess/service_ondemand_runner.go
briancain: Hm, it looks like most of the service files have tests. It shouldn't be too bad to write these, could probably copy the `service_config_test.go` to get started?

internal/server/proto/server.proto
briancain: `UpsertOnDemandRunnerConfig` per our slack convo 😄

internal/cli/main.go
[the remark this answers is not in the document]
evanphx: Decided in slack that when using it in a string context, it should be `"on-demand"` or `"On-Demand"` and as a code identifier `OnDemand` or `onDemand`

internal/cli/project_apply.go
evanphx: Unsure how to model this, mostly because it really is specific to the concept of on-demand runners. I worry if it's just `runner-config` folks might presume it means something about the static runners? But to your point, it feels like the concept of users worrying about a static runner are going away.

mitchellh: Yeah, I feel the same way too, its a bit confusing either way. We can keep it for now, the CLI flags aren't a part of any of our compat guarantees anyways and this is fairly minor.

internal/server/singleprocess/service_ondemand_runner.go
evanphx: Ah! Dunno what I was looking at! Regardless, test added!

builtin/pack/builder.go
evanphx: I think I do want to up this because :18 is going to fade out of maintenance before too long, better to upgrade to :20 as the default now.

website/content/commands/server-install.mdx
evanphx: Yup, this is just for `server install` and the initial config.

internal/plugin/factory.go
evanphx: Sure, we can that later. The other way to handle it is to refactor how we call Factory to have factory level configuration. I almost did that but the PR was getting too long.

internal/serverinstall/serverinstall.go
evanphx: eh? It's implemented, it's used by internal/cli/install.go

izaaklauer (approved):

briancain (approved):
Worked for me! I was able to get a docker devops project up and running. I had a few hiccups configuring my local registry to work over http, but otherwise it was smooth! 😎 I think we could add a note some where on the website around following the Docker guide for configuring a local insecure registry (although this is mostly a local dev issue, prod wouldn't do this). Otherwise all good! Nice job! 👏🏻 🎉

internal/ociregistry/server.go
```diff
@@ -0,0 +1,995 @@
+package ociregistry
```
mitchellh, line 1: Package comment

And I think we should move this to `internal/pkg/ociregistry` since its not strictly Waypoint-specific (similar to `internal/pkg/epinject`. Now that I type that out loud, maybe we should do `internal/pkg/epinject/ociregistry` since this is really focused on the entrypoint injection? Perhaps.

evanphx: Done!

internal/ociregistry/server.go
```diff
@@ -255,12 +255,10 @@ func (s *Server) procureToken(hdr string) error {
...
+		return "", errors.Wrapf(err, "error decoding www-authenticate header")
 	}
 
-	s.Auth = fmt.Sprintf("%s %s", authType, tokenBody.Token)
-
-	return nil
+	return fmt.Sprintf("%s %s", authType, tokenBody.Token), nil
 }
 
 func (s *Server) Negotiate() error {
```
mitchellh: Do we even need the exported `Negotiate` and to call that from plugins anymore then? Or does this happen automatically on every API call anyways?

evanphx: We do call it from all the plugins.

evanphx: For example, https://github.com/hashicorp/waypoint/pull/2056/commits/4bb27782c706bf9b98a0ac79a521d48e8691f6fc#diff-678d37d13f4a054528633c9b23c7f4161cd73755c1b3145cb65e616b7fc25c72R84. I modified it in 4bb to take the repo to setup the transport.

mitchellh (approved):
SHIP!

PR #2065, opened 2021-08-17
waypoint.hcl: "labels" variable and selector functions

mitchellh:
This PR introduces the `labels` HCL variable that is a mapping of the finalized label set for an operation. "Finalized" here means: it has merged all the labels from various sources and this is the final set that will be stored in the database. This variable has multiple uses:

* It can be used to set configuration. For example, you might want to set annotations on Kubernetes resource based on Waypoint labels.
* It can be used to make contextual decisions. For example, you can change the namespace of a Kubernetes deployment based on some set of labels.

To support these use cases, this PR also introduces the `selectormatch` and `selectorlookup` functions. Please see the docs/examples for their usage.

### DIFF
--- a/internal/core/component_creator.go
+++ b/internal/core/component_creator.go
@@ -149,7 +149,7 @@ func (cc *componentCreator) create(
 	// We first get the labels. We use labels to determine the proper use
 	// plugin and other things so we have to grab these first before we do
 	// anything else.
-	hclCtx, _, err := cc.labels(hclCtx, app)
+	hclCtx, labels, err := cc.labels(hclCtx, app)
 	if err != nil {
 		return nil, err
 	}
@@ -181,6 +181,7 @@ func (cc *componentCreator) create(
 			Name: useType,
 		},
 
+		labels:  labels,
 		mappers: pinst.Mappers,
 		plugin:  pinst,
 	}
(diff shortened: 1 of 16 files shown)


internal/core/component_creator.go
```diff
@@ -127,6 +146,14 @@ func (cc *componentCreator) create(
 	loadConfig bool,
 	hclCtx *hcl.EvalContext,
 ) (*Component, error) {
+	// We first get the labels. We use labels to determine the proper use
+	// plugin and other things so we have to grab these first before we do
+	// anything else.
+	hclCtx, labels, err := cc.labels(hclCtx, app)
+	if err != nil {
+		return nil, err
+	}
```
mitchellh, line 155: Note for reviewers: the main reason I did it this way here is because I know we're going to introduce label filters and so on into choosing what plugin to use and various other things. So I set it up so that we already have the labels at the beginning of an operation (rather than say... figuring this out purely in the config package) so we can make those decisions for future features.

mitchellh:
Note also, the docs here are focused on the `labels` var but we do plan on doing a much bigger labels focused docs sweep.

briancain (approved):
🏷️ 👍🏻

krantzinator (approved):
Sick.

PR #2076, opened 2021-08-19
ui: Fixed linting errors and updated API definitions

sabrinako:
~~Notably had to update `Ref.Component`'s namespace to include `Type` in the `ui/lib/waypoint-pb/server_pb.d.ts` file. Type existed in the `Component` namespace, but for some reason did not carry over when under the `Ref` namespace. This was causing some linting errors~~
_See comments from @jgwhite below_

### DIFF
--- a/ui/app/components/project-input-variables/list-item.ts
+++ b/ui/app/components/project-input-variables/list-item.ts
@@ -30,7 +30,7 @@ export default class ProjectInputVariablesListComponent extends Component<Variab
     this.variable = variable;
     this.isEditing = isEditing;
     this.isCreating = isCreating;
-    this.storeInitialVariable();
+    this.initialVariable = JSON.parse(JSON.stringify(this.variable));
   }
 
   get isHcl(): boolean {
--- a/ui/app/components/app-form/project-repository-settings.hbs
+++ b/ui/app/components/app-form/project-repository-settings.hbs
@@ -220,8 +220,8 @@
             spellcheck="false"
             rows="14"
             value={{this.decodedWaypointHcl}}
-            required={{this.this.serverHcl}}
-            {{on "input" (fn this.setWaypointHcl)}}
+            required={{this.serverHcl}}
+            {{on "input" (this.setWaypointHcl)}}
           />
         </div>
       {{else}}
--- a/ui/app/components/operation-logs/index.ts
+++ b/ui/app/components/operation-logs/index.ts
@@ -99,7 +99,7 @@ export default class OperationLogs extends Component<OperationLogsArgs> {
       // If it completes with an error, we should surface that in the UI
       if (event == GetJobStreamResponse.EventCase.COMPLETE && response.getComplete()?.getError()) {
         let error = response.getComplete()?.getError()?.toObject();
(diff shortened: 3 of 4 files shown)


jgwhite:
> Notably had to update `Ref.Component`'s namespace to include `Type` in the `ui/lib/waypoint-pb/server_pb.d.ts` file.

Did you need to do this manually?

jgwhite:
Regarding the `Component.Type` errors:

I think what’s happening here is an issue with how [ts-protoc-gen](https://github.com/improbable-eng/ts-protoc-gen) handles shadowing (or rather, doesn’t handle shadowing).

We have two “Component” types: `Component` (top-level) and `Ref.Component`. In cases like the snippet below, when we say `Component.Type` we mean the top-level `Component.Type`, not the non-existent `Ref.Component.Type`:

https://github.com/hashicorp/waypoint/blob/d2dd71fb3f92a139af6d4447ad5787a338b2701a/ui/lib/waypoint-pb/server_pb.d.ts#L698-L700

Unfortunately, as far as I can tell, TypeScript doesn’t have a convenient way to disambiguate when we’re in a namespace like this. Effectively, when we’re within the `Ref` namespace, we can no longer refer to the top-level `Component` without introducing an alias. Something like this:

```ts
type TopLevelComponent = Component;

export namespace Ref {
  
  export class Component extends jspb.Message {
    getType(): TopLevelComponent.Type
```

In order to robustly solve this problem, we’ll have to report an issue and/or submit a patch to [ts-protoc-gen](https://github.com/improbable-eng/ts-protoc-gen). We could introduce our own patch that runs after `protoc` but it’s likely to be complex in its own right. Fun trivia: this category of problem is in the territory of [macro hygiene](https://en.wikipedia.org/wiki/Hygienic_macro).

sabrinako:
Makes sense, I was wondering why it wouldn't just automatically do this. Updated the PR to just be the other few changes + remove the conflicts from the linting PRs you've been doing as well!

ui/app/components/project-input-variables/list-item.ts
```diff
@@ -30,7 +30,7 @@ export default class ProjectInputVariablesListComponent extends Component<Variab
     this.variable = variable;
     this.isEditing = isEditing;
     this.isCreating = isCreating;
-    this.storeInitialVariable();
+    this.initialVariable = JSON.parse(JSON.stringify(this.variable));
```
jgwhite, line 33: What was the motivation for this change?

jgwhite (approved):
Looks great. One question, but not a blocker.

ui/app/components/project-input-variables/list-item.ts
sabrinako: The linter doesn't seem to follow down to `this.storeInitialVariable()` function to see we're setting `initialVariable`, so it thinks we're not setting `initialVariable` in the constructor. I wasn't in love with the solution, happy to hear other suggestions!

jgwhite: Ooh that's curious. A limitation of typescript I guess. This solution seems fine.

PR #2080, opened 2021-08-19
waypoint on-demand runner inspect and small copy changes

mitchellh:
PR into @evanphx's branch.

### DIFF
--- a/internal/cli/ondemand_runner_inspect.go
+++ b/internal/cli/ondemand_runner_inspect.go
@@ -3,7 +3,7 @@ package cli
 import (
 	"fmt"
 
-	"github.com/golang/protobuf/proto"
+	"github.com/golang/protobuf/jsonpb"
 
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/internal/clierrors"
@@ -14,6 +14,8 @@ import (
 
 type OnDemandRunnerInspectCommand struct {
 	*baseCommand
+
+	flagJson bool
 }
 
 func (c *OnDemandRunnerInspectCommand) Run(args []string) int {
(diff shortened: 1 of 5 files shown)


internal/cli/ondemand_runner_inspect.go
```diff
@@ -0,0 +1,70 @@
...
+	id := c.args[0]
+
+	resp, err := c.project.Client().GetOnDemandRunnerConfig(c.Ctx, &pb.GetOnDemandRunnerConfigRequest{
+		Config: &pb.Ref_OnDemandRunnerConfig{Id: id},
+	})
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+	fmt.Println(proto.MarshalTextString(resp.Config))
```
briancain, line 43: We've been trying to keep `inspect` commands looking the same, so I think this should output with the UI.NamedVars and a header before the config prints

briancain (changes requested):
Minor comment but looks good to me otherwise!

internal/cli/ondemand_runner_inspect.go
mitchellh: Can you show an example? I copied this from `auth-method inspect` haha. I'll copy whatever style.

briancain: Ah, definitely!

I think initially we did it on `waypoint context inspect` and then it was suggested I keep that with https://github.com/hashicorp/waypoint/pull/2080.

mitchellh: CleanShot 2021-08-19 at 16 58 15@2x

Done!

briancain (approved):
Awesome! 👍🏻 ✨

PR #2081, opened 2021-08-20
cli/status: Add a '-refresh' flag to regenerate project applications status on request

briancain:
This pull request introduces a new flag, `-refresh`, which will refresh all of the requested status reports for the current view from running `waypoint status`. This means if you refresh a refresh for a project view, all of that projects applications will be refreshed. If you request a single app, only that app will be refreshed, etc.


internal/cli/status.go
```diff
@@ -146,6 +156,458 @@ func (c *StatusCommand) Run(args []string) int {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// refreshAppStatusRemote takes a single project and application and uses a
+// remote runner through the Waypoint Server API to queue and start a
+// status report refresh job. It takes the queued job id and hooks into the
```
izaaklauer, line 455: Does this need to be performed by a remote runner? I think it would be valuable for projects without a remote runner configured to be able to do this too.

internal/cli/status.go
```diff
@@ -17,18 +16,20 @@ import (
...
 )
 
 type StatusCommand struct {
 	*baseCommand
 
-	flagContextName string
-	flagVerbose     bool
-	flagJson        bool
-	flagAllProjects bool
-	filterFlags     filterFlags
```
krantzinator, line 31: Did you mean to remove this?

internal/cli/status.go
```diff
@@ -146,6 +156,458 @@ func (c *StatusCommand) Run(args []string) int {
 	return 0
 }
 
+// RefreshApplicationStatus takes a project and application target and generates
+// a list of applications to refresh the status on. If all projects are requested
+// to be refreshed, the CLI will do its best to honor the request. However if
+// a project is local and the CLI was not invoked inside that project dir, the
+// CLI won't be able to refresh that projects application statuses.
```
krantzinator, line 163:
```suggestion
// CLI won't be able to refresh that project's application statuses.
```

internal/cli/status.go
```diff
@@ -146,6 +156,458 @@ func (c *StatusCommand) Run(args []string) int {
...
+	// Get our API client
+	client := c.project.Client()
+
+	// Get the entire list of apps
+	// Determine project locality
+	// Do the Work (local or remote)
+	if c.flagAllProjects {
+		resp, err := client.ListProjects(c.Ctx, &empty.Empty{})
+		if err != nil {
+			c.ui.Output("Failed to retrieve all projects:"+clierrors.Humanize(err), terminal.WithErrorStyle())
```
krantzinator, line 174:
```suggestion
			c.ui.Output("Failed to retrieve all projects: "+clierrors.Humanize(err), terminal.WithErrorStyle())
```

internal/cli/status.go
```diff
@@ -146,6 +156,458 @@ func (c *StatusCommand) Run(args []string) int {
...
+					remote = false
+				case *pb.Job_DataSource_Git:
+					remote = true
+				default:
+					remote = false
+				}
+			}
+
+			if !remote {
+				c.ui.Output("Cannot refresh every projects status without command being "+
```
krantzinator, line 207:
```suggestion
				c.ui.Output("Cannot refresh every project's status without command being "+
```

krantzinator: I'm not able to hit this error. I have a project dir that is not a git repo, navigate one layer up so I'm not inside the project dir, run `waypoint status -refresh ex2`, then get stuck on `Refreshing status for app "ex2-app" in project "ex2". Refreshing could take a while...`
A sigkill gives me this https://gist.github.com/krantzinator/b899a5b7d7f4cb5e39f6286c4e54bfc6

internal/cli/status.go
```diff
@@ -146,6 +156,458 @@ func (c *StatusCommand) Run(args []string) int {
...
+					serverAddress = c.serverCtx.Server.Address
+				}
+
+				c.ui.Output(wpProjectNotFound, projectTarget, serverAddress, terminal.WithErrorStyle())
+			}
+			return err
+		}
+		project := projectResp.Project
+
+		// local or remote project?
```
mitchellh, line 238: I don't think we should be making this distinction here. We should just schedule a status report job and let the CLI handle this (in the same way `waypoint build` does). The command itself shouldn't worry about this.

mitchellh (changes requested):
I think my big comments here is that the CLI subcommand is painfully aware and logic-divergent on whether this is "local" or "remote". All our other tasks also have this distinction, but this is handled at the CLI layer (does it go to a local runner, or remote?).

My opinion is that the CLI implementation should probably just schedule a status report refresh job _no matter what_ and let the local vs remote runner handle that properly. The machinery all gets wired through automatically. This is how every other operation works afaik.

krantzinator (changes requested):
There's an error case that needs handled, I believe. See comments.

internal/cli/status.go
briancain: Yep, in the original implementation I thought about using filter flags with what's built in, but it turns out those flags are about filtering on the operations status. I removed the flag definition but never removed the bool from the struct!

internal/cli/status.go
briancain: The automatic runner allocation must be achieved through the `DoApp` function? If I remember right I had trouble making that work when the CLI is configured to have an optional config like this one, but I can give it a shot! If that works then this just turns into a loop calling DoApp on each app and avoiding the `ExpediteStatusReport` endpoint all together.

mitchellh: I don’t remember the details anymore but also happy to help change how any of that works to make it work for you. I think that logic just makes more sense at the core CLI layer rather than the subcommand business logic.

briancain: Agreed...I think that'll make this implementation a lot shorter and cleaner too. And in the future when we fix remote runners being able to refresh statuses this command doesn't change. I'll give it a go and see what I was missing last time and if I can fix it! 👍🏻

internal/cli/status.go
briancain: See this convo! https://github.com/hashicorp/waypoint/pull/2081#discussion_r694087217 I think this is essentially your question

internal/cli/status.go
briancain: You'll only hit this error if you run `waypoint status -refresh -all-projects`. It should fail pretty quickly though for a local project so there's something to investigate and fix here it sounds like

briancain: Well, we axed refreshing statuses on all projects so this is no longer seen. I've also reworked how app statuses are refreshed, so you might not hit this error. If you do let me know!

internal/cli/status.go
briancain: Alright, got this updated to do `DoApp` in both instances!

internal/cli/status.go
```diff
@@ -101,6 +102,15 @@ func (c *StatusCommand) Run(args []string) int {
 		c.ui.Output(wpAppFlagAndTargetIncludedMsg, terminal.WithWarningStyle())
 	}
 
+	// Optionally refresh status
+	if c.flagRefreshAppStatus {
+		if err := c.RefreshApplicationStatus(projectTarget, appTarget); err != nil {
```
izaaklauer, line 107: Potential optimization: if they specify a -app flag, we don't need to refresh status on every app in the project.

briancain: That should happen already!

mitchellh (approved):

krantzinator (approved):
Gave it a whirl again, and my main points of confusion are more related to how Waypoint manages apps and projects. Which is something I need to fix 😅 
`-refresh` works well!

Issue #1479, opened 2021-05-17
List releases from the CLI


izaaklauer:
It would be nice to be able to list releases from the CLI. It would enable me to learn which deployment is currently released without visiting the UI.

Ideally, I think it would include the following:
- The release ID, for cross-referencing to a deployment
- The release platform (alb, k8s, etc).
- The date/time the release occured
- The release hash (i.e. `01F5XRRQ2CDSK8DJQ0GM1PFMTR`)? The UI gives this, but i'm not sure what value it has, and `deployment list` doesn't give you the hash either
- The global release horizon url? I don't know if it belongs here, but I don't think can be obtained anywhere else from the CLI currently.

It's not clear to me what the noun should be. We have `waypoint deploy` and `waypoint deployment list`. If we choose `waypoint release list`, we're overloading the release command unlike deploy. If we choose `waypoint releases list`, then we have inconsistent pluralization with deployment.

evanphx:
This is related to #1478 because if we list releases, we should be listing them accurately.

briancain:
We have an `Unimplemented` field now. So I think when ever this work starts, we can just omit any unimplemented releases since they didn't actually occur.

briancain:
In terms of what we show, I think it should be similar to a `waypoint deployments list`. Something like:

- Id (the release sequence id)
- Deployment id (the sequence id)
- Platform
- Application
- Workspace
- Preload details (if any)
- Started
- Completed
- Health
- Url (omit if no release URL)

PR #2082, opened 2021-08-20
cli/release: Add `release list` command

xiaolin-ninja:
cli/release: Add `release list` command
In response to: https://github.com/hashicorp/waypoint/issues/1479

This PR adds `waypoint release list` functionality to the CLI. Opted to not show releases at all if they are unimplemented pseudo-releases (e.g. Docker).


internal/cli/release_list.go
```diff
@@ -0,0 +1,391 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/hashicorp/waypoint/internal/version"
```
briancain, line 11: This might be something you'll want to configure Goland to do, but typically we group package imports together like:

- stdlib pkgs
- external
- hashicorp specific

I know my `vim-go` sorts them automatically so there's probably some switch that golang has for it!

internal/cli/release_list.go
```diff
@@ -0,0 +1,391 @@
...
+			Workspace:   c.project.WorkspaceRef(),
+		})
+		if status.Code(err) == codes.NotFound {
+			err = nil
+			release = nil
+		}
+		if err != nil {
+			app.UI.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+			return ErrSentinel
+		}
```
briancain, line 69: I think we can delete this `GetLatestRelease` API call all together. The `UI_ListReleases` gives us the latest release. I don't remember what order `UI_ListReleases` does by default but there's an Order field too! Then we can grab the latest release off of the `Releases` slice to be used later on line 138 (not sure if there's another place this is used to check against the latest release but that's one of them).

internal/cli/release_list.go
```diff
@@ -0,0 +1,391 @@
...
+				url := "n/a"
+				if b.Url != "" {
+					url = b.Url
+				} else if releaseBundle.Release.Url != "" {
+					url = releaseBundle.Release.Url
+				}
+				columns = append(columns, url)
+			}
+
+			if !b.Unimplemented {
```
briancain, line 262: We should add a brief doc comment note here saying we won't actually list any releases that didn't actually happen.

internal/cli/release_list.go
```diff
@@ -0,0 +1,391 @@
...
+
+func (c *ReleaseListCommand) Synopsis() string {
+	return "List releases."
+}
+
+func (c *ReleaseListCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint release list [options]
+
+  Lists the releases that were created.
```
briancain, line 388:
```suggestion
  Lists the releases that were created if the platform includes a releaser.
```

.changelog/2082.txt
```diff
@@ -0,0 +1,4 @@
+```
+release-note:improvement
```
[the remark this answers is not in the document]
briancain: This has to be in line with the backticks since this is a markdown codeblock. Not sure if `go-changelog` will parse this right otherwise

internal/cli/release_list.go
```diff
@@ -0,0 +1,391 @@
...
+
+		i["id"] = rel.Release.Sequence
+		i["deploymentId"] = rel.Release.Preload.Deployment.Sequence
+		i["application"] = rel.Release.Application
+		i["workspace"] = rel.Release.Workspace.Workspace
+		i["url"] = rel.Release.Url
+		i["labels"] = rel.Release.Labels
+		i["component"] = rel.Release.Component.Name
+		i["status"] = c.statusJson(rel.Release.Status)
+		i["latestStatusReport"] = rel.LatestStatusReport
```
briancain, line 309: Maybe under this we also include the `Unimplemented` field? We're also omitting any release that didn't actually happen, maybe we should do that for the json view too?

internal/cli/release_list.go
```diff
@@ -0,0 +1,391 @@
...
+			return ErrSentinel
+		}
+		sort.Sort(serversort.ReleaseBundleCompleteDesc(resp.Releases))
+
+		if c.flagJson {
+			return c.displayJson(resp.Releases)
+		}
+
+		headers := []string{
+			"", "ID", "Deployment ID", "Platform", "Details", "Started", "Completed", "Health", "URL",
```
briancain, line 108:
```suggestion
			"", "ID", "Deployment ID", "Platform", "Details", "Started", "Completed", "Health",
```

The URL header gets added later!

briancain (changes requested):
This is great! 🎉 Works well for me too with docker (no releaser) and k8s (releaser)! Just had a few minor comments

internal/cli/release_list.go
xiaolin-ninja: Hm I copied this part straight from deployment_list.go so this could go on the list of hack week things to clean up

internal/cli/release_list.go
xiaolin-ninja: Good catch! I modified to have the field for now, but maybe we should omit altogether in JSON too. Since if those are just Waypoint internal releases that the user doesn't care about ever, they'd just add noise to the JSON output. Unless it's important to have _some_ record of it?

briancain (approved):

Issue #2090, opened 2021-08-23
Specifying ECS sidecar containers without health checks causes a panic


izaaklauer:
**Describe the bug**

Specifying an ECS sidecar container without configuring a HealthCheck causes a panic.

**Steps to Reproduce**

Add a sidecar block to an ECS deployment, something like this:
```
      sidecar {
        name = "db01"
        image = "postgres:13"
        memory  = 512
        container_port = "5432"
        host_port = "5432"
        static_environment = {
          POSTGRES_PASSWORD = "password"
        }
      }
```

Run a `waypoint deploy -v`, and observe the panic

**Expected behavior**
It should not panicl. Also, healthchecks should not required for sidecar containers.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable: 0.5.1

* Waypoint CLI Version: 0.5.1
* Waypoint Server Platform and Version: 0.5.1
* Waypoint Plugin: aws/ecs

**Additional context**

Initially found by user dwanek on Hashicorp discuss, thank you! https://discuss.hashicorp.com/t/ecs-sidecar-problem/28548

Problem is here: https://github.com/hashicorp/waypoint/blob/28ef3eac6560b6be2f238414271a9dced1cc292b/builtin/aws/ecs/platform.go#L1040-L1046

Issue #2061, opened 2021-08-17
plugin/ecs: Update to use resource manager and status functions


izaaklauer:
Once this is complete, status poll operations will work on ECS deployments and releases, and ECS will report the resources that it creates in the `waypoint status -app=<app>` view.

Issue #1645, opened 2021-06-11
Resource-level awareness in the plugins


krantzinator:
**Is your feature request related to a problem? Please describe.**
Previously, Waypoint did not have a way to track resources created by plugins. This led to issues such as #805 and #1641.

**Describe the solution you'd like**
With the addition of resource awareness added to the plugin SDK in 0.4.0, we can now update plugins to know what resources it is creating, and thus handle cleanup.

Plugins to be updated:
- [ ] Nomad
- [ ] AWS
  - [ ] ALB (in progress)
  - [ ] EC2
  - [ ] ECS
  - [ ] Lambda
- [ ] Azure ACI
- [ ] GCR

krantzinator:
We'll circle back to this after the work is done for #1605

Issue #1577, opened 2021-06-02
Use aws/alb releaser for ecs


izaaklauer:
**Is your feature request related to a problem? Please describe.**

Currently, ECS uses it's own releaser, which has ALB manipulation logic very similar to the aws/alb plugin releaser. The alb releaser also has implemented release status, where ecs has not.

It would be nice if ECS could use the ALB releaser. Currently, trying to specify the ALB releaser on an ECS deployment yields the following error at release time:

```
» Releasing...
! 1 error occurred:
  	* argument cannot be satisfied: type *anypb.Any
```

It seems this is because the [ALB releaser function wants a target group](https://github.com/hashicorp/waypoint/blob/main/builtin/aws/alb/releaser.go#L39), while the [ecs releaser wants a deployment](https://github.com/hashicorp/waypoint/blob/main/builtin/aws/ecs/releaser.go#L38).

catsby:
I looked into this and @izaaklauer and I discussed it some; the ECS _deployment_ optionally creates a load balancer, security group, and target group and then the ECS releaser simply sets the [weight to 100][weight]. The ALB releaser expects a target group to already exist and does the work (optional) of creating the security group, load balancer, et. al. 

In order to get the ECS platform to _not_ create the load balancer (and target group) users need to specify `disable_alb` in the `use-ecs` configuration... which contradicts the follow up `use "alb"` we would add in the `release` block of the configuration.

Ultimately 1 or both of these plugins would require some rework to make this work as described. 

If the goal is to add a `Status()` method to ECS, we may want to instead look at reusing ALB's implementation via a shared utility or something. 


[weight]: https://github.com/hashicorp/waypoint/blob/c8e5b0ba25a4f6e9f804499495f288e66b0b5c21/builtin/aws/ecs/releaser.go#L81

izaaklauer:
Is there a reason that the ECS deployment plugin needs to create the load balancer? If the deployment doesn't need the ALB, then it seems like it should be the the releaser's job to create it.

briancain:
Just writing a note here that fixing this could be potentially backwards incompatible when users upgrade to the version that has this fix. I removed the `0.5.x` milestone and placed it in the Next Major project board for this reason.

izaaklauer:
There is a reason the ECS plugin needs to create the load balancer. As it turns out, ECS refuses to attach a target group to a service if the target group isn't behind a load balancer. It fails with this error: 
```
[...]

PR #2098, opened 2021-08-24
Convert ECS platform plugin to use resource manager

izaaklauer:
This converts the ECS plugin from its own resource lifecycle logic to use resource manager, and implements the Status plugin.

Addresses https://github.com/hashicorp/waypoint/issues/1645
Closes https://github.com/hashicorp/waypoint/issues/2061

## What changes

From a user perspective, there are three changes:

### More detailed deployment UI

Previous `waypoint deploy` console output:

Screen Shot 2021-08-24 at 10 47 01 AM

New console output:

Screen Shot 2021-08-24 at 10 47 43 AM

Users can now see the progress on creating or discovering all of the resources that go into an ECS deployment.

### Better rollback behavior on deployment failure

If a deployment fails partway through, waypoint will now call destroy on each resource that was created during that deployment, leaving fewer orphaned resources. Example:

Screen Shot 2021-08-24 at 10 55 22 AM

Note that after the rate limit exception, waypoint destroyed the ALB listener and target group before exiting.

### Status

I've implemented status functions on `cluster` and `service` resources. The `service` resource produces additional `task` resources. You can view them with `waypoint status -app=<app>`

Example of a service coming online: 
Screen Shot 2021-08-24 at 10 12 00 AM

One drawback of this change: we run a status check _immediately_ after a deployment, at which point generally the service exists, but the tasks do not yet exist. The initial status check will generally look like this, until a user enables app polling or runs the upcoming `waypoint status -refresh` command:

Screen Shot 2021-08-24 at 11 11 52 AM

## Technical notes

There are _so many_ of individual resources that an ECS deployment creates or needs to be directly aware of. The full list is:
- cluster
- execution IAM role
- task IAM role (optional)
- internal security group
- external security group
- log group
- subnets
- target group
- ALB
- ALB Listener
- route53 record (optional)
- task definition
- service

This is also a bit of a nerve-wracking change, as it touches the entire surface area of the plugin and there aren't any tests. I tested:
- many config param combinations
- backwards compat (i.e. this can destroy 0.5.1 deployments, and 0.5.1 can destroy and release these deployments)
- app destroy
- mid-deploy failure rollback
- deleting AWS resources out of band before running a deployment destroy
- status checks performed locally and by a remote runner

That said, if anyone has any specific trial workflows in mind, please give them a whirl or let me know.

### Improved destroy logic

We only have destroy implemented for ALB listener, target group, and service, which was the state before this change. Other resources are either app-scoped (security groups, etc), or globally scoped (log group). We have a `DestroyWorkspace` plugin func, which could be useful for this, but I feel like we might need a `DestroyApp` plugin func too, as I don't think most of these resources are workspace scoped.

Relevant issue: https://github.com/hashicorp/waypoint/issues/805

### More status functions

I've only implemented status on the `cluster` and `service` resources, which I think are the most dynamic and valuable. Implementing more status functions would result in more aws API usage, which bring us closer to hitting api rate limits for large waypoint installations. AWS rate limits to 20 RPS per region. With this change, each status check will result in 4 GET api calls, so with the default check interval of 30 seconds on the latest deploy only, waypoint won't be able to support more than 150 apps. 

Once we have a plan to address the rate limit problem, we can implement status functions on more resources.

### ECS Releaser plugin

The ECS releaser plugin cannot be easily replaced by the ALB plugin (context: https://github.com/hashicorp/waypoint/issues/1577#issuecomment-902288517). The ECS plugin doesn't create any resources though - only modify existing deployment resources - so I don't think it needs resource manager.

## Incidental changes

Apologies for not making these smaller separate PRs.

- Closes https://github.com/hashicorp/waypoint/issues/2090
- Improved error messages. Amazon errors will often be something vague like "ARN validation failed". If you're making three or four AWS calls as part of one resource, that doesn't tell you which call failed. This now wraps errors to give that context.
- Internal security group now authorizes ingress from the external security group, instead of 0.0.0.0/0. This improves the security of ECS deploys, as it prevents the public from accessing disabled tasks, or DDOSing individual tasks.
- New config parameter for ecs: `ingress_port`, that allows you to configure the external port the load balancer uses. It still defaults to 443 if a cert is configured, and 80 otherwise.
- Checking more aws calls for "404" equivalent responses, and doing something appropriate instead of erroring.
- All aws calls are now called `withContext`, so if the aws api is slow or hanging it should be possible to cancel the operation.
- Security group creation doesn't call AuthorizeIngress to add ingress rules if the security group already exists. This reduces the prodigious number of API calls we make to create a deployment.


builtin/aws/ecs/platform.go
```diff
@@ -119,650 +135,845 @@ func (p *Platform) ValidateAuth() error {
...
+	)
 }
 
-type lStatus struct {
-	ui   terminal.UI
-	sg   terminal.StepGroup
-	step terminal.Step
-}
+// DeploymentId is a unique ID to be consistently used throughout our deployment
+type DeploymentId string
```
izaaklauer, line 279: Using this as an argmapper marker type

builtin/aws/ecs/platform.go
```diff
@@ -814,145 +1026,238 @@ func createALB(
...
-	if lo != nil {
-		// We receive the error `Log driver awslogs disallows options: awslogs-endpoint`
-		// when setting `awslogs-endpoint`, so that is not included here of the
-		// available options
-		// https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html
-		result["awslogs-datetime-format"] = aws.String(lo.DateTimeFormat)
-		result["awslogs-multiline-pattern"] = aws.String(lo.MultilinePattern)
-		result["mode"] = aws.String(lo.Mode)
-		result["max-buffer-size"] = aws.String(lo.MaxBufferSize)
+	// NOTE(izaak): The random part of ULIDs seems to be near the end, so for long app names, we might not get unique names here.
```
izaaklauer, line 1175: Leaving this for now because it's not new behavior, and because the future task to replace internal IDs with sequence IDs should help this considerably.

builtin/aws/ecs/platform.go
```diff
@@ -1037,36 +1346,38 @@ func (p *Platform) Launch(
...
 				Timeout:     aws.Int64(container.HealthCheck.Timeout),
 				Retries:     aws.Int64(container.HealthCheck.Retries),
 				StartPeriod: aws.Int64(container.HealthCheck.StartPeriod),
-			},
-			Secrets:           secrets,
-			Environment:       env,
-			Memory:            utils.OptionalInt64(int64(container.Memory)),
-			MemoryReservation: utils.OptionalInt64(int64(container.MemoryReservation)),
+			})
 		}
```
izaaklauer, line 1363: This bit fixes #2090

builtin/aws/ecs/platform.go
```diff
@@ -1102,23 +1411,26 @@ func (p *Platform) Launch(
...
+	var taskOut *ecs.RegisterTaskDefinitionOutput
+	var err error
 	// AWS is eventually consistent so even though we probably created the resources that
 	// are referenced by the task definition, it can error out if we try to reference those resources
 	// too quickly. So we're forced to guard actions which reference other AWS services
 	// with loops like this.
-	for i := 0; i < 30; i++ {
-		taskOut, err = ecsSvc.RegisterTaskDefinition(&registerTaskDefinitionInput)
+
+	for i := 0; i <= awsCreateRetries; i++ {
```
izaaklauer, line 1432: Longer term it probably makes sense to have some kind of exponential backoff with a context cancel.

builtin/aws/ecs/platform.go
```diff
@@ -1127,305 +1439,756 @@ func (p *Platform) Launch(
...
+}
+
+func (p *Platform) resourceAlbCreate(
+	ctx context.Context,
+	sg terminal.StepGroup,
+	sess *session.Session,
+	log hclog.Logger,
+	src *component.Source,
+	securityGroups *Resource_ExternalSecurityGroups,
+	subnets *Resource_Subnets, // Required because we need to know which VPC we're in, and subnets discover it.
```
izaaklauer, line 1471: I don't love having vpcId live inside the subnet state. I tried having subnets return a "VpcId" marker type, but we're not currently incorporating the resource return values into the argmapper soup. I think this is good enough for now, and if this problem crops up repeatedly we can change the plugin sdk.

builtin/aws/ecs/plugin.proto
```diff
@@ -4,16 +4,116 @@ package ecs;
...
+    string name = 1;
+    string arn = 2;
+  }
+
+  message SecurityGroup {
+    string name = 1;
+    string id = 2;
+
+    // Indicates if a security group was created by waypoint, and should be destroyed by waypoint
+    bool managed = 3;
```
izaaklauer, line 66: We're only setting (not using) this for now, but it'll be valuable in the future when we decide to implement delete on security groups.

builtin/aws/ecs/platform.go
```diff
@@ -1127,305 +1439,756 @@ func (p *Platform) Launch(
...
-			})
+	svc := iam.New(sess)
 
-			if err != nil {
-				return err
-			}
-		} else if len(def) > 0 && def[0].ForwardConfig != nil && len(def[0].ForwardConfig.TargetGroups) > 1 {
-			// Multiple target groups means we can keep the listener
-			var active bool
+	getOut, err := svc.GetRoleWithContext(ctx, queryInput)
```
izaaklauer, line 1978: With the improved resource visibility, we can see that _this_ turns out to be the slowest part of the ecs deployment most of the time. Go figure.

builtin/aws/ecs/platform.go
```diff
@@ -119,650 +133,845 @@ func (p *Platform) ValidateAuth() error {
...
+				// Warn if we encounter waypoint clusters in other odd states (i.e. DEPROVISIONING, FAILED, etc.)
+				// I think it's ok to try to create a new cluster if one exists in a non-active non-provisioning state
+				log.Warn("Ignoring cluster named %q in state %q", cluster, *c.Status)
+			}
 		}
 	}
 
 	if p.config.EC2Cluster {
-		return "", fmt.Errorf("EC2 clusters can not be automatically created")
+		return fmt.Errorf("EC2 clusters can not be automatically created")
```
briancain, line 506: Since we're changing this in this PR....we should probably opt for the `status.Error` + `codes.CodeID` everywhere here when we're crafting our own error (Versus just adding some extra text and just returning an error we received).

builtin/aws/ecs/platform.go
```diff
@@ -41,7 +55,7 @@ func (p *Platform) ConfigSet(config interface{}) error {
 	c, ok := config.(*Config)
 	if !ok {
 		// this should never happen
-		return fmt.Errorf("Invalid configuration, expected *cloudrun.Config, got %T", config)
+		return fmt.Errorf("invalid configuration, expected *cloudrun.Config, got %T", config)
```
briancain, line 58:
```suggestion
		return status.Errorf(codes.FailedPrecondition, "invalid configuration, expected *ecs.Config, got %T", config)
```

briancain (changes requested):
:whoa: there's a lot here! This really illustrates how nice that StatusAll helper is back over on the sdk side! Nice. I haven't tried this yet but spent some time reading through the code and gave some feedback here.

builtin/aws/ecs/platform.go
```diff
@@ -1102,330 +1409,785 @@ func (p *Platform) Launch(
...
-		if aerr, ok := err.(awserr.Error); ok {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "ResourceConflictException" {
 			switch aerr.Code() {
 			case "ResourceConflictException":
-				return nil, err
+				break OUTER
 			}
 		}
 
+		s.Update("Failed to register ecs task definition. Will retry in %d seconds (up to %d more times)\nError: %s", awsCreateRetryIntervalSeconds, awsCreateRetries-i, err)
```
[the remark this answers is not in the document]
izaaklauer: Thanks! Forgot about StatusWarn.

builtin/aws/ecs/platform.go
izaaklauer: I actually started by doing that! But I realized that these errors will return to the top-level Deploy function, which will wrap it in `status.Errorf` before the plugin returns.

builtin/aws/ecs/platform.go
```diff
@@ -119,650 +133,845 @@ func (p *Platform) ValidateAuth() error {
...
 
-func (p *Platform) SetupLogs(ctx context.Context, s LifecycleStatus, L hclog.Logger, sess *session.Session) (string, error) {
-	// e.updateStatus("setting up CloudWatchLogs")
+		// if we encounter an unrecoverable error, exit now.
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "AccessDeniedException", "UnsupportedFeatureException",
+				"PlatformUnknownException",
+				"PlatformTaskDefinitionIncompatibilityException":
+				break OUTER
```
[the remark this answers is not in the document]
izaaklauer: It would if not for the loop label. If we hit one of those known bad errors, we break out of the for loop altogether.

briancain (approved):
🚀

PR #2112, opened 2021-08-24
K8S Task Launcher Implementation (enables ODR)

mitchellh:
This implements the `TaskLauncher` plugin type for Kubernetes, which enables On-Demand Runners to exist on Kubernetes. This also modifies `waypoint install` targeting K8S to automatically configure the ODR configuration for K8S. Things like remote Docker builds should now work out of the box with Kubernetes-based installations.


mitchellh:
Note that pods are currently not cleaned up because `StopTask` isn't called. There are better solutions to this that I am working on now, but would welcome a review I can always open a follow up PR.

builtin/k8s/plugin.proto
```diff
@@ -28,3 +28,8 @@ message Resource {
     string name = 1;
   }
 }
+
+// Task represents the state of the TaskLaunch implementation.
+message Task {
```
catsby, line 33: The Docker plugin defines this as `TaskInfo`, should we be consistent or is that not very relevant?

mitchellh: Doesn't hurt to match it.

builtin/k8s/task.go
```diff
@@ -0,0 +1,271 @@
...
+	)
+
+	doc.SetField(
+		"context",
+		"the kubectl context to use, as defined in the kubeconfig file",
+	)
+
+	doc.SetField(
+		"image_secret",
+		"name of the Kubernetes secrete to use for the image",
```
catsby, line 105:
```suggestion
		"name of the Kubernetes secret to use for the image",
```

builtin/k8s/task.go
```diff
@@ -0,0 +1,271 @@
...
+		docs.Summary(
+			"this references an existing secret, waypoint does not create this secret",
```
catsby, line 107:
```suggestion
			"this references an existing secret; waypoint does not create this secret",
```

Nit: _I'm not totally sure_ my suggestion here is correct

builtin/k8s/task.go
```diff
@@ -0,0 +1,271 @@
...
+	clientSet, ns, _, err := clientset(p.config.KubeconfigPath, p.config.Context)
+	if err != nil {
+		return err
+	}
+
+	// Get our jobs client
+	jobsClient := clientSet.BatchV1().Jobs(ns)
+	err = jobsClient.Delete(ctx, ti.Id, metav1.DeleteOptions{})
+	if errors.IsNotFound(err) {
+		// If it doesn't exist then that's fine, its already stopped then.
```
catsby, line 149:
```suggestion
		// If it doesn't exist then that's fine, it's already stopped then.
```

builtin/k8s/task.go
```diff
@@ -0,0 +1,271 @@
...
+	tli *component.TaskLaunchInfo,
+) (*TaskInfo, error) {
+	// Get our client
+	clientSet, ns, _, err := clientset(p.config.KubeconfigPath, p.config.Context)
+	if err != nil {
+		return nil, err
+	}
+
+	// Generate an ID for our pod name.
+	id, err := ulid.New(ulid.Now(), rand.Reader)
```
catsby, line 169: We already import `sdk/component`, should we use [`component.Id()`][id] here instead? 


[id]: https://github.com/hashicorp/waypoint-plugin-sdk/blob/f88287ee9c998bdb3bcb626bc1ef9f195a53f043/component/id.go#L14

builtin/k8s/task.go
```diff
@@ -0,0 +1,271 @@
...
+	env := []corev1.EnvVar{}
+	for k, v := range tli.EnvironmentVariables {
+		env = append(env, corev1.EnvVar{
+			Name:  k,
+			Value: v,
+		})
+	}
+
+	// If the user is using the latest tag, then don't specify an overriding pull policy.
+	// This by default means kubernetes will always pull so that latest is useful.
```
catsby, line 196:
```suggestion
	// This by default means kubernetes will always pull so that latest is used.
```

internal/serverinstall/k8s.go
```diff
@@ -1353,6 +1487,27 @@ func (i *K8sInstaller) InstallFlags(set *flag.Set) {
...
+		Target:  &i.config.odrImage,
+		Usage:   "Docker image for the Waypoint On-Demand Runners",
+		Default: defaultODRImage,
+	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "k8s-runner-service-account",
+		Target: &i.config.odrServiceAccount,
+		Usage: "Service account to assign to the on-demand runner. If this is blank, " +
+			"a service account will be created automatically with the correct permissions.",
```
catsby, line 1501: nit: could add `Default: "waypoint-runner"` here and omit the `if i.config.odrServiceAccount == ""` check above. Or would that cause problems with the config returned in `OnDemandRunnerConfig()`?

catsby (approved):
LGTM, just minor nits

builtin/k8s/task.go
mitchellh: We could, but I think this will change in a bit anyways once I start working on https://github.com/hashicorp/waypoint/issues/2116 so lets leave it for now (this is also how Docker does it).

internal/serverinstall/k8s.go
```diff
@@ -1353,6 +1484,28 @@ func (i *K8sInstaller) InstallFlags(set *flag.Set) {
...
+		Default: "waypoint-runner",
+	})
+
+	set.BoolVar(&flag.BoolVar{
+		Name:    "k8s-runner-service-account-init",
+		Target:  &i.config.odrServiceAccountInit,
+		Usage:   "Create the service account if it does not exist.",
+		Default: true,
+	})
+
```
catsby, line 1508: Probably need this in `UpgradeFlags` as well?

mitchellh: Hm! I didn't think about upgrades at all. Yes. Let me take a swing at that.

PR #2119, opened 2021-08-24
ui: de-lint remaining routes

jgwhite:
## Why the change?

A few steps closer to running the linters in CI.

## How do I test it?

These are mostly changes to type annotations with the notable exception of the first commit in the batch, which removes a piece of non-functioning logic. To verify this change:

1. Check out the branch
2. Visit a project with only app (by directly entering the URL into the browser)
3. Verify you see the application list (as we currently do on main)


sabrinako (approved):
Looks good to me!

PR #2121, opened 2021-08-25
ui: bugfix and de-lint `ApiService`

jgwhite:
## Why the change?

Primarily to remove linter errors and warning, but the one error turned out to be a legitimate bug. See the commit message for more details. Here’s what the little test looks like in the browser:

CleanShot 2021-08-25 at 13 30 11@2x

## How do I test it?

1. Check out the branch
2. Try editing input variables for a project
3. Verify no other settings are lost in the process

### DIFF
--- a/ui/app/services/api.ts
+++ b/ui/app/services/api.ts
@@ -277,9 +277,7 @@ export default class ApiService extends Service {
     }
 
     // Application list settings
-    let appList = project.applicationsList.map((app: Application.AsObject) => {
-      return new Application(app);
-    });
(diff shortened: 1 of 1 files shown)


ui/app/services/api.ts
```diff
@@ -306,3 +304,18 @@ declare module '@ember/service' {
     api: ApiService;
   }
 }
+
+function applicationFromObject(object: Application.AsObject): Application {
```
gregone, line 308: this is much cleaner and safer 💯

gregone (approved):

PR #2146, opened 2021-08-26
ui: add types to `components/actions/*.ts`

jgwhite:
## Why the change?

One step closer to running the linters

## How do I test this?

These are all changes to type annotations and are inherently safe. There is one place where I removed a constructor override — I believe this is self-evidently safe.


sabrinako (approved):
🤠 👍

PR #2168, opened 2021-08-27
ui: Update HCL field to codemirror

### LINKED ISSUE #2074: ui: Improve the server-located HCL file text input field
The current field for a user's HCL file in the git settings is a simple text input field, so some nice features like using the tab button and syntax highlighting are not available. This makes the experience frustrating and even potentially hazardous, leading to unexpected behavior in a user's pipeline.

We would like a more robust text editor when inputting HCL files through the settings menu on the UI.

We are in the process of evaluating the idea of an HCL wizard to guide users through the process of making an HCL file, but the UI team came to a consensus that having a more pleasing experience with this text input field is a good middle step towards this goal that will lead to immediate positive results.

If successful, this can be extracted out as an add-on for other HC products to use.

sabrinako:
### TODO:

- [x] Add tests (modifier + service tests)
### Updated HCL field screenshot
Screen Shot 2021-08-27 at 2 01 29 PM


ui/tests/integration/modifiers/code-mirror-test.ts
```diff
@@ -0,0 +1,65 @@
...
+import { fillIn, render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Modifier | code-mirror', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    let valueToChange = 'initial value in editor';
+    this.set('value', valueToChange);
+    this.set('onInput', () => {});
```
jgwhite, line 12: I don’t get why, but ESLint is complaining about these void functions. I say we just disable that rule of this file by adding this at the top of the file:

```ts
/* eslint-disable @typescript-eslint/no-empty-function */
```

ui/tests/integration/modifiers/code-mirror-test.ts
```diff
@@ -0,0 +1,65 @@
...
+    // if set as value on initial render, it won't get deleted on the second fillIn call
+    let firstValue = 'first value';
+    await fillIn(textArea, firstValue);
+
+    assert.dom('.CodeMirror-code').matchesText(firstValue);
+
+    let newValue = 'second value';
+    await fillIn(textArea, newValue);
+
+    assert.equal(this.get('value'), newValue);
```
jgwhite, line 43: ESLint is complaining about this, but I think it’s actually a bug in the rule. I say we disable this rule for this file too:

```ts
/* eslint-disable ember/no-get */
```

jgwhite (changes requested):
This is looking wonderful!!

The overall implementation is rock solid. I’ve left a flock of suggestions. Sorry it’s so many. They’re all pretty minor.

gregone:
Could we check in with @almonk on the editor theme? It feels a little strange to have something completely different than the logs output. I'd also point out we may want to swap themes in dark mode too?

gregone (changes requested):
Added a few other comments, but I agree with @jgwhite that the `code-mirror` service is probably unnecessary

ui/app/modifiers/code-mirror.ts
```diff
@@ -0,0 +1,59 @@
...
+  positional: never;
+  named: {
+    value: string;
+    onInput: (value: string) => void;
+    options: Record<string, unknown>;
+  };
+}
+
+// call to create the CodeMirror waypoint HCL mode for syntax highlighting
+waypointHclMode();
```
jgwhite, line 28: I think this is fine, though I’d be tempted to call the function something like `registerWaypointHclMode`.

An alternative approach would be to use this kind of import:

```ts
import './utils/waypoint-hcl-mode';
```

And then in the module, it can perform the `defineMode` stuff directly (rather than exporting a function):

```ts
import codemirror from 'codemirror';

codemirror.defineSimpleMode('waypointHCL', {
  // ...
});
```

ui/app/modifiers/utils/waypointHclMode.ts
```diff
@@ -0,0 +1,35 @@
+import codemirror from 'codemirror';
```
jgwhite, line 1: Just for consistency, I’d tempted to use kebab case for this file (i.e. `utils/waypoint-hcl-mode.ts`)

ui/app/modifiers/code-mirror.ts
```diff
@@ -0,0 +1,59 @@
...
+export default class CodeMirrorModifier extends Modifier<Args> {
+  _editor!: codemirror.Editor;
+
+  didInstall(): void {
+    this._setup();
+  }
+
```
jgwhite, line 35: A thought for the future evolution of this component (no need to address in this PR)…

At some point we may want to implement the [`didUpdateArguments`](https://github.com/ember-modifier/ember-modifier#api) hook. This would mainly be so that if the calling code decides to send in a new `value` then the editor updates accordingly (i.e. `this._editor.setValue(newValue)`). Imagine, for example, a “reset” button.

It’d probably look something like this:

```ts
didUpdateArguments() {
  // this test might not be necessary if codemirror is smart about this sort of thing
  if (this.args.value !== this._editor.getValue()) {
    this._editor.setValue(this.args.value);
  }

  if (this.args.options) {
    for (let [key, value] of Object.entries(this.args.options)) {
      this._editor.setOption(key, value);
    }
  }
}
```

ui/app/modifiers/code-mirror.ts
```diff
@@ -0,0 +1,59 @@
...
+  matchBrackets: true,
+  autoCloseBrackets: true,
+  styleActiveLine: true,
+};
+interface Args {
+  positional: never;
+  named: {
+    value: string;
+    onInput: (value: string) => void;
+    options: Record<string, unknown>;
```
jgwhite, line 23: I reckon it’s worth marking this as optional to keep us out of trouble in future.

```suggestion
    options?: Record<string, unknown>;
```

ui/app/components/code-editor-field.hbs
```diff
@@ -0,0 +1,4 @@
+<div
+  class="codemirror-container"
+  {{code-mirror value=@value onInput=@onInput options=this.options}}
```
jgwhite, line 3: Where is `this.options` defined? Should it be `@options` instead?

jgwhite (changes requested):
A few more minor suggestions, but overall looking superb.

ui/app/modifiers/code-mirror.ts
```diff
@@ -0,0 +1,55 @@
+import Modifier from 'ember-modifier';
+import codemirror from 'codemirror';
+
+import './utils/register-waypoint-hcl-mode';
+import 'codemirror/addon/edit/matchbrackets';
+import 'codemirror/addon/edit/closebrackets';
+import 'codemirror/addon/selection/active-line';
```
jgwhite, line 7: These are pleasingly symmetrical now 😺

jgwhite (approved):
This PR rules.

outstanding

gregone (approved):

PR #2174, opened 2021-08-30
ui: fix ember-cli-flash types

jgwhite:
## Why the change?

Helps us trust type information for ember-cli-flash.

### Before

CleanShot 2021-08-30 at 12 08 39@2x

### After

CleanShot 2021-08-30 at 12 09 04@2x

## How do I test it?

This only affects type definitions, so is safe. You may want to test it in your editor to see that the types show up.

PR #2178, opened 2021-08-31
ui: de-lint `project-repository-settings.ts`

jgwhite:
## Why the change?

Brings us one step closer to running linters in CI.

**Note:** There are a few remaining TypeScript errors which are addressed by #2174.

## How do I test it?

These changes are mostly type annotations and therefore safe. The exceptions are the additional `isFormField` guard clauses in `setBasicAuth` and `setSshAuth` which are introduced to perform [type narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). These are runtime changes though, so I’d recommend smoke-testing the data source form to ensure nothing has regressed.


sabrinako (approved):
LGTM! 👍

PR #2186, opened 2021-09-01
Project, app, workspace, and label scoping for Runner Config

mitchellh:
This hooks up the deep core work of #2169 to one aspect of real functionality: runners now properly look up config for their scope. So if you set a config that is for runners running jobs for the "foo" project, on-demand runners will now properly get that configuration.

Note this only works for on-demand runners. This was a limitation we expected from the beginning.

## The Journey

We're not quite done, I'm continuing to keep my PRs small and focused. Our journey so far:

1. [x] #2169 set up our core internal data model. 
2. [x] **This PR** hooks it up to runners so once set, runners actually see it.
3. [ ] Future PR will add CLI work so that users can actually set this.
4. [ ] Future PR (not by me) will add UI work so users can set this in the UI.
5. [ ] Future PR: docs on the new config scoping


internal/server/singleprocess/service_runner.go
```diff
@@ -110,22 +109,65 @@ func (s *service) RunnerConfig(
...
+			return status.Errorf(codes.FailedPrecondition,
+				"no pending job for this on-demand runner. A pending job "+
+					"must be registered prior to registering the runner.")
+		}
+
+		// Nack the job cause we're not going to take it ourselves,
+		// the runner will later with JobStream.
+		if _, err := s.state.JobAck(sjob.Id, false); err != nil {
+			return err
+		}
```
evanphx, line 136: Should we have a `JobPeekForRunner` that doesn't perform the assignment?

evanphx (changes requested):
Initial review, mostly wondering about having an API to peek at the job queue.

internal/server/singleprocess/service_runner.go
evanphx: Didn't mean this to be a change request, just a comment.

mitchellh: Yeah, this is probably a better idea.

mitchellh: Done

evanphx (approved):

PR #2189, opened 2021-09-02
ui: de-lint `notification/index.ts`

jgwhite:
## Why the change?

One step closer to running the linters in CI.

## How do I test this?

I believe these changes are self-evidently safe. But please feel free to contradict me!


sabrinako (approved):
LGTM! So close to no errors popping up when running on local 🤩

Issue #805, opened 2020-11-16
waypoint destroy on ECS does not delete all AWS resources created by waypoint


aruntomar:
**Describe the bug**
`waypoint up` on ECS creates lot of AWS resources eg: ALB, security groups, iam role, ECS cluster, task & execution role,  etc
`waypoint destroy` only destroys the ECS cluster & task definition, leaving behind ALB, security groups, iam roles etc. 

**Steps to Reproduce**
https://github.com/hashicorp/waypoint-examples/blob/main/aws-ecs/nodejs/waypoint.hcl

**Expected behavior**
I would expect all the AWS resources created by waypoint to be destroyed when `waypoint destroy` is executed.

opsline-jrios:
It's any update about it? I'm having the same issue, I need to go manually to destroy later the ALB, SGs and other stuff on AWS after I run the destroy command.

krantzinator:
We plan to address this in an upcoming release.

wesky93:
In my case, even the ECS cluster and task have not been deleted.

PR #2197, opened 2021-09-02
Backport of Convert ECS platform plugin to use resource manager into release/0.5.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #2098 to be assessed for backporting due to the inclusion of the label backport/0.5.x.

---

This converts the ECS plugin from its own resource lifecycle logic to use resource manager, and implements the Status plugin.

Addresses https://github.com/hashicorp/waypoint/issues/1645
Closes https://github.com/hashicorp/waypoint/issues/2061

## What changes

From a user perspective, there are three changes:

### More detailed deployment UI

Previous `waypoint deploy` console output:

Screen Shot 2021-08-24 at 10 47 01 AM

New console output:

Screen Shot 2021-08-24 at 10 47 43 AM

Users can now see the progress on creating or discovering all of the resources that go into an ECS deployment.

### Better rollback behavior on deployment failure

If a deployment fails partway through, waypoint will now call destroy on each resource that was created during that deployment, leaving fewer orphaned resources. Example:

Screen Shot 2021-08-24 at 10 55 22 AM

Note that after the rate limit exception, waypoint destroyed the ALB listener and target group before exiting.

### Status

I've implemented status functions on `cluster` and `service` resources. The `service` resource produces additional `task` resources. You can view them with `waypoint status -app=<app>`

Example of a service coming online: 
Screen Shot 2021-08-24 at 10 12 00 AM

One drawback of this change: we run a status check _immediately_ after a deployment, at which point generally the service exists, but the tasks do not yet exist. The initial status check will generally look like this, until a user enables app polling or runs the upcoming `waypoint status -refresh` command:

Screen Shot 2021-08-24 at 11 11 52 AM

## Technical notes

There are _so many_ of individual resources that an ECS deployment creates or needs to be directly aware of. The full list is:
- cluster
- execution IAM role
- task IAM role (optional)
- internal security group
- external security group
- log group
- subnets
- target group
- ALB
- ALB Listener
- route53 record (optional)
- task definition
- service

This is also a bit of a nerve-wracking change, as it touches the entire surface area of the plugin and there aren't any tests. I tested:
- many config param combinations
- backwards compat (i.e. this can destroy 0.5.1 deployments, and 0.5.1 can destroy and release these deployments)
- app destroy
- mid-deploy failure rollback
- deleting AWS resources out of band before running a deployment destroy
- status checks performed locally and by a remote runner

That said, if anyone has any specific trial workflows in mind, please give them a whirl or let me know.

### Improved destroy logic

We only have destroy implemented for ALB listener, target group, and service, which was the state before this change. Other resources are either app-scoped (security groups, etc), or globally scoped (log group). We have a `DestroyWorkspace` plugin func, which could be useful for this, but I feel like we might need a `DestroyApp` plugin func too, as I don't think most of these resources are workspace scoped.

Relevant issue: https://github.com/hashicorp/waypoint/issues/805

### More status functions

I've only implemented status on the `cluster` and `service` resources, which I think are the most dynamic and valuable. Implementing more status functions would result in more aws API usage, which bring us closer to hitting api rate limits for large waypoint installations. AWS rate limits to 20 RPS per region. With this change, each status check will result in 4 GET api calls, so with the default check interval of 30 seconds on the latest deploy only, waypoint won't be able to support more than 150 apps. 

Once we have a plan to address the rate limit problem, we can implement status functions on more resources.

### ECS Releaser plugin

The ECS releaser plugin cannot be easily replaced by the ALB plugin (context: https://github.com/hashicorp/waypoint/issues/1577#issuecomment-902288517). The ECS plugin doesn't create any resources though - only modify existing deployment resources - so I don't think it needs resource manager.

## Incidental changes

Apologies for not making these smaller separate PRs.

- Closes https://github.com/hashicorp/waypoint/issues/2090
- Improved error messages. Amazon errors will often be something vague like "ARN validation failed". If you're making three or four AWS calls as part of one resource, that doesn't tell you which call failed. This now wraps errors to give that context.
- Internal security group now authorizes ingress from the external security group, instead of 0.0.0.0/0. This improves the security of ECS deploys, as it prevents the public from accessing disabled tasks, or DDOSing individual tasks.
- New config parameter for ecs: `ingress_port`, that allows you to configure the external port the load balancer uses. It still defaults to 443 if a cert is configured, and 80 otherwise.
- Checking more aws calls for "404" equivalent responses, and doing something appropriate instead of erroring.
- All aws calls are now called `withContext`, so if the aws api is slow or hanging it should be possible to cancel the operation.
- Security group creation doesn't call AuthorizeIngress to add ingress rules if the security group already exists. This reduces the prodigious number of API calls we make to create a deployment.


izaaklauer:
On it.

izaaklauer:
Unnecessary

PR #2199, opened 2021-09-02
Upgrading plugin SDK version to latest for the 0.5.x branch.

izaaklauer:
After the auto-backport of the ecs resource manager overhaul succeeded (here: https://github.com/hashicorp/waypoint/pull/2196), I tested the release branch, and it wouldn't compile because it didn't backport a required SDK upgrade.

This change brings the release/0.5.x branch up to the latest version of the plugin sdk. The only wrinkle is that the ODR additions were breaking changes to the API, so I also needed to modify our ClientConfig in core waypoint to disable ODR. I think this is fine, as we're not planning to backport any of the ODR stuff. If we do, this will produce a small conflict that will need to be resolved.


briancain (approved):

krantzinator (approved):
Nice workaround

Issue #2113, opened 2021-08-24
Procedure for recovering from failed `waypoint upgrade` is not documented


ragesoss:
**Describe the bug**
The automatic server upgrade process can fail and leave the system in an unusable state, but the error message and upgrade documentation don't provide enough info on how to recover from the situation.

In particular, if the automatic upgrade process (`waypoint server upgrade -platform=nomad -auto-approve`) fails after upgrading the server but before setting up the new Waypoint runner job (at the "Retrieving new auth token for runner" step), the [upgrading documentation](https://www.waypointproject.io/docs/upgrading#automated-upgrade)  does not provide enough guidance on how uninstall the newly installed server and restore the snapshot (or how to retry placing the runner job, or some other way to recover).

**Steps to Reproduce**
* Run `waypoint server upgrade -platform=nomad -auto-approve` and make it fail at the "Retrieving new auth token for runner" step.

**Expected behavior**
Documentation of upgrade process and/or error messages provide explicit instructions for recovering from the problem.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: 0.5.1
* Waypoint Server Platform and Version: nomad, 0.5.1
* Waypoint Plugin: -

**Additional context**
 I spun up a test instance of my Waypoint/Nomad cluster ([repo](https://github.com/WikiEducationFoundation/wikieduinfra)) to try out the upgrade process. With this fresh setup, it had server version 0.5.1 (the latest release as of now) so I expected running the upgrade procedure to basically leave the system unchanged. (I have this in production with a server that's still on 0.4.0.)

I ran the upgrade command, and this happened:

```
 !  2.7.1  ~/h/dashboard_TESTING   *-  waypoint server upgrade -platform=nomad -auto-approve                                                                                                                        141ms  Mon 23 Aug 2021 04:14:12 PM PDT
✓ Context "wikiedu-testing" validated and connected successfully.
✓ Snapshot of server written to: 'waypoint-server-snapshot-1629760461'

» Upgrading...
  Waypoint server will now upgrade from version "v0.5.1"
✓ Detected existing Waypoint server
✓ Upgrade of Waypoint server on Nomad complete!

» Verifying upgrade...
✓ Server connection verified!

» Upgrading runner if required...
✓ Previous runner uninstalled
✓ Waypoint runner job and allocations purged
❌ Retrieving new auth token for runner...
[...]

PR #2200, opened 2021-09-02
docs: add details about remote runners on server upgrade

xiaolin-ninja:
As the title suggests :)

Though in the process of investigating this, we uncovered that the `-runner` flag is mentioned in the documentation, but missing from the flag list. Another PR coming to fix that.

Follow-up to https://github.com/hashicorp/waypoint/issues/2113

### DIFF
--- a/website/content/docs/upgrading/index.mdx
+++ b/website/content/docs/upgrading/index.mdx
@@ -115,7 +115,7 @@ in the future.
 2. Check the help text from running `waypoint server upgrade -help`.
 
 3. Ensure you are in the right server context before upgrading. This can be
-   verified through `waypoint context`.
+   verified through `waypoint context verify`.
 
 4. When ready, run the upgrade command for your given platform. For example:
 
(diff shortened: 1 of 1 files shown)


website/content/docs/upgrading/index.mdx
```diff
@@ -115,7 +115,7 @@ in the future.
 2. Check the help text from running `waypoint server upgrade -help`.
 
 3. Ensure you are in the right server context before upgrading. This can be
-   verified through `waypoint context`.
+   verified through `waypoint context verify`.
```
briancain, line 118: The original wording here is confusing, so we should clarify what the step is trying to tackle.  I wrote it so that's on me 😅 

The idea here was to look at which context you are currently using to make sure you are going to be upgrading to the right server before invoking the upgrade command (since you could have multiple valid contexts and server installs).

So this step should really say that you'll want to `waypoint context inspect` to make sure you're on the right context intended to be upgraded, and if not, do a `waypoint context list` to find the right one and switch to it using `waypoint context use`. You could end it by doing a `waypoint context verify` although the upgrade command does this too before it attempts to upgrade.

briancain (changes requested):

website/content/docs/upgrading/index.mdx
xiaolin-ninja: Ok, yeah I thought about putting `context inspect`, but `context verify` also returns the context it tried to verify, so I figured that was two birds with one stone and more user-friendly. Changed it to be more explicit.

briancain (approved):
Awesome! Looks great! 🎉

PR #2201, opened 2021-09-02
Runners support config files and dynamic config

mitchellh:
Runner configuration has always supported env vars (using the `-runner` flag on `waypoint config`). This PR enhances runners to also support dynamic config and config files, which apps have supported for a couple releases. This brings parity between runner and app config.

## Implementation Details

Months ago I extracted `internal/appconfig` from our entrypoint so that the "app config" logic is reusable. We never refactored runners to be built on top of that, so I did that in this PR. The rest mostly comes "for free" from that.

I did have to update `internal/appconfig` to support detecting unset env vars and to replace unset env vars with their original value. This is important for runners because we run the jobs directly in the same process. We never had to do this before because entrypoints subprocess, so keeping track of unset env vars was unnecessary and so was replacing old values, since if they don't exist anymore, the next subprocess would just pick up the parent process old values.


briancain (approved):
Nice, makes sense! 👍🏻

internal/runner/config.go
```diff
@@ -2,102 +2,193 @@ package runner
...
+		if key == envLogLevel {
+			value := pair[idx+1:]
+			level := hclog.LevelFromString(value)
+			if level == hclog.NoLevel {
+				// We warn this
+				log.Warn("log level provided in env var is invalid", value)
+			} else {
+				// We set the log level on the root logger so it
+				// affects all runner logs.
+				r.logger.SetLevel(level)
```
evanphx, line 191: bitmoji

internal/server/singleprocess/state/config.go
```diff
@@ -102,14 +100,6 @@ func (s *State) configSet(
...
 		}
 	} else {
-		// If this is a runner, we don't support dynamic values currently.
-		if value.Target.Runner != nil {
-			if _, ok := value.Value.(*pb.ConfigVar_Static); !ok {
-				return status.Errorf(codes.FailedPrecondition,
-					"runner-scoped configuration must be static")
-			}
-		}
-
```
evanphx, line 112: bitmoji

evanphx (approved):
:shipit:

PR #2204, opened 2021-09-02
server/singleprocess: poll exit logic

### LINKED ISSUE #2202: server/singleprocess: Update error messaging for polling
**Describe the bug**
The error message here could be updated to be more specific when the poll job fails because there are no active deployments or releases.
This should rarely happen, but for those edge cases, it shouldn't return an error.

**Steps to Reproduce**
* `waypoint init` a project
* Open the UI, and go to the project settings
* Fill out the git repo connection settings, and check "Enable builds on changes"
* The project won't deploy
* You'll eventually see the below in your server logs:
```
2021-09-02T18:52:13.972Z [TRACE] waypoint.server.singleprocess.poll_queuer.application_statusreport: returning peek for apps: project=go
2021-09-02T18:52:13.972Z [TRACE] waypoint.server.singleprocess.poll_queuer.application_statusreport: next poll time: time="2021-09-02 18:52:13.967891267 +0000 UTC m=+76.932768908"
2021-09-02T18:52:13.972Z [TRACE] waypoint.server.singleprocess.poll_queuer.application_statusreport: waiting on watchset and contexts
2021-09-02T18:52:13.972Z [TRACE] waypoint.server.singleprocess.poll_queuer.application_statusreport: queueing poll jobs
2021-09-02T18:52:13.972Z [TRACE] waypoint.server.singleprocess.poll_queuer.application_statusreport.go.go: looking at latest deployment and release to generate status report on
2021-09-02T18:52:13.972Z [WARN]  waypoint.server.singleprocess.poll_queuer.application_statusreport: error building a poll job request. This should not happen repeatedly. If you see this in your log repeatedly, report a bug.: err="rpc error: code = NotFound desc = No application named "go" is available!"
```

I used https://github.com/briancain/go-gitops on Waypoint 0.5.1

**Expected behavior**
Waypoint should not return an error on polling if there are no deploys/releases to poll.

applicable:

$ waypoint version
CLI: v0.5.1 (76d7e17f)
Server: v0.5.1

**Additional context**
This is low priority.

krantzinator:
Fixes #2202

### DIFF
--- a/internal/server/singleprocess/poll_application.go
+++ b/internal/server/singleprocess/poll_application.go
@@ -113,13 +113,13 @@ func (a *applicationPoll) buildPollJobs(
 
 	log.Trace("looking at latest deployment and release to generate status report on")
 	latestDeployment, err := a.state.DeploymentLatest(appRef, &pb.Ref_Workspace{Workspace: a.workspace})
+	// If the deployment isn't found, it's ok
 	if err != nil && status.Code(err) != codes.NotFound {
-		// If the deployment isn't found, it's ok
 		return nil, err
 	}
 	latestRelease, err := a.state.ReleaseLatest(appRef, &pb.Ref_Workspace{Workspace: a.workspace})
+	// If the release isn't found, it's ok.
 	if err != nil && status.Code(err) != codes.NotFound {
-		// If the release isn't found, it's ok.
 		return nil, err
 	}
 	// Some platforms don't release, so we shouldn't error here if we at least got a deployment
--- a/internal/server/singleprocess/poll_application.go
+++ b/internal/server/singleprocess/poll_application.go
@@ -113,20 +113,14 @@ func (a *applicationPoll) buildPollJobs(
 
 	log.Trace("looking at latest deployment and release to generate status report on")
 	latestDeployment, err := a.state.DeploymentLatest(appRef, &pb.Ref_Workspace{Workspace: a.workspace})
-	if err != nil {
+	if err != nil && status.Code(err) != codes.NotFound {
 		// If the deployment isn't found, it's ok
-		s, ok := status.FromError(err)
-		if !ok && s.Code() != codes.NotFound {
-			return nil, err
-		}
+		return nil, err
 	}
 	latestRelease, err := a.state.ReleaseLatest(appRef, &pb.Ref_Workspace{Workspace: a.workspace})
-	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/server/singleprocess/poll_application.go
```diff
@@ -116,15 +116,15 @@ func (a *applicationPoll) buildPollJobs(
 	if err != nil {
 		// If the deployment isn't found, it's ok
 		s, ok := status.FromError(err)
-		if !ok || s.Code() != codes.NotFound {
+		if !ok && s.Code() != codes.NotFound {
```
mitchellh, line 119: Hm, this doesn't seem right to me. I pulled up the `FromError` impl to try to figure out what we're trying to do here: https://github.com/grpc/grpc-go/blob/41e044e1c82fcf6a5801d6cbd7ecf952505eecb1/status/status.go#L81

I think we can just change this to `status.Code`? (https://pkg.go.dev/google.golang.org/grpc/status#Code)

On L116:

```
if err != nil && status.Code(err) != codes.NotFound {
```

If there was an error and it wasn't a not found error, then we have to return an error.

Is that correct?

mitchellh (changes requested):
I was surprised to see this struct has no tests. We don't need to solve it in this PR, but I don't see a reason why this strut isn't testable and I think for scenarios like this it would benefit greatly.

internal/server/singleprocess/poll_application.go
krantzinator: Q1: I believe there are tests under [poll_tests.go](https://github.com/hashicorp/waypoint/blob/main/internal/server/singleprocess/poll_test.go#L462) for this?

Q2: Good call on the error check updates! Change pushed.

Q3: We poll status per project, so if we exit here on an error because one app has no deployments or releases to poll, then returning an error would cause us to exit with error out of the entire polling loop. We do check on line 126 for nil deployments and releases and return `nil, nil` to not queue a job, but to still continue polling any additional apps in the project.

briancain (approved):
Looks good 👍🏻

internal/server/singleprocess/poll_application.go
mitchellh: > Q1: I believe there are tests under poll_tests.go for this?

Ah, so there are... we should probably move those to `poll_application_test.go`. And I would expect a fix like this to have a regression test associated with it. We can add those later though in the interest of just getting it working.

mitchellh (approved):

PR #2217, opened 2021-09-03
internal/config: file vars not parsing complex HCL types

### LINKED ISSUE #2208: internal/config: not parsing complex types from varfiles
**Describe the bug**
`var-file` flag only parses simple types, not complex types.

**Steps to Reproduce**
Using the example from [waypoint-examples/learn/input-variables](https://github.com/hashicorp/waypoint-examples/tree/main/learn/input-variables), add this to your deploy stanza:
```
deploy {
    use "docker" {
      command = var.testlist
    }
  }
```
Define a variable block like so:
```
variable "testlist" {
  default = []
  type    = list(string)
}
```
create a `test.wpvars` file with the entry
```
testlist = ["echo", "hi"]
```
Run a waypoint up/deploy with the flag `var-file=test.wpvars`.
Observe the panic.
Supply the value through either setting the default value in the variable definition, or via the UI, or via the `-var` flag, and observe no panic.

**Expected behavior**
No panic!

* Waypoint CLI Version: 0.5.1

**Additional context**
We are missing a test case around this.

krantzinator:
Fixes #2208

### DIFF
--- a/.changelog/2217.txt
+++ b/.changelog/2217.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+internal/config: Fix parsing of complex HCL types in `-var-file`
+```
--- a/internal/config/variables/testdata/complex.wpvars
+++ b/internal/config/variables/testdata/complex.wpvars
@@ -1,0 +1,1 @@
+testlist = ["waffles", "more waffles"]
--- a/internal/config/variables/variables.go
+++ b/internal/config/variables/variables.go
@@ -1,6 +1,7 @@
 package variables
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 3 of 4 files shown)


internal/server/proto/server.proto
```diff
@@ -491,8 +491,12 @@ message Variable {
     int64 num = 10;
     // hcl stores values of any complex type in a raw string format, and
     // converts it to hcl when we evaluate variables. This is used when
-    // loading values from a file or from the server/UI.
+    // loading simple typed values from a file or simple/complex types 
+    // from the server/UI.
     string hcl = 3;
+    // complexhcl stores any complex typed hcl value provided via a values file
+    // in bytes, converting it back to its hcl expression when we evaluate variables.
+    bytes complexhcl = 11;
```
mitchellh, line 499: Why do we need two separate types for this? I'm looking at the PR but not seeing the fundamental difference...

krantzinator: Because I am tired and my brain skipped right over making a buffer. 😔

krantzinator: Update incoming

briancain (approved):
Nice! 🧇

catsby (approved):
👍 confirmed the test included panics without this change

mitchellh (approved):

Issue #1875, opened 2021-07-15
UI: Report when a release was not executed


briancain:
**Is your feature request related to a problem? Please describe.**

Now that Waypoint has a way to mark a release phase as unimplemented, the UI should reflect this so it's easier to tell which lifecycle phases were executed for an application. https://github.com/hashicorp/waypoint/pull/1873

**Describe the solution you'd like**

When viewing a release in the UI, mark it as skipped when a plugins releaser message reports as `Unimplemented`. There is probably an elegant way to reflect this in the UI so it's clear.

sabrinako:
After a conversation with @almonk, we decided to hold off on further work for this issue until probably the 0.7 release

PR #2219, opened 2021-09-03
UI indicator of unimplemented releases

sabrinako:
#1875 
ON HOLD: check issue for further info


sabrinako:
Does not align with the redesign work

PR #2220, opened 2021-09-03
Feature/ecs task

catsby:
This implements the TaskLauncher plugin type for AWS ECS, which enables On-Demand Runners to exist on ECS. This also modifies waypoint install targeting ECS to automatically configure the ODR configuration. Things like remote Docker builds should now work out of the box with Kubernetes-based installations. 

This PR is analogous to https://github.com/hashicorp/waypoint/pull/2112 , and has the same caveat of not calling `StopTask` at this time. 

I'm opening this in draft for an early look, while I still want refactor the creation of the two IAM roles because there is a lot of duplicated code there.

Usage:

1. Build and install `waypoint` from this branch
2. Build the docker images for the server and the odr worker with `make docker/server`
3. Tag and push those images to AWS ECR ([instructions here](https://docs.aws.amazon.com/AmazonECR/latest/public/getting-started-cli.html))
4. Install Waypoint server referencing the new docker images: 
> waypoint install -accept-tos -platform ecs -ecs-server-image=<your ecr path>:waypoint-server -ecs-odr-image=<your ecr path>:waypoint-server-odr

When that is complete, try it out with this example `waypoint.hcl` file, which is setup to read a nodejs project from a public repo (note the branch `catsby-us-west-2` if you clone it locally):

- https://github.com/catsby/waypoint-examples/blob/60e54d8c049266941e8123dcd433643690db1fad/aws/aws-ecs/nodejs/waypoint.hcl

1. run `waypoint init` somewhere with that hcl file
2. run `waypoint up -remote` 
3. success


builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,245 @@
...
+
+	subnetStrings := strings.Split(p.config.Subnets, ",")
+	subnets := aws.StringSlice(subnetStrings)
+
+	if _, err := ecsSvc.RunTask(&ecs.RunTaskInput{
+		LaunchType:           aws.String(ecs.LaunchTypeFargate),
+		Cluster:              &p.config.Cluster,
+		Count:                aws.Int64(1),
+		TaskDefinition:       &taskDefArn,
+		EnableECSManagedTags: aws.Bool(true),
```
catsby, line 208: `EnableECSManagedTags` ensures any tasks / containers created by this service will inherit the tags of this service.

builtin/aws/ecr/registry.go
```diff
@@ -154,14 +152,13 @@ func (r *Registry) setupRepo(
 func (r *Registry) AccessInfo(
 	ctx context.Context,
 	log hclog.Logger,
-	img *docker.Image,
```
catsby, line 157: We discovered that the ECR registry will not have any `docker.Image` available to it, which actually prevented this method from ever being called. This method did not use the input either.

builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// StopTask signals to docker to stop the container created previously. This
+// method is currently unimplemented.
+func (p *TaskLauncher) StopTask(
+	ctx context.Context,
+	log hclog.Logger,
+	ti *TaskInfo,
+) error {
+	return nil
```
briancain, line 112: Maybe we should do this for the other ODR task plugins too, but I think we should log a warning that the StopTask function is unimplemented.

briancain (commented):
Looks dope @catsby ! I mostly have a few comments at the moment 😄

internal/serverinstall/ecs.go
```diff
@@ -1735,6 +1795,156 @@ const rolePolicy = `{
   ]
 }`
 
+const odrRolePolicy = `{
```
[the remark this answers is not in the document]
catsby: Great idea, added in 8f3758ce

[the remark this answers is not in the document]
catsby: Feedback welcome on the documentation I added

builtin/aws/ecs/task.go
catsby: Good idea; added a `log.Warn` here in 687b6222

builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	registerTaskDefinitionInput := ecs.RegisterTaskDefinitionInput{
+		ContainerDefinitions:    []*ecs.ContainerDefinition{&def},
+		ExecutionRoleArn:        &exRoleArn,
+		TaskRoleArn:             &taskRoleArn,
+		Cpu:                     aws.String("1024"),
+		Memory:                  aws.String("2048"),
```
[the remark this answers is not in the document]
catsby: It's not currently configurable in the other task implementations. I suppose we could configure it, or at least inherit what the server uses? What do you think?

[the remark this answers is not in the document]
briancain: Hmm, probably would want to be separate. I guess for now we can leave it! I imagine people will want to configure it in the future, we can do a full configuration PR for all the task plugins.

internal/serverinstall/ecs.go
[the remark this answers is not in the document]
briancain: Looks great!

website/content/commands/server-upgrade.mdx
```diff
@@ -52,6 +52,8 @@ manually installed runners will not be automatically upgraded.
 - `-ecs-region=<string>` - Configures which AWS region to install into.
 - `-ecs-cpu=<string>` - Configures the requested CPU amount for the Waypoint server task in ECS.
```
izaaklauer, line 53: Might want to mention this configures the waypoint ODR task cpu too, or make another config field for it. same with mem.

builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,249 @@
...
+Launch an ECS task for on-demand tasks from the Waypoint server.
+
+This will use the standard AWS environment variables and IAM Role information to
+source authentication information for AWS, using the configured task role.
+If no task role name is specified, Waypoint will create one with the required
+permissions.
+`)
+
+	doc.SetField(
+		"odr_task_role_name",
```
evanphx, line 87: Given this is just the generic task plugin, I think let's just change the name of this to `task_role_name` since in the future, it might be used for things other than odr.

evanphx: Oh, Also, I see that the field is actually called task_role_name in the config too, so the docs are just out of date.

builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,249 @@
...
+	Subnets string `hcl:"subnets,optional"`
+
+	// SecurityGroupId is the security group used for the Waypoint tasks.
+	SecurityGroupId string `hcl:"security_group_id,optional"`
+
+	// LogGroup is the CloudWatch log group name to use.
+	LogGroup string `hcl:"log_group,optional"`
+}
+
+func (p *TaskLauncher) Documentation() (*docs.Documentation, error) {
```
evanphx, line 68: Looks like we need to just copy over the docs for the other fields here.

builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,249 @@
...
+
+	taskDefArn := *taskDef.TaskDefinitionArn
+
+	subnetStrings := strings.Split(p.config.Subnets, ",")
+	subnets := aws.StringSlice(subnetStrings)
+
+	if _, err := ecsSvc.RunTask(&ecs.RunTaskInput{
+		LaunchType:           aws.String(ecs.LaunchTypeFargate),
+		Cluster:              &p.config.Cluster,
+		Count:                aws.Int64(1),
```
izaaklauer, line 210: Totally unrelated to this PR, but just a thought - we'll probably need to put some thought into making the waypoint runner HA relatively soon, as people will be relying on it more now.

internal/serverinstall/ecs.go
```diff
@@ -1110,13 +1140,27 @@ func (i *ECSInstaller) InstallFlags(set *flag.Set) {
...
+		Target: &i.config.OdrImage,
+		Usage: "Docker image for the Waypoint On-Demand Runners. This will " +
+			"default to the server image with the name (not label) suffixed with '-odr'.",
+	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "ecs-task-role-name",
+		Target: &i.config.TaskRoleName,
+		Usage: "IAM Execution Role to assign to the on-demand runner. If this is blank, " +
+			"an IAM role will be created automatically with the correct permissions.",
```
evanphx, line 1161:
```suggestion
			"an IAM role will be created automatically with the default permissions.",
```

internal/serverinstall/ecs.go
```diff
@@ -1144,13 +1188,27 @@ func (i *ECSInstaller) UpgradeFlags(set *flag.Set) {
...
```
evanphx, line 1209:
```suggestion
			"an IAM role will be created automatically with the default permissions.",
```

builtin/aws/ecs/task.go
```diff
@@ -0,0 +1,249 @@
...
+	taskRoleArn, err := roleArn(p.config.TaskRoleName, sess)
+	if err != nil {
+		return nil, err
+	}
+
+	registerTaskDefinitionInput := ecs.RegisterTaskDefinitionInput{
+		ContainerDefinitions:    []*ecs.ContainerDefinition{&def},
+		ExecutionRoleArn:        &exRoleArn,
+		TaskRoleArn:             &taskRoleArn,
+		Cpu:                     aws.String("1024"),
```
izaaklauer, line 190: Might want to either use use the `p.config.{cpu,mem}` here, or make a new config entry for this. I can imagine someone needing higher values for complex build jobs.

evanphx (changes requested):
A couple of super minor things, then ship it!

builtin/aws/utils/ecs.go
```diff
@@ -0,0 +1,41 @@
...
+		taskOut, err = ecsSvc.RegisterTaskDefinition(def)
+		if err != nil {
+			return nil, err
+		}
+
+		if taskOut != nil && taskOut.TaskDefinition != nil {
+			break
+		}
+
+		// otherwise sleep and try again
```
izaaklauer, line 27: Can we log something here?

izaaklauer:
Looks really good!

krantzinator (approved):
You already answered the questions I had! 🚀

website/content/commands/server-upgrade.mdx
briancain: It doesn't I think! I had the same question, I think cpu/mem for tasks are currently not configurable. Since you had the same feedback it's probably worth adding the fields.

builtin/aws/ecs/task.go
catsby: Yep good catch, cleaned up in 45509b65

builtin/aws/ecs/task.go
catsby: I documented `odr_image` in bacae6c8, were there others? Not sure what else is needed here

website/content/commands/server-upgrade.mdx
catsby: ODR cpu/mem is not configurable in this PR, which is consistent with other initial ODR implementations if I recall... we can add configuration in a follow up PR

builtin/aws/ecs/task.go
catsby: We can discuss internally. I'm going to resolve this for now.

builtin/aws/ecs/task.go
evanphx: The documentation here will show up in the autogeneration so that's why I figured we should fill it out.

builtin/aws/utils/ecs.go
catsby: We could, but we would need to inject in a `log` or `stepgroup` into the function. I wonder if instead a better idea would be to lower the imposed timeout in the `for` loop, to surface any errors quicker. To my knowledge the only retry-able error would be something eventually consistent with a new IAM role. Thoughts?

izaaklauer: It could also be waiting for a security group or a log group. Even if it's 15 seconds, that combined with slow API responses every iteration could leave someone scratching their head as to what waypoint is up to while it's trapped in this loop. Not a hill I want to die on though.

izaaklauer (approved):

briancain (approved):
Gotta fix the website looks like from the CI, but great job! ✨

builtin/aws/utils/ecs.go
catsby: I added a debug log in eaf767d0

builtin/aws/ecs/task.go
catsby: ODR memory and CPU are configurable as of c5c92919

website/content/commands/server-upgrade.mdx
catsby: ODR memory and CPU are configurable as of c5c92919

builtin/aws/ecs/task.go
catsby: I added documentation for the remaining fields in c5c92919

PR #2223, opened 2021-09-06
ui: de-lint `<LogStream>`

jgwhite:
## Why the change?

Another step closer to running eslint in CI.

These are mostly fixes to add type annotations or satisfy TypeScript, with the exception of the template change which genuinely fixes a bug.

## How do I test it?

1. Check out the branch
2. Point the dev server at a local waypoint server (`yarn ember serve local`)
3. Watch the logs for a running deployment
4. Verify the scroll-following works correctly

### DIFF
--- a/ui/app/components/log-stream/index.hbs
+++ b/ui/app/components/log-stream/index.hbs
@@ -20,7 +20,7 @@
             onEnter=(fn (mut this.isFollowingLogs) true)
             onExit=(fn (mut this.isFollowingLogs) false)
           }}
-          {{did-update this.updateScroll this.logLines.length}}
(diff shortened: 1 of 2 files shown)


ui/app/components/log-stream/index.hbs
```diff
@@ -20,7 +20,7 @@
             onEnter=(fn (mut this.isFollowingLogs) true)
             onExit=(fn (mut this.isFollowingLogs) false)
           }}
-          {{did-update this.updateScroll this.logLines.length}}
```
gregone, line 23: what was that `logLines` doing here? 😄

gregone (approved):

PR #2225, opened 2021-09-06
Ignore node_modules directory

gregone:
This could have been a VSCode issue too, but I started getting `template-lint` errors for addons from the `node_modules` dir. This explicitly ignore the directory.
Screen Shot 2021-09-06 at 16 51 02

### DIFF
--- a/ui/.template-lintrc.js
+++ b/ui/.template-lintrc.js
@@ -2,4 +2,5 @@
 
 module.exports = {
   extends: 'recommended',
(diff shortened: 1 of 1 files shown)


jgwhite (approved):
Seems good. I think this must be an issue with the VS Code extension because `yarn lint:hbs` is nice and quiet.

izaaklauer:
@gregone says this issue is gone now since upgrading vscode and we can close it.

PR #2237, opened 2021-09-07
Scoped config in waypoint.hcl

mitchellh:
This introduces the ability to scope configuration in `waypoint.hcl` to workspaces, labels, and runners. This is the frontend-implementation of the already-merged backend functionality in #2186.

The following shows a single `waypoint.hcl` example that is using _all the functionality_ introduced in this PR. This is unlikely to be practical, I expect people will use bits and pieces at a time.

```hcl
project = "foo"

config {
  env = { "ENV" = "default" }

  workspace "dev" {
    env = { "ENV" = "dev" }
  }

  // Could've just used "workspace" here but showing off...
  label "waypoint/workspace is 'production'" {
    env = { "ENV" = "prod" }
  }

  // These apply to on-demand runners only when working on things related
  // to this project.
  runner {
    env = { "BUILD_ENV" = "dev" }
  }
}

app "api" {
  // can use config in here too
}
```


evanphx (approved):
Look at all this topological soooorting! /cc @izaaklauer

briancain (approved):

izaaklauer (approved):

PR #2241, opened 2021-09-07
internal/server: peeking a job must close the memdb txn

mitchellh:
This fixes a regression caused in #2186 where ODR would never get assigned a job.


catsby (approved):
👍 Thank you!

PR #2244, opened 2021-09-07
contrib/kind-k8s: Use NGINX ingress controller

briancain:
This commit updates the k8s helper to install the NGINX ingress
controller instead of envoy and contour. The contour controller setup is
left commented out if users wish to automatically set that up instead.
They can also apply the configs outside of this script if they wish.


briancain:
I found the NGINX controller easier to work with, so we should install that instead. It also seems to be the "default" ingress controller.

izaaklauer (approved):

Issue #2150, opened 2021-08-26
ui: Leaky state in project settings form


jgwhite:
**Describe the bug**
Values entered in the project settings form appear to “leak” between different projects. This is a UI-only problem.

**Steps to Reproduce**
There are probably a few ways to reproduce this, but is the one I had to hand:

1. Create a new project (with “connect a repository” checked)
2. Enter valid repo details and save
3. Return to “Projects”
4. Create another project (also with “connect a repository” checked)
5. Observe that the fields erroneously contain the data from the first project

https://user-images.githubusercontent.com/34030/130976444-66c4bcd1-26ff-45a3-b4a3-e50a38143677.mp4

**Expected behavior**
The fields should be in their default empty state. Data should not appear to “leak” between projects.

**Waypoint Platform Versions**
* Waypoint Version: 0.5.1

krantzinator:
As an added piece of information, the waypoint server logs include
```
2021-09-02T18:53:07.015Z [TRACE] waypoint.server.singleprocess.poll_queuer.application_statusreport.go.go: looking at latest deployment and release to generate status report on
2021-09-02T18:53:07.015Z [WARN]  waypoint.server.singleprocess.poll_queuer.application_statusreport: error building a poll job request. This should not happen repeatedly. If you see this in your log repeatedly, report a bug.: err="rpc error: code = NotFound desc = No application named "go" is available!"
```

koesbong:
This may be a regression because [a similar issue](https://github.com/hashicorp/waypoint/issues/1397) was created before and solved in [this PR](https://github.com/hashicorp/waypoint/pull/1623).

krantzinator:
An additional piece here is, if you select `Enable automated deployment on changes`, then the Waypoint server freezes and it is impossible to add new projects via the UI.
This might be a completely separate issue or related, let me know if I should make a separate issue here.

Here's a gist of the full stack trace when I SIGQUIT a stuck command after I do the above: https://gist.github.com/krantzinator/d1907654759d22ffb8184d02c8b067a0

For ease of seeing the most relevant bits, Ctrl+F and search for `waypoint` and select the `Highlight All` option next to the search box.

PR #2250, opened 2021-09-08
Fix overwriting default project model

gregone:
Fixes: #2150

`populateExistingFields` was mutating the `DEFAULT_PROJECT_MODEL` properties, thus causing the values to be retained when the component was reused. 

This PR sets the `defaultProject` on the component as a static property, and then reads keys from that object in `populateExistingFields`.

### DIFF
--- a/.changelog/2250.txt
+++ b/.changelog/2250.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: fix leaky project repository settings being reused when creating a new project
+```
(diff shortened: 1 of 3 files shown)


jgwhite (approved):

jgwhite:
Do we need to backport this?

briancain:
Should this issue be linked to this PR? https://github.com/hashicorp/waypoint/issues/2150

> Do we need to backport this?

I think so yes! 🤔

PR #2251, opened 2021-09-08
Vscode Customization - Status Bar Background

pbortnick:
🎟️ [Asana Task](https://app.asana.com/0/1200392284811676/1200498586643633/f)

In an attempt to make it easier when working with multiple vscode windows open at once, we have set out to customize the status bar color of each project to its theme color.

This is a preview of what it looks like:

status_bar_waypoint

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,4 @@ tmp/
 # Default Waypoint database file
 data.db
 
-.vscode
+# .vscode - commented for customization
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,6 @@ tmp/
 # Default Waypoint database file
 data.db
 
-# .vscode - commented for customization
+# .vscode - We intentionally commit .vscode in order to share project-specific settings
+
+
--- a/website/.vscode/settings.json
+++ b/website/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "workbench.colorCustomizations": {
+    "statusBar.foreground": "#000",
+    "statusBar.background": "#19C6CB"
+  },
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "workbench.colorCustomizations": {
+    "statusBar.foreground": "#000",
+    "statusBar.background": "#19C6CB"
+  },
+}


.gitignore
```diff
@@ -13,4 +13,4 @@ tmp/
 # Default Waypoint database file
 data.db
 
-.vscode
+# .vscode - commented for customization
```
jgwhite, line 16: Could we elaborate on this comment? Perhaps “We intentionally commit .vscode in order to share project-specific settings”

website/.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "workbench.colorCustomizations": {
+    "statusBar.foreground": "#000",
+    "statusBar.background": "#19C6CB"
+  },
+}
```
jgwhite, line 6: Given that these are the same as the top-level ones, maybe we stick with just those for now?

jgwhite (changes requested):
I have a couple of comments but overall I’m supportive of this change. Aside from making VS Code look a bit more snazzy when working on Waypoint, it opens the door to adding more shared VS Code config to the project, such as `launch.json` configurations and recommended extensions.

.gitignore
pbortnick: Yes definitely!

website/.vscode/settings.json
pbortnick: Good point but I believe some people might work out of just the `/website` directory so it was suggested to add it in here too

pbortnick: I will confirm this and report back

.gitignore
```diff
@@ -13,4 +13,6 @@ tmp/
 # Default Waypoint database file
 data.db
 
-.vscode
+# .vscode - We intentionally commit .vscode in order to share project-specific settings
```
krantzinator, line 16: But what about all of my personal vscode settings? Have you shared any how-tos on how to keep them separate? I can look it up of course but it'd be nice if that advice was part of this :)

website/.vscode/settings.json
jgwhite: Ah OK, that makes sense, that hadn’t occurred to me.

.gitignore
krantzinator: Although, maybe this would be an OK thing to have. It would encourage us to centralize some of the tooling with our project-specific settings. My project vscode settings right now just have a `launch.json` file that @jgwhite shared with me, and that might be OK to get committed. What do you think Jamie?

jgwhite: Agreed. @krantzinator I’d love to learn some tips and tricks from your local `.vscode` 🙂

jgwhite: And yeah, I think sharing that `launch.json` would be very useful indeed. It’s akin to sharing the `hack` directory.

website/.vscode/settings.json
jgwhite: Given that rationale, I’m going to resolve this.

jgwhite (approved):
This gets my 👍 

If @krantzinator also approves, then I vote we proceed.

krantzinator (approved):
🎉

PR #2253, opened 2021-09-08
Add API to job stream to detect configuration metadata

mitchellh:
Concretely, this introduces a new `config` field to the `Job` structure which includes information about the loaded Waypoint configuration within a job. For now, this structure only contains where the waypoint.hcl came from (file or server). This came up as feedback when using server-hosted waypoint.hcl where it wasn't clear what would take priority.

In the future, I think we can use this same structure to add information about what input variable values were used and/or where they came from. This is something I noticed during complex usage where it wasn't clear if they were coming from env vars, files, server side, etc.

**For client-side API users:** this also adds a new job stream event "job change" that is sent an updated Job structure anytime any field is updated. This can be used to update the CLI, UI, etc. The use case is initially to just detect this `config` field is set but clients can really use it to detect anything.

**TODO:** I will be adding unit tests, but wanted to get this up for @briancain to see and to solicit any feedback.


internal/server/proto/server.proto
```diff
@@ -1426,6 +1433,25 @@ message Job {
     InitResult init = 14;
   }
 
+  message Config {
+    // Source is the location where the configuration was loaded from.
+    Source source = 1;
+
+    // Source is the location where the configuration was loaded from.
+    enum Source {
```
briancain, line 1441: Should this be a `oneof` in case we ever support any other source? 🤔

briancain (approved):
Looks good to me! Had a question about the config source types, but otherwise makes sense to me as far as the backend is concerned.

gregone:
Looks good, I think the stream addition is useful for the logs UI. 

Probably unnecessary in this PR, but I wonder if there wouldn't also be some value in adding the `config` field to the `Operation` struct itself, to be able to display historical config changes in the application view?

gregone (approved):

internal/server/proto/server.proto
mitchellh: We can add fields to the enum! The only reason to do `oneof` is if we want to attach data to the sources. I thought about that but wasn't sure since today both would probably be `google.protobuf.Empty` and I couldn't think of anything here.

mitchellh:
> Probably unnecessary in this PR, but I wonder if there wouldn't also be some value in adding the config field to the Operation structs themselves. That would enable some useful UI features, e.g. be able to display historical config changes in the application view

All operations should point back to jobs (all operations _must_ be operated on by jobs) so I tend to view them as sharing their data quite a lot already. And I think we recently added a way to quickly get the job for an operation? I worry about just copying many fields.

internal/server/proto/server.proto
briancain: Ah, roger that. Makes sense 👍🏻

PR #2254, opened 2021-09-08
restructure scss to use index files for imports

gregone:
Moves all imports in `app.scss` to the top of their respective directories into [`_index` files](https://sass-lang.com/documentation/at-rules/import#index-files).


ui/app/styles/app.scss
```diff
@@ -4,9 +4,7 @@
 @use 'pds/utils/layout' as Layout;
 @use 'pds/tokens/font';
 @use 'pds/core/typography/config' as Typography;
-@use 'utilities/animations';
-
-@use 'utilities/focus-ring';
+@use 'utilities/index';
```
amandakalk, line 7: Probably nitpicky: Do you need to specify index here? I _think_ you can do @use 'utilities'; and it will use the index file.  

Not sure if we have a preferred way of doing this type of import.

gregone: thanks. probably showing my age, but I've worked with older versions of Sass that didn't support the implicit index, so I didn't even try doing that 🧓

amandakalk: And here I was thinking that maybe they'd removed the implicit index since I last used it 😅

jgwhite (approved):
Another breath of fresh air for the styles directory.

PR #2261, opened 2021-09-08
builtin/k8s: Introduce Ingress resource support for Kubernetes Releaser

### LINKED ISSUE #889: support ingress resource in kubernetes plugin
n/a

Currently kubernetes plugin is using Loadbalancer or Node port. In case where we have multiple developers working on different projects Node port opening over network is not feasible, from second side creating multiple Loadbalancers  on Kubernetes is creating quite big cost for infrastructure (for example each AWS ELB is ~$15 per month).

Better and desired feature will be usage of ingress resource which is become now stable in Kubernetes (https://kubernetes.io/docs/concepts/services-networking/ingress/).

I am imagine that there will be additional variable in kubernetes releasemanager: ingress (boolean) which will be activating this. Probably this will be require some additional variable to setup a ingress main FQDN 🤔 under which will be dynamically generated ingress route. 

Some unknown here is how to address dynamic URL  generated by waypoint, as this will be not working in this setup as there will be no IP hosting service directly and generating this to ingress IP will be rejected by missing virtual host in ingress controller.

Above are just ideas and doesn't reflect any desired design as I am new beginner in Waypoint, so feel free to adjust this to most suitable implementation way.

We can still use Loadbalancer or Node port approach but this doesnt fit to secure or economically reasonable Development environment design.

This will open whole bunch of use cases in multiple companies where exists shared development clusters which have some basic security control or are used as a dev consolidated environments

briancain:
This pull request adds a new config option `ingress` within the Kubernetes release manager that allows users to have their releases be configured behind an ingress resource that accepts traffic from an ingress controller in their cluster.

Fixes #889 

This doesn't have to be gated to 0.6.0, it can be released with the next minor.

Some example apps:

- https://github.com/hashicorp/waypoint-examples/pull/77
- https://github.com/hashicorp/waypoint-examples/pull/78


builtin/k8s/releaser.go
```diff
@@ -339,10 +353,366 @@ func (r *Releaser) resourceServiceDestroy(
...
+		}
+
+		if r.config.IngressConfig.ClassName != "http" {
+			return status.Error(codes.FailedPrecondition, "An ingress stanza must be "+
+				"of type \"http\".")
+		}
+	} else {
+		// No ingress config was set, we're not configuring an ingress resource
+		return nil
+	}
```
mitchellh, line 389: Little "boolean Zen" thing here but I would just change this whole thing to:

```go
if r.config.IngressConfig == nil {
    return nil
}

// Do checks
```

So you can lower the overall depth of this by one.

builtin/k8s/releaser.go
```diff
@@ -339,10 +353,367 @@ func (r *Releaser) resourceServiceDestroy(
...
+			Name: serviceState.Name,
+			Port: networkingv1.ServiceBackendPort{
+				Name:   serviceBackend.Spec.Ports[0].Name,
+				Number: serviceBackend.Spec.Ports[0].Port,
+			},
+		},
+	}
+
+	// NOTE, the k8s go api doesn't seem to like being set directly from the
+	// string config in our releaser, so we manually set it with this switch
```
evanphx, line 476: You could get around this by casting the string in `PathType` to a `networkingv1.PathType` when assigning it.

builtin/k8s/releaser.go
```diff
@@ -339,10 +353,367 @@ func (r *Releaser) resourceServiceDestroy(
...
+	var pathType networkingv1.PathType
+	switch r.config.IngressConfig.PathType {
+	case "Exact":
+		pathType = "Exact"
```
evanphx, line 480: Are there constants in the networkingv1 package we can use instead of strings here? Since this appears to be an enum I'd rather use something stronger typed than a string.

evanphx (changes requested):
Just a minor thing about some enum-like values.

builtin/k8s/releaser.go
briancain: I believe the values were constants yeah! I'll give that a shot instead

briancain: Yeah that worked! 👍🏻

builtin/k8s/releaser.go
```diff
@@ -65,6 +67,13 @@ func (r *Releaser) resourceManager(log hclog.Logger, dcr *component.DeclaredReso
 			resource.WithDestroy(r.resourceServiceDestroy),
 			resource.WithStatus(r.resourceServiceStatus),
 		)),
+		resource.WithResource(resource.NewResource(
```
izaaklauer, line 70: Just a note - I feel like it might make sense make resource manager aware that some resources are optional. Right now, it calls create, destroy, and status on every resource, and we just return nil quickly if it isn't enabled. It might be nice to add `resource.withOptional(func)`, rm would call that before create, and then not call any other resource function if the resource isn't enabled.

builtin/k8s/releaser.go
```diff
@@ -65,6 +67,13 @@ func (r *Releaser) resourceManager(log hclog.Logger, dcr *component.DeclaredReso
...
+			resource.WithName("ingress"),
+			resource.WithState(&Resource_Ingress{}),
+			resource.WithCreate(r.resourceIngressCreate),
+			resource.WithDestroy(r.resourceIngressDestroy),
```
izaaklauer, line 74: Does this mean that when we destroy one deployment, it'll destroy the whole ingress resource that's in use by all deployments??

builtin/k8s/releaser.go
briancain: That would be nice. We'd need that function to basically asses the config passed in to figure out whether or not it should be called or maybe just accept some kind of anonymous function so the caller can dictate when the optional resource should be used.

I wouldn't gate this PR on that feature existing, but maybe that's something we could tackle soon and remove the early return check at the beginning of some of these functions once it's in place. I am sure larger services like ECS will have optional resources too (if not already)!

builtin/k8s/releaser.go
briancain: Nope, it should stick around the same as the Service does that gets created on release. Good question tho!

builtin/k8s/releaser.go
briancain: Done!

izaaklauer (approved):

PR #2262, opened 2021-09-08
Set and get new config scopes from the CLI

mitchellh:
This PR finishes up the coding part of #2169. This modifies the CLI to support all the new scoping options. 

I admit this isn't the cleanest UX possible. I'd love to challenge this a bit more with folks in the future and figure out how we can maybe make the `waypoint config` CLIs a little easier to use perhaps. Right now you gotta balance a bunch of flags.My counter argument is people probably won't use the CLI for more advanced config setting and will prefer to use the config file or UI (tbd). 

In terms of the bigger picture #2169, I need to work up docs still. But this should finish all the main coding parts.

### DIFF
--- a/internal/cli/config_set.go
+++ b/internal/cli/config_set.go
@@ -16,7 +16,10 @@ import (
 type ConfigSetCommand struct {
 	*baseCommand
 
-	flagRunner bool
+	flagGlobal  bool
+	flagRunner  bool
+	flagProject string
+	flagScope   string
 }
 
 func (c *ConfigSetCommand) Run(args []string) int {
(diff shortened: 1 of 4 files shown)


evanphx (approved):
Looks good, I agree, we should probably do a little CLI UX work soon to see about better ways to configured some of these things without a sea of flags.

internal/cli/config_set.go
```diff
@@ -128,13 +167,51 @@ func (c *ConfigSetCommand) Flags() *flag.Sets {
 	return c.flagSet(0, func(set *flag.Sets) {
 		f := set.NewSet("Command Options")
 
+		f.StringVar(&flag.StringVar{
+			Name:   "scope",
+			Target: &c.flagScope,
+			Usage: "The scope for this configuration. The configuration will only " +
+				"appear within this scope. This can be one of 'global', 'project', or" +
```
catsby, line 174:
```suggestion
				"appear within this scope. This can be one of 'global', 'project', or " +
```

catsby (approved):
👍

briancain (approved):

Issue #2191, opened 2021-09-02
Input variable value should truncate in list view


jgwhite:
**Describe the bug**
In the list of input variables, long values overrun their containers:

<img width="1009" alt="CleanShot 2021-09-02 at 12 08 19@2x" src="https://user-images.githubusercontent.com/34030/131825640-6e4a8b46-386e-43ec-b960-e3ed339ac6a3.png">

**Steps to Reproduce**
1. Add an input variable with a long value
2. Observe the value overrunning as pictured above
 
**Expected behavior**
The value should be elegantly truncated.

**Versions**

* Waypoint CLI Version: 0.5.1
* Safari Version 14.1.2 (16611.3.10.1.6)

PR #2267, opened 2021-09-09
truncate display of long variable values

gregone:
Fixes https://github.com/hashicorp/waypoint/issues/2191 

Screen Shot 2021-09-09 at 18 30 37


jgwhite (approved):
LGTM

Issue #2260, opened 2021-09-08
tests/Frontend/ember-build-tests: Unhandled error event


briancain:
**Describe the bug**

The ember build frontend tests seem to be failing often. I think this test specifically has failed for me about ~5 times today.

https://app.circleci.com/pipelines/github/hashicorp/waypoint/9896/workflows/0f6d5e84-1e59-47ef-9916-75f8f0d22a36/jobs/87707

```
WARNING: Detected collisions between .js and .ts files of the same name. This can result in nondeterministic build output; see https://git.io/JvIwo for more information.
  - waypoint/services/flash-messages.{js,ts}
[BABEL] Note: The code generator has deoptimised the styling of /home/circleci/project/ui/node_modules/axe-core/axe.js as it exceeds the max of 500KB.
[BABEL] Note: The code generator has deoptimised the styling of /home/circleci/project/ui/node_modules/waypoint-pb/server_pb.js as it exceeds the max of 500KB.
events.js:291
      throw er; // Unhandled 'error' event
      ^

Error [ERR_IPC_CHANNEL_CLOSED]: Channel closed
    at ChildProcess.target.send (internal/child_process.js:680:16)
    at Object.sendMessage (/home/circleci/project/ui/node_modules/stagehand/lib/adapters/child-process.js:30:39)
    at CommandCoordinator.sendMessage (/home/circleci/project/ui/node_modules/stagehand/lib/command-coordinator.js:81:23)
    at CommandCoordinator.sendCommand (/home/circleci/project/ui/node_modules/stagehand/lib/command-coordinator.js:36:14)
    at Stagehand.<anonymous> (/home/circleci/project/ui/node_modules/stagehand/lib/stagehand.js:86:44)
    at Generator.next (<anonymous>)
    at /home/circleci/project/ui/node_modules/stagehand/lib/stagehand.js:7:71
    at new Promise (<anonymous>)
    at __awaiter (/home/circleci/project/ui/node_modules/stagehand/lib/stagehand.js:3:12)
    at Stagehand.call (/home/circleci/project/ui/node_modules/stagehand/lib/stagehand.js:82:16)
Emitted 'error' event on ChildProcess instance at:
    at internal/child_process.js:684:35
    at processTicksAndRejections (internal/process/task_queues.js:79:11) {
  code: 'ERR_IPC_CHANNEL_CLOSED'
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command
[...]

PR #2272, opened 2021-09-09
ui: update to node@14 (current LTS)

jgwhite:
This is another attempt to address #2260 🤞


jgwhite:
Well, this PR actually suffered the #2260 error, so I guess that invalidates this hypothesis 😅

PR #2273, opened 2021-09-09
internal/server: Add many validations to authenticated endpoints

### LINKED ISSUE #2270: internal/server: Add API validations to all authenticated endpoints
**Describe the bug**

Similar to #1700, we should make sure all of our authenticated endpoints have API validations to avoid issues like https://github.com/hashicorp/waypoint/issues/1682.

**Steps to Reproduce**

Certain endpoints could cause the server to panic if a malformed request was given. I don't have an extensive list of which those could be beyond the `GetDeployments` endpoint. So part of this work will be going through every service endpoint and making sure it responds properly.

**Expected behavior**

The server doesn't panic if a bad request was sent.

briancain:
This pull request adds many validations to various authenticated endpoints that would panic if given an empty request body.

Fixes https://github.com/hashicorp/waypoint/issues/2270

### DIFF
--- a/internal/server/ptypes/project.go
+++ b/internal/server/ptypes/project.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"strings"
 
-	"github.com/go-ozzo/ozzo-validation/v4"
+	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	hcljson "github.com/hashicorp/hcl/v2/json"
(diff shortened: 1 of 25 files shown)


izaaklauer (approved):
So good!

krantzinator (approved):
3 issues in 1! 🚀

internal/server/ptypes/project.go
```diff
@@ -81,6 +81,20 @@ func ValidateUpsertProjectRequest(v *pb.UpsertProjectRequest) error {
...
 }
 
+// ValidateGetProjectRequest
+func ValidateGetProjectRequest(v *pb.GetProjectRequest) error {
+	return validationext.Error(validation.ValidateStruct(v,
+		validation.Field(&v.Project, validation.Required),
+	))
+}
+
+// ValidateGetProjectRequest
```
jgwhite, line 91: [sand] This comment doesn’t match the function below it. But also: these comments don’t add much, could we drop them altogether?

jgwhite (approved):
Looks air-tight to me. I tested it quickly against the web UI and the errors come through pretty nicely:

CleanShot 2021-09-13 at 12 56 00@2x

(In reality you can’t submit the form without entering something in the name field, so users shouldn’t ever get into this state, but useful to see what these validation messages look like in general)

internal/server/ptypes/project.go
briancain: Oops, yeah probably from all the copy paste to get the bare functions around before adding the content :D I can just drop the comment.

briancain: If I were to do this again I'd probably delete them but now that they are there I'll probably leave them be 😅

PR #2275, opened 2021-09-10
Update ember-a11y-testing to 4.2.0

gregone:


gregone:
closing for now, we could still upgrade, but this was just to test if the `axe-core` dep could be causing the build failures

PR #2280, opened 2021-09-10
ui: Updating empty state messages for projects page

### LINKED ISSUE #2279: Update empty project screen content
When there are no projects, the UI currently advises the user to go to the CLI to create a new project when you can create a project in the UI. Seems like we should be advising the user to continue in the UI rather than jump back to the CLI.

Screen Shot 2021-09-10 at 1 18 00 PM

Update the text to tell the user to click the new project button. Or just delete the message instructing the user to go to the CLI as it is likely intuitive that they can click the new project button.

amandakalk:
This pull request is a small edit to the empty state message for the projects page. The new text includes a link to the "New Project" page.

Previously:
Screen Shot 2021-09-10 at 1 18 00 PM

Now:
Screen Shot 2021-09-10 at 2 21 14 PM

Fixes #2279

### DIFF
--- a/ui/app/templates/workspace/projects/index.hbs
+++ b/ui/app/templates/workspace/projects/index.hbs
@@ -42,12 +42,7 @@
   </Card>
   {{else}}
   <EmptyState>
-    <p>There are no projects to display yet</p>
-    <p>To create your first project, create a new application by running
-      <CopyableCode @ref="empty-init" @inline="true">
-        <code id="empty-init">waypoint init</code>
-      </CopyableCode>
-      from the CLI</p>
+    <p>There are no projects to display yet. Create your first project <LinkTo @route="workspace.projects.new">here</LinkTo>.</p>
     <p>For more help getting started, refer to the <ExternalLink href="https://waypointproject.io/docs/getting-started">Waypoint documentation</ExternalLink></p>
   </EmptyState>
   {{/each}}
--- a/ui/app/templates/workspace/projects/index.hbs
+++ b/ui/app/templates/workspace/projects/index.hbs
@@ -42,7 +42,12 @@
   </Card>
   {{else}}
   <EmptyState>
-    <p>There are no projects to display yet. Create your first project <LinkTo @route="workspace.projects.new">here</LinkTo>.</p>
+    <p>There are no projects to display yet. <LinkTo @route="workspace.projects.new">Create your first project here</LinkTo>.</p>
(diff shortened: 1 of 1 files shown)


gregone:
👍 on updating this. 

That said, before approving, I think we should keep the cli command, as that's also quite helpful for people getting started with Waypoint. As much as I appreciate the idea of UI-only workflows for people who already have command of waypoint, not telling them about the basics could hurt adoption.

It probably could be displayed on a line below. 

e.g. "Alternatively, you can run `waypoint init` in your application directory to generate a `waypoint.hcl` file"

briancain:
> That said, before approving, I think we should keep the cli command

Agreed, I think showing both is a good idea!

ui/app/templates/workspace/projects/index.hbs
```diff
@@ -42,12 +42,7 @@
   </Card>
   {{else}}
   <EmptyState>
-    <p>There are no projects to display yet</p>
-    <p>To create your first project, create a new application by running
-      <CopyableCode @ref="empty-init" @inline="true">
-        <code id="empty-init">waypoint init</code>
-      </CopyableCode>
-      from the CLI</p>
+    <p>There are no projects to display yet. Create your first project <LinkTo @route="workspace.projects.new">here</LinkTo>.</p>
```
jgwhite, line 45: I’d be inclined to make the whole phrase the link, rather than just `here`, as per as [WCAG Success Criterion 2.4.9 Link Purpose (Link Only)](https://www.w3.org/TR/WCAG21/#link-purpose-link-only). Link-in-context is meets the A criteria, but link-only gets us all the way to AAA. It becomes clear why link-only gets more points when you brink up the “Links” list in VoiceOver’s rotor:

CleanShot 2021-09-13 at 11 39 05@2x

jgwhite (changes requested):
Great initiative, just one comment regarding accessibility.

amandakalk:
Thanks all! I've updated to include CLI instructions and fixing the accessibility issue.

@gregone Does this text look ok to you? I just stuck to the original text but swapped the start with "Alternatively." I decided against including the bit about the `waypoint.hcl` because it seemed like more information than we needed to give.

Screen Shot 2021-09-13 at 9 41 15 AM

jgwhite (approved):
Looks good to me.

briancain (approved):

gregone (approved):

Issue #2173, opened 2021-08-27
`waypoint deploy` to Nomad fails with CSI volumes


konsti:
**Describe the bug**
Deploying a service with Waypoint to Nomad fails when the Nomad job file contains CSI volumes. The nomad API responds with an 500 error:
`
```sh
➜ waypoint deploy

» Deploying...
❌ Registering job "example"...
! Unexpected response code: 500 (1 error occurred:
  	* Task group example validation failed: 1 error occurred:
  	* Task group volume validation for example_volume failed: 2 errors
  occurred:
  	* CSI volumes must have an attachment mode
  	* CSI volumes must have an access mode
```
This seems related to the recent Nomad 1.1.0 change that now defines a capabilities stanza. Our volume has an attachment mode and an access mode.

**More important:** if we schedule the job file directly in Nomad, the job schedules & runs correctly. The volume is mounted.

We mount an **EFS** volume in this case, but the same bug happens also with **EBS** volumes. We have not tested other CSI plugins.

Nomad Version: `v1.1.3`
EFS CSI Plugin `v1.3.3`

**Steps to Reproduce**

##### `waypoint.hcl`
```
project = "example"

app "example" {
  build {
    use "docker" {}
  }

  deploy {
    use "nomad-jobspec" {
      jobspec = templatefile("${path.app}/job.nomad")
    }
  }
}

```

##### `job.nomad`
```
job "example" {
  region      = "eu-west-1"
  datacenters = ["eu-west-1a", "eu-west-1b", "eu-west-1c"]
  type        = "service"

  group "example" {
    count = 1

    volume "example_volume" {
      type            = "csi"
      source          = "example_volume"
      read_only       = false
      attachment_mode = "file-system"
      access_mode     = "multi-node-multi-writer"

      mount_options {
        fs_type     = "ext4"
        mount_flags = ["noatime"]
      }
    }

    task "example" {
      driver = "docker"

      config {
        image = "${artifact.image}:${artifact.tag}"
      }

      volume_mount {
        volume = "example_volume"
        destination = "/opt/volume"
      }

      env {
        %{ for k,v in entrypoint.env ~}
        %{ if k != "WAYPOINT_SERVER_ADDR" ~}
        ${k} = "${v}"
        %{ endif ~}
        %{ endfor ~}
        WAYPOINT_SERVER_TLS_SKIP_VERIFY = "1"
      }

      template {
        data = <<EOH
[...]

Issue #1348, opened 2021-04-16
waypoint install (nomad) has no persistence


nateberkopec:
Had a waypoint server installed by `waypoint install` on Nomad, which was then rescheduled for some reason or another (not really sure).

This wiped out the `context` I had because the IP changed (which I could fix by editing the `context` HCL manually) but to my surprise, the `auth_token` no longer worked.

I'm guessing this is because the job created by the `waypoint install` task has no volumes or other persistence. 

Not sure what the right solution is here (looks like k8s has some `-k8s-storage-*` options). I think the current situation is dangerous, because if you don't realize there's no persistence configured, you could easily run with this in prod and get woken up to a bad outcome when all your waypoint state is wiped out and you can't deploy.

I think I'm going to solve this for myself by just deploying my own Waypoint servers and runners using Nomad jobs.

briancain:
Hey there @nateberkopec ! Thanks for bringing this to our attention. As far as persistence and volumes go, our server job _should_ be configured with ephemeral sticky disks and should migrate to a new job: https://github.com/hashicorp/waypoint/blob/cae85b6ce14431e633de6222caa7c2ea56541884/internal/serverinstall/nomad.go#L632-L637

Starting a new Nomad job happens intentionally when we upgrade, but from my testing I didn't lose the `auth_token` or any other server state when the new job was submitted (although a new IP is apparently a known issue in Nomad). So it's strange the `auth_token` didn't work.

Is there anything in the Nomad logs that might give us a clue to what could of happened with that task? Thanks!

nateberkopec:
I think I lost the node in question, which means that everything got lost... I can't find the allocation that got lost anymore, IDK why.

So I guess there's really not any improvement to be had by specifying my own volume then (unless that volume is actually backed by EBS or the equivalent block storage).

@briancain So, how am I _supposed_ to recover from situations where my token is lost or no longer working? e.g. where I have a running waypoint server, but for some reason my local CLI context breaks (IP changes, whatever) and maybe my token doesn't work (IDK?). If I `docker exec` into the running waypoint server container, `waypoint token new` returns `! failed to create client: open data.db: permission denied`

briancain:
@nateberkopec - it doesn't look like we have a great way to support recovering tokens at the moment. From talking to @mitchellh it sounds like there's a potential work around to recover a token:

- Shut down the server
- Copy the data.db just in case things go wrong, but stay with the production one.
- run `waypoint token new`. You should get a token.
[...]

PR #2282, opened 2021-09-10
plugin/nomad: Persistence when installing Waypoint on Nomad

xiaolin-ninja:
This began with https://github.com/hashicorp/waypoint/issues/1348,
and should also resolve https://github.com/hashicorp/waypoint/issues/2173.

Persistence with host volume is fully tested and functional.
Persistence with CSI volume still needs some work...
To-do: resolve issue of Nomad not connecting to the public IP of the EC2 Instance.


internal/serverinstall/nomad.go
```diff
@@ -133,6 +139,39 @@ func (i *NomadInstaller) Install(
...
+					AccessMode:     "single-node-writer",
+					AttachmentMode: "file-system",
+				},
+			},
+			MountOptions: &api.CSIMountOptions{
+				FSType:     "xfs",
+				MountFlags: []string{"noatime"},
+			},
+			RequestedCapacityMin: 1073741824,
+			RequestedCapacityMax: 2147483648,
```
briancain, line 163: It seems like these should be configurable too? We should also abstract them out as defaults and include a comment on the size in MB or GB

internal/serverinstall/nomad.go
```diff
@@ -890,6 +990,24 @@ func (i *NomadInstaller) InstallFlags(set *flag.Set) {
 		Target: &i.config.consulServiceBackendTags,
 		Usage:  "Tags for the Waypoint backend service generated in Consul.",
 	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "nomad-volume-type",
+		Target: &i.config.volumeType,
+		Usage:  "Nomad volume type for the Waypoint server.",
```
briancain, line 997: We should include possible valid values here if we can. Is it just `csi` and `host`?

internal/serverinstall/nomad.go
```diff
@@ -410,6 +452,18 @@ func (i *NomadInstaller) Uninstall(ctx context.Context, opts *InstallOpts) error
...
+
+	vols, _, err := client.CSIVolumes().List(&api.QueryOptions{Prefix: "waypoint"})
+	for _, vol := range vols {
+		if vol.ID == "waypoint" {
+			s.Update("Destroying persistent CSI volume")
+			err = client.CSIVolumes().Deregister(vol.ID, false, &api.WriteOptions{})
+			if err != nil {
+				return err
+			}
+			s.Update("Successfully destroyed persistent volumes")
```
briancain, line 464: Will there ever be more than 1 volume for Waypoint? maybe we could do a `break` here so we don't keep looping once we find the volume id we're after.

internal/serverinstall/nomad.go
```diff
@@ -661,14 +715,63 @@ func waypointNomadJob(c nomadConfig, rawRunFlags []string) *api.Job {
...
+	volumeRequest := api.VolumeRequest{
+		Type:     c.volumeType,
+		ReadOnly: false,
 	}
+
+	if strings.ToLower(c.volumeType) == "csi" {
+		volumeRequest.Source = "waypoint"
+		volumeRequest.AccessMode = "single-node-writer"
+		volumeRequest.AttachmentMode = "file-system"
+	} else {
```
briancain, line 729: We should be explicit here and check for `c.VolumeType == "host"`. Then the else could be an error saying we got an unrecognized or unsupported volume type. Alternatively we could also do this near the beginning when we're checking for missing flags so we don't go through the whole process and error here 🤔

briancain (changes requested):
Nice! Left a few requests, will test after but overall looks great! 😃

internal/serverinstall/nomad.go
xiaolin-ninja: Yeah the checks to make sure the flags match up are up https://github.com/hashicorp/waypoint/blob/d42221d58f1c481f139731d077caa1e9c31f0036/internal/serverinstall/nomad.go#L142

briancain: Ah yeah, I mean we should error if they give us an unsupported volume type, like if they accidentally type `-nomad-volume-type = "something-not-real"`

briancain: Hm, now that I've tried this locally, I wonder if it's clearer to eliminate the `-nomad-volume-type` all together? So if a user requests `-nomad-csi-volume-provider=<string>` or ` -nomad-host-volume=<string>` the type is implied by the flag requested? What do you think? 🤔

internal/serverinstall/nomad.go
```diff
@@ -133,6 +141,54 @@ func (i *NomadInstaller) Install(
 		}, nil
 	}
 
+	if strings.ToLower(i.config.volumeType) == "csi" {
+		var (
+			// bytes
+			defaultCSIVolumeCapacityMin = int64(1073741824)
+			defaultCSIVolumeCapacityMax = int64(2147483648)
+		)
```
evanphx, line 149: These should probably be `const` at the file level, since it looks like they're establishing our expected defaults, and we generally put our expected defaults as consts at the top of the file.

internal/serverinstall/nomad.go
```diff
@@ -133,6 +141,54 @@ func (i *NomadInstaller) Install(
...
+			ID:   "waypoint",
+			Name: "waypoint",
+			RequestedCapabilities: []*api.CSIVolumeCapability{
+				{
+					AccessMode:     "single-node-writer",
+					AttachmentMode: "file-system",
+				},
+			},
+			MountOptions: &api.CSIMountOptions{
+				FSType:     "xfs",
```
evanphx, line 167: Does this need to be configurable?

internal/serverinstall/nomad.go
evanphx: I agree with @briancain, I'd set the type based on what other options they pass in (note also if you do that, you need to error out if they pass in both csi-volume-provider and a host-volume flags).

evanphx (changes requested):
Just a couple minor things and then you're there!

internal/serverinstall/nomad.go
xiaolin-ninja: oooh, yeah I didn't make them global because I thought they weren't used outside this method, but this just reminded me I forgot to put them as defaults in the flags

internal/serverinstall/nomad.go
```diff
@@ -3,6 +3,8 @@ package serverinstall
 import (
 	"context"
 	"fmt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
```
izaaklauer, line 7: I'd put these in the middle import chunk

izaaklauer (approved):

izaaklauer (approved):
Looks good!

internal/serverinstall/nomad.go
```diff
@@ -661,14 +733,63 @@ func waypointNomadJob(c nomadConfig, rawRunFlags []string) *api.Job {
...
+	}
+	cpu := defaultResourcesCPU
+	mem := defaultResourcesMemory
+
+	preTask := api.NewTask("pre_task", "docker")
+	// Observed WP user and group IDs in the published container, update if those ever change
+	waypointUserID := 100
+	waypointGroupID := 1000
+	preTask.Config = map[string]interface{}{
+		// TODO(xx): pin busybox image
```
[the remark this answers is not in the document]
briancain: That makes sense! If pulling `latest` becomes an issue down the road we can always add another flag to pin this specific image, but defaulting to pull latest for now seems fine 👍🏻

evanphx (approved):

briancain (approved):

PR #2286, opened 2021-09-13
disable sourcemaps in non-dev builds

gregone:
Enables source maps only for development builds, to speed up the build times. Part of proposed changes in the investigation on CI failures: https://github.com/hashicorp/waypoint/issues/2260

### DIFF
--- a/ui/ember-cli-build.js
+++ b/ui/ember-cli-build.js
@@ -1,11 +1,13 @@
 'use strict';
-
+let environment = process.env.EMBER_ENV || 'development';
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
 module.exports = function (defaults) {
+  // Disable sourceMaps except in dev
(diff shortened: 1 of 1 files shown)


jgwhite:
Might be interesting to rebase this atop #2291 to see some basic memory stats.

gregone:
@jgwhite Indeed, it doesn't have an impact on memory usage, only on build time: 

Screen Shot 2021-09-14 at 14 49 11

jgwhite:
Hmm. What do you think? Should we proceed with this change if it doesn’t help the memory usage issue?

gregone:
@jgwhite I was on the fence, it will have a minor impact on build time, which is nice. But the downside is that when I run tests locally with `ember test` (so not in dev env), I often appreciate having sourcemaps to debug, so that makes me think we shouldn't.

PR #2290, opened 2021-09-13
ui: fix for hcl field not showing up when initial value is undefined

sabrinako:
## Explanation
Upon implementing the new CodeMirror based HCL text field (#2168), a regression was introduced on the project settings page when the project's value for their HCL file was `undefined`. The CodeMirror field would not render at all when given this unexpected value.

## Buggy UI
Screen Shot 2021-09-13 at 3 09 12 PM

## UI After This PR
Screen Shot 2021-09-13 at 3 09 47 PM

### DIFF
--- a/ui/app/modifiers/code-mirror.ts
+++ b/ui/app/modifiers/code-mirror.ts
@@ -17,7 +17,7 @@ const _PRESET_DEFAULTS: codemirror.EditorConfiguration = {
 interface Args {
   positional: never;
   named: {
-    value: string;
+    value?: string;
     onInput: (value: string) => void;
     options?: Record<string, unknown>;
   };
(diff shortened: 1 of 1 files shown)


ui/app/modifiers/code-mirror.ts
```diff
@@ -42,7 +42,7 @@ export default class CodeMirrorModifier extends Modifier<Args> {
     let editor = codemirror(this.element, {
       ..._PRESET_DEFAULTS,
       ...this.args.named.options,
-      value: this.args.named.value,
+      value: this.args.named.value ? this.args.named.value : '',
```
jgwhite, line 45: [sand] I’d encourage use of `??` in cases like this, just cause it’s slightly easier to scan than a ternary.

```suggestion
      value: this.args.named.value ?? '',
```

jgwhite (approved):
👍

PR #2295, opened 2021-09-13
Plugin status report implementations can get access to the last report

mitchellh:
This makes it so that plugin status report implementations can access the directly previous report. The previous report will be for the exact deployment/release (by ID). 

This **also** changes (fixes?) the `GetLatestStatusReport` API to support specifying a target so you can get the latest depolyment report, release report, etc.

The primary motivation for this is to able to selectively cache reports to avoid API rate limits, or to selectively only refresh a partial set of resources. There isn't SDK _framework_ support for this but you can immediately add a `last *sdk.StatusReport` argument to your StatusFuncs and get access to the last report and use that however you want.

The first status call will have a `nil` last report, which means it is the first report.


briancain (approved):
Dope! ✨ 📝

izaaklauer (approved):

PR #2296, opened 2021-09-14
builtin/k8s - Use sequence # in k8s deploy name instead of ID

### LINKED ISSUE #2060: plugin/k8s: include sequence number in resource names
Currently, it is difficult to look at a kubernetes pod or deployment and understand which waypoint deployment it corresponds to. We encode the internal operation ULID on the deployment, but we don't expose that it users anywhere else.

We should use the waypoint deployment sequence number instead of (or in addition to if necessary), the internal operation ULID as part of the deployment name.

sl1pm4t:
This PR adds support in waypoint core to pass the deployment sequence number to plugins, and also modifies the k8s plugin to use sequence number in k8s deployment resource names instead of the ID. 

I opted to entirely replace the Waypoint ID with Sequence # in the k8s deployment name because k8s already assigns several random strings to pods derived from deployments so having the Waypoint ID is noise in my opinion. After this change, deployment names become:
```
NAME              READY   UP-TO-DATE   AVAILABLE   AGE
httpbin-v1        1/1     1            1           34m
httpbin-v2        1/1     1            1           7m1s
```
Where `httpbin` is the waypoint app name.
Pods therefore become:
```
NAME                          READY   STATUS    RESTARTS   AGE
httpbin-v1-7c6bb9f655-n4f6j   1/1     Running   0          36m
httpbin-v2-6fb5898b7d-drvct   1/1     Running   0          9m2s
```

This PR depends on a [change to the waypoint-plugin-sdk](https://github.com/hashicorp/waypoint-plugin-sdk/pull/58).

Closes #2060

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ data.db
 # .vscode - We intentionally commit .vscode in order to share project-specific settings
 
 
+.idea
--- a/internal/core/app_deploy.go
+++ b/internal/core/app_deploy.go
@@ -144,11 +144,15 @@ type deployOperation struct {
 	// id is populated with the deployment id on Upsert
 	id string
 
+	// sequence is the monotonically incrementing version number of
+	// the deployment
+	sequence uint64
+
 	// cebToken is the token to set for the deployment to auth
 	cebToken string
 
 	// result is either a component.Deployment or component.DeploymentWithUrl
-	result interface{}
+	result   interface{}
 }
 
 func (op *deployOperation) Close() error {
@@ -272,6 +276,7 @@ func (op *deployOperation) Upsert(
 	if op.id == "" {
 		// Set our internal ID for the Do step
 		op.id = resp.Deployment.Id
+		op.sequence = resp.Deployment.Sequence
 
 		// We need to get our token that we'll give this deployment
 		resp, err := client.GenerateInviteToken(ctx, &pb.InviteTokenRequest{
@@ -297,6 +302,7 @@ func (op *deployOperation) Upsert(
 	// Set the new values on our deployment config
 	dconfig := *op.DeploymentConfig
 	dconfig.Id = op.id
+	dconfig.Sequence = op.sequence
 	dconfig.EntrypointInviteToken = op.cebToken
 	op.DeploymentConfig = &dconfig
 
--- a/internal/core/app_deploy.go
+++ b/internal/core/app_deploy.go
@@ -152,7 +152,7 @@ type deployOperation struct {
 	cebToken string
 
 	// result is either a component.Deployment or component.DeploymentWithUrl
-	result   interface{}
+	result interface{}
 }
 
 func (op *deployOperation) Close() error {
--- a/.changelog/2296.txt
+++ b/.changelog/2296.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/k8s: Use sequence number in k8s deployment name for improved traceability to waypoint deployments.
+```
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -16,3 +16,4 @@ data.db
 # .vscode - We intentionally commit .vscode in order to share project-specific settings
 
 
+.idea
```
izaaklauer, line 19: I think this is a good idea - will make the lives of goland developers better.

internal/core/app_deploy.go
```diff
@@ -144,11 +144,15 @@ type deployOperation struct {
...
+	// sequence is the monotonically incrementing version number of
+	// the deployment
+	sequence uint64
+
 	// cebToken is the token to set for the deployment to auth
 	cebToken string
 
 	// result is either a component.Deployment or component.DeploymentWithUrl
-	result interface{}
+	result   interface{}
```
izaaklauer, line 155: - Can you run `make format`? That should fix the linting test.

izaaklauer (changes requested):
Nice! This touches the plugin, the SDK, and the core, and your wiring all looks great and straightforward to me. Backwards-compat looks good too.

My only concern with this change is that it introduces some potential for collisions with the deployment name. For example:

- If the waypoint DB is lost/corrupted with no backups and is replaced with a new DB, waypoint could attempt to create deployments with the same sequence number as existing deployments, which would fail (I presume). This should be rare, can be remedied by deleting the existing deployment from the cluster before the collision occurs. It can also be addressed in the future by waypoint auto-discovering and adopting existing cluster resources.

- If multiple waypoint servers are using the same cluster and have the same app name, they would have been able to coexist previously, and won't anymore. I don't think that's a common use-case, or one we want to support - teams within the same company should probably have unique app names.

I don't think those edge-cases outweigh the clear benefit of improved traceability between k8s and waypoint though - just noting for posterity.

A few more things we'd request:

- Can you add the updated SDK version that this depends on to the go.mod? Should be as easy as `go get github.com/hashicorp/waypoint-plugin-sdk@<latest-commit-sha-on-your-branch>`. That should fix another test too.

- Can you add a changelog entry? Guide here. Something like this maybe:
~~~
```release-note:improvement
plugin/k8s: Use sequence number in k8s deployment name for improved tracibliity to waypoint deployments.
```
~~~

Again, nice job on this, thank you!

.gitignore
sl1pm4t: 👍🏽 - I'm looking forward to the nix dev environment blog post from you folks.

internal/core/app_deploy.go
sl1pm4t: Done.

sl1pm4t:
I added the changelog entry and resolved the other comments.
However, for updating the waypoint-plugin-sdk go module, I had to use a `replace`, because my commit doesn't exist in the `hashicorp/waypoint-plugin-sdk` repo (unless you are willing to merge).
```
replace github.com/hashicorp/waypoint-plugin-sdk => github.com/sl1pm4t/waypoint-plugin-sdk v0.0.0-20210914072340-bd8e2aac4058
```
This, however, fails for a new reason I think.

izaaklauer:
Oh that makes sense. I merged the waypoint-plugin-sdk PR, so you should be good to just `go get github.com/hashicorp/waypoint-plugin-sdk` now.

sl1pm4t:
`go.mod` has been updated

sl1pm4t:
Pushed another change. I realised the commit I used when updating SDK module was not the latest (it was the code commit, not the merge commit). 
I also did a `go mod tidy` to remove old entries.

izaaklauer (approved):

PR #2297, opened 2021-09-14
Bump the `ember-build-tests` resource class

gregone:
Fixes: https://github.com/hashicorp/waypoint/issues/2260

starting with `medium+`, we'll see if that works, otherwise we could go to `large` to be sure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -317,6 +317,7 @@ jobs:
   ember-build-tests:
     docker:
       - image: *EMBER_IMAGE
+    resource_class: medium+
     environment:
       JOBS: 2 # limit parallelism for broccoli-babel-transpiler
     steps:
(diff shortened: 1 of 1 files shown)


jgwhite (approved):

jgwhite:
The process still uses all available memory 🤦‍♀️ but this should give us some headroom.

image

gregone:
😭

gregone:
@jgwhite setting `JOBS=1` actually brings the memory usage back down slightly under 5GB 🥲 
Screen Shot 2021-09-14 at 15 15 00

jgwhite:
> @jgwhite setting `JOBS=1` actually brings the memory usage back down slightly under 5GB 🥲
> Screen Shot 2021-09-14 at 15 15 00

It also doesn’t seem increase the build time much. I guess Babel isn’t embarrassingly parallel.

PR #2303, opened 2021-09-14
internal/server/state: peek should return blocked jobs

mitchellh:
We use Peek to check if there is any potential candidate jobs for a
runner, not a guarantee that it will immediately be assigned that job.
In the scenario where a job is blocked (defined below), peek would
return nil which broke the on-demand runner use case that used peek to
verify a job exists for a runner.

A job is "blocked" when another job is already assigned for the same
`(app, project, workspace)` tuple. This serializes operations for that
tuple to prevent concurrency issues but still allow concurrency across
projects/apps/workspaces.

### DIFF
--- a/internal/server/singleprocess/state/job.go
+++ b/internal/server/singleprocess/state/job.go
@@ -322,7 +322,7 @@ RETRY_ASSIGN:
 	runnerRec := newRunnerRecord(r)
 
 	// candidateQuery finds candidate jobs to assign.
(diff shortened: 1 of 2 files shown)


izaaklauer (approved):

krantzinator (approved):
Ahh, I think this addresses the secondary bug we were seeing as part of the UI's frozen state issue! 🙌🏻

mitchellh:
Let's get it in!

PR #2305, opened 2021-09-14
ui: updated flash message service file name to avoid collisions

sabrinako:
### Reasoning for PR
Every time you boot up the ember local dev environment with `ember serve`, there's a warning message about colliding names of `flash-messages` ts/js files.

>  WARNING: Detected collisions between .js and .ts files of the same name. This can result in nondeterministic build output; see https://git.io/JvIwo for more information.
>  - waypoint/services/flash-messages.{js,ts}

This is because we created a service `flash-messages.ts` that builds upon the 3rd party Ember add-on `ember-cli-flash`, which also has a file `flash-messages.js` within it.

This PR stops this warning from occurring without affecting the flash message functionality by simply renaming the custom service, and updating the imports.


ui/app/components/project-input-variables/list.ts
```diff
@@ -14,7 +14,7 @@ interface ProjectSettingsArgs {
 export default class ProjectInputVariablesListComponent extends Component<ProjectSettingsArgs> {
   @service api!: ApiService;
   @service router!: RouterService;
-  @service flashMessages!: PdsFlashMessages;
+  @service flashMessages!: FlashMessagesService;
```
jgwhite, line 17: This is super subtle, but strictly speaking we also need to update these `@service` decorators to tell Ember exactly which service we want to inject.

```suggestion
  @service('pds-flash-messages') flashMessages!: FlashMessagesService;
```

The service decorator doesn't normally need this extra bit of info because it can infer the name of the service from the name of the field it’s decorating (i.e. `flashMessages` → `flash-messages`). Because the name of the service now has `pds-` at the front, `@service` on its own would infer the wrong thing.

jgwhite (changes requested):
Great initiative, I have one super-subtle bit of feedback.

jgwhite (approved):
This PR works well and I think we should merge it but I do want to express one concern:

If we use just `@service` instead of `@service('pdsFlashMessages')` then the app will silently use the wrong service — it’ll use the one from the add-on. This isn’t a problem right now because we don’t actually add anything but it would be a problem in future if we ever did. So: how can we make sure we always use the correct `@service(...)` expression? I _think_ writing a custom linter might be the way to go. To be explored in future horde time.

gregone (approved):

PR #2309, opened 2021-09-15
builtin/k8s: Introduce config to autoscale kubernetes deployments

### LINKED ISSUE #2292: builtin/k8s: Enable autoscaling kubernetes deployments

Currently, when Waypoint creates a Kubernetes deployment, it only creates a single pod per deployment with no autoscaling. Users can still enable autoscaling out of band from Waypoint, but now this would have to be done for each future deployment as well leading to some messy deployment dependencies.

Waypoint should use the horizontal pod autoscaler to autoscale deployments and have a configuration interface for enabling users to define max and min replicas.

briancain:
This PR adds a new config to the Kubernetes plugin called
`autoscale`. When requested, Waypoint will create a Horizontal Pod
Autoscaler to manage the scale of the current deployment based on the
configured metrics.

Fixes #2292 

How to test:

- You need a kubernetes cluster with a `metrics-server` installed. Be mindful of the TLS requirement. Set metrics-server up properly with the required certs, or start it insecurely for dev purposes
- Your deployment stanza needs to specify resource limits and requirements, otherwise the `metrics-server` can't properly do its calculations for the pod for scaling.


builtin/k8s/platform.go
```diff
@@ -736,11 +752,234 @@ func (p *Platform) resourceDeploymentDestroy(
...
+		s.Status(terminal.StatusError)
+		s.Done()
+		return err
+	}
+
+	if len(metricsPods.Items) == 0 {
+		ui.Output("There were no pods recognized in the cluster as a metrics-server "+
+			"with the label '%s'. This means your autoscaler might "+
+			"not be functional until a metrics-server exists and can provide metrics to "+
+			"the horizontal pod autoscaler.",
```
jgwhite, line 810: Is there a guide to installing a metric server we could link to from here? I guess it’d have to be a pretty dependable URL.

briancain: Hmm, I tend to not like using live URLs to link people to something external if it's a domain we don't control... i.e. k8s docs change and we have to put out a release to fix the URL (or old clients get stuck with a broken URL for those who refuse to update).

Maybe we can wordsmith this to be clearer, i.e. give them enough of a nudge to know to look for how to install metrics-server.

jgwhite: Yeah, as I user I know that I’d appreciate being led exactly to where I need to be to take the next step.

briancain: I just pushed an additional sentence here, let me know what you think. This is tricky 😅

evanphx (approved):
LOVE IT.

builtin/k8s/platform.go
briancain: @evanphx also brought up to include some kind of URL, so given there's a couple asks for it I added the github project URL!

jgwhite: Seems great to me!

mitchellh (approved):

PR #2312, opened 2021-09-16
ui: tell vscode to use workspace version of typescript

jgwhite:
## Why the change?

This ensures VS Code is using the project’s own version of Typescript, and will display info that jibes with ember-cli-typescript.

## How do I test it?

Load up the branch in VS Code, open some Typescript files, verify they still behave as you’d expect. Try making some deliberate type errors.


gregone (approved):
Works great! 🚢

PR #2318, opened 2021-09-17
builtin/aws: add create acm certificate

psihachina:
This commit adds the ability to creating certificate, for example

```
app "vole" {
  ...
  ...
  ...
  deploy {
    use "aws-ecs" {
      region = var.aws_region
      memory = var.memory
      cluster = "${var.env}-${var.namespace}"
      subnets = ["subnet-XXXXXXXXXXXXXX", "subnet-XXXXXXXXXXXXXX", "subnet-XXXXXXXXXXXXXX"]
      alb {
        zone_id = var.zone_id
        domain_name = "${app.name}.${var.domain_name}"
        create_acm_certificate = true
      }
    }
  }
}
  ...
  ...
  ...
```

## Tests
[asciicast](https://asciinema.org/a/gadK1JLoUfkTzsTSqxFgeG0pO)


evanphx:
Hi @psihachina, thanks so much for the PR.

The team and I have been talking about your PR and deciding how we'd like to proceed. One thing that we're increasingly worried about is the amount of resources that are created and destroy within the context of a single deployment, even though those resources are very long lived.

This isn't a knock on your work, it looks good and straightforward. We're just thinking that at this time, we want to take a pause on adding more things to this section while we come up with a better way to manage resources like load balancers, certificates, etc.

briancain:
I'm going to go ahead and close this PR. Thank you for taking the time to contribute to Waypoint! Given that we didn't think this was the right approach, we don't want PRs to stay open and become so stale. For what it's worth, the team has been actively working on a way to provision and manage these kinds of resources through Terraform rather than a per-deployment approach. We're hoping to share that feature soon. Thanks!

PR #2319, opened 2021-09-17
k8s: Improve CPU and Memory resource setting ergonomics

evanphx:


briancain (approved):
Looks great! Thanks for the fixups on my branch 👍🏻

PR #2320, opened 2021-09-17
internal/clicontext: default to gRPC port for login addr

mitchellh:
Before we would not set a port at all, and I'm not sure what port it was
trying to connect to. Given our installation method on all platforms
defaults our gRPC to port 9701, it makes sense to default to port 9701
if none is specified.

This makes the initial installation login experience smoother cause most
installations can just do:

    $ waypoint login <IP or hostname>

And it'll just work, rather than looking up the port, which is most
likely 9701.

### DIFF
--- a/internal/clicontext/config.go
+++ b/internal/clicontext/config.go
@@ -62,6 +62,12 @@ func (c *Config) FromURL(v string) error {
 	// Override
 	if u.Host != "" {
 		c.Server.Address = u.Host
+
+		// If no port is specified, default to port 9701 which is our default
+		// gRPC port for Waypoint installations.
+		if idx := strings.IndexByte(u.Host, ':'); idx < 0 {
+			c.Server.Address += ":9701"
+		}
 	}
 
 	// Specifically http will override TLS
--- a/internal/clicontext/config.go
+++ b/internal/clicontext/config.go
@@ -66,7 +66,7 @@ func (c *Config) FromURL(v string) error {
 		// If no port is specified, default to port 9701 which is our default
 		// gRPC port for Waypoint installations.
 		if idx := strings.IndexByte(u.Host, ':'); idx < 0 {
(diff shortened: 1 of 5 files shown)


internal/clicontext/config.go
```diff
@@ -62,6 +62,12 @@ func (c *Config) FromURL(v string) error {
 	// Override
 	if u.Host != "" {
 		c.Server.Address = u.Host
+
+		// If no port is specified, default to port 9701 which is our default
+		// gRPC port for Waypoint installations.
+		if idx := strings.IndexByte(u.Host, ':'); idx < 0 {
+			c.Server.Address += ":9701"
```
briancain, line 69: Should we do something like....

```suggestion
			c.Server.Address += fmt.Sprintf(":%s", serverinstall.defaultGrpcPort)
```

Maybe the source of truth should move outside of serverinstall but it would be nice if we had a place that defined these defaults for things like this to use in case we change it in the future 🤔

mitchellh: Note I can't accept your suggestion because its not exported. We can export it and perhaps use that though.

briancain: Ah yeah, I should of done a code block 😄 I figured it might break too since the suggestion doesn't import `severinstall` too

mitchellh: Actually no, this creates an import cycle so this can't be the source of truth. We use this port ALL OVER so I'd say we just keep it hardcoded and figure it out later.

briancain: What if we place it in `internal/serverconfig/config`? https://github.com/hashicorp/waypoint/blob/main/internal/serverconfig/config.go

mitchellh: Done!

briancain (approved):
Legit! 🎉

PR #2324, opened 2021-09-20
fix other occurence of "negoate"

gregone:
Follow up to https://github.com/hashicorp/waypoint/pull/2323 I missed another typo


briancain (approved):
😆 looks like that's the last one

Issue #1514, opened 2021-05-19
Allow `waypoint install` to take and pass raw flags to `server run`


evanphx:
**Is your feature request related to a problem? Please describe.**
`server run` provides more functionality than `waypoint install` exposes, so there should be a meta-flag on `install` to pass additional flags directly to `server run`.

Relates to #1491

PR #2325, opened 2021-09-20
cli, server: set default log level to debug on server install

krantzinator:
Set default log level to debug rather than trace for the server during an install.
As part of the work for #1514, users will be able to set the log level to trace if desired (for example: `waypoint install -platform=docker -accept-tos -run-flags="-vvv"`)


briancain (approved):
We paired on this so I saw it work in action 👍🏻 It'll be nice to keep our logs to more relevant messages, and it's always good to know users can get the full firehose log experience if they so choose.

PR #2327, opened 2021-09-20
"waypoint k8s bootstrap" hidden command used by the Helm chart

mitchellh:
This introduces a `waypoint k8s bootstrap` command. This command is used by the Helm chart to bootstrap the cluster and set up all the initial secrets. This should never be executed manually by a user so it is hidden.

### DIFF
--- a/internal/cli/k8s_bootstrap.go
+++ b/internal/cli/k8s_bootstrap.go
@@ -147,7 +147,7 @@ func (c *K8SBootstrapCommand) Run(args []string) int {
 	_, err = client.SetServerConfig(ctx, &pb.SetServerConfigRequest{
 		Config: &pb.ServerConfig{
 			AdvertiseAddrs: []*pb.ServerConfig_AdvertiseAddr{
-				&pb.ServerConfig_AdvertiseAddr{
+				{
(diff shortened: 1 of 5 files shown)


internal/cli/k8s_bootstrap.go
```diff
@@ -0,0 +1,366 @@
...
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:   "runner-token-secret",
+			Target: &c.flagRunnerTokenSecret,
+			Usage:  "The name of the Kubernetes secret to write the runner token to.",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:   "odr-image",
```
catsby, line 267: Nit: should we prefix these with `k8s-` to be consistent with the normal [K8s installation flags](https://www.waypointproject.io/commands/install#kubernetes-options)? Maybe we aren't concerned so much because this is an internal thing, so I'm fine with either way.

catsby: Makes sense thanks 👍

briancain (approved):
Looks good 👍🏻

Issue #1238, opened 2021-03-26
Waypoint Up from browser UI


gregone:
**Is your feature request related to a problem? Please describe.**
In our quest to bring Waypoint CLI functionality to the browser UI, we should be adding the ability to trigger the `up` command from the the CLI

**Describe the solution you'd like**
The application view should have a button to allow the user to run `waypoint up` from their browser

jgwhite:
This work was partially blocked by #1684 but is now unblocked thanks to #1685

PR #1500, opened 2021-05-18
ui: add “waypoint up” button where applicable

jgwhite:
Addresses #1238

PR #2331, opened 2021-09-21
waypoint up in the browser

gregone:
Fixes: #1238 

This PR basically reopens https://github.com/hashicorp/waypoint/pull/1500 with a few modifications.Because of some technical constraints, the "Waypoint Up" button will not be available to everyone. 
Users will only see it if: 
- they have a Git `DataSource` set up
- `DataSourcePoll` is not enabled (otherwise "Up" automatically runs)
- there are no existing releases

Example (fils to deploy for other reasons, but you get the idea): 

https://user-images.githubusercontent.com/1416421/134381988-41e22f32-e19e-438d-9698-4459b5955868.mov


ui/app/routes/workspace/projects/project/app.ts
```diff
@@ -56,6 +56,7 @@ export default class App extends Route {
     appRef.setProject(proj.name);
 
     return hash({
+      project: proj,
```
jgwhite, line 59: Can we add this field to the `Model` interface defined at the top of the file?

jgwhite (changes requested):
This is looking great 👍 I’m just testing it locally to be double-sure.

Aside from the one inline comment, could I also suggest we rebase and squash this?

ui/app/routes/workspace/projects/project/app.ts
gregone: done

gregone: I'll use the `squash and merge` button on GH

jgwhite (approved):

PR #2334, opened 2021-09-21
waypoint login -from-kubernetes

mitchellh:
This introduces `waypoint login -from-kubernetes` that performs automatic authentication after a Helm-based install. This looks up the Waypoint service, gets the token from the Kubernetes secret, and completes the login.

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -510,7 +510,7 @@ for one of two reasons. First, the Waypoint server may not be bootstrapped.
 After installing Waypoint on Kubernetes, it takes a few minutes for Waypoint
 to bootstrap itself.
 
-If Waypoint is already bootstrapped, its possible the server administrator
+If Waypoint is already bootstrapped, it's possible the server administrator
 already deleted the secret. Future users should not use this authentication
 method and should instead ask another Waypoint user to generate an invite token
 for them.
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -38,6 +38,7 @@ type LoginCommand struct {
 
 func (c *LoginCommand) Run(args []string) int {
 	ctx := c.Ctx
+	log := c.Log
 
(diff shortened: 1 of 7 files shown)


internal/cli/login.go
```diff
@@ -319,6 +502,26 @@ waypoint login https://example.com
...
 double-check the token and try again.
+`
+
+	errK8STokenEmpty = `
+The Waypoint token in the Kubernetes secret is empty. This is usually
+for one of two reasons. First, the Waypoint server may not be bootstrapped.
+After installing Waypoint on Kubernetes, it takes a few minutes for Waypoint
+to bootstrap itself.
+
+If Waypoint is already bootstrapped, its possible the server administrator
```
briancain, line 513:
```suggestion
If Waypoint is already bootstrapped, it's possible the server administrator
```

internal/cli/login.go
```diff
@@ -239,6 +316,81 @@ func (c *LoginCommand) loginOIDC() (string, int) {
 	return respToken.Token, 0
 }
 
+func (c *LoginCommand) loginK8S(ctx context.Context) (string, int) {
```
briancain, line 319: Adding this makes me think we should have some kind of Interface for our authentication functions here for a more rigid way to define these auth funcs, thoughts? I could see us adding more in the future.

briancain (approved):
🔒 🎉  Make sense! 👍🏻 Had a minor comment that we could maybe address here, or maybe leave as a TODO here for the future if we add more authentication functions to our login.

internal/cli/login.go
mitchellh: Its already typed to `func(context.Context) (string,int)` and in the switch above, so it sort of already is? I don't think we need a separate full Go interface cause they can't stand on their own (they require the login command metadata).

briancain: Cool cool, makes sense!

PR #2335, opened 2021-09-21
Log warnings if some things we expect to work actually don't

catsby:
@evanphx this came up in discussion a few weeks back; I wasn't sure what else we should do here so we at least log a warning. Is there any recovery available here?

### DIFF
--- a/internal/core/app_build.go
+++ b/internal/core/app_build.go
@@ -171,7 +171,8 @@ func (op *buildOperation) Do(ctx context.Context, log hclog.Logger, app *App, _ 
 					log.Warn("unexpected response type from callDynamicFunc")
 				}
 			} else {
-				log.Warn("error calling dynamic func", "error", err)
+				log.Error("error calling dynamic func", "error", err)
+				return err
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


internal/core/app_build.go
```diff
@@ -167,7 +167,11 @@ func (op *buildOperation) Do(ctx context.Context, log hclog.Logger, app *App, _
 					// ... which we make available to build plugin.
 					args = append(args, plugin.ArgNamedAny("access_info", any))
 					log.Debug("injected access info")
+				} else {
+					log.Warn("unexpected response type from callDynamicFunc")
 				}
+			} else {
+				log.Warn("error calling dynamic func", "error", err)
```
evanphx, line 174:
```suggestion
				log.Error("error calling dynamic func", "error", err)
				return err
```

internal/core/app_build.go
```diff
@@ -167,7 +167,11 @@ func (op *buildOperation) Do(ctx context.Context, log hclog.Logger, app *App, _
 					// ... which we make available to build plugin.
 					args = append(args, plugin.ArgNamedAny("access_info", any))
 					log.Debug("injected access info")
+				} else {
+					log.Warn("unexpected response type from callDynamicFunc")
```
evanphx, line 171:
```suggestion
					log.Error("unexpected response type from callDynamicFunc", "type", hclog.Fmt("%T", raw))
					return nil, errors.New("AccessInfoFunc didn't provide a typed any")
```

evanphx (approved):

internal/core/app_build.go
catsby: applied with minor edit in f8e5c606

PR #2336, opened 2021-09-21
Helm Deployment Plugin

mitchellh:
This introduces a Helm deployment plugin. This lets Waypoint users use Helm charts to deploy their applications. Both local paths to charts and charts published in repositories work.

Here is an example:

```hcl
deploy {
  use "helm" {
    name  = "myrelease"
    chart = "${path.app}/chart"

    set {
      name  = "deployment.image"
      value = artifact.name
    }
  }
}
```

And an example of `waypoint status` which shows all the resources the Helm chart created:

CleanShot 2021-09-22 at 10 17 20@2x

## Impl Notes

This uses Helm programmatically via a Go library and does not require `helm` the CLI to be installed. Authentication for Helm is done the same way as the Kubernetes providers with both in-cluster and out-of-cluster auth supported using well known environment variables and file paths. Net: if `kubectl` works, this works.

Helm has _a lot of options_. I tried to pick reasonable defaults (based mostly on Terraform, some divergences). As users run into different challenges we can introduce new config options but for now I didn't want to overwhelm people and this keeps in line with Waypoint being opinionated out of the box.


evanphx (approved):
Looks good! One question, do/should we provide real status for the resources that we can do status for?

mitchellh:
> Looks good! One question, do/should we provide real status for the resources that we can do status for?

Yes! We don't in this PR mainly because I didn't want to bloat it, but what I did was create a package `builtin/k8s/internal/k8sstatus` which will let us centralize all our logic for building resource status reports. For now, all resources are "unknown" so you get the minimal in the screenshot above. Following this merge, I'm going to add deployments, services, etc. so we can show REAL status per resource. Unknown resources (CRDs and stuff) will continue to default to the behavior in this PR.

evanphx:
Sounds good!

Issue #2339, opened 2021-09-22
ui: Renaming a config variable does not remove the original


jgwhite:
**Describe the bug**
When I rename a config variable then I get two versions of the variable, one with the old name and one with the new:

https://user-images.githubusercontent.com/34030/134333304-3c9b5247-a196-433b-abd1-3d7bb30a37e6.mp4

**Steps to Reproduce**

1. Rename a config variable in the UI (Edit -> Change name -> Save)
2. Observe the duplicate with the original name

**Expected behavior**
Only the config variable with the new name should exist.

**Waypoint Platform Versions**

* Waypoint CLI Version: 55f28f1e83b04ccea7463f385d3ba44aa671aa34

gregone:
I'll look into the issue later in this cycle, but with the ability to compare the `initialVariable` we should be able to create the renamed var, then unset the old one when that case happens.

jgwhite:
@gregone alternatively we could have the variable name be read-only (this is what Heroku’s config vars UI does).

PR #2340, opened 2021-09-22
ui/mirage: improve `SetConfig` simulation

jgwhite:
## Why the change?

The previous implementation was pretty wonky and would create duplicates every time any property of a variable was edited.

This version tries to be faithful to core and, as such, exhibits the behavior described in #2339.

## How do I test this?

1. Check out the branch
2. Boot the UI dev server in Mirage mode (`yarn start`)
3. Visit [http://localhost:4200](http://localhost:4200)
4. Try adding, editing, removing config variables
5. Verify it works just like prod


gregone (approved):
Looks good. left some notes on the other issue

PR #2346, opened 2021-09-23
Server automatically reloads TLS cert files on change

mitchellh:
This makes it so that the Waypoint server will automatically rotate the TLS certificate when it detects a file change on disk. This happens with no disruption to any network connections. 

This is particularly useful in a Kubernetes environment where the TLS certs may be coming from a mounted secret. When the secret changes, Kubernetes will update the files. In this scenario, the Waypoint server will begin using the new TLS certificate as soon as it can.

I brought in an `internal/pkg/cert` package that I've been copying around for years in various projects. 😛


krantzinator (approved):
copy/paste seems fine 😄

PR #2347, opened 2021-09-24
Waypoint server: Support insecure HTTP listener 

mitchellh:
This adds a `-listen-http-insecure` flag to the Waypoint server to create an insecure (non-TLS) HTTP listener. 

This insecure HTTP listener behaves in two ways:

1. If there is an `X-Forwarded-Proto` of `https`, it routes the traffic.
2. Otherwise, it redirects to https.

The insecure HTTP listener is primarily to help with the Kubernetes use case where resources such as `Ingress` do not typically talk to their backends with TLS. Further, it helps the basic `LoadBalancer Service` case because port 80 will redirect to TLS versus not connecting today. 

This is opt-in functionality. If the flag isn't specified, an insecure listener is not created. 

There is no way to get this today with `waypoint install` (unless you specify a manual flag). The plan is to integrate this into the Waypoint Helm chart directly.

### DIFF
--- a/.changelog/2347.txt
+++ b/.changelog/2347.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+cli: `server run` can now create a non-TLS HTTP listener. This listener
+redirects to HTTPS unless X-Forwarded-Proto is https.
+```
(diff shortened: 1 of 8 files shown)


krantzinator (approved):
Seems reasonably protected

PR #2351, opened 2021-09-24
internal/server: config resolution needs to account for workspaces

mitchellh:
The config resolution rules omitted workspace targeting. This ended up
in behavior that worked as "intended" in this broken case: sometimes
you'd get workspace-scoped prioritizing over non-workspace-scoped and
vice versa. It was undefined!

This clarifies the resolution to note that a variable with a workspace
set takes higher priority over one without a variable set, and encodes
this rule into the sort logic. This also adds tests for this.


mitchellh:
@evanphx When this is merged, you just have to replace the server process. Previous deployments should magically get the proper config on entrypoint reconnect.

mitchellh:
Evan approved in Slack haha

PR #2352, opened 2021-09-24
ui: Add Docker image/tag information and Overview section

### LINKED ISSUE #1520: Display Docker Image Version

Currently, Waypoint considers a "version" to be an instance of a deployment.
Whilst this is great for many use cases, it is not fascinating as it is basically removing information such as:
- which commit is this build based on?
- which Docker tag is being used?
- what was the last change?

Examples of the current UI:

Screenshot 2021-05-20 at 12 42 35

Screenshot 2021-05-20 at 12 44 34

Examples of the current CLI:

```
$ waypoint deployment list -workspace-all
    | ID |   PLATFORM   |  DETAILS   |   STARTED    |  COMPLETED
----+----+--------------+------------+--------------+---------------
  ✔ | 14 | cloudfoundry | artifact:6 | 18 hours ago | 18 hours ago
  ✖ | 13 | cloudfoundry | artifact:5 | 18 hours ago | 18 hours ago
```

The Waypoint UI (and the CLI) are not showing any relevant information to the user about what has been deployed. A vague ID / version is provided instead.

We should be able to see exactly which artifact has been deployed / released in a certain workspace and maybe add relevant information such as "which commit did generate this change".

Without this improvement, it is really hard to tell with a quick glance, which Docker Image (repository + tag) was used or which commit caused the build to be triggered.

N/A

- Seeing the version (as in software version, not Waypoint version) of the deployment that is currently deployed / released
- Getting with a quick glance the reason why a certain deployment / release was performed (`git log --format="%h: %s" | head -n1`)
- Reverting to a specific version by performing a GitOps revert commit directly from waypoint (write access to the GitOps repo)

**Additional context**

Our use case is mainly focused on Docker images and their tags. We would like to know, at any time, which version are we running in dev, staging or prod.

sabrinako:
#1520
## Original UI
Screen Shot 2021-09-30 at 5 24 20 PM

## Updated UI
Screen Shot 2021-09-30 at 5 24 02 PM

## How to Test
1. Pull this branch down `ui/docker-version`
2. Run a local Waypoint server ([see here](https://github.com/hashicorp/waypoint/tree/main/ui))
3. Create a project
4. `waypoint up`
5. Check the deployment details page


jgwhite:
I tested this with a kubernetes deploy config and got the following:

CleanShot 2021-10-01 at 12 06 55@2x

Does the Kubernetes plugin still need updated to thread this info through?

jgwhite:
What this looks like when the status report is still pending:

CleanShot 2021-10-01 at 12 07 46@2x

ui/app/components/container-image-tag/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    return;
+  }
+
+  parseImageAndTag(): void {
+    if (!this.args.statusReport || !this.args.statusReport.resourcesList) {
+      return;
+    }
+
+    let container = this.args.statusReport.resourcesList.find((r) => r.type === 'container');
```
jgwhite, line 50: I’m realizing that this method of digging up the image reference isn’t general enough. On the resource detail page (which I wrote) it only works for Kubernetes pods, and here it only works for Docker containers. We should collaborate a general approach to this.

jgwhite: I tried traversing everything in `statusReport.resourcesList` in search of `image` fields, this does get us what we want, but also some stuff we don’t want.

image

Here’s roughly what that approach would look like: https://github.com/hashicorp/waypoint/commit/cfbef23817ec7e6eabd9e30c9c8ec6ac5f93da98

ui/app/components/meta-table/index.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="artifact-overview">
+  <table>
+    <tbody>
+      {{#if @statusReport}}
```
jgwhite, line 4: Given that this component is representing data from a status report, what do you think about calling it something more explicit like `<StatusReportMetaTable>`?

ui/app/components/meta-table/index.hbs
```diff
@@ -0,0 +1,29 @@
...
+        <tr>
+          <td>
```
jgwhite, line 6: You can use a [row-scoped table header](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/th#attr-scope) here, if you like.

```suggestion
          <th scope="row">
```

ui/app/components/meta-table/index.ts
```diff
@@ -0,0 +1,92 @@
...
+interface OverviewTableArgs {
+  model: Deployment.AsObject & WithStatusReport;
+  artifactType: string;
+}
+
+interface WithStatusReport {
+  statusReport?: StatusReport.AsObject;
+}
+
+export default class OverviewTable extends Component<OverviewTableArgs> {
```
jgwhite, line 26: If you’d rather not keep this class name in sync with the filename, you can export an anonymous class.

```suggestion
export default class extends Component<OverviewTableArgs> {
```

ui/app/components/meta-table/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  @service api!: ApiService;
+  @service('pdsFlashMessages') flashMessages!: FlashMessagesService;
+  @tracked isRefreshRunning = false;
+
+  constructor(owner: unknown, args: OverviewTableArgs) {
+    super(owner, args);
+  }
```
jgwhite, line 33: I’d suggest deleting this, since it’s not doing any additional work.

```suggestion
```

jgwhite:
By the way: I love how you’ve structured the “How to test” instructions. Going to steal that format going forward. Hope you don’t mind!

ui/app/components/meta-table/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  Deployment,
+  Ref,
+  ExpediteStatusReportRequest,
+  StatusReport,
+  GetJobStreamRequest,
+  GetJobStreamResponse,
+  Job,
+} from 'waypoint-pb';
+
+interface OverviewTableArgs {
```
jgwhite, line 17: I quite like to call these just `Args`, that way I don’t need to rename the type if I rename the component.

internal/server/gen/server.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.26.0
-// 	protoc        v3.15.8
+// 	protoc-gen-go v1.27.1
+// 	protoc        v3.17.3
```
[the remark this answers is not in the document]
sabrinako: Whoops, no I didn't, I'll fix this!

ui/app/components/container-image-tag/index.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+    } else {
+      findImageRefs(value, result);
+    }
+  }
+
+  return result;
+}
+
+class ImageRef {
```
gregone, line 46: For consistency with the structure of other files, could we move the `ImageRef` class definition above the Component class?

gregone (approved):
LGTM, just one minor comment

ui/app/components/container-image-tag/index.ts
```diff
@@ -0,0 +1,64 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import ApiService from 'waypoint/services/api';
+import { StatusReport } from 'waypoint-pb';
+
+interface Args {
+  statusReport: StatusReport.AsObject;
+}
+
+export default class DockerImageBadge extends Component<Args> {
```
jgwhite, line 10: Shall we make the class name match the filename?

```suggestion
export default class ContainerImageTag extends Component<Args> {
```

Or we could drop the class name altogether like:

```suggestion
export default class extends Component<Args> {
```

ui/tests/integration/components/status-report-meta-table-test.ts
```diff
@@ -0,0 +1,49 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { getUnixTime, subDays } from 'date-fns';
+import hbs from 'htmlbars-inline-precompile';
+
+module('Integration | Component | meta-table', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
```
jgwhite, line 10: I’d be tempted to split this test in two: one for the happy path and one for the empty state.

ui/app/components/status-report-meta-table/index.hbs
```diff
@@ -0,0 +1,29 @@
+<div class="artifact-overview">
```
jgwhite, line 1: There are a few localizable strings in this file. Could we add them to `en-us.yaml`?

jgwhite (changes requested):
Looking good. Just a few minor suggestions.

jgwhite (approved):
🥳

PR #2353, opened 2021-09-24
Add `context set` subcommand

catsby:
This PR adds a subcommand `set` to the `context` command, allowing users to set a property of the current context. The only supported property at this time is `-workspace`. This enables users to set the workspace once, and avoid having to repeatedly set the `-workspace` with every CLI invocation. 

Additionally this all introduces support for reading the desired workspace from the environment, using the environment variable `WP_WORKSPACE`. I considered `WAYPOINT_WORKSPACE` and can go either way.

The CLI flag `-workspace` is still supported however it itself no longer has a default value; the value is now derived from either the flag, the environment, the stored value in the context storage, or the default value of `"default"`, in that order, with the first non-empty value being used.

TODO: 
- [x] website content

### DIFF
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -25,7 +25,10 @@ import (
 	"github.com/hashicorp/waypoint/internal/server/grpcmetadata"
 )
 
-const defaultWorkspace = "default"
+const (
+	defaultWorkspace        = "default"
+	defaultWorkspaceEnvName = "WP_WORKSPACE"
+)
 
 // baseCommand is embedded in all commands to provide common logic and data.
 //
@@ -209,25 +212,14 @@ func (c *baseCommand) Init(opts ...Option) error {
 	}
 	c.contextStorage = contextStorage
 
-	/// load workspace
-
-	// compute the workspace based on available values, in this order of
-	// precedence:
-	// - value stored in the CLI context
-	// - vaule from the environment variable WP_WORKSPACE
-	// - value set in the CLI flag -workspace
-	//
-	// The default value is "default"
-	// TODO (clint): re-work as a switch
+	// load workspace from cli/env/storage
 	workspace, err := c.workspace()
 	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 		return err
 	}
 
-	// q.Q("+>=> final workspace value: ", workspace)
 	c.refWorkspace = &pb.Ref_Workspace{Workspace: workspace}
-
-	/// end load workpace
 
 	// Parse the configuration
 	c.cfg = &config.Config{}
@@ -480,8 +472,7 @@ func (c *baseCommand) flagSet(bit flagSetBit, f func(*flag.Sets)) *flag.Sets {
 		f.StringVar(&flag.StringVar{
 			Name:   "workspace",
 			Target: &c.flagWorkspace,
-			// Default: "default",
-			Usage: "Workspace to operate in.",
+			Usage:  "Workspace to operate in.",
 		})
 	}
 
@@ -628,10 +619,17 @@ func checkFlagsAfterArgs(args []string, set *flag.Sets) error {
 	return nil
 }
 
+// workspace computes the workspace based on available values, in this order of
+// precedence (last value wins):
 //
+// - value stored in the CLI context
+// - vaule from the environment variable WP_WORKSPACE
+// - value set in the CLI flag -workspace
+//
+// The default value is "default"
 func (c *baseCommand) workspace() (string, error) {
 	// load env for workspace
-	workspaceENV := os.Getenv("WP_WORKSPACE")
+	workspaceENV := os.Getenv(defaultWorkspaceEnvName)
 	switch {
 	case c.flagWorkspace != "":
 		return c.flagWorkspace, nil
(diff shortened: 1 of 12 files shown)


internal/cli/base.go
```diff
@@ -24,6 +25,11 @@ import (
 	"github.com/hashicorp/waypoint/internal/server/grpcmetadata"
 )
 
+const (
+	defaultWorkspace        = "default"
+	defaultWorkspaceEnvName = "WP_WORKSPACE"
```
krantzinator, line 30: This should probably be `WAYPOINT_WORKSPACE` to fit the pattern of our other env vars.
```suggestion
	defaultWorkspaceEnvName = "WAYPOINT_WORKSPACE"
```

internal/cli/base.go
```diff
@@ -608,6 +619,47 @@ func checkFlagsAfterArgs(args []string, set *flag.Sets) error {
 	return nil
 }
 
+// workspace computes the workspace based on available values, in this order of
+// precedence (last value wins):
+//
+// - value stored in the CLI context
+// - vaule from the environment variable WP_WORKSPACE
```
krantzinator, line 626:
```suggestion
// - value from the environment variable WAYPOINT_WORKSPACE
```

internal/cli/base.go
```diff
@@ -608,6 +619,47 @@ func checkFlagsAfterArgs(args []string, set *flag.Sets) error {
...
+		return workspaceENV, nil
+	default:
+		// attempt to load from CLI context storage
+		defaultName, err := c.contextStorage.Default()
+		if err != nil {
+			return "", err
+		}
+
+		// If we have no context name, then we just return the default
+		if defaultName != "" && defaultName != "-" {
```
krantzinator, line 646: When would the context workspace get set to `-`?

internal/cli/base.go
```diff
@@ -608,6 +619,47 @@ func checkFlagsAfterArgs(args []string, set *flag.Sets) error {
...
+				return "", err
+			}
+			if cfg.Workspace != "" {
+				return cfg.Workspace, nil
+			}
+		}
+	}
+
+	// default value
+	return defaultWorkspace, nil
```
krantzinator, line 660: I think this is a bit easier to read if the default return value is inside of the `default` case.

internal/cli/base_test.go
```diff
@@ -91,3 +93,111 @@ func TestCheckFlagsAfterArgs(t *testing.T) {
...
+func TestWorkspacePrecedence(t *testing.T) {
+	cases := []struct {
+		Name             string
+		Args             []string
+		Env              string
+		ContextWorkspace string
+		Expected         string
+	}{
+		{
+			"default no inpupts",
```
krantzinator, line 106:
```suggestion
			"default no inputs",
```

internal/cli/context_create.go
```diff
@@ -37,7 +37,7 @@ func (c *ContextCreateCommand) Run(args []string) int {
 
 	name := args[0]
 
-	// Get our contexts
+	// Set the context
```
krantzinator, line 40: Nice clarification

internal/cli/status.go
```diff
@@ -1179,6 +1179,8 @@ func (c *StatusCommand) getWorkspaceFromProject(pr *pb.GetProjectResponse) (stri
 	var workspace string
 
 	if len(pr.Workspaces) != 0 {
+		// TODO (clint): verify this is fine with the removal of default
```
krantzinator, line 1182: Did you do this? 😄

krantzinator (commented):
A few nit changes; take them or leave them!

I did have a question though -- the `workspace` in the actual config file doesn't get set until the user runs `context set -workspace="foo"`. Up until then, it is an empty string. Running `context inspect [context-name]` _does_ print out `default` for the Workspace Name in this case, but the actual context file still has an empty string.
Is this intentional? It doesn't matter code-wise, but it's an (albeit very small) inconsistency for UX.

internal/cli/base.go
catsby: It allows operations that do not expect a context (https://github.com/hashicorp/waypoint/pull/48), so in this situation we don't want to attempt to load the context.

internal/cli/base.go
catsby: 👍 update in 7886733d

internal/cli/status.go
catsby: I have now! Thank you for the reminder. I had to make some adjustments in dbf8ed63 please take a look when you can :D

internal/cli/base.go
krantzinator: TIL!

krantzinator (approved):
I still have [this question](https://github.com/hashicorp/waypoint/pull/2353#pullrequestreview-764342387) but this LGTM.

catsby:
@krantzinator apologies I forgot to circle back and reply :D 

It is intentional, yes. I felt that populating the context file with `"workspace":"default"` would be akin to an upgrade path of sorts, and wasn't sure when I should slip that operation in. As you mentioned, the net effect in the code is the same, so I didn't feel it worth it to do so. I don't anticipate many people directly inspecting the context file itself (as opposed to using `context inspect`) as well so it didn't seem worth the effort for code that would probably have to run every time to make sure everyone is updated, but only actually do something just the first time. That was my thinking at least. Is there some UX I'm missing, other than a difference if someone inspects the context file directly?

krantzinator:
> Is there some UX I'm missing, other than a difference if someone inspects the context file directly?

Nope that was it! Thanks for expanding on that.

internal/cli/context_set.go
```diff
@@ -0,0 +1,99 @@
...
+	// set new workspace
+	cfg.Workspace = c.flagWorkspace
+
+	// store updated context
+	if err := c.contextStorage.Set(contextName, cfg); err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	c.ui.Output("Context %q updated to use %s workspace.", contextName, cfg.Workspace, terminal.WithSuccessStyle())
```
briancain, line 65: I think it's worth also including Server Address here, since most of the time context name will be pretty generic and hard to differentiate between servers (we do this on `waypoint status too).

briancain (approved):
Dope @catsby 😄 ! I had a minor comment to address but I will approve. Looks good! ✨

internal/cli/context_set.go
catsby: Good idea, I changed it in f0c2f42d to look like so:

```
$ waypoint context set -workspace dev

Context "install-1633109546" (localhost:9701) updated to use dev workspace.
```

briancain: Looks good!

Issue #2198, opened 2021-09-02
Automatically configure the alb to redirect HTTP to HTTPS


psihachina:
**Is your feature request related to a problem? Please describe.**
When I deployment the ECS app in Waypoint with a certificate does not redirect from HTTP to HTTPS.

**Describe the solution you'd like**
If a cert has been configured waypoint to automatically configure the alb to redirect HTTP to HTTPS

**Describe alternatives you've considered**

**Explain any additional use-cases**

**Additional context**

izaaklauer:
This is a great idea! If you're interested in implementing this, feel free to assign yourself to it (or let us know so we can assign you).

PR #2354, opened 2021-09-27
builtin/aws: add redirect listener

psihachina:
This commit adds a redirect listener to the HTTPS listener that is created when you configure the certificate, for example

```
app "vole" {
  ...
  ...
  ...
  deploy {
    use "aws-ecs" {
      region = var.aws_region
      memory = var.memory
      cluster = "${var.env}-${var.namespace}"
      alb {
        zone_id = var.zone_id
        domain_name = "${app.name}.${var.domain_name}"
        certificate = var.acm_certificate
      }
    }
  }
}
  ...
  ...
  ...
```

### Test
[asciicast](https://asciinema.org/a/9HTTXQJuMaU1reLZnVOKiC6Iz)


psihachina:
https://github.com/hashicorp/waypoint/issues/2198

AutomationD:
Any update on this? It would be quite useful. 🙏🏻

builtin/aws/ecs/platform.go
```diff
@@ -910,6 +918,78 @@ func (p *Platform) resourceServiceDestroy(
...
+
+	s := sg.Add("Initiating ALB Redirect Listener creation")
+	defer s.Abort()
+
+	var HTTPListener *elbv2.Listener
+
+	if externalIngressPort == 443 {
+		elbsrv := elbv2.New(sess)
+
+		listeners, err := elbsrv.DescribeListenersWithContext(ctx, &elbv2.DescribeListenersInput{
```
izaaklauer, line 940: This means that we'll make two calls `DescribeListenersWithContext` in many cases - once when we create the normal listener, and once here. There isn't a great way to share this information between these two resources at this moment. Instead of creating a new resource for the redirect listener, can you add this logic to the existing listener resource?

builtin/aws/ecs/platform.go
```diff
@@ -910,6 +918,78 @@ func (p *Platform) resourceServiceDestroy(
...
+
+		for _, listener := range listeners.Listeners {
+			if *listener.Protocol == "HTTP" {
+				HTTPListener = listener
+			}
+		}
+
+		if HTTPListener != nil && *HTTPListener.DefaultActions[0].Type == "forward" {
+			s.Update("HTTP listener use for forwarding. Please destroy previus deployment or try again deploy app for create redirect listener")
+		} else {
```
izaaklauer, line 956: Currently, it looks like if we already created a redirect listener, we'll try to create it again. I don't think we should ever try creating a listener if it already exists. 

We should also check to see if the settings of the existing listener match what we would create, and give some nice output in that case, like "Using existing redirect ALB listener".

builtin/aws/ecs/platform.go
```diff
@@ -910,6 +918,78 @@ func (p *Platform) resourceServiceDestroy(
...
+			s.Update("Creating redirect listener")
+
+			out, err := elbsrv.CreateListenerWithContext(ctx, &elbv2.CreateListenerInput{
+				LoadBalancerArn: &alb.Arn,
+				Port:            aws.Int64(int64(80)),
+				Protocol:        aws.String("HTTP"),
+				DefaultActions: []*elbv2.Action{
+					{
+						RedirectConfig: &elbv2.RedirectActionConfig{
+							Host:       &p.config.ALB.FQDN,
```
izaaklauer, line 966: This can be nil - ALB config isn't required. Do we need a "Host" set for the redirect listener to work?

builtin/aws/ecs/platform.go
```diff
@@ -910,6 +918,78 @@ func (p *Platform) resourceServiceDestroy(
...
+						Type: aws.String("redirect"),
+					},
+				},
+			})
+
+			if err != nil {
+				return status.Errorf(codes.Internal, "failed to create redirect listener: %s", err)
+			}
+
+			s.Update("Redirect listener is created")
```
izaaklauer, line 981: Let's give some more verbose output: something like: "Created an ALB listener redirecting from port 80 to 443"

AutomationD:
I'd like to bump this issue, @izaaklauer @psihachina, is there anything else required to get this merged please? Thank you!

evanphx:
Hi @AutomationD,

Yup! There is feedback left that needs to be addressed. If you'd like to pick up the changes, feel free! We're not yet ready to jump in and handle it ourselves.

evanphx:
Hi all, due to inactivity we're going to close this PR. If you do get back to it, we can reopen it.

PR #2355, opened 2021-09-27
ui: dedupe ember-cli-babel

jgwhite:
## Why the change?

In the course of making other changes, I noticed we had a rather outrageous number of different versions of ember-cli-babel.

So, I followed [these instructions](https://docs.ember-cli-typescript.com/upgrade-notes#update-ember-cli-babel) to try and improve the situation.

Technically this dedupes *everything* but ember-cli-babel is a major cause of duplication.

## What does it look like?

image

Running `yarn why ember-cli-babel | grep -C 0 'ember-cli-babel@'`…

<details>
<summary>Before</summary>

```
❯ yarn why ember-cli-babel | grep -C 0 'ember-cli-babel@' 
=> Found "ember-cli-babel@7.23.0"
=> Found "ember-cli-favicon#ember-cli-babel@7.7.3"
=> Found "ember-maybe-import-regenerator#ember-cli-babel@6.18.0"
=> Found "ember-cli-page-object#ember-cli-babel@6.18.0"
=> Found "ember-inflector#ember-cli-babel@6.18.0"
=> Found "ember-truth-helpers#ember-cli-babel@6.18.0"
=> Found "ember-cli-typescript-blueprints#ember-cli-babel@7.21.0"
=> Found "ember-auto-import#ember-cli-babel@7.21.0"
=> Found "ember-cli-cjs-transform#ember-cli-babel@7.21.0"
=> Found "@ember-decorators/component#ember-cli-babel@7.22.1"
=> Found "@ember/render-modifiers#ember-cli-babel@7.22.1"
=> Found "ember-classic-decorator#ember-cli-babel@7.22.1"
=> Found "ember-composable-helpers#ember-cli-babel@7.22.1"
=> Found "ember-load-initializers#ember-cli-babel@7.23.1"
=> Found "ember-cli-flash#ember-cli-babel@7.26.1"
=> Found "ember-set-helper#ember-cli-babel@7.26.1"
=> Found "ember-concurrency-async#ember-cli-babel@7.22.1"
=> Found "ember-concurrency-decorators#ember-cli-babel@7.22.1"
=> Found "ember-concurrency-ts#ember-cli-babel@7.22.1"
=> Found "ember-named-blocks-polyfill#ember-cli-babel@7.22.1"
=> Found "ember-cli-clipboard#ember-cli-babel@7.22.1"
=> Found "ember-page-title#ember-cli-babel@7.23.1"
=> Found "ember-qunit#ember-cli-babel@7.23.1"
[...]

PR #2356, opened 2021-09-27
UI: Application & Operation logs using XTerm.js terminal

gregone:
Screen Shot 2021-09-24 at 16 12 33

dark mode: 

https://user-images.githubusercontent.com/1416421/134953199-77a64bf2-c43a-464e-b3aa-70c468818032.mov


gregone:
Looks like there's some deps issues following my last rebase after https://github.com/hashicorp/waypoint/pull/2355 I'lll investigate tomorrow.

ui/app/utils/terminal.ts
```diff
@@ -0,0 +1,33 @@
+import { ITerminalOptions, Terminal } from 'xterm';
```
sabrinako, line 1: Still in the process of reviewing this PR, but could we change the file name of this to `createTerminal`? I'm guessing there's a chance of a namespace collision with this util and the component

gregone: good idea! i'll change that

gregone: done in https://github.com/hashicorp/waypoint/pull/2356/commits/e405966f2cc83aaed577e9fdc27b557650cde444 !

ui/tests/integration/components/terminal-test.ts
```diff
@@ -0,0 +1,28 @@
...
+    await render(hbs`<Terminal @terminal={{this.terminal}}/>`);
+    // write line and see if it renders
+    terminal.writeln('Welcome to Waypoint!');
+    // We have to use the runloop as writeln isn't async
+    // Note that even the xterm.js rendering tests use polling to evaluate rendering
+    later(() => {
+      assert.equal(terminal?.buffer?.active?.getLine(0)?.translateToString(true), 'Welcome to Waypoint!');
+    }, 300);
+    // Clean up terminal
+    terminal.dispose();
```
jgwhite, line 26: Rather than calling `dispose` in the test, I think we could call `clearRender` and assert that dispose was called correctly by the component.

```suggestion
    // Clean up terminal
    await clearRender();
    assert.equal(terminal.element, undefined);
```

ui/app/components/log-stream/index.hbs
```diff
@@ -1,34 +1 @@
...
-  {{else}}
-    <EmptyState>
-      <p>There are no logs to display for this app yet</p>
-      <p>Logs will be displayed when you run actions in the CLI such as <code class="inline">build</code>,
-        <code class="inline">deploy</code> or <code class="inline">release</code></p>
-    </EmptyState>
-  {{/if}}
-</div>
\ No newline at end of file
+<Terminal @terminal={{this.terminal}}/>
```
jgwhite, line 1: What do you think about calling the component `<RenderTerminal>` rather than just `<Terminal>`? That might make it slightly more explicit that the component expects to be given a `Terminal` object.

```suggestion
<RenderTerminal @terminal={{this.terminal}}/>
```

ui/tests/integration/components/terminal-test.ts
```diff
@@ -0,0 +1,28 @@
+import { module, test } from 'qunit';
+
+import { createTerminal } from 'waypoint/utils/terminal';
```
jgwhite, line 3:
```suggestion
import { createTerminal } from 'waypoint/utils/create-terminal';
```

jgwhite (commented):
This is looking awesome so far! Everything is so much simpler/cleaner and I think the split between the terminal object and the terminal component feels pretty intuitive.

ui/ember-cli-build.js
```diff
@@ -37,10 +37,15 @@ module.exports = function (defaults) {
...
-      //   plugins: [new (require('webpack-bundle-analyzer').BundleAnalyzerPlugin)()],
-      // },
+
+      webpack: {
+        // uncomment this to see a tree view of what ember-cli-auto-import builds
+        // plugins: [new (require('webpack-bundle-analyzer').BundleAnalyzerPlugin)()],
+        // required for ansi-colors to work in the browser
+        node: {
+          process: true,
+        },
```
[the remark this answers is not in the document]
gregone: yeah, within reason. In this case, there was [an unguarded property read on `process.env` in a simple boolean chec](https://github.com/doowb/ansi-colors/blob/a4794363369d7b4d1872d248fc43a12761640d8e/index.js#L13)k. So if native objects like `process`, `fs` etc. are just used for configuration, this helps alleviate errors, but any more than that, and packages probably won't work in the browser, even if they can be imported.

See https://github.com/ef4/ember-auto-import#global-is-undefined-or-cant-find-module-path-or-cant-find-module-fs & https://v4.webpack.js.org/configuration/node/

ui/tests/integration/components/terminal-test.ts
```diff
@@ -0,0 +1,28 @@
...
+    let terminal = createTerminal({ inputDisabled: true });
+    this.set('terminal', terminal);
+    // pass terminal and render
+    await render(hbs`<Terminal @terminal={{this.terminal}}/>`);
+    // write line and see if it renders
+    terminal.writeln('Welcome to Waypoint!');
+    // We have to use the runloop as writeln isn't async
+    // Note that even the xterm.js rendering tests use polling to evaluate rendering
+    later(() => {
+      assert.equal(terminal?.buffer?.active?.getLine(0)?.translateToString(true), 'Welcome to Waypoint!');
```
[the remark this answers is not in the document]
gregone: Yeah I obviously considered that. It uses canvas with WebGL, so there isn't really a DOM to access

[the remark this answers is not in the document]
gregone: Update: i added a second test using the dom Renderer and asserting content in https://github.com/hashicorp/waypoint/pull/2356/commits/d645464ee5ea68593960df58f309df613a511c98

ui/tests/integration/components/terminal-test.ts
gregone: The element is retained on the object on `dispose`, so now I'm using [`assignedSlot`](https://developer.mozilla.org/en-US/docs/Web/API/Element/assignedSlot) to assess whether it's been removed from the DOM.

ui/tests/integration/components/terminal-test.ts
```diff
@@ -0,0 +1,28 @@
...
+  test('basic rendering', async function (assert) {
+    // Setup terminal externally, as expected by the component
+    let terminal = createTerminal({ inputDisabled: true });
+    this.set('terminal', terminal);
+    // pass terminal and render
+    await render(hbs`<Terminal @terminal={{this.terminal}}/>`);
+    // write line and see if it renders
+    terminal.writeln('Welcome to Waypoint!');
+    // We have to use the runloop as writeln isn't async
+    // Note that even the xterm.js rendering tests use polling to evaluate rendering
```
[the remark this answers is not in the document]
gregone: So they use a 10ms polling interval, but poll continuously. I've found that 10ms works fine with the canvas renderer, but had to bump it to 50ms in the DOM test

ui/ember-cli-build.js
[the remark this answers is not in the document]
jgwhite: Reminds me of the days of browserify

ui/tests/integration/components/terminal-test.ts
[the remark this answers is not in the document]
jgwhite: Oh wow, I totally didn’t realize it uses canvas. That’s cool!

jgwhite (approved):
🎉

sabrinako (approved):
LGTM! 👍  Awesome job with this

Issue #2137, opened 2021-08-25
core, cli, ui: log the values of used input variables in output


krantzinator:
**Is your feature request related to a problem? Please describe.**
We should output the list of variables and their values for the user to verify.

**Describe the solution you'd like**
This could be at the start of an operation, or in the output once an operation concludes. This should be in the logs and UI definitely, and possibly in the CLI, though the output could be long so we'd want to do some thinking around that. A user should be able to look at the result of a Build or Deploy and see at a glance what variable values were used in that stage.

**Describe alternatives you've considered**
Keeping them a 🔍 mystery 🕵🏻 

**Additional context**
This was stated in the Future Ideas of the input variables RFC

PR #2362, opened 2021-09-27
Support specifying custom environment variables for input variable default values

mitchellh:
This lets input variables specify any named environment variable as a source for a default value. Example: 

```
variable "region" {
  type    = string
  env     = ["REGION"]
}
```

If the `REGION` env var is set, it will populate the default value of this variable (Note: `WP_VAR_region` takes priority always). If multiple env vars are specified, the first matching takes priority. If none of the defined env vars are set, the `default` value (if set) is used. If no default is specified, it is an error since the variable is required and unset.

The env vars specified here are loaded _only at the runner (config evaluation time)_ currently. I'm unsure if we want to expand this to load at CLI time but I would say no, we only want WP_VAR_ variables there, unless the CLI also happens to be the runner (local mode).

This also fixes a potential bug I found where WP_VAR_ variables weren't loaded at runner-time. This could easily be as intended but I think we want to load those at runner-time too so that runner scoped configs can set variables for builds and so on.

**TODO:** I still have to write docs and validate this a bit more.

### DIFF
--- a/.changelog/2362.txt
+++ b/.changelog/2362.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+config: input variables (`variable`) can now use an `env` key to specify
+alternate environment variable names to source variable values from
+```
--- a/internal/config/variables/testdata/valid.hcl
+++ b/internal/config/variables/testdata/valid.hcl
@@ -17,3 +17,9 @@ variable "whatdoesittaketobenumber" {
   default = 1
   type = number
 }
+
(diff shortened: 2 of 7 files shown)


krantzinator (approved):
Nice! I think this work expedites the need to output what values were used in a given operation #2137, to lend clarity to what source is providing what value.

mitchellh:
I believe there is good consensus around this. Merging!

PR #2363, opened 2021-09-28
ui: upgrade conflicting express type declarations

jgwhite:
## Why the change?

This is a fairly benign change that updates the following libraries by patch increments:

- @types/express
- @types/express-serve-static-core
- @types/serve-static

The benefit is negligible, it just gets rid of this ugly warning introduced by #2355:

```
Interface 'Response<ResBody>' incorrectly extends interface 'ServerResponse'
```


gregone (approved):
LGTM

PR #2371, opened 2021-09-28
gitrefpretty no longer requires `git`, works in ODR

mitchellh:
This makes `gitrefpretty` fall back to using the go-git library if `git` isn't available. We still prefer `git` because its MUCH faster for this particularly use case but we can fallback and have the same functionality if it isn't available. I also added unit tests for both functionality.

This is required for ODR since our ODR image doesn't have Git.


evanphx:
Ironically I have this change locally but hadn't pushed it into a PR, but yours looks better!

evanphx (approved):

PR #2372, opened 2021-09-28
builtin/docker: if host is docker.io, use index.docker.io for registry

mitchellh:
The Docker `reference` library doesn't give us the real registry host in
this special case, so we just set it manually.

If we find a better way to do this, we should do it.


mitchellh:
(Approved by Evan in Slack)

PR #2373, opened 2021-09-28
Update runner config page

koesbong:
Runners now support dynamic config values


mitchellh:
Thanks! We need to make a big pass for 0.6 regardless but this is good.

PR #2378, opened 2021-09-28
internal/core: add project/app to JobInfo

mitchellh:
A little PR that builds on top of recent SDK improvements.

This exposes the project and app name in the component.JobInfo structure
that is already passed to all plugin operations. This will allow plugins
to make use of project/app name as necessary for behavior.

I want to use this information for default release names for Helm.


krantzinator (approved):
beautiful

PR #2383, opened 2021-09-29
[WIP] 🐹 UI: USe ember changeset

gregone:


ui/app/components/project-config-variables/list-item.ts
```diff
@@ -51,23 +57,28 @@ export default class ProjectConfigVariablesListItemComponent extends Component<V
...
-      await this.args.deleteVariable(this.initialVariable);
-      this.storeInitialVariable();
+    if (this.variable.name !== this.changeset.name) {
+      await this.args.saveVariableSettings(this.changeset, false);
+      await this.args.deleteVariable(this.variable);
     } else {
-      await this.args.saveVariableSettings(this.variable, false);
+      await this.args.saveVariableSettings(this.changeset, false);
     }
+    this.changeset?.execute();
```
jgwhite, line 81: Do we need to execute the changeset? What happens if we don’t?

ui/app/components/project-config-variables/list-item.ts
```diff
@@ -51,23 +57,28 @@ export default class ProjectConfigVariablesListItemComponent extends Component<V
...
     this.isCreating = false;
     this.isEditing = false;
```
jgwhite, line 83: Should we discard the changeset at this point?

ui/app/components/project-config-variables/list-item.ts
```diff
@@ -83,6 +94,7 @@ export default class ProjectConfigVariablesListItemComponent extends Component<V
   cancelEdit(): void {
     this.isCreating = false;
     this.isEditing = false;
-    this.variable = this.initialVariable;
+    this.changeset?.unexecute();
+    // this.variable = this.initialVariable;
```
jgwhite, line 98: Similarly, do we need to execute/unexecute the changeset? Or cold we just throw it away at this point?

ui/app/components/project-config-variables/list.ts
```diff
@@ -13,6 +14,8 @@ interface ProjectConfigArgs {
   project: Project.AsObject;
 }
 
+type ConfigVarChangeset = Partial<BufferedChangeset> & ConfigVar.AsObject;
```
jgwhite, line 17: We could probably make use of this type in `list-item.ts` too.

PR #2385, opened 2021-09-29
Add workspace CLI command, list/inspect subcommands

catsby:
This PR introduces a top level command tree `workspace` that is scoped to the current context, with two subcommands:

- `list`: list all workspaces in this context
- `inspect <name>`: inspect the workspace `<name>`, or the current workspace

~This PR builds off of https://github.com/hashicorp/waypoint/pull/2353 so I'm opening it as a draft until 2353 is approved and merged. Until then this PR will appear to be larger than it actually is.~ #2353 has been merged 

UX:
<details>

**List:**
```
$ waypoint workspace list                  
   NAME   |        PROJECTS
----------+--------------------------
  default | example-go,example-ruby
  dev     | example-go
  lab     | example-ruby,example-go
```

**Inspect the default workspace:**
```
$ waypoint workspace inspect               

» Workspace Info:
             Name: default
  Latest activity: 22 minutes ago
         Projects: example-go,example-ruby
```

**Inspect workspace by name:**
```
$ waypoint workspace inspect dev

» Workspace Info:
             Name: dev
  Latest activity: 28 minutes ago
         Projects: example-go
```

</details>

### DIFF
--- a/internal/cli/workspace_inspect.go
+++ b/internal/cli/workspace_inspect.go
@@ -81,7 +81,7 @@ func (c *WorkspaceInspectCommand) Run(args []string) int {
 			Name: "Name", Value: workspace.Name,
 		},
 		{
-			Name: "Latest activity", Value: lastActiveTime,
+			Name: "Last Updated", Value: lastActiveTime,
 		},
 		{
 			Name: "Projects", Value: strings.Join(projects, ","),
--- a/internal/cli/workspace_inspect.go
+++ b/internal/cli/workspace_inspect.go
@@ -104,7 +104,7 @@ func (c *WorkspaceInspectCommand) AutocompleteFlags() complete.Flags {
 }
 
 func (c *WorkspaceInspectCommand) Synopsis() string {
-	return "Output Workspace information."
+	return "Output information for a given Workspace."
 }
 
 func (c *WorkspaceInspectCommand) Help() string {
--- a/internal/cli/workspace_inspect.go
+++ b/internal/cli/workspace_inspect.go
@@ -1,0 +1,118 @@
+package cli
+
+import (
+	"errors"
+	"strings"
+
+	"github.com/dustin/go-humanize"
+	"github.com/golang/protobuf/ptypes"
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/clierrors"
+	"github.com/hashicorp/waypoint/internal/pkg/flag"
+	"github.com/posener/complete"
+)
+
+type WorkspaceInspectCommand struct {
+	*baseCommand
+
+	flagWorkspaceName string
+}
+
+func (c *WorkspaceInspectCommand) Run(args []string) int {
+	// Initialize. If we fail, we just exit since Init handles the UI.
+	flagSet := c.Flags()
+	if err := c.Init(
+		WithArgs(args),
+		WithFlags(flagSet),
+		WithNoConfig(),
+	); err != nil {
+		return 1
+	}
+	args = flagSet.Args()
+
+	if len(args) > 1 {
+		c.ui.Output(c.Help(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	// if a workspace name is given as an argument, use that. This will take
+	// precedence over the -workspace flag as it's an argument to the command
+	var workspaceName string
+	if len(args) > 0 {
+		workspaceName = args[0]
+	}
+
+	if workspaceName == "" {
+		// lookup the default
+		wp, err := c.workspace()
+		if err != nil {
+			c.ui.Output(
+				clierrors.Humanize(errors.New("error loading context default workspace")),
+				terminal.WithErrorStyle(),
+			)
+			return 1
+		}
+		workspaceName = wp
+	}
+
+	workspace, err := getWorkspace(c.Ctx, c.project.Client(), workspaceName)
+	if err != nil {
+		c.ui.Output(
+			clierrors.Humanize(err),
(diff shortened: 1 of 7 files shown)


evanphx (approved):
Looks good!

internal/cli/workspace_inspect.go
```diff
@@ -0,0 +1,118 @@
...
+	if t, err := ptypes.Timestamp(workspace.ActiveTime); err == nil {
+		lastActiveTime = humanize.Time(t)
+	}
+
+	c.ui.NamedValues([]terminal.NamedValue{
+		{
+			Name: "Name", Value: workspace.Name,
+		},
+		{
+			Name: "Latest activity", Value: lastActiveTime,
```
briancain, line 84: It'd be cool if we could actually say what the latest activity was, this looks to be a timestamp but the actual activity was unknown, right? Not sure if we hold onto what the latest operation was. Maybe this header could be `"Last Updated"` since latest activity makes me think I should be getting what the activity was rather than when it happened

internal/cli/workspace_list.go
```diff
@@ -0,0 +1,135 @@
...
+
+func getWorkspace(ctx context.Context, client pb.WaypointClient, name string) (*gen.Workspace, error) {
+	resp, err := client.GetWorkspace(ctx, &pb.GetWorkspaceRequest{
+		Workspace: &pb.Ref_Workspace{
+			Workspace: name,
+		},
+	})
+	if err != nil {
+		if status.Code(err) == codes.NotFound {
+			return nil, fmt.Errorf("no workspace found for name \"%s\"", name)
```
briancain, line 124:
```suggestion
			return nil, fmt.Errorf("no workspace found for name %q", name)
```

internal/cli/workspace_list.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		return nil, err
+	}
+
+	// this is unlikely to happen without first hitting the codes.NotFound error
+	// above, but added here to be defensive.
+	if resp.Workspace == nil {
+		return nil, fmt.Errorf("no workspace returned for name \"%s\"", name)
```
briancain, line 132:
```suggestion
		return nil, fmt.Errorf("no workspace returned for name %q", name)
```

internal/cli/workspace_inspect.go
```diff
@@ -0,0 +1,118 @@
...
+func (c *WorkspaceInspectCommand) Synopsis() string {
+	return "Output information for a given Workspace."
+}
+
+func (c *WorkspaceInspectCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint workspace inspect [<name>]
+
+  Output information about a waypoint workspace, including the projects and
+  last known activity
```
briancain, line 115:
```suggestion
  Output information about a Waypoint workspace, including all projects and
  last known activity timestamp
```

Not sure if "workspace" should be capitalized here 🤔

internal/cli/workspace_list.go
```diff
@@ -0,0 +1,135 @@
...
+	// and want color detection to work.
+	out, _, err := c.ui.OutputWriters()
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	table := tablewriter.NewWriter(out)
+	table.SetHeader([]string{"Name", "Projects"})
+	table.SetBorder(false)
```
briancain, line 64: Not sure if you knew, but we have a built in table writer within UI which would help keep our `waypoint * list` functions looking the same.

You can initialize the table from the SDK terminal package

```
tbl := terminal.NewTable(headers...)
```

And then build up the columns just as you're already doing with `tbl.Rich`, and print using the UI package with `c.ui.Table(tbl)`

internal/cli/workspace_list.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (c *WorkspaceListCommand) Synopsis() string {
+	return "List workspaces for the current context."
+}
+
+func (c *WorkspaceListCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint workspace list
+
+  Lists all the known workspaces available to the CLI for the current sever
```
briancain, line 110:
```suggestion
  Lists all the known workspaces available to the CLI for the current Waypoint server
```

briancain (changes requested):
Nice! Love to get more inspection tools on the CLI. Looks good. Most of my comments are pretty minor fixups 😄 👍🏻

internal/cli/workspace_list.go
catsby: Ah, I missed the part where `%q` does quoting, thanks!

internal/cli/workspace_inspect.go
catsby: I don't believe so, but I'm not the best at these things :D

briancain: Ah yeah, I noticed it was capitalized else where, so maybe the other `Workspace` instances should be lower cased? 🤔

internal/cli/workspace_inspect.go
catsby: > the actual activity was unknown, right?

To my knowledge yes, it is unknown. I agree and updated the label used here to be `Last Updated` in ab3cda44

internal/cli/workspace_list.go
catsby: Thanks! I refactored that in 1880c9ec

catsby: Thanks! I refactored that in 1880c9e

briancain (approved):

PR #2387, opened 2021-09-29
Website: move context and context inspect lower in CLI navigation

catsby:
I believe the current `context` and `context-inspect` are misplaced in the CLI nav. I've moved them lower with the other `context` commands, and also removed the `-` from `context-inspect`. Attached are before/after screen grabs.

**Before:**
Screen Shot 2021-09-29 at 16 37 04

**After:**
Screen Shot 2021-09-29 at 16 37 37

### DIFF
--- a/website/data/commands-nav-data.json
+++ b/website/data/commands-nav-data.json
@@ -8,14 +8,6 @@
     "path": "deploy"
   },
   {
-    "title": "context",
-    "path": "context"
-  },
-  {
-    "title": "context-inspect",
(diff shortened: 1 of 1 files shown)


krantzinator:
Good call. I vote we also move `project apply` from the top-level "most important commands" section down into alphabetical order.

briancain (approved):

catsby:
@krantzinator great idea, done in 0ec65552

Issue #2393, opened 2021-09-30
Upgrade to ember-cli-babel 7.x.x wherever possible


jgwhite:
In the interest of improving build times (and build memory usage) for the UI, we have been advised to upgrade to ember-cli-babel 7.x.x where possible (and away from ember-cli-babel 6.x.x).

In my limited research, I believe the two dependencies that are still pinned to older versions of ember-cli-babel are [ember-cli-page-object](https://github.com/san650/ember-cli-page-object) and [ember-cli-flash](https://github.com/poteto/ember-cli-flash).

PR #2396, opened 2021-09-30
ui: upgrade ember-concurrency to 2.x

jgwhite:
## Why the change?

One step towards #2393 

## “Always run `yarn-dedupe` you say?”

Per Chris Krycho’s advice, I also ran `npx yarn-dedupe` after `yarn install`. This process is pretty weird and awkward, I find you have to:

1. Tweak deps in `package.json`
2. Run `yarn install` (or simply `yarn`)
3. Run `npx yarn-dedupe`
4. Run `yarn install` again (because reasons?)

## How do I test it?

I smoke tested this against a real Waypoint server and everything appeared to work as normally (including the “Deploy & Release” button).


gregone (approved):
LGTM

PR #2398, opened 2021-09-30
Decorate the location of an image better to improve ODR handling

evanphx:
This change lets tools that push their image to a remote registry as part of the build (used by ODR) to indicate as much so that later phases[^1] can know that the image is already there.

[^1]: Mostly this is because in ODR mode, we still run the Push function of Registry and this allows that Push function to know that it doesn't need do do anything more.


mitchellh (approved):

builtin/pack/builder.go
```diff
@@ -262,6 +262,7 @@ RUN mkdir /tmp/cache /tmp/layers && \
 
 	dir := src.Path
 
+	step.Update("Repository is available and ready: %s:%s", ai.Image, ai.Tag)
```
briancain, line 265: Do we print something similar for the docker builder already? 🤔

briancain (approved):

briancain:
I think if you rebase off main you'll get the circle fix for the failing integration test

PR #2399, opened 2021-09-30
Fix a few minor kubernetes papercuts

evanphx:
1. This scales the wait time back so that bad deploys fail much faster.
   Previously we had it hardcoded to 10 minutes, now it's dynamic.
2. When a deployment fails, print out non Info events from the pods to
   help the user diagonise what happened.
3. Provide some small grace period on pod cleanup to allow for the
   potential of debugging the failed pods directly.
4. Handle not being able to search for metric pods
5. Handle destroying an old deployment that didn't have autoscaling
   defined yet.
6. Setup the kubernetes service with the workspace name if the workspace
   is not the default workspace

### DIFF
--- a/.changelog/2399.txt
+++ b/.changelog/2399.txt
@@ -2,6 +2,6 @@
 plugin/k8: Report events on failed pods when a deployment fails
 ```
 
-```release-note:bugfix
+```release-note:bug
 plugin/k8: Setup Kubernetes services for different workspaces properly
 ```
--- a/.changelog/2399.txt
+++ b/.changelog/2399.txt
@@ -1,0 +1,7 @@
+```release-note:feature
+plugin/k8: Report events on failed pods when a deployment fails
+```
+
+```release-note:bugfix
+plugin/k8: Setup Kubernetes services for different workspaces properly
+```
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -703,7 +703,7 @@ func (p *Platform) resourceDeploymentCreate(
 	)
 
 	// We wait the maximum amount of time that the deployment controller would wait for a pod
-	// to start before exitting. We double the time to allow for various kubernetes based
+	// to start before exiting. We double the time to allow for various Kubernetes based
 	// delays in startup, detection, and reporting.
 	timeout := time.Duration((timeoutSeconds*failureThreshold)+initialDelaySeconds) * 2 * time.Second
 
@@ -772,6 +772,7 @@ func (p *Platform) resourceDeploymentCreate(
 	})
 	if err != nil {
 		step.Update("Error detected waiting for Deployment to start.")
+		step.Status(terminal.StatusError)
 		step.Abort()
 
 		ui.Output("The following is events for pods observed while attempting to start the Deployment", terminal.WithWarningStyle())
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	k8sresource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


builtin/k8s/releaser.go
```diff
@@ -724,12 +724,18 @@ func (r *Releaser) Release(
...
 	src *component.Source,
+	job *component.JobInfo,
 	ui terminal.UI,
 	target *Deployment,
 	dcr *component.DeclaredResourcesResp,
 ) (*Release, error) {
 	var result Release
-	result.ServiceName = src.App
+
+	if job.Workspace == "default" {
```
mitchellh, line 734: I think we should throw this in the labels too for the label selector.

.changelog/2399.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:feature
+plugin/k8: Report events on failed pods when a deployment fails
+```
+
+```release-note:bugfix
```
briancain, line 5:
```suggestion
```release-note:bug
```

Should just be `bug`

builtin/k8s/platform.go
```diff
@@ -698,20 +702,25 @@ func (p *Platform) resourceDeploymentCreate(
 		reportedError bool
 	)
 
-	timeout := 10 * time.Minute
+	// We wait the maximum amount of time that the deployment controller would wait for a pod
+	// to start before exitting. We double the time to allow for various kubernetes based
```
briancain, line 706:
```suggestion
	// to start before exiting. We double the time to allow for various Kubernetes based
```

builtin/k8s/platform.go
```diff
@@ -698,20 +702,25 @@ func (p *Platform) resourceDeploymentCreate(
...
+	err = wait.PollImmediate(time.Second, timeout, func() (bool, error) {
 		dep, err := dc.Get(ctx, result.Name, metav1.GetOptions{})
 		if err != nil {
 			return false, err
 		}
 
 		if time.Since(lastStatus) > 10*time.Second {
 			step.Update(fmt.Sprintf(
-				"Waiting on deployment to become available: %d/%d/%d",
+				"Waiting on deployment to become available: requested=%d running=%d ready=%d",
```
briancain, line 721: I noticed this change in your demo and I like it! 👍🏻 ✨

builtin/k8s/platform.go
```diff
@@ -760,6 +771,28 @@ func (p *Platform) resourceDeploymentCreate(
 		return false, nil
 	})
 	if err != nil {
+		step.Update("Error detected waiting for Deployment to start.")
```
briancain, line 774:
```suggestion
		step.Update("Error detected waiting for Deployment to start.")
		step.Status(terminal.StatusError)
```

builtin/k8s/platform.go
```diff
@@ -760,6 +771,28 @@ func (p *Platform) resourceDeploymentCreate(
...
+			events, err := ev.List(ctx, metav1.ListOptions{
+				FieldSelector: sel.String(),
+			})
+			if err == nil {
+				ui.Output("Events for %s", name, terminal.WithHeaderStyle())
+				for _, ev := range events.Items {
+					if ev.Type == "Normal" {
+						continue
+					}
+					ui.Output("  %s: %s (%s)", ev.Type, ev.Message, ev.Reason)
```
briancain, line 791: I also noticed this change in your demo, excited for this! 😄

briancain (changes requested):
I've got a few minor requests and comments, but this looks great! ✨ 💯

builtin/k8s/releaser.go
evanphx: Not sure what ya mean....

briancain (approved):

builtin/k8s/platform.go
evanphx: bitmoji

builtin/k8s/platform.go
evanphx: bitmoji

PR #2403, opened 2021-10-01
Moving all runner config flags under `runner config`

izaaklauer:
This changes these commands:

```
waypoint runner on-demand inspect <id>
waypoint runner on-demand set -id=<id> -thing=<thing>
waypoint runner list (returns config IDs)
```

To these:
```
waypoint runner config inspect <id>
waypoint runner config set -id=<id> -thing=<thing>
waypoint runner config list (returns config IDs)
```

This also de-emphasises on the "on-demand" nature of this config across the output of these commands.


izaaklauer:
no changelog, because these commands weren't released yet.

catsby (changes requested):
It looks like the old command files are now empty, when I think they should be removed?

izaaklauer:
@catsby Which files do you mean in particular? It looks to me like all the old command files were deleted or renamed.

catsby (approved):
LGTM 👍

evanphx:
So, I think we need to be careful with wording here. There is already `waypoint config set -runner` and this now adds `waypoint runner config`. We already have annoying overlapping terminology around "app config" between the waypoint.hcl and the values created by `waypoint config`. That's one reason that I left it with `on-demand` in there, so it was clear WHAT config it was.

evanphx (changes requested):
<read comment about overlapping naming of 'runner config'>

izaaklauer:
That's a good point. The main thing that that I was trying to fix was just that `inspect`, `set`, and `list` weren't all under the same parent. Two other ideas then:

We use a common prefix, like `on-demand` from earlier:
```
waypoint runner on-demand inspect <id>
waypoint runner on-demand set -id=<id> -thing=<thing>
waypoint runner on-demand list (returns config IDs)
```

Instead of calling this "config" at all, we call them runner "profiles".
```
waypoint runner profile inspect <id>
waypoint runner profile set -id=<id> -thing=<thing>
waypoint runner profile list (returns config IDs)
```

izaaklauer:
After discussion, we decided we don't need to rename the internal plumbing from ondemand runner config to `runner profiles` - it's ok to have a friendly name for the command that hints at the internal names. It also be a fair amount of work to rename for minimal benefit.

internal/cli/runner_profile_set.go
```diff
@@ -186,37 +186,37 @@ func (c *OnDemandRunnerConfigApplyCommand) Run(args []string) int {
...
 			Default: "",
-			Usage:   "The name of an existing on-demand runner to update.",
+			Usage:   "The name of an existing runner profile to update.",
 		})
 
 		f.StringVar(&flag.StringVar{
 			Name:    "oci-url",
 			Target:  &c.flagOCIUrl,
-			Default: "hashicorp/waypoint:stable",
+			Default: "hashicorp/waypoint:stable", // TODO: update to published prod ODR image before 0.6.0 release
```
evanphx, line 219: TODO?

izaaklauer: Ah, that's unrelated to this change - just something I noticed that we'll need to get to before the release. I figure it'll be `hashicorp/waypoint-odr`, but I don't see us referencing that anywhere - it seems like we don't have the circleCI conf yet to publish it?

PR #2409, opened 2021-10-04
ui: ensure ember-cli-flash doesn’t clobber pds-flash-messages

jgwhite:
This is a pretty subtle bug. By default, ember-cli-flash [automatically injects][1] a `flashMessages` service into various types such as routes and controllers. This is a problem for us because it clobbers our custom `PdsFlashMessages` service. As a result, you could inject the service into a route and it would refuse to work for no apparent reason.

It looks like the ember-cli-flash maintainers intended to remove the automatic injections for v2.0.0 but didn’t actually do so.

[1]: https://github.com/poteto/ember-cli-flash/blob/dde3da9d1ee3fae24e910a7353bb9f88d7fb1690/app/initializers/flash-messages.js#L20-L22

### DIFF
--- a/ui/config/environment.js
+++ b/ui/config/environment.js
@@ -30,6 +30,7 @@ module.exports = function (environment) {
       extendedTimeout: 700,
       type: 'info',
       types: ['error', 'success', 'info', 'warning'],
+      injectionFactories: [],
     },
 
     // Ember-toggle addon settings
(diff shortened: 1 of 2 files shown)


ui/config/environment.js
```diff
@@ -30,6 +30,7 @@ module.exports = function (environment) {
       extendedTimeout: 700,
       type: 'info',
       types: ['error', 'success', 'info', 'warning'],
+      injectionFactories: [],
```
gregone, line 33: 👍

gregone (approved):

PR #2411, opened 2021-10-04
fix text overflow in flash notifications with long words

### LINKED ISSUE #2265: ui: Long errors overflow the notification box
**Describe the bug**
Long errors overflow the notification box:

CleanShot 2021-09-09 at 15 15 50@2x

**Steps to Reproduce**
1. From the UI, take some action that triggers a long error, such as clicking the “re-run health check” button on a project that does not have a remote source configured
2. Observe the overflow error as pictured above 

**Expected behavior**
The box should resize itself to contain the error.

* Waypoint CLI Version: 7b4ab193a817d7d904de913c602c267ff07600e0

gregone:
Fixes: #2265 

before: 

Screen Shot 2021-10-04 at 17 51 17

after: 
Screen Shot 2021-10-04 at 17 51 10

### DIFF
--- a/.changelog/2411.txt
+++ b/.changelog/2411.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: notification messages display nicely when containing long words such as URLs
+```
--- a/ui/app/styles/components/notifications.scss
+++ b/ui/app/styles/components/notifications.scss
@@ -42,5 +42,10 @@
         opacity: 0;
       }
     }
+    // Overrides for the flash-messages content
+    .pds-popup__header {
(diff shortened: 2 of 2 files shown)


amandakalk:
@gregone Did you check this in Safari? Looks like `overflow-wrap: anywhere` [isn't supported in Safari](https://caniuse.com/?search=overflow-wrap), but not sure if the issue also exists in Safari.

gregone:
@amandakalk good call, I just added a fallback to use `word-break` instead

jgwhite (approved):
Yessss 🙌

Issue #2075, opened 2021-08-18
Include UI output in runner logs at DEBUG level or higher


izaaklauer:
**Is your feature request related to a problem? Please describe.**
Currently, UI logs (i.e. calls to stepGroup.Status) do not appear in runner logs. As usage of remote and on-demand runners becomes more common in waypoint, it will be difficult to determine what runners are doing from their logs alone if the logs do not include this UI output.

To work around this right now, we _occasionally_ emit nearly identical UI and Log messages in plugins, but that's onerous for plugin developers and not implemented consistently. Example below. https://github.com/hashicorp/waypoint/blob/d6d385024e2af40c9cf31b2bbe1e0ba683746ad7/builtin/aws/ecs/platform.go#L1056-L1061

**Describe the solution you'd like**
It would be great if we could give our logger to the UI, and have it print every message to the logger at level `debug`.

briancain:
I'm actually not sure if this is a waypoint thing or waypoint-plugin-sdk thing! But probably good to have it live here for now, since there's probably a once or twice over that will need to be done to see what logs and stepgroup output we can simplify once it's in place.

PR #2412, opened 2021-10-04
Papercuts in kubernetes server install

izaaklauer:
- Fixes a problem where deployments would be marked as "Degraded", but were actually fine
- Adds a new clusterrole and clusterrolebinding to the install flow to allow the runner to get and list nodes, which it needs to discover the IP of nodeport type services.
- If for some reason the nodeport ip discovery fails, don't fail the whole release (per @evanphx 's suggestion)

### DIFF
--- a/builtin/k8s/releaser.go
+++ b/builtin/k8s/releaser.go
@@ -320,11 +320,13 @@ func (r *Releaser) resourceServiceCreate(
 		nodeclient := clientSet.CoreV1().Nodes()
 		nodes, err := nodeclient.List(ctx, metav1.ListOptions{})
 		if err != nil {
-			return err
+			// Rather than fail the whole release, report the error and then complete.
+			step.Status(terminal.StatusError)
+			step.Update("Cannot determine release URL for nodeport service due to failure to list nodes: %s", err)
+		} else {
+			nodeIP := nodes.Items[0].Status.Addresses[0].Address
+			result.Url = fmt.Sprintf("http://%s:%d", nodeIP, service.Spec.Ports[0].NodePort)
 		}
-
-		nodeIP := nodes.Items[0].Status.Addresses[0].Address
-		result.Url = fmt.Sprintf("http://%s:%d", nodeIP, service.Spec.Ports[0].NodePort)
 	} else {
 		result.Url = fmt.Sprintf("http://%s:%d", service.Spec.ClusterIP, service.Spec.Ports[0].Port)
 	}
--- a/internal/serverinstall/k8s.go
+++ b/internal/serverinstall/k8s.go
@@ -10,6 +10,9 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	"github.com/ghodss/yaml"
 	appsv1 "k8s.io/api/apps/v1"
 	apiv1 "k8s.io/api/core/v1"
@@ -309,6 +312,8 @@ func (i *K8sInstaller) Install(
 		// Ensure the service is ready to use before returning
 		_, err = net.DialTimeout("tcp", httpAddr, 1*time.Second)
 		if err != nil {
+			// Depending on the platform, this can take a long time. On EKS, it's by far the longest step. Adding an explicit message helps
+			s.Update("Service %s exists and is configured, but isn't yet accepting incoming connections. Waiting...\n", serviceName)
 			return false, nil
 		}
 		log.Info("http server ready", "httpAddr", addr)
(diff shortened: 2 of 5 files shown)


krantzinator (approved):
Nice docs updates as well

builtin/k8s/releaser.go
```diff
@@ -320,11 +320,13 @@ func (r *Releaser) resourceServiceCreate(
 		nodeclient := clientSet.CoreV1().Nodes()
 		nodes, err := nodeclient.List(ctx, metav1.ListOptions{})
 		if err != nil {
-			return err
+			// Rather than fail the whole release, report the error and then complete.
+			step.Status(terminal.StatusError)
+			step.Update("Cannot determine release URL for nodeport service due to failure to list nodes: %s", err)
```
briancain, line 325: Maybe this message should just be on its own with a `c.ui.Output` and a `terminal.WithWarningStyle()`? Otherwise I think this updates whatever the last stepgroup line was, and could potentially be updated over if we add any other stepgroup output later. It could also be its own step if we do the `sg.Add()` and close it with done right away.

internal/serverinstall/k8s.go
```diff
@@ -309,6 +311,8 @@ func (i *K8sInstaller) Install(
 		// Ensure the service is ready to use before returning
 		_, err = net.DialTimeout("tcp", httpAddr, 1*time.Second)
 		if err != nil {
+			// Depending on the platform, this can take a long time. On EKS, it's by far the longest step. Adding an explicit message helps
+			s.Update("Service %s exists and is configured, but isn't yet accepting incoming connections. Waiting...\n", serviceName)
```
briancain, line 315:
```suggestion
			s.Update("Service %q exists and is configured, but isn't yet accepting incoming connections. Waiting...", serviceName)
```

Not sure if you need the `\n`, `Update` should get a new line by default unless you were hoping for an extra one. We should probably also update the stepgroup for when the service is ready and we're done waiting?

briancain (changes requested):
Some minor requests! 🧻 ✂️ Looks good though, agree on making the deployments status simpler, hopefully the real issues get surfaced in its pods status.

briancain (approved):

internal/serverinstall/k8s.go
```diff
@@ -1358,6 +1364,43 @@ func newServiceAccount(c k8sConfig) (*apiv1.ServiceAccount, error) {
 	}, nil
 }
 
+// newServiceAccountClusterRoleWithBinding creates the cluster role and binding necessary to create and verify
+// a nodeport type services.
+func newServiceAccountClusterRoleWithBinding(c k8sConfig) (*rbacv1.ClusterRole, *rbacv1.ClusterRoleBinding, error) {
+	roleName := "waypoint-runner"
```
catsby, line 1370: > roleName := "waypoint-runner"

Should this default to the same value as the `k8s-runner-service-account` flag (which defaults to `waypoint-runner`)?

internal/serverinstall/k8s.go
```diff
@@ -1690,6 +1733,31 @@ func (i *K8sInstaller) initServiceAccount(
...
+		}
+		if _, err := crClient.Create(ctx, cr, metav1.CreateOptions{}); err != nil {
+			return status.Errorf(codes.Internal, "Failed to create cluster role %s: %s", cr.Name, err)
+		}
+	}
+	if crb != nil {
+		crbClient := clientset.RbacV1().ClusterRoleBindings()
+		_, err = crbClient.Get(ctx, crb.Name, metav1.GetOptions{})
+		if err != nil && !errors.IsNotFound(err) {
+			return err
```
catsby, line 1754: Should we wrap the 2 `IsNotFound` errors in `status.Errorf` here, to be consistent with the other error scenarios from the `newServiceAccountClusterRoleWithBinding`? Or is directly returning the error more appropriate?

catsby (approved):
LGTM, just minor nit/questions that I don't consider blockers

builtin/k8s/releaser.go
```diff
@@ -320,11 +320,15 @@ func (r *Releaser) resourceServiceCreate(
 		nodeclient := clientSet.CoreV1().Nodes()
 		nodes, err := nodeclient.List(ctx, metav1.ListOptions{})
 		if err != nil {
-			return err
+			// Rather than fail the whole release, report the error and then complete.
+			// Print in a standalone step, so the output won't get overwritten if we add more step output later in the future.
+			errStep := sg.Add("Cannot determine release URL for nodeport service due to failure to list nodes: %s", err)
+			errStep.Status(terminal.StatusError)
+			errStep.Done()
```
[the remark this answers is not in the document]
izaaklauer: There are a bunch of places where I think ui messages would be super helpful in logs - honestly, i've been meaning to solve this eventually with https://github.com/hashicorp/waypoint/issues/2075, so i've been reticent to duplicate messages between ui and the logs. If you think this is an especially important one though, I can see putting a log here.

internal/serverinstall/k8s.go
```diff
@@ -309,8 +311,12 @@ func (i *K8sInstaller) Install(
 		// Ensure the service is ready to use before returning
 		_, err = net.DialTimeout("tcp", httpAddr, 1*time.Second)
 		if err != nil {
+			// Depending on the platform, this can take a long time. On EKS, it's by far the longest step. Adding an explicit message helps
+			s.Update("Service %q exists and is configured, but isn't yet accepting incoming connections. Waiting...", serviceName)
```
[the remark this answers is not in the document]
izaaklauer: Oh interesting thought - I honestly can't think of an error that we wouldn't want to retry on here (for 10 minutes at least). Everything I can think of (socket timeout, dns miss, etc.) could be transient as the ALB comes up.

internal/serverinstall/k8s.go
izaaklauer: I've seen us do this both ways - sometimes we include the type of the resource in it's name, and sometimes we don't. My personal preference is not to include it - i.e. name a role `waypoint-runner` instead of `waypoint-runner-role`, because the resource type is almost always implied by the context.

The service account and role to have the same name won't cause a conflict.

internal/serverinstall/k8s.go
izaaklauer: Oh nice catch, thank you! That check was a late addition copy-paste, and I missed customizing the error.

PR #2413, opened 2021-10-04
cli: Project ops part 1 - CLI updates

krantzinator:
This is a short PR adding a `ProjectTargetRequired` config item and corresponding `WithMultipleApp` option to set some commands to target all applications with the targeted project by default. Users can still specify `-app=APPNAME` to target a specific app.

This change also removes the `-project` flag from the `context X` commands and makes it a global flag. Various other documentation updates are included here as well to amend some of our language from being app-specific to project-specific.

This is a surface layer change, as it only supports functionality that we already had on the CLI side, and does not add any project-scoped awareness to our server or job system. It seemed like a good first pass to start with.

With this functionality, all commands updated with the new config option will now run that command against each application in the project in sequence. A failure in one app is non-blocking; all applications will finish their jobs individually.

Example output:
image


krantzinator:
Pause: there is some weirdness with `release` that I'm seeing after further testing. Will mark as Draft.

internal/cli/artifact_list.go
```diff
@@ -62,6 +62,9 @@ func (c *ArtifactListCommand) Run(args []string) int {
 		}
 		sort.Sort(serversort.ArtifactStartDesc(resp.Artifacts))
 
+		// Output the app's name before the json or table data
+		app.UI.Output("%s", app.Ref().Application, terminal.WithHeaderStyle())
```
evanphx, line 66: This will make the output not proper json though... so probably just do this in the non-json case.

internal/cli/deployment_list.go
```diff
@@ -71,6 +71,7 @@ func (c *DeploymentListCommand) Run(args []string) int {
 	client := c.project.Client()
 
 	err := c.DoApp(c.Ctx, func(ctx context.Context, app *clientpkg.App) error {
+		app.UI.Output("%s", app.Ref().Application, terminal.WithHeaderStyle())
```
evanphx, line 74: Same as other about the output not being valid json in this case.

internal/cli/option.go
```diff
@@ -34,6 +34,17 @@ func WithSingleApp() Option {
 	}
 }
 
+// WithMultipleApp configures the CLI to expect a configuration with
+// one or more apps defined in a project and a project targeted with `-project`.
+// TODO krantzinator: better doc
```
evanphx, line 39: Missed a TODO

internal/cli/release_list.go
```diff
@@ -48,6 +48,9 @@ func (c *ReleaseListCommand) Run(args []string) int {
 	client := c.project.Client()
 
 	err := c.DoApp(c.Ctx, func(ctx context.Context, app *clientpkg.App) error {
+		// UI -- this should happen at the top so that the app name shows clearly
+		// for any errors we may encounter prior to the actual table output
+		app.UI.Output("%s", app.Ref().Application, terminal.WithHeaderStyle())
```
evanphx, line 53: Ditto again about the json.

internal/cli/artifact_list.go
```diff
@@ -62,6 +62,9 @@ func (c *ArtifactListCommand) Run(args []string) int {
...
+
```
briancain, line 67: Hm, leaving a comment here since this is the command I tried, but I think the `WithMultipleApp` CLI option needs some tweaking. Before this change I was able to list artifacts for remote projects, but now I can't:

```
brian@localghost:waypoint λ waypoint artifact list go-gitops-0/go                                                                                                                                                                                                                                                   ±[PROJECT-OPS]
! This command requires a targeted project.
brian@localghost:waypoint λ waypoint artifact list -project=go-gitops-0                                                                                                                                                                                                                                             ±[PROJECT-OPS]
! failed to load config: A Waypoint configuration file (waypoint.hcl) is required but wasn't found.
  The command can use configuration stored on the server by adding project/app
  as a final argument on the command line.
  For instance: `waypoint deployment list webapp/web` for the app 'web' in
  the project 'webapp'
brian@localghost:waypoint λ waypoint artifact list -project=go-gitops-0 -app=go                                                                                                                                                                                                                                     ±[PROJECT-OPS]
! failed to load config: A Waypoint configuration file (waypoint.hcl) is required but wasn't found.
  The command can use configuration stored on the server by adding project/app
  as a final argument on the command line.
  For instance: `waypoint deployment list webapp/web` for the app 'web' in
  the project 'webapp'
```

Before this change:

```
brian@localghost:waypoint λ waypoint artifact list go-gitops-0/go                                                                                                                                                                                                                                                   ±[PROJECT-OPS]
    | ID | REGISTRY | DETAILS |    STARTED     |   COMPLETED
----+----+----------+---------+----------------+-----------------
  ✔ |  1 | docker   | build:1 | 27 minutes ago | 27 minutes ago
```

internal/cli/deployment_list.go
```diff
@@ -127,6 +128,13 @@ func (c *DeploymentListCommand) Run(args []string) int {
 			c.project.UI.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 			return ErrSentinel
 		}
+		if len(resp.Deployments) == 0 {
+			c.project.UI.Output(
+				fmt.Sprintf("No deployments found for application %q", app.Ref().Application),
```
briancain, line 133: small nit, I believe `UI.Output` let's you format strings, so I think you can remove the `fmt.Sprintf` and just do `"your formatted string goes %s", "here"`?

website/content/docs/projects/index.mdx
```diff
@@ -113,3 +113,11 @@ applications inside a single project. For monorepo users, a project MUST
...
 _unique_ repository, so monorepo users may create multiple projects pointing
 to the same repository.
+
+One simple example of this one-to-many pattern is the [waypoint-examples](https://github.com/hashicorp/waypoint-examples/)
+repository. The repository contains many waypoint projects, and each project
+in a sub-dir of the `waypoint-examples` repo maps to this single repository.
+
+Support for this is seen in the gitops workflow, where a user specifies
+the parent repo to map to the project and can optionally specify
+a [Git source path](https://www.waypointproject.io/commands/project-apply#git-path).
```
briancain, line 123:
```suggestion
One simple example of this one-to-many pattern is the [waypoint-examples](https://github.com/hashicorp/waypoint-examples/)
repository. The repository contains many Waypoint projects, and each project
in a sub-dir of the `waypoint-examples` repo maps to this single repository.

Support for this is seen in the GitOps workflow, where a user specifies
the parent repo to map to the project and can optionally specify
a [Git source path](https://www.waypointproject.io/commands/project-apply#git-path).
```

.changelog/2413.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli: Update base commands to default to all apps within the project
```
briancain, line 2:
```suggestion
cli: Update base commands to default to all apps within the project if project has more than one application
```

I think in the single app case this is still true, but the big change is now if your project has multiple apps you're no longer required to specify an app flag.

internal/cli/artifact_list.go
briancain: This is true for remote `waypoint up`s too.

```
brian@localghost:waypoint λ waypoint up -remote go-gitops-0/go                                                                                                                                                                                                                                                      ±[PROJECT-OPS]
! This command requires a targeted project.
brian@localghost:waypoint λ waypoint up -remote -project=go-gitops-0 -app=go                                                                                                                                                                                                                                        ±[PROJECT-OPS]
! failed to load config: A Waypoint configuration file (waypoint.hcl) is required but wasn't found.
  The command can use configuration stored on the server by adding project/app
  as a final argument on the command line.
  For instance: `waypoint deployment list webapp/web` for the app 'web' in
  the project 'webapp'
brian@localghost:waypoint λ waypoint up -remote -project=go-gitops-0                                                                                                                                                                                                                                                ±[PROJECT-OPS]
! failed to load config: A Waypoint configuration file (waypoint.hcl) is required but wasn't found.
  The command can use configuration stored on the server by adding project/app
  as a final argument on the command line.
  For instance: `waypoint deployment list webapp/web` for the app 'web' in
  the project 'webapp'
```

internal/cli/deployment_list.go
briancain: I'm also thinking for the json case, for all projects, instead of returning two separate json structures separated by newlines, we should probably collect and wrap them into a single array for each item listed? Maybe that's a future enhancement 🤔

briancain (changes requested):
🎉 noice! Looking good. My main issue was getting `project/app` targeting for remote projects working again, I think it's just an issue with the new `WithMultipleApp` CLI option and it'll work for all commands that use that option.

internal/cli/deployment_list.go
krantzinator: Hm, we could do that by storing the json on the command, returning nothing in the `DoApp` if flag.json, and then returning the collected json outside of that loop?

PR #2416, opened 2021-10-05
Enabling dynamic config files

izaaklauer:
Prior to this, it appears that dynamic config targeting files did not work - our logic was treating all dynamic config as env vars, and not respecting [`name_is_path` ](https://github.com/hashicorp/waypoint/blob/33fa916c8fa2e295e8913efba91de7e086760cd6/internal/server/proto/server.proto#L3433).

This enables stanzas like this to work:

```
 config {
   file = {
    "/shared-config/nginx.conf" = configdynamic("vault", {
    path = "secret/data/apps/sample-app"
     key = "data/nginx.conf"
    })
   }
 }
```

I can't claim to have deeply understood all of the surrounding logic, so if there's a better place or way to do this, I'm open to feedback.

This also incidentally adds some better error messages to the vault config-sourcer plugin.

### DIFF
--- a/.changelog/2416.txt
+++ b/.changelog/2416.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+config: Fix dynamic config vars targeting files.
+```
+
--- a/builtin/vault/config_sourcer.go
+++ b/builtin/vault/config_sourcer.go
@@ -2,6 +2,7 @@ package vault
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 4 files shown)


mitchellh (commented):
Only comment I’d have is that `internal/appconfig` is built to be relatively easy (or at least, possible) to test. It’s one of the design goals of the package and separating it from plugins and so on so we can just simulate any behavior. I’d recommend getting a failing test in here.

xiaolin-ninja (approved):

PR #2417, opened 2021-10-05
ui: remove unused import in test

jgwhite:
Makes me wonder if we should have ESLint error on warnings


gregone (approved):

gregone:
erroring on warnings may be a little too much IMO

jgwhite:
> erroring on warnings may be a little too much IMO

fair

PR #2420, opened 2021-10-05
Install ember-test-selectors

gregone:
While thinking about the removal of `ember-cli-page-object`, I noticed we didn't have `ember-test-selectors` installed, so we were shipping all the `data-test-` attributes to production


jgwhite (approved):
Good spot!

PR #2421, opened 2021-10-05
UI: fix config var renaming

### LINKED ISSUE #2339: ui: Renaming a config variable does not remove the original

gregone:
Fixes #2339 

Because the API does an upsert for config vars, and unicity is enforced via variable names, renaming a variable would cause the creation of a new variable. This PR adds logic to compare the updated variable name with the original one, and deletes the initial variable after save if the name has been changed.

### DIFF
--- a/.changelog/2421.txt
+++ b/.changelog/2421.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: fixed config variable duplication when renaming 
+```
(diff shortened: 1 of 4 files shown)


sabrinako (approved):
Looks good to me! 👍

jgwhite:
This PR taught me the word “unicity”! Thank you!

PR #2426, opened 2021-10-05
cli: release cmd tuneups

krantzinator:
Things done:
* removed the `[id]` arg that isn't used
* amend the deployment ID returned in the CLI message to match the type of ID we use in `deployment list`
* fix the logic on whether or not a 🚀 shows in `release list`

### DIFF
--- a/internal/cli/release_create.go
+++ b/internal/cli/release_create.go
@@ -239,7 +239,7 @@ Usage: waypoint release [options]
   Open a deployment to traffic.
 
   This defaults to the latest deployment. Other deployments can be
-  specified by ID.
+  specified by ID using the '-deployment' flag.
 
 ` + c.Flags().Help())
 }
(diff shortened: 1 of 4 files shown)


internal/cli/release_create.go
```diff
@@ -233,12 +234,12 @@ func (c *ReleaseCreateCommand) Synopsis() string {
...
 func (c *ReleaseCreateCommand) Help() string {
 	return formatHelp(`
-Usage: waypoint release [options] [id]
+Usage: waypoint release [options]
 
   Open a deployment to traffic.
 
   This defaults to the latest deployment. Other deployments can be
-  specified by sequence number or long ID.
+  specified by ID.
```
briancain, line 242:
```suggestion
  specified by ID using the '-deployment' flag.
```

Might be worth being explicit for how to do this

briancain (approved):
🌮

Issue #2059, opened 2021-08-17
plugin/k8s: Support sidecar containers


izaaklauer:
Waypoint should allow Kubernetes users to configure and run sidecar containers.

PR #2428, opened 2021-10-05
plugin/k8s: support sidecar containers

xiaolin-ninja:
[Issue](https://github.com/hashicorp/waypoint/issues/2059)
[RFC](https://docs.google.com/document/d/14vTndgn4vCvxZ1ZRcyxj9A2nzWSyUdwo5LOYmQfks1Q)

cc: @izaaklauer 

This PR introduces a new feature, allowing advanced Kubernetes users to add sidecar containers to their apps by adding a new “sidecar” stanza to the k8s plugin config.

This PR introduces several new structs, and breaks backwards compatibility by refactoring the "Ports" config.

See usage examples in RFC.

### DIFF
--- a/.changelog/2428.txt
+++ b/.changelog/2428.txt
@@ -1,5 +1,5 @@
 ```release-note:feature
-plugin/k8s: Allows advanced users to add sidecar containers to apps using the k8s plugin config.
+plugin/k8s: Allows users to add sidecar containers to apps using the k8s plugin config.
 ```
 
 ```
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -143,7 +143,7 @@ func (p *Platform) ConfigSet(config interface{}) error {
 			),
 		))
 		if err != nil {
-			return status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid kubernetes platform plugin config: %s", err.Error()))
+			return status.Errorf(codes.InvalidArgument, "Invalid kubernetes platform plugin config: %s", err.Error())
 		}
 	}
 
@@ -621,7 +621,7 @@ func (p *Platform) resourceDeploymentCreate(
 
 	// App container must have some kind of port
 	if len(appContainerSpec.Ports) == 0 {
-		ui.Output(fmt.Sprintf("No ports defined - defaulting to http on port %d", DefaultServicePort), terminal.WithWarningStyle())
+		ui.Output("No ports defined - defaulting to http on port %d", DefaultServicePort, terminal.WithWarningStyle())
 		appContainerSpec.Ports = append(appContainerSpec.Ports, &Port{Port: DefaultServicePort, Name: "http"})
 	}
 


builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
+			key := strings.Split(k, "_")
+			resourceName := corev1.ResourceName(key[1])
 
 			quantity, err := k8sresource.ParseQuantity(v)
 			if err != nil {
-				return err
+				return nil,
+					status.Errorf(codes.InvalidArgument, "failed to parse resource %s to k8s quantity: %s", v, err)
 			}
 			resourceLimits[resourceName] = quantity
```
briancain, line 429: The `||` or change means we never set `resourceRequests`, just `resourceLimits`

builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
 
-	_, cpuLimit := resourceLimits[corev1.ResourceCPU]
-	_, cpuRequest := resourceRequests[corev1.ResourceCPU]
-
-	if p.config.AutoscaleConfig != nil && !(cpuLimit || cpuRequest) {
-		ui.Output("For autoscaling in Kubernetes to work, a deployment must specify "+
-			"cpu resource limits and requests. Otherwise the metrics-server will not properly be able "+
-			"to scale your deployment.", terminal.WithWarningStyle())
-	}
-
```
briancain, line 479: This shouldn't of been deleted. Notice we still need to check the cpu limit and requests check.

builtin/k8s/platform.go
```diff
@@ -771,29 +839,6 @@ func (p *Platform) resourceDeploymentCreate(
...
-				ui.Output("Events for %s", name, terminal.WithHeaderStyle())
-				for _, ev := range events.Items {
-					if ev.Type == "Normal" {
-						continue
-					}
-					ui.Output("  %s: %s (%s)", ev.Type, ev.Message, ev.Reason)
-				}
-			}
-		}
-
```
briancain, line 796: Hm, I think all of this (686, then 705 to here) was accidentally deleted? Same with the following diff with `deleteGrace` and deleting gracefully in the destroy func.

builtin/k8s/platform.go
```diff
@@ -1245,7 +1278,8 @@ type Config struct {
 
 	// A full resource of options to define ports for your service running on the container
 	// Defaults to port 3000.
-	Ports []map[string]string `hcl:"ports,optional"`
+	// Todo(XX): add in HCL parse logic to warn if defining ports the old way, & update docs
+	Ports []*Port `hcl:"port,block"`
```
briancain, line 1282: Agreed, we should get this sorted out in this PR if we can. I know some people are using this already so removing it will break their waypoint.hcl. Maybe we can get an auto-upgrade a `ports` stanza to `port` structs and display a depreciation warning?

builtin/k8s/platform.go
```diff
@@ -1309,6 +1343,30 @@ type AutoscaleConfig struct {
 type Pod struct {
 	SecurityContext *PodSecurityContext `hcl:"security_context,block"`
 	Container       *Container          `hcl:"container,block"`
+	Sidecars        []*SidecarContainer `hcl:"sidecar,block"`
+}
+
+// SidecarContainer describes the configuration for the sidecar container
+type SidecarContainer struct {
```
briancain, line 1350: Hm, rather than having a `SidecarContainer` struct, we should just move all of these over to the existing `Container` struct. I don't see anything specific to sidecars that requires it to be separate.

builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
+		p.config.Probe,
+		p.config.ProbePath,
+		p.config.CPU,
+		p.config.Memory,
+		p.config.Resources,
+		command,
+		args,
+		p.config.ScratchSpace,
+		volumes,
+		log,
```
briancain, line 613: If we update `SidecarContainer` struct to just be `Container`, we can initialize a container struct before this and just pass it through rather than having a lot of params.

builtin/k8s/platform.go
```diff
@@ -669,6 +739,13 @@ func (p *Platform) resourceDeploymentCreate(
 
 	dc := clientSet.AppsV1().Deployments(ns)
 
+	// TODO(izaak) delete me
+	j, err := json.Marshal(deployment)
+	if err != nil {
+		return fmt.Errorf("Failed to marshal deploy json")
+	}
+	log.Info(fmt.Sprintf("Deployment json \n\n %s \n\n", j))
```
briancain, line 747: Delete!

builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
+	for k, v := range deployConfig.Env() {
+		envVars[k] = v
+	}
+
+	// Check autoscaling
+	if p.config.AutoscaleConfig != nil && p.config.CPU == nil {
+		ui.Output("For autoscaling in Kubernetes to work, a deployment must specify "+
+			"cpu resource limits and requests. Otherwise the metrics-server will not properly be able "+
+			"to scale your deployment.", terminal.WithWarningStyle())
+	}
```
briancain, line 575: This can still be set through the `resource_limits` and `resource_requests` map, so I'd recommend removing this and putting it back where we create the container

.changelog/2416.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+config: Fix dynamic config vars targeting files.
+```
+
```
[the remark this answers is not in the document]
briancain: Once we figure out how we want to handle the breaking change, it should be included in this changelog too as `release-note:breaking-change`.

builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
+		tag := splitImage[1]
+		if tag == "latest" {
+			// Always pull a latest image, so the k8s workers don't use their local cache and ignore new changes.
+			pullPolicy = corev1.PullAlways
+		} else {
+			// A tag is present, we can use k8s worker caching
+			pullPolicy = corev1.PullIfNotPresent
+		}
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "image %s is in an invalid format", image)
```
briancain, line 330:
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "image %q is in an invalid format", image)
```

builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
+				sidecarConfig.Probe,
+				sidecarConfig.ProbePath,
+				sidecarConfig.CPU,
+				sidecarConfig.Memory,
+				sidecarConfig.Resources,
+				sidecarConfig.Command,
+				sidecarConfig.Args,
+				p.config.ScratchSpace,
+				volumes,
+				log,
```
briancain, line 646: Same here, we can pass through a `Container` struct and simplify this (it should already be a Container struct so we won't have to create it again like we would for the app container earlier)

briancain (changes requested):
Yo! This looks like a good start 🎉 🚗 I haven't given this a shot yet, so giving some early code review feedback right now. Let me know if you have any questions.

Some high level things, it looks like maybe the PR for [enabling dynamic configs](https://github.com/hashicorp/waypoint/pull/2416) got mixed into this and the commits got duplicated in the history? Rebasing to fix that up would be good to do before merge. Then I think some of Evans demo fixups and UI improvements got deleted which should go back as well

builtin/k8s/platform.go
```diff
@@ -296,318 +294,392 @@ func (p *Platform) resourceDeploymentStatus(
...
+	} else if len(splitImage) == 2 {
+		tag := splitImage[1]
+		if tag == "latest" {
+			// Always pull a latest image, so the k8s workers don't use their local cache and ignore new changes.
+			pullPolicy = corev1.PullAlways
+		} else {
+			// A tag is present, we can use k8s worker caching
+			pullPolicy = corev1.PullIfNotPresent
+		}
+	} else {
```
briancain, line 329: I'm not sure we should fail here, we can probably let K8S do its thing if the image is wrong and let it surface the error instead. The alternative is getting a really good regex to match container registry URLs in the image name which might be a pain. This is one reason why the integration tests are failing. An image name can have a container registry in front of it:

```shell
# a local registry
localhost:5000/sinatra:latest
# or a live one
docker.io/webapp:1.2.3
```

.changelog/2428.txt
```diff
@@ -0,0 +1,8 @@
+```release-note:feature
+plugin/k8s: Allows advanced users to add sidecar containers to apps using the k8s plugin config.
```
mitchellh, line 2: Nitpick but I don't think we need to say "advanced users" here, just "Allows users to".

builtin/k8s/platform.go
```diff
@@ -97,6 +99,57 @@ func (p *Platform) DefaultReleaserFunc() interface{} {
...
+			validation.Field(&c.StaticEnvVars,
+				validation.Empty.When(container.StaticEnvVars != nil).Error(fmt.Sprintf(containerOverlayErrStr, "StaticEnvVars")),
+			),
+			validation.Field(&c.ServicePort,
+				validation.Empty.When(len(container.Ports) > 0).Error("Cannot define both 'service_port' and container 'port'. Use"+
+					" container 'port' multiple times for configuring multiple container ports"),
+			),
+		))
+		if err != nil {
+			return status.Errorf(codes.InvalidArgument, fmt.Sprintf("Invalid kubernetes platform plugin config: %s", err.Error()))
```
briancain, line 146: Small nit: since it's already `Errorf` you can just do this, I think 🤔 
```suggestion
			return status.Errorf(codes.InvalidArgument, "Invalid kubernetes platform plugin config: %s", err.Error())
```

builtin/k8s/platform.go
```diff
@@ -476,132 +499,230 @@ func (p *Platform) resourceDeploymentCreate(
...
+	if err != nil {
+		return status.Errorf(codes.InvalidArgument, "Failed to parse container config: %s", err)
 	}
 
-	// Update the deployment with our spec
-	deployment.Spec.Template.Spec = corev1.PodSpec{
-		Containers: []corev1.Container{container},
+	// App container must have some kind of port
+	if len(appContainerSpec.Ports) == 0 {
+		ui.Output(fmt.Sprintf("No ports defined - defaulting to http on port %d", DefaultServicePort), terminal.WithWarningStyle())
```
briancain, line 624: I think ui.Output should also let you format without `fmt.Sprintf` if I remember right

```suggestion
		ui.Output("No ports defined - defaulting to http on port %d", DefaultServicePort, terminal.WithWarningStyle())
```

briancain (approved):
Just a minor comment from me, but these updates look great! 🎉

PR #2437, opened 2021-10-06
Improve documentation for path-based hcl functions

izaaklauer:
In practice, the file hcl functions require absolute paths, so they are really only useful when combined with an hcl variable that gives a relative path (like path.app). I struggled for a while to find a way to make `filebase64` useful before stumbling into the answer.

This will help point users in the right direction.

Screen Shot 2021-10-07 at 11 10 15 AM


internal/config/funcs/doc.go
```diff
@@ -19,9 +19,9 @@ func Docs() map[string]string {
 		"dirname":         "return all except the last component of the string interpreted as a filesystem path",
 		"distinct":        "removes any duplicate elements from a list",
 		"element":         "return the value at given numeric index in the list",
-		"file":            "return the data in the file located at the given path",
-		"filebase64":      "return the data in the file located at the given path after base64 decoding it",
-		"fileexists":      "indicate if a file exists at the given path",
+		"file":            "return the data in the file located at the given path. This can be paired with the `${path.app}` variable to get a path relative to your app folder, i.e. `filebase64(\"${path.app}/file.txt\") string`\"",
+		"filebase64":      "return the data in the file located at the given path after base64 decoding it. This can be paired with the `${path.app}` variable to get a path relative to your app folder, i.e. `filebase64(\"${path.app}/file.txt\") string`",
+		"fileexists":      "indicate if a file exists at the given path. This can be paired with the `${path.app}` variable to get a path relative to your app folder, i.e. `filebase64(\"${path.app}/file.txt\") string`\"",
```
briancain, line 24: I think there's a trailing `\"` that's not needed here...I think we can remove `string` too and move the backtick to the end of the function example?

PR #2444, opened 2021-10-07
`waypoint user modify` command that will not modify a user returns help.

izaaklauer:
Previous behavior:

Running `modify` with no flags returned success. This was confusing for people who expected this to return help text.
```
$ waypoint user modify
User modification successful.
```

Running `modify` with flags to specify (but not modify) a user also returned `user modification successful`, even though no real modifications were performed.
```
$ waypoint user modify -username=foo
User modification successful.
```

As of this change, `waypoint user modify` without a flag that will do a modification (i.e. `-new-username` or `-display-name`) will return an error and help text, similar to other commands like `project inspect`:

```
$ ./waypoint user modify
! At least one user modification flag must be specified...
  
  Usage: waypoint user modify [options]
  
    Modify details about a user.
  
    Some details such as username and display name may be updated.
    Use "waypoint user inspect" to see the current attributes for a user.
  
  Global Options:
  
    -app=<string>
        App to target. Certain commands require a single app target for Waypoint
        configurations with multiple apps. If you have a single app, then this
        can be ignored.
  
    -plain
        Plain output: no colors, no animation. The default is false.
  
    -workspace=<string>
        Workspace to operate in.
  
  Command Options:
  
    -display-name=<string>
        The display name for a user. If this is set, this is used in some places
        in the CLI and UI. This does not have to be unique.
  
    -new-username=<string>
        Set a new username for this user. This must be unique to the server.
  
    -username=<string>
        The user to modify. This defaults to the currently logged in user.
```

```
$ ./waypoint user modify -username=bar
! At least one user modification flag must be specified...
  
  Usage: waypoint user modify [options]
  
    Modify details about a user.
  
    Some details such as username and display name may be updated.
    Use "waypoint user inspect" to see the current attributes for a user.
  
  Global Options:
  
    -app=<string>
        App to target. Certain commands require a single app target for Waypoint
        configurations with multiple apps. If you have a single app, then this
        can be ignored.
  
    -plain
        Plain output: no colors, no animation. The default is false.
  
    -workspace=<string>
        Workspace to operate in.
  
  Command Options:
  
    -display-name=<string>
        The display name for a user. If this is set, this is used in some places
        in the CLI and UI. This does not have to be unique.
  
    -new-username=<string>
        Set a new username for this user. This must be unique to the server.
  
    -username=<string>
        The user to modify. This defaults to the currently logged in user.
```

### DIFF
--- a/.changelog/2444.txt
+++ b/.changelog/2444.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-cli/user-modify: Return help on malformed command
+cli: Return help on malformed command
 ```


internal/cli/user_modify.go
```diff
@@ -50,11 +50,20 @@ func (c *UserModifyCommand) Run(args []string) int {
...
 		user.Username = v
+		willModify = true
 	}
 	if v := c.flagDisplay; v != "" {
 		user.Display = v
+		willModify = true
+	}
+
+	if !willModify {
+		c.ui.Output("At least one user modification flag must be specified...\n\n"+
```
catsby, line 64:
```suggestion
		c.ui.Output("at least one user modification flag must be specified"+
```

Nit: After a quick check of `internal/cli/` I believe "most" of our CLI errors have a lowercase error message and don't have the `...` or new lines, so I suggest we modify this to be consistent.

izaaklauer:
In a related ux nitpick, I was really surprised to find that `-username` _selects_ a user, but `-display-name` modifies a user. I was expecting to be able to use `-display-name` to find a user (makes sense that I can't - display names aren't globally unique).

It feels to me like the username should be an argument, and `-display-name` should be `-new-display-name`.

krantzinator (approved):
🎉

.changelog/2444.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli/user-modify: Return help on malformed command
```
mitchellh, line 2: I think our prefix is usually just `cli:` and specify the "user modify" in the text

briancain (approved):
The go tests might be failing from a flake, rerunning them before merge would be good! 👍🏻

Issue #2381, opened 2021-09-29
ECS multiple waypoint-entrypoint and odd behaviors


marcotuna:
**Describe the bug**
When updating/changing the current running ECS container it will start getting weird behaviors on the Waypoint side such as duplicating log lines. After a bit of diagnostics I found out that multiple waypoint-entrypoint is getting launched
```
user@machine:/home/user/project/server$ waypoint exec sh
/opt/service # ps aux

PID   USER     TIME  COMMAND
    1 root      0:00 /waypoint-entrypoint /waypoint-entrypoint /waypoint-entrypoint /waypoint-entrypoint /opt/service/binary
   15 root      0:00 /waypoint-entrypoint /waypoint-entrypoint /waypoint-entrypoint /opt/service/binary
   24 root      0:00 /waypoint-entrypoint /waypoint-entrypoint /opt/service/binary
   33 root      0:00 /waypoint-entrypoint /opt/service/binary
   40 root      0:46 /opt/service/binary
   52 root      0:00 /bin/sh
   57 root      0:00 ps aux
```
I've attemped to kill one of those but the service would restart.  

**Steps to Reproduce**
I don't know exactly how it happened but I believe it was a mixed of different things.  
I deployed a different `app` to the same `project` as the one I am reporting, one thing I noticed is that the configuration environment variables begun to get mixed between each `app` despite using `waypoint config set -app=app_name -workspace=production`, at the end I decided to delete the waypoint deployment and deploy again. For some reason the `app` that I am describing here with issues (I wasn't touching it) was restarting due to changes in the configuration (which I didn't change for this app) and started to show duplicated, triplicated, ... log lines  

**Expected behavior**
Changing an `app` in the same project shouldn't change the behavior of the other `app` and also, shouldn't start to spawn multiple `waypoint-entrypoint`

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: 0.5.2
* Waypoint Server Platform and Version: `aws/ecs`
* Waypoint Plugin: `aws/ecs`

**Additional context**

izaaklauer:
Hmm, it looks like there there might be two bugs here:

1: It sounds from your experience like the CEB isn't filtering only it's individual `-app` scoped config when it decides to hup the app process.

2: There might be an underlying problem with the CEB hup feature that somehow spawns multiple processes.

We'll look into it, thank you!

marcotuna:
[...]

PR #2447, opened 2021-10-07
Only inject/append Entrypoint if it's not already there

catsby:
Currently the `waypoint-entrypoint` binary is injected into the container and appended to the front of its `Entrypoint` definition (as long as `disable_entrypoint` is `false`). 

In the case of using `docker-pull` in the `build` phase, if the image we pull already has the entrypoint included, this will result in a chain of entrypoints _as well as additional layers in the image_ (see #2381).

In this PR, if a container already includes an entry for `/waypoint-entrypoint` in its `Entrypoint` definition, then don't inject it again. We assume that if it is already included in the `Entrypoint` definition, that the binary also exists in the container. 

Fixes https://github.com/hashicorp/waypoint/issues/2381


internal/pkg/epinject/ociregistry/server.go
```diff
@@ -600,8 +600,19 @@ func (s *Server) updateManifest(w http.ResponseWriter, r *http.Request) {
...
 	})
 
-	config.Config.Entrypoint = append([]string{"/waypoint-entrypoint"}, config.Config.Entrypoint...)
-	s.Logger.Debug("injected entrypoint", "value", config.Config.Entrypoint)
+	var found bool
+	for _, entryPoint := range config.Config.Entrypoint {
+		if entryPoint == "/waypoint-entrypoint" {
+			found = true
+			s.Logger.Debug("waypoint-entrypoint found, not injecting")
+			break
```
catsby, line 608: Intentionally didn't use the new `containsEntrypoint` function here to avoid including a new package (which at one point gave me a circular dependency issue, but that may be resolved)

internal/pkg/epinject/ociregistry/server.go
```diff
@@ -600,8 +600,19 @@ func (s *Server) updateManifest(w http.ResponseWriter, r *http.Request) {
 		Author:    "waypoint",
 	})
 
-	config.Config.Entrypoint = append([]string{"/waypoint-entrypoint"}, config.Config.Entrypoint...)
-	s.Logger.Debug("injected entrypoint", "value", config.Config.Entrypoint)
+	var found bool
+	for _, entryPoint := range config.Config.Entrypoint {
```
evanphx, line 604: Ditto about only checking `Entrypoint[0]` here.

internal/pkg/epinject/epinject.go
```diff
@@ -195,11 +195,5 @@ func withConnectionHelper(c *client.Client) error {
...
 
 func containsEntrypoint(entrypoint []string) bool {
-	for _, ep := range entrypoint {
-		if ep == "/waypoint-entrypoint" {
-			return true
-		}
-	}
-
-	return false
+	return entrypoint[0] == "/waypoint-entrypoint"
```
evanphx, line 198: You probably have to check that entrypoint is at least 1 element to avoid a panic.

internal/pkg/epinject/ociregistry/server.go
```diff
@@ -600,16 +600,7 @@ func (s *Server) updateManifest(w http.ResponseWriter, r *http.Request) {
...
-	for _, entryPoint := range config.Config.Entrypoint {
-		if entryPoint == "/waypoint-entrypoint" {
-			found = true
-			s.Logger.Debug("waypoint-entrypoint found, not injecting")
-			break
-		}
-	}
-
-	if !found {
+	if config.Config.Entrypoint[0] != "/waypoint-entrypoint" {
```
evanphx, line 603: Ditto on the length check.

internal/pkg/epinject/epinject.go
```diff
@@ -185,3 +193,10 @@ func withConnectionHelper(c *client.Client) error {
 
 	return nil
 }
+
+func containsEntrypoint(entrypoint []string) bool {
+	if len(entrypoint) > 0 {
+		return entrypoint[0] == "/waypoint-entrypoint"
+	}
+	return false
```
mitchellh, line 201: This is fine but could also just do:

```go
return len(entrypoint) > 0 && entrypoint[0] == "/waypoint-entrypoint"
```

mitchellh (approved):
I actually thought of this back in 0.1 days but thought that the string comparison would be too weird. I guess the pros outweigh the cons and the likelihood someone names their binary `waypoint-entrypoint` is very low :)

internal/pkg/epinject/epinject.go
catsby: I like it, added in 3ba1a5f5, thanks

PR #2457, opened 2021-10-11
UI: headings refactor

### LINKED ISSUE #1854: Out of order headings

We currently have occurrences of both out-of-order headings and headings that skip levels.

https://user-images.githubusercontent.com/34030/125523753-5e13726b-592e-4322-a8d2-d3d9e20d66f4.mp4

**Any other background?**

We’ve deferred addressing this issue by [excluding certain elements from automated a11y testing](https://github.com/hashicorp/waypoint/blob/v0.4.1/ui/tests/helpers/a11y.ts#L16), and by merit of the fact that we don’t yet have exhaustive test coverage (some pages such as empty states and error pages don’t get audited as part of the suite).

**Describe the solution you’d like**

We should correct the out-of-order headings and remove the exclusion if possible (though there may be other a11y violations that stand in the way of this).

gregone:
Fixes: #1854 

This PR decouples the heading levels (`h1`, `h2`, `h3`, `h4`) from their styling. I'm using `.h1` etc. but also open to more descriptive `heading-1` classes if anyone thinks it's clearer. 

Screen Shot 2021-10-11 at 17 55 25


jgwhite (approved):
With this change, I think we can remove `['.card-header']` from the exclude list in `a11y.ts` too.

I’m not completely sold on stuff like `<h3 class="h3">` but I’m sure we’ll get used to it. We can always revisit if a better idea comes up.

sabrinako (approved):
🤠 👍

sabrinako:
> With this change, I think we can remove `['.card-header']` from the exclude list in `a11y.ts` too.
> 
> I’m not completely sold on stuff like `<h3 class="h3">` but I’m sure we’ll get used to it. We can always revisit if a better idea comes up.

I agree about classes as `h#` but can't think of a better solution off the top of my head. I'm happy with it for now until we think of something else!

johncowen:
👏 I saw this flying past and thought I'd just mention that Consul pretty much uses this approach also if that helps with the 'feel' of this.

We use HTML tags to mean something in the markup, but then use similarly named CSS classes (we use something ever so slightly different thats not classes but for the sake of it I'm just saying classes) so they look like something else using the styles - so exactly the same approach as what you have here.

Naming wise we use a `h100, h200` etc etc scheme for when we get one of those inevitable/sods law 'in-between' headers, and means we can make a `h150` if we need to. We also have a set of `p100` etc for a similar approach for body copy.

jgwhite:
We think the original issue is likely moot following the 0.7 redesign, and therefore we can close this.

briancain:
@gregone I'm going to go ahead and close this!

PR #2459, opened 2021-10-11
Generate and include the task launcher docs

evanphx:
The new OnDemand Runner work is built on plugin-specific task types. That ODR config includes the ability to specify plugin config, which corresponds to the task config for the plugin type.

This includes the docs for Kubernetes, Docker, and ECS task types for users to configure.

### DIFF
--- a/website/content/partials/components/task-aws-ecs.mdx
+++ b/website/content/partials/components/task-aws-ecs.mdx
@@ -28,7 +28,7 @@ These parameters are used in the [`use` stanza](/docs/waypoint-hcl/use) for this
 
 Cluster name to place On-Demand runner tasks in.
 
-ECS Cluster to place On-Demand runners in. This defaults to the cluster used by the waypoint server.
+ECS Cluster to place On-Demand runners in. This defaults to the cluster used by the Waypoint server.
 
 - Type: **string**
 - **Optional**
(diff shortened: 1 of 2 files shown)


website/content/partials/components/task-docker.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+### Optional Parameters
+
+These parameters are used in the [`use` stanza](/docs/waypoint-hcl/use) for this plugin.
+
+#### binds
+
+A 'source:destination' list of folders to mount onto the container from the host.
+
+A list of folders to mount onto the container from the host. The expected format for each string entry in the list is `source:destination`. So for example: `binds: ["host_folder/scripts:/scripts"].
```
krantzinator, line 54: This probably has to be fixed in the source
```suggestion
A list of folders to mount onto the container from the host. The expected format for each string entry in the list is `source:destination`. So for example: `binds: ["host_folder/scripts:/scripts"]`.
```

website/content/partials/components/task-kubernetes.mdx
```diff
@@ -0,0 +1,76 @@
...
+Pull policy to use for the task container image.
+
+- Type: **string**
+- **Optional**
+
+#### image_secret
+
+Name of the Kubernetes secret to use for the image.
+
+This references an existing secret; waypoint does not create this secret.
```
krantzinator, line 47: Another source-edit
```suggestion
This references an existing secret; Waypoint does not create this secret.
```

website/content/partials/components/task-aws-ecs.mdx
```diff
@@ -0,0 +1,114 @@
...
+AWS region to use. Defaults to the region used for the Waypoint Server.
+
+- Type: **string**
+- **Optional**
+
+#### security_group_id
+
+Security Group ID to place the On-Demand Runner task in.
+
+Security Group ID to place the On-Demand Runner task in. This defaults to the security group used for the waypoint server.
```
briancain, line 85: There's a few lower case `waypoint`s here, but I think you have to fix it on the plugin and then regenerate the docs

briancain (approved):
🎉

PR #2477, opened 2021-10-13
internal/cli: Ensure project target is set for AppOptional CLI option

briancain:
Prior to this commit, the `waypoint status` CLI would set the project
target if it was inside a Waypoint project, using the AppOptional CLI
option for initialization. This commit ensures the project target is set
if its invoked inside a project dir.


mitchellh (approved):

krantzinator (approved):
Neat!

PR #2479, opened 2021-10-13
init: disable plugin check for now

mitchellh:
With the introduction of input variables, it is possible to use input
variables in a way that `waypoint init` cannot actually load the
configuration. This is because we don't actually render input variables
in the CLI (because we may not have all values).

The solution is to move this into a runner operation. We've already
started this process with the work we did in 0.6 around InitOp, but we
don't yet call this from the CLI.

For now, for 0.6, we just disable this init check. This was a safety
check anyways and has no practical implications so we can turn it off.
We'll turn it back on once we switch to remote operations.

### DIFF
--- a/internal/cli/init.go
+++ b/internal/cli/init.go
@@ -151,7 +151,13 @@ func (c *InitCommand) Run(args []string) int {
 		c.validateConfig,
 		c.validateServer,
 		c.validateProject,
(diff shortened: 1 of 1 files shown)


krantzinator (approved):
Can you open a corresponding issue so we don't forget?

PR #2483, opened 2021-10-13
internal/ceb/ssh: wait on exit until stdout copy is complete

mitchellh:
**Do not merge before 0.6 because this changes non-test code.**

There is a race condition between the command exiting and the copy to
stdout being complete. On command completion, we want to wait for the
stdout copy to finish before actually exiting.

This also fixes our second most flaky test, which I believe was not in fact false
positive flaky, but was REAL positive flaky; I think this was a bug.

### DIFF
--- a/internal/ceb/ssh/server.go
+++ b/internal/ceb/ssh/server.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"strconv"
 	"syscall"
+	"time"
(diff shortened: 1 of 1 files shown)


mitchellh:
Also noting for me: we may want to backport this 0.6.1 once we verify this is correct, cause this is a bug.

PR #2486, opened 2021-10-14
Increase failure threshold from 5 to 30.

izaaklauer:
Currently, by default, waypoint is timing out the deployment if the pod isn't ready in under a minute. This often isn't enough time, especially for an initial deployment that needs to download a docker image, or for large apps that take a while to initialize. Our timeout in the previous release was 10 minutes.

Here's what that looks like (a simple next.js app deployed for the first time to eks, care of @jgwhite):
image (1)

Increasing the failure threshold to 30 will increase the timeout to about 5 minutes, which should be enough time the majority of apps to pull and start. Users with quick-starting apps (or _very_ slow apps) can change this in their k8s config by modifying `pod.container.failure_threshold`: https://tip.waypointproject.io/plugins/kubernetes#pod-container-probe-failure_threshold

This is a tough balance - it isn't uncommon to deploy an app that won't ever become healthy, and users often won't start investigating until waypoint times out and reports the failure. If we time out sooner, users can start debugging sooner, and they would appreciate that cycle being faster.

I think that waypoint giving up too early on a working app is a pretty bad failure mode though - it isn't obvious what the problem is, or how to fix it by changing config. I suspect a new user is likely to stop using waypoint if it repeatedly times out on deploying their app.

### DIFF
--- a/builtin/k8s/platform.go
+++ b/builtin/k8s/platform.go
@@ -560,6 +560,7 @@ func configureContainer(
 			Handler:             handler,
 			InitialDelaySeconds: initialDelaySeconds,
 			TimeoutSeconds:      timeoutSeconds,
+			FailureThreshold:    failureThreshold,
 		}
 	}
 


builtin/k8s/platform.go
```diff
@@ -560,6 +560,7 @@ func configureContainer(
 			Handler:             handler,
 			InitialDelaySeconds: initialDelaySeconds,
 			TimeoutSeconds:      timeoutSeconds,
+			FailureThreshold:    failureThreshold,
```
izaaklauer, line 563: This was a bug - we were getting the default failureThreshold for readiness, which was `3`. This is why Jamie's log shows a 40s timeout instead of 60s like you would expect from our defaults.

PR #2488, opened 2021-10-14
internal/server: fix flaky test TestServiceGetLogStream_depPlugin

mitchellh:
The issue here is that we can't force a flush and wait for a recv on the
client/server. Instead of sleeping 100ms and hoping for the best, we use
Eventually with a timeout to accumulate log entries. This should make
this deterministic in even the slowest practical cases.

### DIFF
--- a/internal/server/singleprocess/service_entrypoint.go
+++ b/internal/server/singleprocess/service_entrypoint.go
@@ -233,6 +233,7 @@ func (s *service) EntrypointLogStream(
 						return err
 					}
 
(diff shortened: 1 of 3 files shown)


evanphx (approved):

mitchellh:
There's actually another flake present in this test I haven't figured out yet. I'm sure this alleviates some issues but I'm still hunting down more...

mitchellh:
I found it, I think it was a real bug. See my comment @evanphx

PR #2490, opened 2021-10-14
Increase ECS ODR memory

izaaklauer:
1024 isn't enough to build https://github.com/hashicorp/waypoint-examples/tree/main/kubernetes/java


krantzinator (approved):
2048!

PR #2519, opened 2021-10-19
remove ember-in-viewport

gregone:
The code that was using this addon (logs components) was removed or updated to work without it in https://github.com/hashicorp/waypoint/pull/2356


jgwhite (approved):
Woohoo! Good spot

PR #2527, opened 2021-10-20
Add size-limit action

gregone:
~~Uses the [size-limit](https://github.com/marketplace/actions/size-limit-action) action to report asset sizes.~~

Now uses ember-asset-size instead


sabrinako (approved):
Nice! LGTM but I haven't worked with GH actions much so I'd wait for @jgwhite's approval too

jgwhite:
@sabrinako don’t sell yourself short. Your approval is more than enough!

jgwhite (approved):
This rules 🥳

PR #2532, opened 2021-10-20
core/app_build: Fix panic where Waypoint plugin missing AccessInfoFunc

briancain:
This pull request fixes an issue where a compiled Waypoint plugin fails to
register an AccessInfoFunc(). Instead of directly passing nil, causing a
panic during build, we look to see if the function is defined first. If
not, return the proper errors.

### DIFF
--- a/internal/core/app_build.go
+++ b/internal/core/app_build.go
@@ -155,7 +155,7 @@ func (op *buildOperation) Do(ctx context.Context, log hclog.Logger, app *App, _ 
 
 	// If there is a registry defined and it implements RegistryAccess...
 	if op.Registry != nil {
-		if ra, ok := op.Registry.Value.(component.RegistryAccess); ok {
+		if ra, ok := op.Registry.Value.(component.RegistryAccess); ok && ra.AccessInfoFunc() != nil {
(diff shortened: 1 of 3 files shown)


internal/core/app_build.go
```diff
@@ -176,6 +176,15 @@ func (op *buildOperation) Do(ctx context.Context, log hclog.Logger, app *App, _
...
 			}
+		} else {
+			if ra != nil && ra.AccessInfoFunc() == nil {
+				return nil, status.Error(codes.Internal, "The plugin requested does not "+
+					"define an AccessInfoFunc() in its Registry plugin. This is an internal "+
+					"error and should be reported to the author of the plugin.")
+			} else {
+				return nil, status.Error(codes.Internal, "The given build operation value "+
+					"is not a component.RegistryAccess type.")
+			}
```
evanphx, line 187: Pretty sure this is wrong because it will mean that any plugin that doesn't implement AccessInfoFunc fails to build, which isn't what you want. Basically you should try to run the build anyway.

evanphx (changes requested):

PR #2548, opened 2021-10-21
internal/runner: unit test for accept timeout

mitchellh:
This adds a unit test to verify accept timeouts work (they do).

At the same time, I modified our accept timeout to not cancel the entire
runner, and to avoid some data races that we had. In practice, neither
was a real issue due to the way we use accept timeouts currently. But
stepping back, I felt this was the correct behavior. Now, accept
timeouts are timeouts _per-accept_ call and does not cancel the _entire
runner_.

### DIFF
--- a/internal/runner/accept.go
+++ b/internal/runner/accept.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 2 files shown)


internal/runner/accept.go
```diff
@@ -121,27 +127,19 @@ func (r *Runner) accept(ctx context.Context, id string) error {
...
 
-			// If we raced to this point and timed out BUT the job was accepted, don't cancel
-			// the context.
-			if accepted {
-				return
-			}
+			// Mark that we canceled
+			atomic.StoreInt32(&canceled, 1)
 
-			canceled = true
```
mitchellh, line 142: This was a data race in the prev code. `canceled` read/write was not protected by a mutex below. Not a huge issue and never caused any issue in practice, but worth noting it isn't strictly correct.

briancain (approved):

PR #2549, opened 2021-10-21
website: bump deps to fix print styles

zchsh:
This PR bumps dependencies in order to address `@media print` style issues. Specifically:
- Bumps to latest `@hashicorp/react-consent-manager`, which hides itself in print media
- Bumps to latest `@hashicorp/react-subnav`, which hides itself in print media
- Removes now redundant project-level CSS intended to achieve the same.

Underlying component changes were made in https://github.com/hashicorp/react-components/pull/406. Also bumps to updated consent-manager with changes here: https://github.com/hashicorp/react-components/pull/407.


zchsh:
### Before

before

### After

after

zchsh:
I _believe_ the failing `website-mdx` test here can be ignored... I think it's detecting a false positive because `package-lock.json` seems to be modified during CircleCI job:

CleanShot 2021-12-03 at 09 57 41@2x

To avoid this, I think in https://github.com/hashicorp/waypoint/blob/e7866945e629fd7094ba485939a28f4e2acea5a0/.circleci/config.yml#L105 we might want to change the last `npm install` to `npm ci`, which will [do a more strict install and avoid writing to package-locks](https://docs.npmjs.com/cli/v6/commands/npm-ci).

briancain:
> I believe the failing website-mdx test here can be ignored...

Yep, no worries on this. This test is mainly to ensure we Waypoint devs are generating our auto-generating docs on any changes. Weird that it failed for you there! But yeah, for website updates like this, feel free to ignore 😄

jescalan (approved):

PR #2562, opened 2021-10-22
ui: avoid loading *all* status reports

### LINKED ISSUE #2561: ui: `ListStatusReports` is returning far more data than we need
**Describe the bug**

For a deployment or release that’s been around for a few days, `ListStatusReports` returns *a lot* of data:

list-status-reports

**Potential solutions**
1. Use `UI_ListDeployments` (currently a little slow)
2. Use `GetLatestStatusReport` (though we’ll have to call this once for each deployment/release).

e3de6f4e5d64e7eca73629bd51ebe786112d5f5c

jgwhite:
## Why the change?

Closes #2561

## What does it look like?

Hopefully identical to before.

## How do I test it?

This is going to require some smoke testing against a real Waypoint server or two.

### DIFF
--- a/.changelog/2562.txt
+++ b/.changelog/2562.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: avoid loading *all* status reports
+```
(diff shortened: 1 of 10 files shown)


gregone (approved):
LGTM. Just a small comment: I wonder if it would make sense to extract the bundle mapping into separate functions to make maintenance (and potentially future unit testing) easier.

sabrinako (approved):
👍

PR #2571, opened 2021-10-22
internal/runner: Set runner state to `stateExit` when a runner fails to receive configuration

catsby:
With configuration updates, [Runners will launch 2 go-routines][routines] handling project configuration changes, one to receive new configuration and another to act on it[^1], and then [blocks][blocks] until the acting go-routine sets a specific state or exits. If the receiving go-routine fails to receive or otherwise errors on a configuration, that go-routine will log an error and return, which leaves the acting go-routine just hanging, and the main go-routine still blocking. As a result, tasks/pods/containers can pile up and be left in a stuck state and not shutting down. In the case of AWS ECS, this results in a growing number of ECS tasks sitting idle.

Due to the way polling works, if the `poll-interval` is frequent enough such that the server cannot create on-demand runners fast enough to process the job before the next interval, the job could be overwritten before the on-demand runner originally launched for it even gets a chance to do the work. As a result the tasks errored with this message:

> [ERROR] waypoint.runner.agent.runner.config_recv: error receiving configuration, exiting: err="rpc error: code = FailedPrecondition desc = no pending job for this on-demand runner. A pending job must be registered prior to registering the runner."

In this PR, if the receiving go-routine fails/errors, set the Runner state to `stateExit`, which will unblock the main go-routine and cause the runner to exit.  

A new error message is added, in addition to the one above:

> waypoint.runner.agent: error starting runner: err="rpc error: code = Internal desc = early exit while waiting for first config processing"

Note that the problem of jobs being created and overwritten faster than runners can process them remains, however we no longer leak tasks/containers/pods.

Also note that I've only reproduced this on AWS ECS, but it's assumed to affect other platforms, though those platforms may have builtin processes for reaping them[^2]

[routines]: https://github.com/hashicorp/waypoint/blob/7055c149ce8927e309f82f99bc401c2bc5b18714/internal/runner/runner.go#L205-L206
[blocks]: https://github.com/hashicorp/waypoint/blob/7055c149ce8927e309f82f99bc401c2bc5b18714/internal/runner/runner.go#L210-L212
[^1]: basically
[^2]: possibly

### DIFF
--- a/internal/runner/config.go
+++ b/internal/runner/config.go
@@ -16,17 +16,15 @@ import (
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 )
 
-var (
-	// appConfigRefreshPeriod is the interval between checking for new
-	// config values. In a steady state, configuration NORMALLY doesn't
-	// change so this is set fairly high to avoid unnecessary load on
-	// dynamic config sources.
-	//
-	// NOTE(mitchellh): In the future, we'd like to build a way for
-	// config sources to edge-trigger when changes happen to prevent
-	// this refresh.
-	appConfigRefreshPeriod = 15 * time.Second
-)
+// appConfigRefreshPeriod is the interval between checking for new
+// config values. In a steady state, configuration NORMALLY doesn't
+// change so this is set fairly high to avoid unnecessary load on
+// dynamic config sources.
+//
+// NOTE(mitchellh): In the future, we'd like to build a way for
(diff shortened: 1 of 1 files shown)


mitchellh:
With #3087, this fix was reverted because the config stream exiting no longer exits the runner. This PR didn't have a regression test so I can't be 100% sure that is correct, so I'm just making this note in case my future self or some future Waypointer sees this behavior again and needs a hint at when/where/why that might've changed.

PR #2597, opened 2021-10-26
serverinstall/nomad: Automatically setup Consul service on install

briancain:
Prior to this commit, the consul service integration was an optional
opt-in service that users could request when using the server install
helper CLI with Nomad. This commit changes that by instead setting up
the Consul service automatically, given how many Nomad workflows require
Consul for any kind of networking.


internal/serverinstall/nomad.go
```diff
@@ -219,11 +221,20 @@ func (i *NomadInstaller) Install(
...
-			"This could change in the future if Nomad creates a new allocation for the Waypoint server, \n"+
-			"which would break all existing Waypoint contexts.", terminal.WithWarningStyle(),
-	)
+	if i.config.consulService {
+		s = sg.Add("The CLI has been configured to automatically install a Consul service for\n" +
+			"the Waypoint service backend and ui service in Nomad.")
+		s.Done()
+	}
+
+	s = sg.Add(
```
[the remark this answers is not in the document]
briancain: Done!

internal/cli/install.go
```diff
@@ -116,13 +116,47 @@ func (c *InstallCommand) Run(args []string) int {
...
+	maxRetries := 12 // attempts to retry for a full minute
+	sr := sg.Add("Attempting to make connection to server...")
+	for {
+		log.Info("connecting to the server so we can set the server config", "addr", contextConfig.Server.Address)
+		conn, err = serverclient.Connect(ctx,
+			serverclient.FromContextConfig(contextConfig),
+			serverclient.Timeout(5*time.Second),
+		)
+		if err != nil {
+			c.ui.Output(
```
briancain, line 131: Todo: going to turn this into its own step group

internal/serverinstall/nomad.go
```diff
@@ -1131,6 +1211,50 @@ func (i *NomadInstaller) UpgradeFlags(set *flag.Set) {
 		Target: &i.config.hostVolume,
 		Usage:  "Nomad host volume name.",
 	})
+
+	set.BoolVar(&flag.BoolVar{
+		Name:    "nomad-consul-service",
+		Target:  &i.config.consulService,
+		Usage:   "Create service for Waypoint UI in Consul.",
+		Default: true,
+	})
```
evanphx, line 1220: Does our flag system work with `true` by default bools?

briancain: It does! And `-nomad-consul-service=false` also works for turning it off (note that this flag was introduced prior to this PR)

evanphx: Excellent!

evanphx (approved):

mitchellh (approved):
This is great! We may want to update some docs somewhere to note Consul is required by default and how to use single-node Nomad instead (and the big WARNINGS with that).

briancain:
Sounds good, I think we're overdue for a proper `waypoint install` page along side the [manual run a server docs](https://tip.waypointproject.io/docs/server/run) so I think I'll do a followup PR today to cover that an include info about consul + nomad.

briancain:
Going to backport this, and when I fixup the server install docs, I'll be sure to include a note about server upgrades and the consul requirement change.

PR #2604, opened 2021-10-27
ui: improve mirage handler types

jgwhite:
## Why the change?

Currently, we’re not actually type-checking files in the `mirage` directory, due to this:

https://github.com/hashicorp/waypoint/blob/dddaff9375d713bf0416bf00ad81b4f70e8ce810/ui/tsconfig.json#L33

This is the first in a series of PRs to address type errors in the mirage directory, building up to a PR which will add `mirage/**/*.ts` to the config line above.

## How do I test it?

If CI succeeds then we should be good.

## What’s the next step?

Address type errors in the `mirage/models` directory.


sabrinako (approved):
LGTM! 👍

Issue #2037, opened 2021-08-12
Cannot disable project data source polling with CLI


izaaklauer:
**Describe the bug**
If polling is enabled for a project, it isn't possible to disable it with the CLI

**Steps to Reproduce**
- Clone https://github.com/hashicorp/waypoint-examples/tree/main/docker/ruby

- Run `waypoint init`

- Observe that datasource polling is disabled:

```
$ grpcurl -d '{"project": { "project":"example-ruby"}}' --insecure -H "client-api-protocol: 1,1" -H "Authorization: `waypoint token new`" localhost:9711 hashicorp.waypoint.Waypoint/GetProject

{
  "project": {
    "name": "example-ruby",
    "applications": [
      {
        "name": "example-ruby",
        "project": {
          "project": "example-ruby"
        }
      }
    ]
  }
}
```

- Enable datasource polling with `waypoint project apply -poll -data-source=git -git-url=https://github.com/hashicorp/waypoint-examples -git-path=docker/ruby example-ruby`

- Observe polling is enabled

```
$ grpcurl -d '{"project": { "project":"example-ruby"}}' --insecure -H "client-api-protocol: 1,1" -H "Authorization: `waypoint token new`" localhost:9711 hashicorp.waypoint.Waypoint/GetProject | jq .

{
  "project": {
    "name": "example-ruby",
    "applications": [
      {
        "name": "example-ruby",
        "project": {
          "project": "example-ruby"
        }
      }
    ],
    "remoteEnabled": true,
    "dataSource": {
      "git": {
        "url": "https://github.com/hashicorp/waypoint-examples",
        "path": "docker/ruby"
      }
    },
    "dataSourcePoll": {
      "enabled": true,
      "interval": "30s"
    }
  }
}
```

- Attempt to disable datasource polling with `waypoint project apply -poll=false example-ruby`

- Observe datasource polling is still enabled by re-running the previous `grpcurl`.

Please include any `waypoint.hcl` files if applicable, as well as a
[GitHub Gist](https://gist.github.com/) of any relevant logs or steps to
reproduce the bug. Running `waypoint` commands with `-v` up to `-vvv` will
include any additional debugging info in the log.

**Expected behavior**

I would expect something like `waypoint project apply -poll=false example-ruby` to disable polling. It appears that this isn't possible to do at all via the CLI though: https://github.com/hashicorp/waypoint/blob/main/internal/cli/project_apply.go#L168-L175

**Waypoint Platform Versions**
[...]

PR #2610, opened 2021-10-28
WIP/Discussion: Change project apply flags to accept false and unset values

catsby:
Currently with `waypoint project apply`, users [need to provide several flags][flags] in order to enable git or app status polling. Two flags are problematic (`flagPoll` and `flagAppStatusPoll`) because they are declared as `bool` values, and our flag parsing package cannot discern between a `false` boolean value supplied versus the zero value `false` if either flag is simply omitted. That, coupled with a [logical truthy check][check], currently prevent users from turning either polling or app status polling off with the CLI once they are enabled (users can still disable git polling from the UI).

This PR adds a new flag value `StringBoolVar`, for boolean values that need to discern between `true`,`false`, and simply not specified. `StringBoolVar` is a string value which `true`, `false`, and "empty string" `""` are the only valid values. This enables users to omit either flag and not have that be interpreted as supplying `false`, at the cost of requiring developers to treat a string like a boolean and check for `== "true"` or `== "false"`. 

Fixe #2037

I don't find this solution to be ideal but my attempts to modify [`BoolVar`][boolvar] to use a `*bool` [ended in frustration/dissatisfaction][project-apply-booleans] and I found this to be an acceptable compromise. 

Example error message if providing the CLI a non-truthy value:

```
$ waypoint project apply -app-status-poll-interval="120s" -data-source="git" -git-url="https://github.com/some/repo" -poll= -app-status-poll=false hello-world
! invalid value "" for flag -poll: please use 'true' or 'false'
```

*Note:* I added `pr/no-changelog` label for now while this is discussed/considered

[flags]: https://github.com/hashicorp/waypoint/blob/751129bd6718b5172135738baa9440c96798b3a7/internal/cli/project_apply.go#L162-L185
[check]: https://github.com/hashicorp/waypoint/blob/751129bd6718b5172135738baa9440c96798b3a7/internal/cli/project_apply.go#L169
[boolvar]: https://github.com/hashicorp/waypoint/blob/751129bd6718b5172135738baa9440c96798b3a7/internal/pkg/flag/flag_bool.go
[project-apply-booleans]: https://github.com/hashicorp/waypoint/compare/project-apply-booleans?expand=1

### DIFF
--- a/internal/cli/project_apply.go
+++ b/internal/cli/project_apply.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"io/ioutil"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
@@ -34,9 +35,9 @@ type ProjectApplyCommand struct {
 	flagGitKeyPassword        string
 	flagFromWaypointHcl       string
 	flagWaypointHcl           string
-	flagPoll                  bool
+	flagPoll                  string
 	flagPollInterval          string
-	flagAppStatusPoll         bool
+	flagAppStatusPoll         string
 	flagAppStatusPollInterval string
 	flagOndemandRunner        string
 }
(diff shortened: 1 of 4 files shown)


evanphx:
@catsby I think the reason that your BoolPtr bit wasn't working is that the target type should actually be `**bool`, not `*bool`, as you want to see where a pointer will be set.

catsby:
@evanphx thanks for the suggestion, I'll timebox a stab at `**bool` and see what I come up with

catsby:
closing in favor of #2673

PR #2619, opened 2021-10-29
UI: Use ember simple auth for Authentication

gregone:
Preliminary work before the OIDC implementation. 

Switches the authentication system from our homegrown `session` service to using  [ember-simple-auth](https://github.com/simplabs/ember-simple-auth).

This PR adds a single `token` authenticator and does not change any of the existing behavior, but we'll add a second authenticator for `oidc`, to be able to handle the logic separately. 

The [`session-migrate` initializer](https://github.com/hashicorp/waypoint/compare/ui/ember-simple-auth?expand=1#diff-c252585c419612b5d8157272c81ee6ad3aa522f909e11bc496ff78e9f4ea61d7) is used to convert existing tokens to the new format for a seamless transition for our users.


sabrinako (approved):
LGTM 👍 Excited for all the auth changes to land!

ui/app/initializers/migrate-session.ts
```diff
@@ -0,0 +1,17 @@
...
+export function initialize(/*application: Application*/): void {
+  let waypointToken = window.localStorage.waypointAuthToken;
+  if (waypointToken) {
+    window.localStorage.setItem(
+      'ember_simple_auth-session',
+      JSON.stringify({ authenticated: { authenticator: 'authenticator:token', token: waypointToken } })
+    );
+    window.localStorage.removeItem('waypointAuthToken');
+  }
+}
```
jgwhite, line 13: This is great.

ui/tests/acceptance/build-detail-test.ts
```diff
@@ -1,15 +1,16 @@
...
 import { setupMirage } from 'ember-cli-mirage/test-support';
-import login from 'waypoint/tests/helpers/login';
 
 module('Acceptance | build detail', function (hooks) {
   setupApplicationTest(hooks);
   setupMirage(hooks);
-  login();
 
   test('redirects from the ID to the sequence route', async function (assert) {
+    await login();
```
jgwhite, line 13: Just spitballing, but I wonder if we could add a `setupSession(hooks)` function to the `login` module and use that to reduce the repetition a bit.

```ts
// tests/helpers/login.ts

import { authenticateSession } from 'ember-simple-auth/test-support';

export default async function login(waypointToken = 'default-test-token-value'): Promise<void> {
  return await authenticateSession({ token: waypointToken });
}

export function setupSession(hooks: NestedHooks): void {
  hooks.beforeEach(() => login());
}
```

```ts
// ui/tests/acceptance/build-detail-test.ts

import login, { setupSession } from 'waypoint/tests/helpers/login';
import { setupApplicationTest } from 'ember-qunit';
import { setupMirage } from 'ember-cli-mirage/test-support';
import login from 'waypoint/tests/helpers/login';

module('Acceptance | build detail', function (hooks) {
  setupApplicationTest(hooks);
  setupMirage(hooks);
  setupSession(hooks);
```

ui/tests/helpers/login.ts
```diff
@@ -1,3 +1,5 @@
-export default function login(token?: string): void {
-  window.localStorage.waypointAuthToken = token || 'default-test-token-value';
+import { authenticateSession } from 'ember-simple-auth/test-support';
+
+export default async function login(waypointToken = 'default-test-token-value'): Promise<void> {
+  return await authenticateSession({ token: waypointToken });
 }
```
jgwhite, line 5: See comments above about potentially adding `setupSession` here.

ui/types/ember-simple-auth/services/session.d.ts
```diff
@@ -0,0 +1,14 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    authenticate(authenticator: string, params: any): Promise<void>;
+    isAuthenticated: boolean;
+    invalidate(): Promise<void>;
+    data: SessionData;
+  }
+
+  interface SessionData {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    authenticated?: any;
```
jgwhite, line 12: Could we use `unknown` rather than `any` in the spots? Or does that just add clutter?

jgwhite (commented):
Loves the core changes. I’ve left some feedback about the test changes. Keen to hear what you think.

ui/types/ember-simple-auth/services/session.d.ts
gregone: I think that would work. I'm still sometimes confused by the difference between those two in ts.

ui/tests/acceptance/build-detail-test.ts
gregone: yeah that's a good idea. Updating

gregone: done in https://github.com/hashicorp/waypoint/pull/2619/commits/ac7800ebf0fb64c8f4995725004591d9b24a8581

jgwhite (approved):
Awesome!

ui/types/ember-simple-auth/services/session.d.ts
jgwhite: Yeah me too. I think it’s that `unknown` forces you to check what type a thing is before you use it (or at least check that it conforms to some interface).

PR #2623, opened 2021-10-29
Extract state package test suite

mitchellh:
This extracts the state package test suite into a new package `internal/serverstate/statetest`. For future state package implementations, this allows them to run a broad implementation verification suite of tests to ensure they are functioning.

All I basically did was make our inmem tests generic, while keeping only the tests that specifically poke at the inmemory behavior (i.e. pruning, certain parts of the job system, etc.) in the singleprocess state package. 

Over time, we can add more test cases to the `statetest` package and verify them across all implementations.

This is still WIP but opening a PR sooner than later so we can get these into main and not have divergence as core work is going on too. @evanphx This also gives you some of the more basic CRUD-y tests.

## TODO:

* operation tests (rely on AppOperation which is an impl detail)


evanphx (approved):
This looks excellent, a perfect basis for the alternate state impl.

briancain (approved):
Makes sense!

One thing I'm worried about with this change is it looks like only the tests initialized in `init` will run? Can we fix init pick up the tests inside the file or warn about tests not added to init in the file? I could see it being easy down the road to add a new test, run `go test` and think it passed without knowing you also need to add it to `init` to be evaluated

PR #2634, opened 2021-11-01
internal/core: log variable value source

mitchellh:
Add a debug-level log that shows up in runners that shows where variables are sourced by job. This can be used by callers to more easily debug Waypoint configurations. I think we can make this even easier later but this is a bandaid for now to have SOMETHING to let us know.

### DIFF
--- a/internal/core/project.go
+++ b/internal/core/project.go
@@ -150,6 +150,16 @@ func NewProject(ctx context.Context, os ...Option) (*Project, error) {
 	}
 
 	p.logger.Info("project initialized", "workspace", p.workspace)
+
+	// Output all the variables that this project will used along with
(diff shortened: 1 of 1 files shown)


internal/core/project.go
```diff
@@ -150,6 +150,16 @@ func NewProject(ctx context.Context, os ...Option) (*Project, error) {
 	}
 
 	p.logger.Info("project initialized", "workspace", p.workspace)
+
+	// Output all the variables that this project will used along with
```
catsby, line 154:
```suggestion
	// Output all the variables that this project will use along with
```

catsby (approved):

izaaklauer (approved):
thank you

Issue #2647, opened 2021-11-02
`waypoint server upgrade` fails on k8s due to `Failed to create cluster role "waypoint-runner"`


jgwhite:
**Describe the bug**
When I run `waypoint server upgrade -platform=kubernetes` I see the following error:

```
❌ Initializing role bindings for on-demand runner...
! Error upgrading server on kubernetes: Failed to create cluster role "waypoint-runner":
  "clusterroles.rbac.authorization.k8s.io \"waypoint-runner\" already exists"
```

**Steps to Reproduce**

0. Given an existing Waypoint installation on Kubernetes
1. Run `waypoint server upgrade -platform=kubernetes -auto-approve`
2. Observe the error noted above

**Expected behavior**
Waypoint should detect and re-use the existing `waypoint-runner` role.

**Waypoint Platform Versions**
```
$ waypoint version
CLI: v0.6.0-174-gc7942dc6 (c7942dc6)
Server: v0.6.0-212-g283a474a
```

```
$ kuberctl get nodes
NAME             STATUS   ROLES                  AGE   VERSION
docker-desktop   Ready    control-plane,master   21d   v1.21.4
```

Issue #2452, opened 2021-10-08
`waypoint server uninstall -platform=kubernetes` leaves `waypoint-runner` role and binding


jgwhite:
**Steps to Reproduce**

1. Install the waypoint server to a Kubernetes cluster
   ```
   $ waypoint install \
     -platform=kubernetes \
     -accept-tos \
     -k8s-server-image='ghcr.io/hashicorp/waypoint/alpha:latest' \
     -k8s-odr-image='ghcr.io/hashicorp/waypoint/alpha-odr:latest'
   ```
2. Uninstall the waypoint server
   ```
   $ waypoint server uninstall -platform=kubernetes -auto-approve`
   ```
3. Observe that the `waypoint-runner` clusterrole and clusterrolebinding resources are still present
   ```
   $ kubectl get clusterrole/waypoint-runner clusterrolebinding/waypoint-runner
   
   NAME                                                    CREATED AT
   clusterrole.rbac.authorization.k8s.io/waypoint-runner   2021-10-08T13:14:40Z
   
   NAME                                                           ROLE                          AGE
   clusterrolebinding.rbac.authorization.k8s.io/waypoint-runner   ClusterRole/waypoint-runner   7m20s
   ```
4. Try to reinstall the server and observe the following failure:
   ```
   $ waypoint install \
     -platform=kubernetes \
     -accept-tos \
     -k8s-server-image='ghcr.io/hashicorp/waypoint/alpha:latest' \
     -k8s-odr-image='ghcr.io/hashicorp/waypoint/alpha-odr:latest'
   
   ✓ Gathering information about the Kubernetes cluster...
   x Initializing role bindings for on-demand runner...
   ! Error installing server into kubernetes: Failed to create cluster role "waypoint-runner":
     "clusterroles.rbac.authorization.k8s.io \"waypoint-runner\" already exists"
   ```

**Expected behavior**
The `waypoint-runner` `clusterrole` and `clusterrolebinding` should be removed as part of the uninstall process.

**Waypoint Platform Versions**
* Waypoint CLI Version: 990dbce851a585b24f80b850bd8696367ecd473a
* Waypoint Server Platform and Version: `kubernetes`
* Waypoint Plugin: `kubernetes`

PR #2654, opened 2021-11-02
Do not error on existing cluster roles, and clean up rbac on uninstall

izaaklauer:
Fixes https://github.com/hashicorp/waypoint/issues/2452
Fixes https://github.com/hashicorp/waypoint/issues/2647

It's not beautiful - could be dryer, and it would be better long-term if we discovered what rbac resources we were managing based on tags instead of names I think. This is good enough though, especially seeing as we're likely replacing the this functionality with the helm chart eventually.


briancain (approved):
The code makes sense to me 👍🏻

catsby (approved):

PR #2660, opened 2021-11-02
Backport of Feature: Nomad On-Demand Runner Support into release/0.6.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #2593 to be assessed for backporting due to the inclusion of the label backport/0.6.x.

**WARNING** automatic cherry-pick of commits failed. Commits will require human attention.

The below text is copied from the body of the original PR.

---

This pull request introduces support for launching on-demand runner tasks with Nomad. It launches a nomad job using the ODR image to build containers. Now that StopTask is implemented, it also cleans up the nomad jobs once complete.

Fixes #2503

PR #2661, opened 2021-11-02
Backport/nomad/odr/fix

briancain:
Clean up of https://github.com/hashicorp/waypoint/pull/2660 because it failed

### DIFF
--- a/builtin/nomad/task.go
+++ b/builtin/nomad/task.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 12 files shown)


builtin/nomad/task.go
```diff
@@ -0,0 +1,265 @@
...
+	job := api.NewServiceJob(taskName, taskName, p.config.Region, 10)
+	job.Datacenters = []string{p.config.Datacenter}
+	tg := api.NewTaskGroup(taskName, 1)
+	tg.Networks = []*api.NetworkResource{
+		{
+			Mode: "host",
+		},
+	}
+
+	job.Namespace = &p.config.Namespace
```
paladin-devops, line 205: @briancain I think this line could be causing some unexpected behavior. In my runner profile, I set the Nomad namespace to be "ops" (where I keep my infra/devops-related Nomad jobs, such as a Waypoint runner). The Nomad jobspec that the runner is deploying is written to deploy to the "dev" namespace, since it's an application. However, the result that I'm seeing is that the Nomad job which is deployed ends up in the namespace of the runner, which is "ops".

Runner profile:
```
{
  "datacenter": "dc1",
  "namespace": "ops",
  "nomad_host": "http://nomad.service.dc1.consul:4646",
  "region": "global",
  "resources_cpu": "2000",
  "resources_memory": "4000"
}
```
Job snippet:
```
job "my-job" {
  name        = "my-job"
  namespace   = "${namespace}"  //the value being passed here via templatefile in waypoint.hcl is "dev", which is also the Waypoint workspace name
```

PR #2667, opened 2021-11-03
Feature destroy additional resources for ECS applications

nicholasjackson:
This PR removes additional resources such as security groups, ALBs, and tasks to the `waypoint destroy` lifecycle. The motivation is that should practitioners wish to try waypoint almost all of the resources created during the `waypoint server install` or `waypoint up` phases are now cleanly removed. Prior to this PR, a waypoint user would manually have to remove resources using the AWS console.


briancain (commented):
Thanks for opening this pull request @nicholasjackson ! Could you also include a changelog? We detail the process here in case you haven't seen it.... https://github.com/hashicorp/waypoint/blob/main/.github/CHANGELOG_GUIDE.md#changelog-entry-examples

internal/serverinstall/ecs.go
```diff
@@ -1886,6 +1887,7 @@ const odrRolePolicy = `{
         "ecs:ListTasks",
         "ecs:DescribeTasks",
         "ecs:RegisterTaskDefinition",
+				"ecs:DeregisterTaskDefinition",
```
jgwhite, line 1890: Awesome PR. I couldn’t help but notice this extra indentation…

izaaklauer:
Currently, we're treating all resources as if they belong to the deployment - including resources like the ALB, that are actually necessary for every deployment. As a result, every time we prune an old deployment, we call Destroy on every resource that deployment owns, and by adding destroy funcs to global resources like the ALB, we will destroy those too.

Here's how it goes down over three subsequent `waypoint deploy` runs:
- The first deployment creates its service, target group, etc. It also creates the ALB, ALB Listener, task definition, etc.
- The second deployment creates its service and target group. It also calls create on the ALB, but the create func is idempotent, so nothing happens.
- The third deployment runs and completes, then waypoint goes to prune the first deployment. The first deployment calls destroy on everything it knows about, including the ALB. The third (released) deployment can no longer accept traffic.

We've been punting on this until we get around to building app/workspace scoped resources. We do have the [WorkspaceDestroyer](https://github.com/hashicorp/waypoint-plugin-sdk/blob/main/component/component.go#L182) interface, but it hasn't been widely used, and I think we need a corresponding `WorkspaceCreate` to spool up the resources and store them so we know what to destroy when the time comes.

We might be able to hack around this without a core change that big - it looks like we get a reference to the most recent deployment in the existing `WorkspaceDestroy` func, and we can probably throw some marker object into the DestroyAll argmapper soup that the ALB resource destroy function can use to noop if it's being called by a deployment destroy operation. I'll play with that idea a little to see how viable it is.

nicholasjackson:
Awesome thank you @izaaklauer, totally understand, I saw the comments in the original code and was not sure what the full implications of the PR would be. Please give me a shout if I can be of any help with any testing or anything like that regarding the workspaces stuff.

@jgwhite and @briancain, will get those bits sorted today. Thanks again.

izaaklauer:
@nicholasjackson Thank you for the PR and for resurfacing this! Implementing the delete funcs for the resources is a huge help - when we get the core wiring sorted, the cleanup logic should come together really quickly.

internal/serverinstall/ecs.go
nicholasjackson: I have no idea why my editor decided to indent that with tabs, I think it is trolling me

PR #2671, opened 2021-11-03
cli/server: use filter flags; update UI_ListReleasesRequest

### LINKED ISSUE #2666: The limit filter flag does not work for deployments or releases
**Describe the bug**
The `leimit=` filter flag does not appear to work.

**Steps to Reproduce**
Create multiple deployments/releases and run `waypoint deployment/release list -limit=1`

**Expected behavior**
Only one deployment/release is shown

* Waypoint CLI Version: 0.6.1
* Waypoint Server Platform and Version: kubernetes 0.6.0
* Waypoint Plugin: n/a

**Additional context**
The flag appears to function for `waypoint artifact list`

krantzinator:
Fixes #2666 
Adds `Status` and `PhysicalState` filters to the `UI_ListReleasesRequest`, to match the `UI_ListDeploymentsRequest`, as well as the original `ListReleasesRequest`.

Note 1: The backport for this would fail on the second commit, so I'll go ahead and open that separately.

Note 2: There is still some funkiness going on with our filters. I didn't have time to dig into just now because I wanted to get this PR up for the release tomorrow. But this PR at least resolves #2666 and also contributes the above additional update, without making any existing functionality worse.


briancain (approved):
Nice! Makes sense to me, looks good 👍🏻 I think this should still have a changelog...otherwise I think it'll get lost and only be present on the release/0.6.x branch and not on main.

izaaklauer (approved):

Issue #2662, opened 2021-11-03
ui: Default workspace selection


jgwhite:
**Is your feature request related to a problem? Please describe.**
When I visit Waypoint’s UI, it automatically drops me into the `default` workspace. But what if I don’t use the default workspace? How should it behave?

**Describe the solution you'd like**
Given I have not specified the workspace in the URL, I want the UI to intelligently choose the workspace I’m most likely to want.

Some thoughts the team discussed in a recent sync.

- Store the most-recently viewed workspace in local storage and try to re-use this
- Try using the most recently active workspace (we bump activity timestamp on every operation)
- Failing that, show the alphabetically first
- Failing all the above, settle for default

Here’s that in flowchart form (there are other ways to draw this, but I think this works):

![image](https://user-images.githubusercontent.com/34030/140068859-42c3423f-091b-4c2b-a76e-687d7270b566.png)

Additionally, @amandakalk suggested the following:

> if your default workspace is blank and you have other workspaces, the empty message could give you a little suggestion to change workspaces

I think this will be very useful when I *do* specify a workspace but the UI can tell that I don’t actually have anything in there.

**Describe alternatives you've considered**
TODO

**Additional context**
This will follow #1339 (currently underway)

Long term, we’ve discussed having a similar model to GitHub’s branch selection, where users can elect a default for each project. This, however, is not in the immediate roadmap.

izaaklauer:
I think the downside of using "most recently active" is it increases the odds that someone will take an action against the prod workspace, thinking they're in dev. I feel like doing something consistent, like alphabetical, might help give people "muscle memory" for jumping to the workspace they're interested in.

jgwhite:
@izaaklauer all good points, and simpler to implement too! I don’t see any reason not to go with this simpler approach.

jgwhite:
Updated flow chart:

![image](https://user-images.githubusercontent.com/34030/140191340-0a053658-59b6-4cf6-8e33-fef7ceea46ba.png)

jgwhite:
Actually, given the special status of “default”, maybe this would be better:

![image](https://user-images.githubusercontent.com/34030/140191961-7dc835cd-98bb-42f9-ab18-6d5fa804a86d.png)

Issue #1339, opened 2021-04-14
Workspace Selection


josephlim75:
**Is your feature request related to a problem? Please describe.**
As `waypoint` CLI allow to use `-workspace`, it would be a good feature to allow user to view workspaces from Waypoint web ui.  Similar to how Vault namespace web ui display

**Describe the solution you'd like**
A drop down list at the header that allow user to select which workspace they would like to view.  Pairing with RBAC, it will help to segregate Team A can view Workspace A, TEam B can view Workspace B

almonk:
@josephlim75 thanks for raising this issue. Below are the proposed designs for the Waypoint UI.

![Application -_ Deployments](https://user-images.githubusercontent.com/51724/119483858-53bda100-bd4d-11eb-9c27-2b6929d3d839.png)

![Application -_ Deployments-1](https://user-images.githubusercontent.com/51724/119483863-54eece00-bd4d-11eb-82f8-bb4c0660326f.png)

jgwhite:
Closed by #2674

PR #2674, opened 2021-11-04
ui: workspace switcher

jgwhite:
## Why the change?

Addresses #1339 

## What’s the plan?

- [x] Add the switcher
- [x] Iron out loading states
- [x] Empty state
- [x] Add high-level test coverage
- [x] Add granular test coverage
  - [x] Happy path
  - [x] Error states
  - [x] Loading states
- [x] Use proper CSS, not cowboy CSS
- [x] Make dark mode look passable
- [x] Add changelog entry

### What’s not the plan?

Default workspace selection will be addressed in a separate PR. See #2662 for problem statement.

## What does it look like?

https://user-images.githubusercontent.com/34030/141003230-648a874e-f1c2-4bf0-a849-bd3d27525bc1.mp4

https://user-images.githubusercontent.com/34030/141003234-43f071d9-b2aa-40bf-b870-0dce60308aab.mp4

(note: the loading state no longer appear in the flow above, just the error)

## How do I test it?

1. `git checkout ui/workspace-switcher`
2. Boot the dev server in [mirage mode](https://github.com/hashicorp/waypoint/blob/main/ui/README.md#running-with-mocks) or [local mode](https://github.com/hashicorp/waypoint/blob/main/ui/README.md#running-with-a-local-waypoint-server)
3. Navigate to an app page
4. Try switching workspaces
5. Verify everything works as you’d expect

### DIFF
--- a/ui/app/components/workspace-switcher.ts
+++ b/ui/app/components/workspace-switcher.ts
@@ -1,0 +1,69 @@
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+import { task, TaskGenerator } from 'ember-concurrency';
+import { taskFor } from 'ember-concurrency-ts';
+import { Workspace } from 'waypoint-pb';
+import ApiService from 'waypoint/services/api';
+
+type Args = {
+  current?: string;
+  route?: string;
+  models?: string[];
+  isOpen?: boolean;
+};
+
+type Result = { workspaces: Workspace.AsObject[] } | { error: Error };
+
+export default class extends Component<Args> {
+  @service api!: ApiService;
+
+  constructor(owner: unknown, args: Args) {
+    super(owner, args);
+
+    taskFor(this.loadWorkspaces).perform();
+  }
+
+  @task({ restartable: true })
+  *loadWorkspaces(): TaskGenerator<Result> {
+    try {
+      let workspaces = yield this.api.listWorkspaces();
+      return { workspaces };
+    } catch (error) {
+      return { error };
+    }
+  }
+
+  get shouldExist(): boolean {
+    let task = taskFor(this.loadWorkspaces);
+
+    if (task.isRunning) {
+      return false;
+    }
+
+    let value = task.last?.value;
+
+    if (!value) {
+      return false;
+    }
+
+    if ('error' in value) {
+      return true;
+    }
+
+    let { workspaces } = value;
+
+    if (workspaces.length > 1) {
+      return true;
+    }
+
+    if (workspaces.length === 0) {
+      return false;
+    }
+
+    if (workspaces[0].name === this.args.current) {
+      return false;
+    }
+
+    return true;
+  }
+}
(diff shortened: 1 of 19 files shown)


ui/app/templates/workspace/projects/project/app.hbs
```diff
@@ -9,7 +9,20 @@
 )}}
 <PageHeader @iconName="git-repository">
   <div class="title">
-    <h1>{{@model.application.application}}</h1>
+    <div class="title-wrapper">
```
jgwhite, line 12: Not sure I entirely needed to introducing a wrapping element here, but I didn’t want to throw off the `title` styling.

amandakalk:
One small thought on the "no active workspaces" state. Assuming a user has not created any workspace and is the default space, maybe we should hide the workspace switcher if you only have a single workspace. Seems like it could cause less confusion if you have no knowledge of workspaces.

I'm also a little confused by the "no active workspaces" message. In your screenshot, isn't "production" considered an active workspace? Should the message say "No **other** active workspaces found" to be more clear?

@almonk Do you have any thoughts on this?

I don't think this change should hold up this PR though, we can change it later if we agree on a different behavior.

almonk:
I agree that we shouldn't show the switcher if there's only 1 workspace (and while you can't create them in the UI). @jgwhite what do you think about populating workspace data when the user navigates to an app rather than just in time when the user clicks the dropdown?

ui/tests/integration/components/workspace-switcher-test.ts
```diff
@@ -0,0 +1,78 @@
+import { render, waitFor, settled } from '@ember/test-helpers';
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { setupRenderingTest } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import { module, test } from 'qunit';
+import { Response } from 'miragejs';
+
+module('Integration | Component | workspace-switcher', function (hooks) {
```
sabrinako, line 9: Appreciate how comprehensive these tests are!

ui/app/components/workspace-switcher.ts
```diff
@@ -0,0 +1,35 @@
...
+export default class extends Component<Args> {
+  @service api!: ApiService;
+
+  constructor(owner: unknown, args: Args) {
+    super(owner, args);
+
+    taskFor(this.loadWorkspaces).perform();
+  }
+
+  @task({ restartable: true })
```
sabrinako, line 26: I think this is the first time I'm seeing `ember-concurrency` and `@task`, super interesting

sabrinako (approved):
Looks good to me! For something so complicated under the hood, this code is super straight forward and clear, great job! 🤠 👍

jgwhite:
> @jgwhite what do you think about populating workspace data when the user navigates to an app rather than just in time when the user clicks the dropdown?

@almonk luckily it already does behave that way, I just artificially slowed things down to make the videos more illustrative.

@almonk + @amandakalk I think I’ll go ahead and _hide_ the switcher altogether until we’ve loaded the workspaces and, as you both suggest, only show it under either of these conditions:

1. There’s more than one workspace
2. The current workspace amongst those returned by the API

PR #2675, opened 2021-11-04
ui: address some ember-simple-auth type warnings

jgwhite:
## Why the change?

Does what it says on the tin, specifically these two warnings:

CleanShot 2021-11-04 at 12 22 20@2x
CleanShot 2021-11-04 at 12 22 05@2x

## How do I test it?

This mostly changes type declarations, but it couldn’t hurt to check out the branch and smoke test it (both against your own editor’s intellisense and against a real Waypoint server).


gregone (approved):

gregone:
Good catch, thanks! works fine in VSCode.

PR #2681, opened 2021-11-04
ui: update icons to use the FlightIcon package

### LINKED ISSUE #2392: UI: migrate the UI Icons from `structure-icons` to `flight`
Use https://github.com/hashicorp/flight to replace the Structure icons

i.e restart the work started in our experiment here https://github.com/hashicorp/waypoint/pull/2194 and replace the icons. 

It would be a good idea to start by adding more screenshots to the Percy test to cover most icons, then update the icons.

sabrinako:
## Related Issue
#2392 

## Description
This PR intends to transfer the Waypoint Ui over from using the `Pds::Icon` components to the new `FlightIcon` ones.

## How to Test
1. Pull this branch down `git checkout ui/flight-icon-update`
2. `yarn` to download the new packages
3. `yarn start` to boot up UI with mock data
4. Click around to ensure all icons are looking as expected!

## Additional Notes
`flight-icons` currently doesn't have any animated icons, so our `loading` spinner is not present. There's [discussion](https://github.com/hashicorp/flight/issues/282) about adding it that's picked up in recent days, so I imagine this will be added soon but I'm not planning on holding this PR back any longer since it's just 1 icon left.

### Deployments Page
Screen Shot 2021-12-08 at 3 40 06 PM

### Auth Page
Screen Shot 2021-12-08 at 3 51 04 PM

_As of making this PR, the method for altering the colors of the icons in `Flight` requires altering the `use` element instead of `path`. This also seems to prevent us from changing the stroke width. Feedback on how we could change the stroke width would be helpful (since I'm pretty sure it was thinner before)_

### DIFF
--- a/ui/app/components/app-item/deployment.hbs
+++ b/ui/app/components/app-item/deployment.hbs
@@ -26,7 +26,7 @@
       <:badge>
         {{#if (eq @deployment.state 4)}}
           <b data-test-destroyed-badge class="badge badge--destroyed">
-            <Pds::Icon @type="trash" class="icon" />&nbsp; {{t "page.deployments.destroyed_label"}}
+            <FlightIcon @name="trash" class="icon" />&nbsp; {{t "page.deployments.destroyed_label"}}
           </b>
         {{/if}}
       </:badge>
--- a/ui/app/components/copyable-code/index.hbs
+++ b/ui/app/components/copyable-code/index.hbs
@@ -7,6 +7,6 @@
    class={{if this.copySuccess 'copy-success'}}
    title="Copy"
   >
-    <Pds::Icon @type={{if this.copySuccess 'copy-success' 'copy-action'}} class="icon" />
+    <FlightIcon @name={{if this.copySuccess 'copy-success' 'copy-action'}} class="icon" />
   </CopyButton>
 </span>
--- a/ui/app/components/copyable-code/index.hbs
+++ b/ui/app/components/copyable-code/index.hbs
@@ -7,6 +7,6 @@
    class={{if this.copySuccess 'copy-success'}}
    title="Copy"
   >
-    <FlightIcon @name={{if this.copySuccess 'copy-success' 'copy-action'}} class="icon" />
+    <FlightIcon @name={{if this.copySuccess 'clipboard-checked' 'clipboard'}} class="icon" />
   </CopyButton>
 </span>
--- a/ui/app/components/git-commit.hbs
+++ b/ui/app/components/git-commit.hbs
@@ -1,6 +1,6 @@
 {{#if @commit}}
   <span class="git-commit" title="Triggered by commit: {{@commit}}">
-    <Pds::Icon @type="git-commit"/>
+    <FlightIcon @name="git-commit"/>
     {{!-- TODO: add git profile picture of user whose commit it was when possible --}}
     <CopyButton @clipboardText={{@commit}}>
      <span class="git-commit__sha">{{truncate-commit @commit}}</span>
--- a/ui/app/components/project-input-variables/list.hbs
+++ b/ui/app/components/project-input-variables/list.hbs
@@ -14,7 +14,7 @@
   </ul>
   <Pds::ButtonSet>
     <Pds::Button data-test-input-variables-add-variable disabled={{this.isCreating}} @variant="primary" {{on "click" this.addVariable}}>
-      <Pds::Icon @type="plus-plain" />{{t "form.project_variables_settings.button_apply"}}
+      <FlightIcon @name="plus" />{{t "form.project_variables_settings.button_apply"}}
     </Pds::Button>
   </Pds::ButtonSet>
 {{else}}
@@ -25,7 +25,7 @@
     </P.Body>
     <P.Footer>
       <Pds::Button data-test-input-variables-add-variable @variant="primary" {{on "click" this.addVariable}}>
-        <Pds::Icon @type="plus-plain" />{{t "form.project_variables_settings.button_apply"}}
+        <FlightIcon @name="plus" />{{t "form.project_variables_settings.button_apply"}}
       </Pds::Button>
     </P.Footer>
   </Pds::EmptyState>
--- a/ui/app/components/project-input-variables/list.hbs
+++ b/ui/app/components/project-input-variables/list.hbs
@@ -24,7 +24,7 @@
       {{t "form.project_variables_settings.empty_state_content"}}
     </P.Body>
     <P.Footer>
-      <Pds::Button data-test-input-variables-add-variable @variant="primary" {{on "click" this.addVariable}}>
+      <Pds::Button data-test-input-variables-add-variable @variant="primary" class="button--add-variable" {{on "click" this.addVariable}}>
         <FlightIcon @name="plus" />{{t "form.project_variables_settings.button_apply"}}
       </Pds::Button>
     </P.Footer>
--- a/ui/app/components/workspace-switcher.hbs
+++ b/ui/app/components/workspace-switcher.hbs
@@ -37,7 +37,7 @@
               >
                 <span class="workspace-switcher__link-contents">
                   {{#if (eq workspace.name @current)}}
-                    <FlightIcon @name="check-plain" />
+                    <FlightIcon @name="check" />
                   {{/if}}
                   {{workspace.name}}
                 </span>
--- a/ui/app/components/workspace-switcher.hbs
+++ b/ui/app/components/workspace-switcher.hbs
@@ -20,7 +20,7 @@
                 data-test-workspace-switcher-error-state
                 class="workspace-switcher__error-state"
               >
-                <Pds::Icon @type="alert-triangle" />
+                <FlightIcon @name="alert-triangle" />
                 {{t "workspace-switcher.error"}}
                 ({{this.loadWorkspaces.last.value.error.message}})
               </div>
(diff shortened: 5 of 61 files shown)


sabrinako:
Just wanted to update on GH as well but this is on hold until we get the HC logos in Flight

jgwhite:
Are we dropping icons from the breadcrumbs?

CleanShot 2021-12-09 at 10 02 12@2x

- Looks like there’s unneeded space for an icon to the left of the app name
- Icon next to “deployments” seems a bit close to the text (maybe it’s not meant to be there?)

ui/app/templates/workspace/projects/project/app/deployment/deployment-seq.hbs
```diff
@@ -8,7 +8,7 @@
             href={{enforce-protocol @model.preload.deployUrl}}
             class="button button--secondary button--external-link">
             <span>{{lowercase @model.preload.deployUrl}}</span>
-            <Pds::Icon @type="exit" class="icon" />
+            <FlightIcon @name="exit" class="icon" />
```
jgwhite, line 11: Maybe `exit` isn’t the icon name?

CleanShot 2021-12-09 at 10 05 58@2x

ui/app/components/copyable-code/index.hbs
```diff
@@ -7,6 +7,6 @@
    class={{if this.copySuccess 'copy-success'}}
    title="Copy"
   >
-    <Pds::Icon @type={{if this.copySuccess 'copy-success' 'copy-action'}} class="icon" />
+    <FlightIcon @name={{if this.copySuccess 'copy-success' 'copy-action'}} class="icon" />
```
jgwhite, line 10: Looks like the icon is missing here:

CleanShot 2021-12-09 at 10 07 20@2x

ui/app/components/workspace-switcher.hbs
```diff
@@ -37,7 +37,7 @@
               >
                 <span class="workspace-switcher__link-contents">
                   {{#if (eq workspace.name @current)}}
-                    <Pds::Icon @type="check-plain" />
+                    <FlightIcon @name="check-plain" />
```
jgwhite, line 40: Doesn’t look like these quite came through (also not sure what’s happened to the alignment):

CleanShot 2021-12-09 at 10 08 29@2x

ui/app/components/git-commit.hbs
```diff
@@ -1,6 +1,6 @@
 {{#if @commit}}
   <span class="git-commit" title="Triggered by commit: {{@commit}}">
-    <Pds::Icon @type="git-commit"/>
+    <FlightIcon @name="git-commit"/>
```
jgwhite, line 3: Vertical alignment looks a little off here:

CleanShot 2021-12-09 at 10 09 36@2x

ui/app/components/project-input-variables/list.hbs
```diff
@@ -25,7 +25,7 @@
     </P.Body>
     <P.Footer>
       <Pds::Button data-test-input-variables-add-variable @variant="primary" {{on "click" this.addVariable}}>
-        <Pds::Icon @type="plus-plain" />{{t "form.project_variables_settings.button_apply"}}
+        <FlightIcon @name="plus" />{{t "form.project_variables_settings.button_apply"}}
```
jgwhite, line 28: Alignment looks a little off here:

CleanShot 2021-12-09 at 10 13 59@2x

ui/app/templates/workspace/projects/new.hbs
```diff
@@ -28,7 +28,7 @@
     <fieldset class="pds-formFieldSet">
       <div class="pds-formField">
         <div class="pds-fieldName" id="create-git-help-text">
-          <Pds::Icon @type="git-repository" class="icon" />
+          <FlightIcon @name="git-repo" class="icon" />
```
jgwhite, line 31: Was this icon meant to grow?

CleanShot 2021-12-09 at 10 18 05@2x

ui/app/components/app-item/deployment.hbs
```diff
@@ -26,7 +26,7 @@
       <:badge>
         {{#if (eq @deployment.state 4)}}
           <b data-test-destroyed-badge class="badge badge--destroyed">
-            <Pds::Icon @type="trash" class="icon" />&nbsp; {{t "page.deployments.destroyed_label"}}
+            <FlightIcon @name="trash" class="icon" />&nbsp; {{t "page.deployments.destroyed_label"}}
```
jgwhite, line 29: This seems a little tight in its box now. Maybe let’s add a touch more padding?

CleanShot 2021-12-09 at 10 18 49@2x

jgwhite (changes requested):
The new icons are looking really nice.

I spotted a few nitpicks, but in general the transition is butter smooth.

ui/app/components/app-item/deployment.hbs
sabrinako: This is eventually going to be removed because the new designs that Al made for the deployment card states (in the redesign Figma). Should I just remove it in this PR?

jgwhite: Ah — I’d say leave it as is. You’d probably have to change some tests too if you remove it.

ui/app/styles/_layout.scss
```diff
@@ -128,16 +128,6 @@ hr {
...
-  width: scale.$lg-2;
-  height: scale.$lg-2;
-  flex-shrink: 0;
-}
-
-.pds-icon {
-  display: inline-flex;
-  align-items: center;
-  justify-content: center;
+button .icon {
```
gregone, line 131: should this style be moved to button.scss? 

also this is just for the `button` element, not the button css classes, so:
- do button-style links with icons still render correctly?
- do `button` elements with icons, used without the button style (for accessibility) still render properly?

gregone (approved):
LGTM. just a few questions about using the `button` element for styling

ui/app/styles/_layout.scss
jgwhite: It looks like this gets overridden by more-specific rules in `button.scss`:

CleanShot 2021-12-14 at 16 37 31@2x

Maybe we don’t need it at all?

jgwhite (approved):
Looking good. The points @gregone raised are worth investigating but apart from that I didn’t spot anything out of place.

ui/app/styles/components/button.scss
```diff
@@ -92,11 +86,23 @@ a[href].button {
   }
 
   &--external-link {
-    > .pds-icon {
+    > svg {
```
didoo, line 89: here you're relying on the component internal structure (expects the SVG to be a direct child) and in case we decide in the future to change it, this would break (unknowingly) the styling.

my suggestion is to use a wrapping container/div, and style that container. in this way you decouple hosting codebase from the components implementation

ui/app/styles/components/button.scss
```diff
@@ -92,11 +86,23 @@ a[href].button {
...
     }
   }
 
+  &--add-variable {
+    > span {
+      display: flex;
+      align-items: stretch;
+      margin-left: calc(-1 * 0.25rem);
+
+      svg {
```
didoo, line 100: potentially, also here you have a coupling (less strong than the other one, but still)

one option could be to apply a class to the FlightIcon component and use that one as a CSS selector here (and also in other cases I see where you have used the DOM element as target)

ui/app/styles/pages/auth-page.scss
```diff
@@ -31,8 +31,10 @@
 
     .icon.brand {
       display: flex;
+      width: scale.$lg-14;
+      height: scale.$lg-14;
 
-      path {
+      use {
```
didoo, line 37: this is not how we would expect color to be applied. you're manipulating directly the internal SVG but we expose a `@color` prop for this (see https://flight-hashicorp.vercel.app/engineering#example-fill)

plus, our icons use `currentColor` by default so they can just inherit the color from the parent container

plus, our icons are flat, they don't contain strokes, so the `stroke: $brand-color;` declaration can be removed

ui/app/styles/pages/auth-page.scss
```diff
@@ -31,8 +31,10 @@
 
     .icon.brand {
       display: flex;
+      width: scale.$lg-14;
```
didoo, line 34: @heatherlarsen for knowledge, one case in which the `24px` icon is scaled up to `96px`

ui/app/styles/components/workspace-switcher.scss
```diff
@@ -31,14 +30,14 @@
   line-height: var(--pds-lineHeight--dense);
   padding-left: scale.$lg--2;
 
-  .pds-icon {
+  svg {
     position: absolute;
     font-size: scale.$lg--1;
```
didoo, line 35: do you need a font-size for an icon?

didoo (commented):
@sabrinako added a few comments, take them for what they are (just suggestions)

ui/app/styles/pages/auth-page.scss
heatherlarsen: I ran into an issue getting this to run locally, but I'm assuming this is either for the Waypoint logo or the HashiCorp logo. I'd advocate for using a logo in these instances instead of an icon.

heatherlarsen:
@sabrinako I saw your note about changing the stroke weight. Stroke weight is not something that should be changed on Flight icons as each icon was intentionally designed with consistent stroke weight based on the size of the icon. More information about how the icons are drawn can be found in the [docs](https://flight-hashicorp.vercel.app/design#drawing-icons). Is there a specific reason you need to adjust the stroke weight?

ui/app/styles/pages/auth-page.scss
jgwhite: > I ran into an issue getting this to run locally

@heatherlarsen I’d like to hear more about this. I’ll reach out to you on Slack :)

PR #2684, opened 2021-11-05
Feature: app destroy for ECS

izaaklauer:
Building off @nicholasjackson's work in https://github.com/hashicorp/waypoint/pull/2667, this implements WorkspaceDestroy to delete costly aws resources when an app is deleted.

## How to verify

Deploy from an ecs project, and then run `waypoint destroy -auto-approve`. Note the following output as waypoint destroys your deployment, and then shared resources:

```
$ waypoint destroy -auto-approve

» Destroying releases for application 'example-nodejs'...

» Destroying deployments for application 'example-nodejs'...
✓ Finished destroying ECS deployment
✓ Deleted service example-nodejs-01FKRGJS16PANKBG4
✓ Deregistered this deployment's target group from ALB listener
✓ Deleting target group example-nodejs-01FKRGJS16PANKBG4
✓ Finished destroying ECS deployment
✓ Deleted service example-nodejs-01FKRG53B45XAQF3T
✓ Deleted ALB Listener
✓ Deleting target group example-nodejs-01FKRG53B45XAQF3T

» Destroying shared deploy resources for application 'example-nodejs'...
✓ Finished destroying ECS deployment
✓ Deleted ECS task definition
✓ Deleted ALB arn:aws:elasticloadbalancing:us-east-1:797645259670:loadbalancer/app/waypoint-ecs-example-nodejs/4cd35175f42fe868
✓ Deleted security group sg-008a68bc1664f7b1a
✓ Deleted security group sg-03d85903c85d5bc25
Destroy successful!
```

## Problems this doesn't solve
- We aren't destroying _all_ the resources associated with the app. We're leaving around global resources that are used by multiple apps and projects (like the waypoint log group), and also some free resources like the execution role, task role, and route53 record. There's nothing stopping us from implementing delete on the non-globally scoped resources in the future.
- Building workspace-awareness into the resource destroy funcs themselves is not a great design - just the easiest way to implement the feature right now. I think in the future we should make a WorkspaceCreate hook that generates these workspace scoped resources, manage them in a separate resourceManager instance from the deployment.

Closes https://github.com/hashicorp/waypoint/pull/2667


builtin/aws/ecs/platform.go
```diff
@@ -286,6 +295,10 @@ func (p *Platform) resourceManager(log hclog.Logger, dcr *component.DeclaredReso
 // DeploymentId is a unique ID to be consistently used throughout our deployment
 type DeploymentId string
 
+// WorkspaceDestroy marks that we're destroying the entire workspace (not just the deployment), and
+// we should destroy global resources too
+type WorkspaceDestroy bool
```
evanphx, line 300: I'm a little unclear why we define this as a type...

izaaklauer: Ah good point - I added a comment clarifying: https://github.com/hashicorp/waypoint/pull/2684/files#diff-20946acdf4caf300d4d679801d8116bde77354aae07f8a988943474ca4351e85R295-R297

They're defined as types so that argmapper inside resource manager can wire them from where we invoke the overall destroy function ([here](https://github.com/hashicorp/waypoint/pull/2684/files#diff-20946acdf4caf300d4d679801d8116bde77354aae07f8a988943474ca4351e85R519)) to where we use them in individual resource funcs (like [here](https://github.com/hashicorp/waypoint/pull/2684/files#diff-20946acdf4caf300d4d679801d8116bde77354aae07f8a988943474ca4351e85R1570)).

I think it would be possible to use argmapper [named parameter matching](https://github.com/hashicorp/go-argmapper#named-and-typed-values) instead, but this felt better to me.

evanphx (approved):

PR #2688, opened 2021-11-05
UI/OIDC auth

### LINKED ISSUE #2390: UI OIDC Authentication
Following up on https://github.com/hashicorp/waypoint/issues/1621, we need to enable signing up and logging into the Waypoint UI using an OIDC provider

gregone:
Closes #2390 

This PR adds the OIDC Authentication flow to the browser UI. 

The docs and Go references to auth links have been updated to reflect the new `auth/oidc-callback` URL.


ui/app/authenticators/oidc.ts
```diff
@@ -0,0 +1,42 @@
...
+  code: string;
+  state: string;
+}
+
+@classic
+export default class OIDCAuthenticator extends OAuth2ImplicitGrantAuthenticator {
+  restore(data: SessionData): Promise<SessionData> {
+    if (data.token) {
+      window.localStorage.removeItem('waypointOIDCAuthMethod');
+      window.localStorage.removeItem('waypointOIDCNonce');
```
jgwhite, line 24: Could we add a comment here to explain why we clean up these now?

ui/app/components/oidc-auth-buttons/index.ts
```diff
@@ -0,0 +1,44 @@
...
+import { inject as service } from '@ember/service';
+import { task } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+
+interface OIDCAuthButtonsArgs {
+  model: ListOIDCAuthMethodsResponse.AsObject;
+}
+
+export default class OIDCAuthButtonsComponent extends Component<OIDCAuthButtonsArgs> {
+  @tracked model!: ListOIDCAuthMethodsResponse.AsObject;
```
jgwhite, line 20: Is this still in use?

ui/app/components/oidc-auth-buttons/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    let urlRequest = new GetOIDCAuthURLRequest();
+    let authMethodrequest = new GetAuthMethodRequest();
+
+    let authMethodRef = new Ref.AuthMethod();
+    authMethodRef.setName(authMethodProviderName);
+    authMethodrequest.setAuthMethod(authMethodRef);
+    urlRequest.setAuthMethod(authMethodRef);
+    let randomArray = new Uint32Array(10);
+    window.crypto.getRandomValues(randomArray);
+    let nonce = randomArray.join('').slice(0, 20);
```
jgwhite, line 35: Could we extract this out into a separate function? Maybe `generateNonce` or something like that. Just to make things a little clearer.

ui/app/routes/auth/index.ts
```diff
@@ -0,0 +1,28 @@
...
+export default class AuthIndex extends Route {
+  @service session!: SessionService;
+  @service api!: ApiService;
+  @tracked authMethods!: ListOIDCAuthMethodsResponse;
+
+  async beforeModel(): Promise<void> {
+    let authMethods = await this.api.client.listOIDCAuthMethods(new Empty(), this.api.WithMeta());
+    this.authMethods = authMethods;
+    return;
+  }
```
jgwhite, line 18: Why do we do this in `beforeModel` rather than directly in `model`?

ui/app/routes/auth/oidc-callback.ts
```diff
@@ -0,0 +1,33 @@
...
+    completeAuthRequest.setAuthMethod(authMethodRef);
+    completeAuthRequest.setRedirectUri(window.location.origin + window.location.pathname);
+    let nonce = window.localStorage.getItem('waypointOIDCNonce');
+    if (nonce) {
+      completeAuthRequest.setNonce(nonce);
+    }
+    completeAuthRequest.setState(oidcParams.state);
+    let resp = await this.api.client.completeOIDCAuth(completeAuthRequest, this.api.WithMeta());
+    let respObject = resp.toObject();
+    await this.session.authenticate('authenticator:oidc', respObject);
```
jgwhite, line 31: Should we clean up any of the stuff in local storage at this point, or is that handled elsewhere?

ui/mirage/models/auth-method.ts
```diff
@@ -0,0 +1,13 @@
+import { Model } from 'miragejs';
+import { OIDCAuthMethod } from 'waypoint-pb';
+
+export default Model.extend({
+  toProtobuf(): OIDCAuthMethod {
+    let result = new OIDCAuthMethod();
+
+    result.setDisplayName(this.displayName);
+    result.setKind(0);
```
jgwhite, line 9: Seems like this data is managed by the factory?

```suggestion
    result.setKind(this.kind);
```

jgwhite (changes requested):
Looking great! I added a few questions and suggestions but the overall architecture is spot on.

ui/app/components/oidc-auth-buttons/index.ts
gregone: Good point I was thinking about doing similar

ui/app/routes/auth/oidc-callback.ts
gregone: I thought about this, and then realized that because ember-simple-auth reuses the same session store and key, it is not necessary as it will overwrite the session, and a single authenticator is used https://github.com/simplabs/ember-simple-auth/blob/master/README.md?plain=1#L372-L373

ui/app/components/oidc-auth-buttons/index.ts
gregone: not used here but used in the template. I'll remove the `@tracked` decorator.

ui/app/routes/auth/index.ts
gregone: I think that's a letfover from the previous prototype, where I was thinking about doing a redirect to an `auth/oidc` route, I'll move everything to  model as this is not happening anymore.

ui/app/authenticators/oidc.ts
gregone: added, also lol this should have been in `authenticate` not `restore` 🤦

gregone:
@jgwhite comments addressed in https://github.com/hashicorp/waypoint/pull/2688/commits/457c62312ac0899a08a31287c027afa7138d3554

ui/app/components/oidc-auth-buttons/index.ts
jgwhite: I see `@model` referenced in the template but not `this.model`. Not a big deal, either way.

ui/app/routes/auth/oidc-callback.ts
jgwhite: Ahhh, that makes sense 👍

jgwhite (approved):

PR #2692, opened 2021-11-05
support arm64 for the entrypoint, inject properly

mitchellh:
This adds a `linux/arm64` entrypoint asset to our Waypoint binary. This
increases binary size by ~30% but I think this is worthwhile as more of
the world's dev and deployment machines move to arm64.

This makes it so that Docker entrypoint injection injects the entrypoint
binary that matches the architecture and errors if it detects an
unsupported architecture.

This also adds detection for cloud native buildpacks to warn if the
Docker daemon isn't x86_64 since Buildpacks currently have issues with
arm64.

This also changes the Docker build to output the image architecture
since @jgwhite has built and deployed mismatching architecutures
many times.

Going to target this for backporting.

### DIFF
--- a/.changelog/2692.txt
+++ b/.changelog/2692.txt
@@ -1,0 +1,7 @@
+```release-note:improvement
+plugin/docker: inject `arm64` Waypoint entrypoints for arm images
+```
+
+```release-note:improvement
+plugin/pack: detect non-Intel Docker server and show a warning
+```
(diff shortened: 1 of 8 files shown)


evanphx (approved):
you say "arm64", docker says "aarch64", let's not call the whole thing off, let's just support both!

PR #2694, opened 2021-11-08
ui: add git commit mock data to mirage

### LINKED ISSUE #2680: ui: add git commit mock data to mirage
With the current testing setup, there's no way to see the Gitops related features (git commit badges, etc) with the preloaded data. You need to make a Waypoint server to see these assets in the UI

Add git information to one of the test data projects so we can see that data in the UI.

jgwhite:
## Why the change?

Closes #2680

## Any other notes?

This is cheap-and-cheerful solution, and an example of where I think it’s worth cutting corners in Mirage. A full-fidelity solution would include modeling and simulating `Job`, which is one of the largest and most complex protobufs in the system, and therefore more work than is currently justified (in my opinion). As we push further into introspecting data about jobs, I’m pretty confident the situation the time to dive into simulating Job will clearly present itself.

## How do I test it?

I would try adding/changing git commit SHAs in our default scenario data, to see if you can get them to show up in the UI.

### DIFF
--- a/ui/mirage/factories/build.ts
+++ b/ui/mirage/factories/build.ts
@@ -5,6 +5,29 @@ export default Factory.extend({
   id: () => fakeId(),
   sequence: (i) => i + 1,
 
+  /**
+   * `gitCommitRef` is a shorthand for setting the commit SHA for this build.
+   * Provide a value here and it will appear in the `preload` field for the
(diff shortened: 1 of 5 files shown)


gregone (approved):

sabrinako (approved):
Looks great! 👍 So glad this is in our mock data now!

PR #2698, opened 2021-11-08
serverinstall/nomad: Spend more time looking at waypoint-runner allocation on install to ensure start up

### LINKED ISSUE #2683: serverinstall/nomad: Spend more time looking at `waypoint-runner` allocation on install to ensure start up
**Describe the bug**

Currently the server install command for the Nomad platform spins waiting for the `waypoint-runner` job to be "created" and "running", and then exits immediately when that job enters these states. One issue with this is with the recent addition of Consul, if the runner cannot properly resolve the Consul DNS the runner job passes the created and running states, but soon after the container exits due to not making a successful connection to Waypoint server because it couldn't resolve the Consul DNS hostname.

Ideally, we should fix up the server install function for Nomad to not only wait until the job is running, but also ensure the jobs container in the allocation successfully connects back to Waypoint server before calling the install "successful". Note that this failure happens almost immediately after the job is running, so waiting a bit longer and doing a few more cycles to check the container is healthy and running should be enough.

**Steps to Reproduce**

Install Waypoint server to Nomad with misconfigured Consul DNS, see that the waypoint-runner exits due to failing to connect back to Waypoint server.

**Expected behavior**

Update the install runner function to also ensure the container is successfully running beyond the first check.

applicable:

* Waypoint CLI Version: 0.6.2
* Waypoint Server Platform and Version: Nomad, 0.6.2+
* Waypoint Plugin: n/a

briancain:
Prior to this commit, the server install command would look for the very
first instance for when an allocation was considered running. This is
generally ok, however if the Nomad job fails to get started a moment
later (like a static runner failing to connect back to Waypoint Server),
the runner allocation would exit later but the server install command
would consider the installation successful.

This is especially important when users are configuring Consul DNS
with Waypoint Server installed to Nomad. The static runner might
fail to properly make a connection through the Consul DNS hostname
and fail, leaving the installation without a static runner but the CLI claiming
the install succeeded.

This commit fixes that by doing a few retries for a few seconds on the scheduled
allocation once its in a "running" state to validate it properly started
up beyond the first few moments of the job.

Fixes #2683

### DIFF
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -408,28 +408,20 @@ func (i *NomadInstaller) Upgrade(
 	}
 
 	s.Update("Waiting for allocation to be scheduled")
-EVAL:
 	qopts := &api.QueryOptions{
 		WaitIndex: resp.EvalCreateIndex,
 	}
 
-	eval, meta, err := client.Evaluations().Info(resp.EvalID, qopts)
+	eval, meta, err := i.waitForEvaluation(ctx, s, client, resp, qopts)
 	if err != nil {
-		return nil, err
-	}
-	qopts.WaitIndex = meta.LastIndex
-	switch eval.Status {
-	case "pending":
-		goto EVAL
-	case "complete":
-		s.Update("Nomad allocation created")
-	case "failed", "canceled", "blocked":
 		s.Update("Nomad failed to schedule the waypoint server ", serverName)
 		s.Status(terminal.StatusError)
-		return nil, fmt.Errorf("nomad evaluation did not transition to 'complete'")
-	default:
-		return nil, fmt.Errorf("unknown eval status: %q", eval.Status)
+		return nil, err
 	}
+	if eval == nil {
+		return nil, fmt.Errorf("evaluation status could not be determined")
+	}
+	qopts.WaitIndex = meta.LastIndex
 
 	var allocID string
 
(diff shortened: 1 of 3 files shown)


internal/serverinstall/nomad.go
```diff
@@ -802,7 +787,12 @@ EVAL:
 		}
 
 		if allocID != "" {
-			return allocID, nil
+			if retries == maxRetries {
+				return allocID, nil
+			} else {
+				s.Update("Ensuring allocation %q has properly started up...", allocs[0].ID)
+				time.Sleep(1 * time.Second)
+			}
```
evanphx, line 795: Because we move the WaitIndex forward before trying again, won't the call to Allocations wait more than 1 second if the allocation is running fine, because there won't be any updates to report, so the call won't return?

internal/serverinstall/nomad.go
```diff
@@ -756,30 +748,22 @@ func (i *NomadInstaller) runJob(
 	}
 
 	s.Update("Waiting for allocation to be scheduled")
-EVAL:
 	qopts := &api.QueryOptions{
```
evanphx, line 751: Related to my later comment, we probably need to be setting the WaitTime (or whatever it's called) on this so that we don't block indefinitely on these calls, but rather wait some amount of time before we figure out if we want to continue waiting.

internal/serverinstall/nomad.go
briancain: From what I could tell, the call to Allocations uses a default `WaitTime` if none specified, but I can set a more explicit one of 500 milliseconds to be sure. From my testing in the positive case, the call always returned after the ~3 seconds this loop retries on.

internal/serverinstall/nomad.go
briancain: The API tells me they use a default `WaitTime`, so we shouldn't be blocking forever. I set a default one anyway to be sure.

krantzinator (approved):
Looks like this will be a bit friendlier UX when errors happen!

internal/serverinstall/nomad.go
evanphx: The change looks good!

evanphx (approved):

PR #2700, opened 2021-11-08
cli: add alias for -app and -workspace flags

krantzinator:
This is piece of foundational work for further
improvements to the CLI's UX. This is a separate
PR from the bulk of that redesign, since this
change affects a large number of files and could
be distracting in PR review of the larger scope
of work.

As a bonus, I also reordered the global "targeting" commands to be in order of precedence, which happens to also match their alphabetical order. This reads nicer in both respects.

### DIFF
--- a/.changelog/2700.txt
+++ b/.changelog/2700.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Add alias for -app and -workspace flags
+```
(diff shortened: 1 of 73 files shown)


briancain (approved):

mitchellh (approved):

izaaklauer (commented):
👍

PR #2702, opened 2021-11-09
ui: add all operation info to OperationStatusIndicator

sabrinako:
## Description
Originally, the `OperationStatusIndicator` just encapsulated the time information related to an operation. This was always paired with the actual operation (ie: `Deployed to Docker`) which was being hard coded in every instance to account for the slight differences in code between each operation type. This PR combines the operation and the time information in `OperationStatusIndicator` and creates child versions for each operation type.
Additionally, this PR lays some groundwork for #2681 

## To Test
1. Pull down this branch `git checkout ui/operation-status-indicator`
2. `yarn start`
3. Check these areas for consistency across the app/with the previous version
   - Build/Deployment/Release lists
   - Build/Deployment/Release details page

### DIFF
--- a/ui/app/components/operation-status-indicator/release.hbs
+++ b/ui/app/components/operation-status-indicator/release.hbs
@@ -1,0 +1,5 @@
+<OperationStatusIndicator @operation={{@operation}} @isDetailed={{false}}>
+  <:status-text>
+    {{if (eq @operation.status.state 1) 'Releasing' 'Released'}} on
+  </:status-text>
+</OperationStatusIndicator>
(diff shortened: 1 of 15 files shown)


ui/app/components/operation-status-indicator/release.hbs
```diff
@@ -0,0 +1,5 @@
+<OperationStatusIndicator @operation={{@operation}} @isDetailed={{false}}>
+  <:status-text>
+    {{if (eq @operation.status.state 1) 'Releasing' 'Released'}} on
```
jgwhite, line 3: Would be nice if we can localize this. Don’t sweat it if it’s a pain though.

If you’re feeling adventurous, you could try some [advanced ICU formatting](https://formatjs.io/docs/core-concepts/icu-syntax/#select-format), though that gets pretty wacky pretty quickly 😅

jgwhite (approved):
This is an incredibly elegant refactor. I love it 😍

PR #2707, opened 2021-11-10
feat(404): helper text for notfound docs

thiskevinwang:
image

### DIFF
--- a/website/pages/404.jsx
+++ b/website/pages/404.jsx
@@ -49,9 +49,9 @@ export default function NotFound() {
       <p>
         Try viewing the&nbsp;
         <Link href={asPath.replace(reg, '')}>
-          <a>latest</a>
+          <a>latest version</a>
         </Link>
-        &nbsp;version instead.
+        &nbsp;instead.
       </p>
     </>
   )
--- a/website/pages/404.jsx
+++ b/website/pages/404.jsx
@@ -19,7 +19,7 @@ export default function NotFound() {
     <div id="p-404" className="g-grid-container">
       <h1 className="g-type-display-1">Page Not Found</h1>
       <p>
-        We&lsquo;re sorry but we can&lsquo;t find the page you&lsquo;re looking
+        We&rsquo;re sorry but we can&rsquo;t find the page you&rsquo;re looking
         for.
       </p>
       <p>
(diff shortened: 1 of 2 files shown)


website/pages/404.jsx
```diff
@@ -15,18 +18,39 @@ export default function NotFound() {
...
       </p>
       <p>
         <Link href="/">
           <a>Back to Home</a>
         </Link>
       </p>
+    </>
+  )
+
+  const reg = /\/v\d+[.]\d+[.](\d+|x)/g
```
BRKalow, line 35: Can we use the util from `docs-page`? https://github.com/hashicorp/react-components/blob/main/packages/docs-page/util/index.ts

website/pages/404.jsx
```diff
@@ -15,18 +18,39 @@ export default function NotFound() {
...
+  const docsMessage = (
+    <>
+      <p>
+        The document you requested is not available for the current version.
```
BRKalow, line 39: Maybe:
```
We&rsquo;re sorry, but this page does not exist for version [x.y.z]. Try viewing this page for version [latest] instead.
```

website/pages/404.jsx
thiskevinwang: `router.query` unfortunately doesn't play nicely with this util, since it's empty (`{}`) on page refresh,

a possible alternative is splitting `asPath` into "params", but that feels over-complicated

website/pages/404.jsx
thiskevinwang: 1dba0ff

image

website/pages/404.jsx
```diff
@@ -15,18 +18,42 @@ export default function NotFound() {
...
+  const matches = reg.exec(asPath)
+  const docsMessage = (
+    <>
+      <p>
+        We&rsquo;re sorry, but this page does not exist for version&nbsp;
+        <b>{matches?.groups?.version}</b>.
+      </p>
+      <p>
+        Try viewing the&nbsp;
+        <Link href={asPath.replace(reg, '')}>
```
BRKalow, line 45: @thiskevinwang This is just showing up as `/` 🤔 : https://waypoint-git-kevin-404-for-docs-hashicorp.vercel.app/plugins/v0.4.x/helm

thiskevinwang: Weird. Clicking it still redirects to `/plugins/helm` for me.
Looking...

BRKalow (approved):

PR #2713, opened 2021-11-10
docs, serverinstall: updates to docs & naming of runner install

krantzinator:
We support ODR in all builting install platforms now,
so our documentation should reflect that.

Also updates the Nomad ODR config to list the Name as the platform name, matching our other builtin install platforms.


briancain (approved):
👍🏻

PR #2724, opened 2021-11-11
Rename `make bin/cli-only` to `make bin/no-ceb`

izaaklauer:
What it's actually doing is skipping embedding the ceb. The image it produces can be used as a cli, runner, or server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,11 @@ bin: # bin creates the binaries for Waypoint for the current platform
 	cd internal/assets && go-bindata -pkg assets -o prod.go -tags assetsembedded ./ceb
 	CGO_ENABLED=$(CGO_ENABLED) go build -ldflags $(GOLDFLAGS) -tags assetsembedded -o ./waypoint ./cmd/waypoint
 
-.PHONY: bin/cli-only
-bin/cli-only: # bin/cli-only only recompiles waypoint
+# bin/cli-only only recompiles waypoint, and doesn't recompile or embed the ceb.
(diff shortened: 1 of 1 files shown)


krantzinator (approved):

Makefile
```diff
@@ -20,8 +20,11 @@ bin: # bin creates the binaries for Waypoint for the current platform
 	cd internal/assets && go-bindata -pkg assets -o prod.go -tags assetsembedded ./ceb
 	CGO_ENABLED=$(CGO_ENABLED) go build -ldflags $(GOLDFLAGS) -tags assetsembedded -o ./waypoint ./cmd/waypoint
 
-.PHONY: bin/cli-only
-bin/cli-only: # bin/cli-only only recompiles waypoint
+# bin/cli-only only recompiles waypoint, and doesn't recompile or embed the ceb.
+# You can use the binary it produces as a server, runner, or CLI, but it won't contain the CEB, so
+# it won't be able to build projects that don't have `disable_entrypoint = true` set in their build hcl.
+.PHONY: bin/no-ceb
+bin/cli-only:
```
briancain, line 27: Does this need to be changed too?

briancain (approved):

mitchellh (approved):
Note that it can't be used as a "runner" unless you're only doing non-entrypoint-injected jobs. If you are, it'll error, which is fine.

Makefile
izaaklauer: oh it sure does.

PR #2726, opened 2021-11-12
.../statetest: Test that list returns multiple projects

evanphx:


mitchellh (approved):

internal/serverstate/statetest/test_project.go
```diff
@@ -61,17 +64,25 @@ func TestProject(t *testing.T, factory Factory, restartF RestartFactory) {
 		// Get case insensitive
 		{
 			resp, err := s.ProjectGet(&pb.Ref_Project{
-				Project: "abcDe",
+				Project: strings.ToLower(name),
 			})
-			require.NoError(err)
+			require.NoError(err, "unable to use case insensitive name for: %s", name)
```
krantzinator, line 69: Friday brain: why can't we use a lowercased name for a Project?

evanphx: Oh, that error is actually saying "This error is because case insensitive lookup didn't work". It's confusing because it's the message on `NoError` and thus is describing the situation when an error is returned.

PR #2731, opened 2021-11-15
ui: use node 14

jgwhite:
## Why the change?

Node v12 enters end-of-life on 2022-04-30 so now seems like a reasonable time to move to Node v14, which will be in maintenance status until 2023-04-30. This also brings us in line with both the website and cloud-ui.

## How do I test it?

CI should give a good indication, but worth checking out the branch, switching to Node v14, and making sure things work as you expect.


gregone (approved):
Looking good!

PR #2732, opened 2021-11-15
make install: make sure go/bin directory exists

kohlerm:
In case you do a fresh checkout and you have never compiled a go project "make install" will fail because the go bin directory does not exist. 

This might be a rare problem for a developer machine, but might cause issues in CI builds and it looks to me it is easy to fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ bin/entrypoint: # create the entrypoint for the current platform
 
 .PHONY: install
 install: bin # build and copy binaries to $GOPATH/bin/waypoint
-	mkdir -p $(GOPATH)/bin/waypoint
+	mkdir -p $(GOPATH)/bin
 	cp ./waypoint $(GOPATH)/bin/waypoint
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -47,6 +47,7 @@ bin/entrypoint: # create the entrypoint for the current platform
 
 .PHONY: install
 install: bin # build and copy binaries to $GOPATH/bin/waypoint
+	mkdir -p $(GOPATH)/bin/waypoint
```
briancain, line 50:
```suggestion
	mkdir -p $(GOPATH)/bin
```

briancain (changes requested):
Thanks for making a pull request @kohlerm ! Just one suggestion, we're moving the `waypoint` binary directly under `bin` rather than a subfolder/binary like `waypoint/waypoint`. Otherwise looks good!

briancain (approved):
Thank you!

PR #2738, opened 2021-11-16
ui: upgrade ember-auto-import to ^2.2.4

jgwhite:
## Why the change?

Just keeping things up-to-date (and because we can now that we’re on Node 14)

## How does it affect build times?

Comparing the ember-build-tests time between `main` (1m17s) and this branch (1m25s) doesn’t show any significant difference.

## What does it look like?

Totally normal, hopefully.

CleanShot 2021-11-17 at 12 48 18@2x

## How do I test it?

1. Check out the branch
2. Verify dev mode works as you’d expect
3. Verify it works when compiled into the server binary

## Anything else to consider?

Yeah, with this change a bunch of stuff moves out of `vendor.js` and into several `chunk.*.js` files. I’m not yet sure if our ember-asset-size action will account for this.


ui/ember-cli-build.js
```diff
@@ -41,9 +41,11 @@ module.exports = function (defaults) {
...
         // uncomment this to see a tree view of what ember-cli-auto-import builds
         // plugins: [new (require('webpack-bundle-analyzer').BundleAnalyzerPlugin)()],
-        // required for ansi-colors to work in the browser
-        node: {
-          process: true,
+        resolve: {
+          fallback: {
+            // required for ansi-colors to work in the browser
+            process: require.resolve('process/browser'),
+          },
```
jgwhite, line 48: This doesn’t seem to work. Not sure why yet. Will take another swing at it tomorrow.

jgwhite:
Woah!

image

ui/ember-cli-build.js
jgwhite: Just realized it’s because ansi-colors expect the `process` global to be around. Going to investigate how to convince Webpack 5 to sneak it in.

jgwhite: Tried using [DefinePlugin](https://webpack.js.org/plugins/define-plugin/) but it doesn’t seem to operator on dependencies.

jgwhite: Figured this out in the end

jgwhite:
OK I think the file size discrepancy is because Webpack 5 automatically breaks up the vendor.js payload into chunks. I’m not sure if ember-asset-size is aware of this and can incorporate the `chunk.*.js` files into the analysis.

gregone:
@jgwhite It should be possible to use `splitChunks` webpack config to avoid the splitting https://github.com/ef4/ember-auto-import/blob/684d97a147fc8d5446c4aa909c129bce438b3d23/test-scenarios/common-chunk-test.ts#L47

jgwhite:
@gregone great suggestion. Just reading the `splitChunks` documentation. It’s complex!

jgwhite:
@gregone judging by https://github.com/ef4/ember-auto-import/issues/426, it doesn’t look like it’s easy to revert to the former chunk-less behavior. We could get it down to one chunk with config alone, but that’d still live outside the `vendor.js` file. Looks like other folks want ember-asset-size to take this into account too: https://github.com/simplabs/ember-asset-size-action/pull/33

jgwhite:
> (e.g https://github.com/ef4/ember-auto-import/blob/684d97a147fc8d5446c4aa909c129bce438b3d23/test-scenarios/common-chunk-test.ts#L47)

Confusingly, this test is actually *enabling* chunk-splitting, even for very tiny files.

jgwhite:
I think for now the benefits of ember-asset-size outweigh the benefits of this upgrade. This PR is still useful, as we’ll need to upgrade eventually and now we have documentation on how to do so. In the meantime, I’m going to make myself a note to lobby for improvements such as https://github.com/simplabs/ember-asset-size-action/pull/33

PR #2742, opened 2021-11-16
ui: start lab boilerplate

mitchellh:
The lab will be a place where anyone on the Waypoint team can throw
non-user-ready functionality to better assist in developing Waypoint.
For example, we can make it easy to list runner profiles or runners
without having to commit to a true shippable UX.

This commit doesn't do anything practical except lay down some basic
boilerplate folder structure.


evanphx (approved):
I love a good lab.

jgwhite (approved):

PR #2746, opened 2021-11-17
list runners api

izaaklauer:
Very basic first implementation of list runners.

The immediate need for this is to determine if any remote runners exist, so we can decide if we can execute projects remotely vs locally.

This doesn't entirely meet that need - because runners aren't currently marked as remote vs cli, there's no way to tell if a runner is remote, or just some other user who happens to be running a CLI command with a local runner. For my immediate purposes though, that kind of false-positive isn't a dealbreaker. We can add more filter capabilities in the future if we need to.


krantzinator (approved):
Makes sense!

internal/server/proto/server.proto
```diff
@@ -195,6 +195,9 @@ service Waypoint {
   // GetRunner gets information about a single runner.
   rpc GetRunner(GetRunnerRequest) returns (Runner);
 
+  // ListRunners gets information about existing runners.
```
mitchellh, line 198: Nitpick but I would change these docs to note that this only lists runners that are currently running and registered, i.e. explicitly omits on-demand runners that are not currently running.

mitchellh (approved):
One small nitpick but the rest good.

briancain (approved):

PR #2747, opened 2021-11-17
Websocket endpoint for exec stream

mitchellh:
This adds a websocket API (over the HTTP port) for `waypoint exec` so that the UI can exec since grpc-web does not support bidirectional gRPC streams. The API still requires auth and goes through our normal HTTP middleware. The stream format is protobuf-encoded binary streams so the API is almost identical to a gRPC stream. 

I had this branch sitting for awhile but @gregone has confirmed it works well enough to try to merge into main!

Not going to do a changelog entry since we don't document our API publicly atm.


gregone (approved):
💯

internal/server/httpapi/exec.go
```diff
@@ -0,0 +1,159 @@
...
+			// We allow connections from anywhere so that the UI can be on another
+			// host and so that we work with proxying. We aren't too worried about
+			// CSRF due to the nature of Waypoint servers and the underlying
+			// protocols and auth tokens, but one day it'd be nice to be stricter
+			// here and accept only known UI hosts (we don't even have this
+			// config at the time of writing).
+			InsecureSkipVerify: true,
+		})
+		if err != nil {
+			log.Info("error accepting websocket connection", "err", err)
```
briancain, line 43: Should this be `log.Error`?

briancain (approved):
Looks pretty good to me. Agreed this implementation is pretty simple and straight forward!

internal/server/httpapi/exec.go
mitchellh: Nah, because there are various reasons triggered by the user this can fail. So operators needn't be concerned. This is user-triggered more than anything.

evanphx (approved):
🎵 Everything's better with a for/select loop!

PR #2749, opened 2021-11-17
Auto local/remote detection

izaaklauer:
This introduces the logic for automatic detection of local vs remote operations. It isn't used anywhere yet. 

I had some fun building tests that mocked the server - there are enough permutations that I didn't build a case for each one, but this covers the highlights.

### DIFF
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -399,12 +399,15 @@ func (c *baseCommand) Init(opts ...Option) error {
 	return nil
 }
 
-// remoteIsPossible attempts to determine if the current waypoint infrastructure will be successful
+// remoteOpPreferred attempts to determine if the current waypoint infrastructure will be successful
 // performing a remote operation against this project. It verifies the project's datasource,
 // it's ODR runner profile, and detects if a remote runner is currently registered.
-func remoteIsPossible(ctx context.Context, client pb.WaypointClient, project *pb.Project, log hclog.Logger) (bool, error) {
-	// Check if remote is disabled in the waypoint.hcl
-
+// If an operation can occur successfully remotely, we prefer the remote environment for consistency
+// and security reasons.
+//
+// Note that this cannot guarantee that an operation will succeed remotely - the remote environment
+// may not have the right auth configured, the right plugins configured, etc.
+func remoteOpPreferred(ctx context.Context, client pb.WaypointClient, project *pb.Project, log hclog.Logger) (bool, error) {
 	if !project.RemoteEnabled {
 		log.Debug("Remote operations are disabled for this project - operation cannot occur remotely")
 		return false, nil
@@ -453,9 +456,8 @@ func remoteIsPossible(ctx context.Context, client pb.WaypointClient, project *pb
 	}
 
 	// Check to see if we have a runner profile assigned to this project
-
 	if project.OndemandRunner != nil {
-		log.Debug("Project has an explicit ODR profile set - operation will work remotely")
+		log.Debug("Project has an explicit ODR profile set - operation is possible remotely")
 		return true, nil
 	}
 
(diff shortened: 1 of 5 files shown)


evanphx (approved):
This would mean that usage of ODR is required. I think that's probably ok since we know that ODR unlocks a lot of functionality now. If we do require that, various docs and things should be updated to reflect it.

internal/cli/base.go
```diff
@@ -397,6 +399,95 @@ func (c *baseCommand) Init(opts ...Option) error {
...
+		log.Debug("Default runner profile exists - operation will work remotely.")
+		return true, nil
+	}
+
+	log.Debug("No runner profile is set for this project and no global default exists - operation should happen locally")
+	// The operation here _could_ still happen remotely - executed on the remote runner itself without ODR.
+	// If it's a container build op it will probably fail (because no kaniko), and if it's a deploy/release op it
+	// very well might fail do to incorrect/insufficient permissions. Because it probably won't work, we won't try,
+	// but the user could force it to happen locally by setting -local=false.
+	return false, nil
```
mitchellh, line 488: Yeah this makes me wonder the difference between two semantics:

- `remoteIsPossible` (this function)
- `remoteIsPreferable` (hypothetical)

As @evanphx pointed out, this forces ODR completely, which isn't unreasonable, but will the CLI allow FORCING remote still? If so, no problem, but it still may be useful to have the `remoteIsPossible` validation which could return `true` even w/o ODR.

mitchellh (changes requested):
Looks great! The tests are great. I personally would move away from using a mock and use an in-mem server instead which we have all the helpers for.

internal/cli/base.go
evanphx: Another thought is, rather than look at the code and hope it works, add a test job to the system that can be launched and watched. So that way we validate the pipeline with an empirical test rather than a theoretical one.

mitchellh: We have the code for this already, we have noop jobs. So that's a good idea, just very slow, but maybe right during preflight for real ops.

evanphx: Yeah, it depends entirely when this code would be run. You wouldn't want to be doing it to decide if every operation could take place. I'd rather have a flag somewhere that says "this project has run remote jobs before" that is set in the past and this tool just checks that flag.

internal/cli/base.go
```diff
@@ -397,6 +399,95 @@ func (c *baseCommand) Init(opts ...Option) error {
...
+	}
+
+	if project.DataSource == nil {
+		log.Debug("Project has no datasource configured - operation cannot occur remotely")
+		// This is probably going to be fatal somewhere downstream
+		return false, nil
+	}
+
+	var hasRemoteDataSource bool
+	switch project.DataSource.GetSource().(type) {
```
briancain, line 420: Should we be checking this in addition to the `RemoteEnabled` bool? https://github.com/hashicorp/waypoint/blob/7ba0386c24a98b3034c2fcb8dee9e397c12845e5/internal/server/proto/server.proto#L584

I think we have an explicit coupling on if a data source is git it's technically remote, but this this option is explicitly telling us if the project is remote or not...assuming it's correct and properly in all cases 🤔

internal/cli/base.go
```diff
@@ -397,6 +399,95 @@ func (c *baseCommand) Init(opts ...Option) error {
...
+	default:
+		hasRemoteDataSource = false
+	}
+
+	if !hasRemoteDataSource {
+		log.Debug("Project does not have a remote data source - operation cannot occur remotely")
+		return false, nil
+	}
+
+	// We know the project can handle remote ops at this point - but do we have runners?
```
briancain, line 432: What happens when you run Waypoint in local mode? i.e. no server. I think we need a check for this too?

briancain (commented):
Just a couple of comments from me....can do a proper review once the other comments are addressed :) 👍🏻

internal/cli/base.go
izaaklauer: Yup, that's the first thing we check in this function!

internal/cli/base.go
izaaklauer: In local server mode, we won't have a remote runner available, which will force us to the local runner.

internal/cli/base.go
briancain: Oops, missed it! 😄

briancain (approved):

internal/cli/base.go
izaaklauer: The CLI will still be able to force a job happen remotely - so if they know something these heuristics don't (i.e. they want it to run on the remote runner with no ODR), they can still queue up the remote job. 

I like the idea of a test-baloon job to see of remote ops are possible, but for full coverage we'd need to run it before every op (too slow), or reevaluate it every time the project gets upserted, every time a runner profile is upserted, and every time a runner registers (without additional heuristics).

 I think this is probably good enough for now, but I can see doing that down the road.

mitchellh (approved):

PR #2750, opened 2021-11-18
run make gen/ts

gregone:


jgwhite (approved):

jgwhite:
This is causing a rather odd failure mode in CI!

jgwhite:
We figured why the tests were hanging forever rather than simply failing. The change in the generated protobuf files meant that imports were failing and thus Mirage handlers were crashing outright rather than returning a rejected promise, which in turn seemed to cause Mirage/Pretender to leave requests unresolved indefinitely.

PR #2751, opened 2021-11-18
statetest: Make the timing of the watchset tests more forgiving

evanphx:
This gives the watchset implementor a little more latitude about how quickly it detects the new data.


mitchellh (approved):

briancain (approved):
⏲️ 🙏🏻

PR #2754, opened 2021-11-18
CLI end-to-end tests

izaaklauer:
This adds some new cli end-to-end tests. They're not hooked up to CI yet - that'll take a bit more work - but they're as useful as the other e2e tests. 

I'm about to make a PR that affects a ton of our CLI's surface area, and rather than _just_ doing a bunch of manual testing, I wrote this.

### DIFF
--- a/test-e2e/cli_smoke.test.go
+++ b/test-e2e/cli_smoke.test.go
@@ -1,23 +1,18 @@
 package test
 
-import (
-	"strings"
-	"testing"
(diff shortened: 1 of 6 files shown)


briancain (approved):
Legit! Let's try to get these hooked up to CI soon 🙏🏻

krantzinator (approved):
Love seeing the progression here.

PR #2756, opened 2021-11-19
Better error handling around missing ODR profiles

izaaklauer:
Catches an extremely corner-case panic.

Without this, if you try to upsert a project with a git ref, data polling disabled, and a runner profile ref that doesn't exist, we panic.


krantzinator (approved):

internal/server/singleprocess/service_job.go
```diff
@@ -166,7 +166,7 @@ func (s *service) queueJobReqToJob(
 		}
 	}
 
-	// If the job can be run any any runner, then we attempt to see if we should spawn
+	// If the job can be run by any runner, then we attempt to see if we should spawn
 	// on on-demand runner for it. We only consider jobs for any runner because ones
```
catsby, line 170:
```suggestion
	// an on-demand runner for it. We only consider jobs for any runner because ones
```

internal/server/singleprocess/service_job.go
```diff
@@ -279,6 +283,10 @@ func (s *service) onDemandRunnerStartJob(
 ) (*pb.Job, string, error) {
 	log := hclog.FromContext(ctx)
 
+	if od == nil {
+		return nil, "", status.Error(codes.FailedPrecondition, "Missing required parameter OnDemandRunnerConfig")
+	}
+
```
catsby, line 289: Nit: I don't see any other way into this method outside of the above `wrapJobWithRunner`, which already performs a `nil` check on the `OnDemandRunnerConfig`, so this seems redundant.

catsby (approved):
👍 suggestion and comment, take or leave them 😄

internal/server/singleprocess/service_job.go
xiaolin-ninja: Came to say the same thing :)

xiaolin-ninja (approved):

mitchellh (commented):
We unit test the ODR launching stuff decently well in the singleprocess package. Any chance we can repro this for a regression test?

mitchellh (approved):
Looks good, the only thing I'd add here is given where the code was changed, this _should_ be easily unit testable and we may want to do that.

krantzinator:
In writing the tests, I wasn't able to recreate the problem exactly as described in this PR, so, since this isn't urgent and holidays mean we won't cross paths for a couple weeks, I'm just going to leave this little test here for @izaaklauer when he gets back. Take it or leave it 😄 I assume I am missing the crux of this edge case. https://github.com/hashicorp/waypoint/compare/odr-profile-not-found...copy-of-odr-profile-not-found?expand=1

briancain:
Ahhh, oops. Looks like we didn't get to this one 😢 I accidentally redid it in https://github.com/hashicorp/waypoint/pull/3054 and added a lil' test so I'm going to close this since they are the same!

PR #2758, opened 2021-11-19
Add UpsertWorkspace GRPC endpoint

catsby:
Prior to this, Workspaces are created dynamically as needed/referenced by Projects. This PR adds an explicit endpoint `UpsertWorkspace` that will create/update a workspace. 

In the future, workspace creation will be done through these methods (either the endpoint, or the Workspace state method `WorkspacePut`), but for now this PR only adds the endpoints and state methods. 

Note that at this time `UpsertWorkspace` **_does not_** modify the associated `Projects`. This may change in the future to do so, or we may add other endpoint(s) similar to updating an Application association with a Project in [`UpsertApplication`][upsertapp]

[upsertapp]: https://github.com/hashicorp/waypoint/blob/eae4441084484d236caf6b635b883637c39ff57a/internal/server/singleprocess/service_project.go#L86


.changelog/2758.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+core/server: add new grpc endpoint UpserWorkspace
```
mitchellh, line 2: don't think we need a changelog entry since our API isn't documented atm

internal/server/ptypes/workspace.go
```diff
@@ -10,6 +12,25 @@ import (
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 )
 
+// WorkspaceNameRegexp is the valid Workspace name regular expression.
+// Copied from UsernameRegexp.
+var WorkspaceNameRegexp = regexp.MustCompile(`(?i)^[a-z\d][a-z\d_-]*[a-z\d]+$`)
```
mitchellh, line 17: This is fine to start with but might be too restrictive cause this is ascii only and not super friendly to foreign languages. We may want to change to unicode character classes at some point: https://www.regular-expressions.info/unicode.html

internal/server/singleprocess/state/workspace.go
```diff
@@ -14,16 +14,43 @@ import (
...
 
+// WorkspacePut creates or updates the given Workspace.
+//
+// Project changes will be ignored
+func (s *State) WorkspacePut(workspace *pb.Workspace) error {
+	memTxn := s.inmem.Txn(true)
+	defer memTxn.Abort()
+
+	err := s.db.Update(func(dbTxn *bolt.Tx) error {
+		prev, err := s.workspaceGet(dbTxn, memTxn, &pb.Ref_Workspace{
```
mitchellh, line 33: I think this logic also should exist in the workspacePut.

mitchellh (changes requested):
Looks really good, only a couple minor comments 👍

internal/server/singleprocess/service_workspace_test.go
```diff
@@ -0,0 +1,70 @@
...
+			resp, err := client.UpsertWorkspace(ctx, &Req{
+				Workspace: serverptypes.TestWorkspace(t, &pb.Workspace{
+					Name: "staging",
+				}),
+			})
+			require.NoError(err)
+			require.NotNil(resp)
+		}
+
+		// Create another
```
xiaolin-ninja, line 38: Can we also add a test for updating an existing workspace?

internal/server/ptypes/workspace.go
evanphx: Do we even really need to restrict the workspace name?

catsby: I would think maybe restrict it to not include spaces, but otherwise I would think not. I borrowed this from Users

.changelog/2758.txt
catsby: 👍 removed in d9f7fc7f4

internal/server/singleprocess/state/workspace.go
catsby: Moved in e025389f8

internal/server/ptypes/workspace.go
evanphx: The easy way to do this is loop over each rune and call `unicode.Is(unicode.Nl, r)`. https://pkg.go.dev/unicode#Is, https://pkg.go.dev/unicode#Nl

mitchellh: That's equivalent to the regexp `\p{Nl}+`

internal/server/singleprocess/service_workspace_test.go
catsby: At this time there is no real updating happening at all. At this time we don't yet support updating the workspace with an associated project (or removing one either), and we don't support renaming workspaces.

catsby: To clarify, we absolutely should have tests on updating workspace and project associations, but I think a discussion needs to be had on if we support that through this Upsert method, or an other endpoint just for adding/removing projects.

xiaolin-ninja: Oh, thanks for clarifying. Yeah, I was a bit confused reading through the code why this was an upsert instead of a create.

internal/server/ptypes/workspace.go
catsby: After some tinkering I updated the regular expression to be `^[\p{L}\p{N}_]*$` in 8bb7e90d13554d5d20bee5f9cdd178a90366a2bb . I used https://regex101.com/r/Ngn8bv/1 to adjust it as I couldn't get Mitchell's suggestion to work as-is, and I don't know regex well enough to know if it was intended as pseudo code or regex that would normally work in some thing other than Golang. 

I also added new tests for this, as well as an additional verification in `workspacePut` where as before we were only verifying in the service.

mitchellh (approved):

briancain (approved):
Looks good to me!

PR #2760, opened 2021-11-19
Refactor CLI init sequence

izaaklauer:
I had two connected goals here:

- Give us a clear place to put the local/remote detection from https://github.com/hashicorp/waypoint/pull/2749
- Apply some of the simplifications laid out in the project UX rfc to simplify this hairy piece of our init sequence: https://github.com/hashicorp/waypoint/blob/eae4441084484d236caf6b635b883637c39ff57a/internal/cli/base.go#L237-L343

A fair amount of the surface area of this change is renaming, either to express subtly different behavior as a result of these changes, or for general clarity, explicitness, or consistency. Bikeshedding welcome.

### What functionality changed

- The project/app syntax is removed. To make an operation happen remotely now, add the -remote flag, and the -project/-app flags if required.
- The `status` command no longer requires a project argument (it uses the flag instead), and takes the operation flags so that the refresh job can happen remotely.

Still TODO:
- [x] One QA pass over all the commands and local/remote permutations (especially the status command r/e multiple apps)

### DIFF
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -11,6 +11,8 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/waypoint/internal/config/variables"
+
 	"github.com/adrg/xdg"
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/hashicorp/go-hclog"
@@ -21,7 +23,6 @@ import (
 	clientpkg "github.com/hashicorp/waypoint/internal/client"
 	"github.com/hashicorp/waypoint/internal/clierrors"
 	"github.com/hashicorp/waypoint/internal/config"
-	"github.com/hashicorp/waypoint/internal/config/variables"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	"github.com/hashicorp/waypoint/internal/runner"
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
@@ -234,6 +235,17 @@ func (c *baseCommand) Init(opts ...Option) error {
 
 	c.refWorkspace = &pb.Ref_Workspace{Workspace: workspace}
 
+	// Collect variable values from -var and -varfile flags,
+	// and env vars set with WP_VAR_* and set them on the job
+	vars, diags := variables.LoadVariableValues(c.flagVars, c.flagVarFile)
+	if diags.HasErrors() {
+		// we only return errors for file parsing, so we are specific
+		// in the error log here
+		c.logError(c.Log, "failed to load wpvars file", errors.New(diags.Error()))
+		return diags
+	}
+	c.variables = vars
+
 	// Parse the configuration
 	c.cfg = &config.Config{}
 
@@ -271,6 +283,15 @@ func (c *baseCommand) Init(opts ...Option) error {
 		return err
 	}
 
+	// Create our client (must be done after the project ref and vars are set)
+	if !baseCfg.NoClient {
+		c.project, err = c.initClient(nil)
+		if err != nil {
+			c.logError(c.Log, "failed to create client", err)
+			return err
+		}
+	}
+
 	// Parse app flags
 	if c.flagApp != "" {
 		// NOTE: we could allow app to be specified multiple times in the future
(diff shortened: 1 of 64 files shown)


internal/cli/base.go
```diff
@@ -354,47 +300,47 @@ func (c *baseCommand) Init(opts ...Option) error {
...
+	return nil
+}
 
-			c.refApp = &pb.Ref_Application{
-				Project:     c.cfg.Project,
-				Application: c.cfg.Apps()[0],
-			}
-		}
+// startLocalRunner starts a local runner, and adds its id to the grpc context.
+func (c *baseCommand) startLocalRunner() (*runner.Runner, error) {
```
izaaklauer, line 324: We need to break this out of the project client's init sequence, because in the near future we will need to _use_ the project client to determine if we should start the local runner.

internal/cli/init.go
```diff
@@ -220,11 +222,19 @@ func (c *InitCommand) validateConfig() bool {
...
 	s := sg.Add("Validating configuration file...")
-	cfg, err := c.initConfig(c.fromProject, false)
+	cfg, err := c.initConfig(c.fromProject)
 	if err != nil {
 		c.stepError(s, initStepConfig, err)
 		return false
 	}
+	if cfg == nil {
+		// This should never happen, because if there is no config, init should have created
+		// it an exited earlier.
```
krantzinator, line 231: This comment confuses me. We are in the middle of `init` here, so when exactly should the config have been created earlier? In a different `init`? This one? Is this whole `validateConfig` function pointless?
Also small typo that may be contributing to my confusion.
```suggestion
		// This should never happen, because if there is no config, init should have created
		// it and exited earlier.
```

internal/client/job.go
```diff
@@ -60,14 +60,14 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
 // The receiver must be careful to not block sending to mon as it will block
 // the job state processing loop.
 func (c *Project) doJobMonitored(ctx context.Context, job *pb.Job, ui terminal.UI, monCh chan pb.Job_State) (*pb.Job_Result, error) {
-	// Be sure that the monitor is closed so the reciever knows for sure the job isn't going
+	// Be sure that the monitor is closedthe reciever knows for sure the job isn't going
```
krantzinator:
```suggestion
	// Be sure that the monitor is closed so the receiver knows for sure the job isn't going
```

internal/client/server.go
```diff
@@ -38,8 +38,8 @@ func (c *Project) initServerClient(ctx context.Context, cfg *config) (*grpc.Clie
 
 	// If we're local, then connection is optional.
 	opts := cfg.connectOpts
-	if c.local {
-		log.Trace("WithLocal set, server credentials optional")
+	if !c.noLocalServer {
+		log.Trace("Local server may be created later - server credentials optional")
```
krantzinator, line 42: I'm still not entirely clear what I should expect my server's status to be. Maybe something slightly more specific?
```suggestion
		log.Trace("In-memory server specified; local server may be created later - server credentials optional")
```

internal/cli/base.go
```diff
@@ -232,114 +236,55 @@ func (c *baseCommand) Init(opts ...Option) error {
...
-				c.refApp == nil &&
-				c.flagApp != "" {
-				c.refApp = &pb.Ref_Application{
-					Project:     project.Project,
-					Application: c.flagApp,
-				}
-			}
-		}
+	if (baseCfg.SingleAppTarget || baseCfg.MultiAppTarget) && len(c.refApps) == 0 {
+		err = fmt.Errorf("This command requires an app to be targeted, but no apps were found in project %q.", c.refProject.Project)
```
krantzinator, line 280: Does this miss the flow where the user specifies a `-project` for a multi-app-target command, we don't have a waypoint.hcl to load apps from, and we haven't yet collected all the apps assigned to the specified project after setting the Project on line 260?

internal/client/project.go
```diff
@@ -89,28 +90,33 @@ func New(ctx context.Context, opts ...Option) (*Project, error) {
...
+		return nil, err
 	}
 
-	return client, nil
+	c.activeRunner = r
+
+	// We spin up the job processing here. Anything that spawns jobs (either locally spawned
+	// or server spawned) will be processed by this runner ONLY if the runner is directly targeted.
+	// Because this runner's lifetime is bound to a CLI context and therefore transient, we don't
+	// want to accept jobs that aren't related to local activities (job's queued or RPCs made)
```
krantzinator, line 109:
```suggestion
	// want to accept jobs that aren't related to local activities (jobs queued or RPCs made)
```

krantzinator (commented):
Looking good! I've just got a few relatively minor comments and questions. I haven't manually tested this yet but I can tomorrow!
Overarching question -- since the base.go/Init func is a bit convoluted, would it be helpful to add over-the-top comments to the order we do things in there? I feel like, we spent so much time trying to parse it for this work, and I know others have spent time stuck in the loops before as well, and while this work does a _lot_ to clear up much of that, there are still some tricky bits with knowing at what point a project is definitely set, or apps are definitely set, etc. (This is related to my question added directly to that file as well.) Just a thought!

internal/client/job.go
krantzinator: Hah, I finished my review just a few seconds after you fixed this 😄

internal/cli/init.go
izaaklauer: Ah yup, my understanding is that if you run `waypoint init` with no config, it'll generate a new config file for you, and then exit (here, I think: https://github.com/hashicorp/waypoint/pull/2760/files#diff-37f93a8cf9c002499294368250c727ee1e6360f84486e987c97514a2b7cf737aR143-R147).

So I don't think you'd be able to make it past that check, but then not find a config file in this stage.

internal/client/server.go
izaaklauer: haha, it's actually not entirely clear at this point - I believe our logic is that if we don't detect a context with a server address, and `noLocalServer` isn't set, we'll create a local server.

So at this point, we don't yet know if we'll end up creating an in-memory server, we just know that it _may_ happen.

internal/cli/base.go
izaaklauer: Good catch! Fixing. We'll need to make an api call to list the project here to fix it.

krantzinator (approved):

PR #2762, opened 2021-11-22
ui: make `<JsonViewer>` test more robust

jgwhite:
## Why the change?

Turns out Safari and Chrome emit different error messages when they can’t parse JSON.

## How do I test it?

Try running the test suite in Safari.


gregone (approved):
Good catch!

PR #2772, opened 2021-11-23
gitdirty package for checking if a repo is dirty

izaaklauer:
Introduces a new package, `gitdirty`, that includes two primary functions:

```
func RepoIsDirty(log hclog.Logger, repoPath string, remoteUrl string, remoteBranch string) (bool, error)
func FileIsDirty(log hclog.Logger, repoPath string, remoteUrl string, remoteBranch string, filePath string) (bool, error)
```

The impetus for building this is that if a user runs `waypoint up` from within a project dir with VCS configured (i.e. they've added a remote url and a ref to their project), we want to be able to execute the operation remotely. If the user happens to be sitting on uncommited changes though, or is on a branch other than the projects, or has _any_ local state that's different from what the remote runner will use, they may expect waypoint to use their local changes, and could become confused when their changes aren't reflected in the deploy.

Enter these utils. They detect a difference between a local repo's current state, and the state of a remote on a particular branch.

In the near future, we could use this to detect if only the `waypoint.hcl` file is dirty, and forward its contents for use in the operation in that case (rather than warn).

NOTE: The hundreds of files in this change are mostly fixtures for tests. We've commited a number of repos with their `.git` folders renamed to `DOTgit` to be used as test fixtures.


internal/pkg/vcsutil/vcs_util.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to list git remotes")
+	}
+	if len(remotes) == 0 {
+		return "", fmt.Errorf("no remotes found for repo at path %q", v.path)
+	}
+
+	var matchingRemoteName string
+
+	for _, line := range strings.Split(remotes, "\n") {
```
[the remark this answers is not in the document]
izaaklauer: Done. Go-get was way better.

izaaklauer:
I also now see the fixture-based testing approach that we've taken with `vcs_git`: https://github.com/hashicorp/waypoint/tree/main/internal/config/funcs/testdata

I think I prefer that approach, but not enough to want to retool all these tests, unless someone objects.

catsby (commented):
Could this be added to the `VCSGit` type we have?

- https://github.com/hashicorp/waypoint/blob/59dc7705140dc3c5bef90a140a0a4e0d14954677/internal/config/funcs/vcs_git.go

Granted that type is nested in `internal/config`, maybe we could find a middle ground for the package if it applies to more than configs? That, or the `GitSource` type

- https://github.com/hashicorp/waypoint/blob/59dc7705140dc3c5bef90a140a0a4e0d14954677/internal/datasource/git.go

internal/pkg/vcsutil/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package vcsutil provides utility functions for inspecting the state of a local git repository.
+//
+// Currently, vcsutil executes the `git` command, so git must be installed for it to work.
+package vcsutil
```
mitchellh, line 4: Nitpick, I tend to dislike "util" in the package names cause its kind of a kitchen sink (i.e. https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common). And given this is git-specific for now and indefinitely, may I suggest `gitdirty`?

Its easy enough to change package names later, especially internal ones.

internal/pkg/vcsutil/vcs_util_test.go
```diff
@@ -0,0 +1,362 @@
...
+
+var testHasGit bool
+
+func init() {
+	if _, err := exec.LookPath("git"); err == nil {
+		testHasGit = true
+	}
+}
+
+func generateGitState(branchName string, t *testing.T) (VCSTester, error) {
```
mitchellh, line 29: I recommend just generating this yourself and loading them up. See `internal/config/funcs` for how we do it for Git there. The reason I recommend that is because its a LOT easier to create more test cases cause you just run normal `git` commands and just basically freeze the state (the `.git` directory) at a certain point, rather than rebuilding it all at runtime.

I'd take the helpers from `internal/config/funcs` that translate `DOTgit` into `.git` (so you can commit to this repo)

internal/pkg/vcsutil/vcs_util_test.go
```diff
@@ -0,0 +1,362 @@
...
+
+	require.NoError(err)
+
+	t.Run("Initial state is clean", func(t *testing.T) {
+		dirty, err := v.RepoHasDiff(branchName)
+		require.NoError(err)
+		require.False(dirty)
+	})
+
+	t.Run("Creating (but not committing) a new file is dirty", func(t *testing.T) {
```
mitchellh, line 124: Given the above note (on `DOTgit`), you can then just make a series of exact-state git repo folders rather than dynamically recreating the state in your test.

internal/pkg/vcsutil/vcs_util.go
```diff
@@ -0,0 +1,276 @@
...
+	httpRemoteRegexp *regexp.Regexp
+)
+
+func init() {
+	// Regex matching http/https remotes, tokenizing the unique components for replacement.
+	sshRemoteRegexp, _ = regexp.Compile(`git@(.*?):(.*\.git)`)                 // Example: git@git.test:testorg/testrepo.git
+	httpRemoteRegexp, _ = regexp.Compile(`http[s]?:\/\/(.*?\..*?)\/(.*\.git)`) // Example: https://git.test/testorg/testrepo.git
+}
+
+type VcsChecker struct {
```
mitchellh, line 25: Given this is a Git-specific, I'd just name this `Git` or `Repo` (I prefer the latter and just making this whole package Git-specific. I've done "VCS abstraction" in the past and its generally not worth it)

internal/pkg/vcsutil/vcs_util.go
```diff
@@ -0,0 +1,276 @@
...
+	// remoteName is the name of the remote that corresponds to the url
+	remoteName string
+
+	// internal go-git repo
+	repo *git.Repository
+}
+
+// NewVcsChecker creates a new configured VcsChecker.
+// NOTE: VcsChecker subprocceses to the `git` command. Git must be installed.
+func NewVcsChecker(log hclog.Logger, path string, remoteUrl string) (*VcsChecker, error) {
```
mitchellh, line 40: `New` alone here, probably.

mitchellh, line 40: How do we plan on getting the "remote URL"? 

My recommendation would be to strike this, take an optional local branch, detect the [upstream tracking branch](https://git-scm.com/book/en/v2/Git-Branching-Remote-Branches) and then use that remote URL. That would be the way that git itself (if you run `git status`) detects if you have unpushed changes.

internal/pkg/vcsutil/vcs_util.go
```diff
@@ -0,0 +1,276 @@
...
+		return false, err
+	}
+
+	return diff, nil
+}
+
+// FileHasDiff checks only the specified file for unstaged, staged, and
+// committed (but not pushed) differences between the local VcsChecker.path
+// repo and the specified remote url and branch
+func (v *VcsChecker) FileHasDiff(remoteUrl string, remoteBranch string, filename string) (bool, error) {
```
mitchellh, line 87: Feels a bit dirty that the `New` function requests a remote but then you have to respecify a remote for this API. Why not just use the same remote that was specified for New?

internal/pkg/vcsutil/vcs_util.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return httpsRemote, nil
+}
+
+// getRemoteName queries the repo at VcsChecker.path for all remotes, and then
+// searches for the remote that matches the provided url, returning an error if
+// no remote url is found.
+// It will also attempt to match against different protocols - if an https protocol is
+// specified, if it can't find an exact match, it will look for an ssh-style match (and vice-versa)
+func (v *VcsChecker) getRemoteName(url string) (name string, err error) {
```
mitchellh, line 145: This whole function is pretty spooky to me. We're trying to perfectly capture a way to match some user input URL to a git remote URL using a variety of regular expression based conversions. And we're assuming that HTTP can be converted to SSH and vice versa when they're not at all regular (only GitHub has a regular pattern afaik), so this is depending on the assumption that the remote Git server follows a GitHub-esque model. I find it hard to believe we'd ever get this exactly right (not a slight on YOU personally, literally anyone of us writing this).

I guess this goes back to: should we not use the remote URL at all, and just use upstream tracking branches to detect unpushed changes? That's how I would've done it probably.

If you really want to compare using a URL, I think the only way to do it would be to fetch that remote and do a diff between the git data rather than a metadata diff like this.

mitchellh (changes requested):

izaaklauer:
Oh holy cats, that was a lot more commits than I expected. Doing some git-depuzzling.

internal/pkg/vcsutil/vcs_util.go
izaaklauer: Agreed that the regex is spooky - it's being used to convert between https and ssh style protocols. The idea is that if a user configures an https git remote on their project, but clones with ssh locally (which is what I did when setting up my waypoint project), we don't have a clear way of figuring out which remote in the locally-cloned repository to diff against.

I could potentially see this falling down if someone is using an new exotic repo provider, that uses their own funky https and ssh formatting, that we can't convert between. The failure case still isn't bad though - we just won't be able to determine how to diff, and we'll output that as a warning.

It's worth noting that this logic is only important to feed a warning to users if we detect their local git repo mismatches their remote _and_ we're about to perform an operation locally. We don't intend to use the diff that we generate directly - just see if any diff exists.

internal/pkg/vcsutil/vcs_util.go
izaaklauer: I'm planning to use the remote url configured on the project settings. I'm reluctant to use tracking branches, because in many cases tracking isn't set at all, and if it is it's unlikely to directly track the remote/branch that we care about (i.e. the one the remote runner is going to use)

izaaklauer: I just turned all the methods into functions. We weren't tracking anything stateful - we were just saving the logger and path for convenience, but I think it might have been hurting readability.

izaaklauer:
@catsby Good thought! I think i'd prefer them to be separate at this point - their purposes and methodologies are just different enough that I don't think combining them is necessary. I also renamed our package to have a less generic name. I'm not wholly against it though.

catsby:
@izaaklauer 👍 just wanted to make sure we're not spreading common functionality across multiple packages.

mitchellh (approved):
One important little change but the rest looks good!

PR #2776, opened 2021-11-25
ui/lab: add runners page

jgwhite:
## Why the change?

`ListRunners` recently landed, so this is a quick experiment to try it out.

## What does it look like?

CleanShot 2021-11-25 at 14 27 20@2x

## How do I test it?

1. Check out the branch
2. Boot the dev server against a real waypoint server (`yarn start local`)
3. Visit [localhost:4200/-lab/runners](http://localhost:4200/-lab/runners)
4. Verify you see some runner info

### DIFF
--- a/ui/app/styles/lab.scss
+++ b/ui/app/styles/lab.scss
@@ -1,0 +1,5 @@
+.lab\/header {
+  padding-bottom: scale.$base;
+  margin-bottom: scale.$base;
+  border-bottom: 1px solid rgb(var(--border));
+}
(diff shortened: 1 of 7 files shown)


ui/app/styles/lab.scss
```diff
@@ -0,0 +1,47 @@
+.lab\/header {
+  padding-bottom: scale.$base;
+  margin-bottom: scale.$base;
+  border-bottom: 1px solid rgb(var(--border));
+}
+
+.lab\/runner-table {
```
gregone, line 7: 😅 good thing this is a lab feature because I wouldn't want to generalize `/`-based classnames

gregone (approved):

ui/app/styles/lab.scss
jgwhite: haha, yeah, this is the lab, where we get to try out weird stuff!

PR #2779, opened 2021-11-29
Refactor: moving job initialization work into the client package

izaaklauer:
### What changed

This PR primarily moves the local runner trigger logic out of the CLI and into the project client. Rather than rely on each command to declare ahead of time if they require a runner, and start the runner before the command can run, now we wait until a CLI command actually attempts to spawn a job, and start the runner there.

Upsides to doing this:
- CLI commands don't need to explicitly declare if they need a runner. This is a big win for the `status` command, which only needs a runner conditionally (if `-refresh` is present). 
- The client package seems like the "right" place for this logic to live - as the primary internal interface to the waypoint APIs, it seems like it should be responsible for abstracting these details. With this change, all of the logic around handling the local runner lives inside the client package.

Downsides to doing this:
- The details of `logs` and `exec` _within_ the client package get a little bit messier. Unlike every other command, `logs` and `exec` don't create jobs from the CLI side. Instead, they need to start a local runner, call a server API, and then the server creates the job and decides if it needs to be assigned back to the local runner. So for logs and exec, we need to become ready to accept local jobs, but _not_ create a job. To make that happen, I've split all of the job spoolup logic into a function called `setupLocalJobSystem`, which is normally called within `doJob` but is _also_ called by the logs and exec client operation funcs.

This also sets the stage for us to add more complexity into the `setupLocalJobSystem` sequence - namely checking if VCS is dirty and issuing a warning or forwarding the `waypoint.hcl` file along with the job.

### Smaller incidental changes:
- Adds the operation set flags to logs and exec. They both could trigger operations, and without them you have no control over if those operations happen locally or remotely. As a result, up until now, it hasn't been possible to run exec or logs non-locally! I tested them both, and they work with remotely ODR.
- Wires in `remoteOpPreferred`, initially introduced in https://github.com/hashicorp/waypoint/pull/2749. Operations will now occur remotely if they're likely to work remotely, unless `-local` is specified.
- Misc typo fixes and improved logging

### DIFF
--- a/builtin/aws/lambda/platform_logs.go
+++ b/builtin/aws/lambda/platform_logs.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/builtin/aws/utils"
+	"github.com/pkg/errors"
 )
 
 // Logs fetches logs from cloudwatch
@@ -48,7 +49,7 @@ func (p *Platform) Logs(
 		})
 
 		if err != nil {
-			return err
+			return errors.Wrapf(err, "failed to describe log stream for group %q in region %q", group, p.config.Region)
 		}
 
 		if len(streams.LogStreams) == 0 {
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -134,14 +134,6 @@ func (c *StatusCommand) RefreshApplicationStatus() error {
 	// Get our API client
 	client := c.project.Client()
 
-	// We now know that we're going to need a runner for the refresh job.
-	if !c.remoteOpPreferred() {
-		_, err := c.startLocalRunner()
-		if err != nil {
-			return err
-		}
-	}
-
 	// Get the entire list of apps
 	// Determine project locality
 	// Do the Work (local or remote)
--- a/internal/client/job.go
+++ b/internal/client/job.go
@@ -6,6 +6,10 @@ import (
 	"io"
 	"time"
 
+	"github.com/hashicorp/waypoint/internal/server/grpcmetadata"
+
+	"github.com/pkg/errors"
+
 	"github.com/hashicorp/go-hclog"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -27,23 +31,11 @@ func (c *Project) job() *pb.Job {
 		Application: &pb.Ref_Application{
 			Project: c.project.Project,
 		},
-
-		DataSource: &pb.Job_DataSource{
-			Source: &pb.Job_DataSource_Local{
-				Local: &pb.Job_Local{},
-			},
-		},
 		DataSourceOverrides: c.dataSourceOverrides,
 
 		Operation: &pb.Job_Noop_{
 			Noop: &pb.Job_Noop{},
 		},
-	}
-
-	// If we're not local, we set a nil data source so it defaults to
-	// wahtever the project has remotely.
-	if !c.useLocalRunner {
-		job.DataSource = nil
 	}
 
 	return job
@@ -60,6 +52,32 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
 // The receiver must be careful to not block sending to mon as it will block
 // the job state processing loop.
 func (c *Project) doJobMonitored(ctx context.Context, job *pb.Job, ui terminal.UI, monCh chan pb.Job_State) (*pb.Job_Result, error) {
+
+	if c.useLocalRunner == nil {
+		// Automatically determine if we should use local or remote
+
+		// NOTE(izaak): If in the future we need this in other places too, we should probably cache it on the parent struct.
+		getProjectResp, err := c.client.GetProject(ctx, &pb.GetProjectRequest{Project: c.project})
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get project %s", c.project.Project)
+		}
+		remotePreferred, err := remoteOpPreferred(ctx, c.client, getProjectResp.Project, c.logger)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to determine if job %s should run locally or remotely", job.Application)
+		}
+
+		// Store this for later operations on this same project
+		useLocalRunner := !remotePreferred
+		c.useLocalRunner = &useLocalRunner
+	}
+
+	if *c.useLocalRunner && c.activeRunner == nil {
+		// we need a local runner and we haven't started it yet
+		if err := c.startRunner(); err != nil {
+			return nil, errors.Wrapf(err, "failed to start local runner for job %s, err")
+		}
+	}
+
 	// Be sure that the monitor is closed so the receiver knows for sure the job isn't going
 	// anymore.
 	if monCh != nil {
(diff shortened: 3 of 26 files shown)


mitchellh:
Historical fact: we used to do this, if you go back in git history this was how it originally worked. It was switched to CLI global at some point by @evanphx for some good reason but I don't exactly recall why. :)

internal/cli/status.go
```diff
@@ -134,14 +134,6 @@ func (c *StatusCommand) RefreshApplicationStatus() error {
 	// Get our API client
 	client := c.project.Client()
 
-	// We now know that we're going to need a runner for the refresh job.
```
izaaklauer, line 137: This was my primary impetus for moving the runner startup logic. Once we know we're going to need to run a job, we need to do a bunch of work:

- Figure out if we should run locally or remotely (based on flags, the project, and the waypoint infra's current state)
- If local, start the local runner
- If remote, check if VCS is dirty, and if so warn

Basically all the work that currently is in `setupLocalJobSystem`. If we don't move that logic into the `doJob` path, we'll need to do it here, which feels messier to me than making the logs and exec operation funcs call `setupLocalJobSystem`.

internal/client/job.go
```diff
@@ -55,19 +45,78 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
 	return c.doJobMonitored(ctx, job, ui, nil)
 }
 
+// setupLocalJobSystem does the pre-work required to run jobs locally. This includes:
```
izaaklauer, line 48: Extra scrutiny here please. This whole function smells, but we're going to need to execute this bundle of logic _somewhere_ - either in the status CLI command (if we don't move the runner startup logic), or in the logs and exec operations (as this pr proposes). This is the least-worst option I've come up with, but i'm certainly open to better ideas.

internal/client/operation.go
```diff
@@ -158,20 +160,19 @@ func (c *App) Destroy(ctx context.Context, op *pb.Job_DestroyOp) error {
 	return err
 }
 
-func (c *App) Exec(ctx context.Context, op *pb.Job_ExecOp, mon chan pb.Job_State) error {
-	if op == nil {
-		op = &pb.Job_ExecOp{}
-	}
+func (c *App) Exec(ctx context.Context, ec *execclient.Client) (exitCode int, err error) {
```
izaaklauer, line 163: This function was unused previously. Now, the exec cli calls it to trigger the op. This gives the client package here the opportunity to call the internal `setupLocalJobSystem` method to spool up a local runner that can receive the job if the server decides it's necessary.

internal/client/runner.go
```diff
@@ -15,13 +25,119 @@ func (c *Project) startRunner() (*runner.Runner, error) {
...
+
+// remoteOpPreferred attempts to determine if the current waypoint infrastructure will be successful
+// performing a remote operation against this project. It verifies the project's datasource,
+// it's ODR runner profile, and detects if a remote runner is currently registered.
+// If an operation can occur successfully remotely, we prefer the remote environment for consistency
+// and security reasons.
+//
+// Note that this cannot guarantee that an operation will succeed remotely - the remote environment
+// may not have the right auth configured, the right plugins configured, etc.
+func remoteOpPreferred(ctx context.Context, client pb.WaypointClient, project *pb.Project, log hclog.Logger) (bool, error) {
```
izaaklauer, line 60: This isn't new - just moved from the CLI's `base.go`.

izaaklauer:
> Historical fact: we used to do this, if you go back in git history this was how it originally worked. It was switched to CLI global at some point by @evanphx for some good reason but I don't exactly recall why. :)

Thank you! I spoke to @evanphx about this, and I think there were two reasons it was moved historically:
- To make it so we only start one local runner per cli invocation, and re-use it if we perform multiple operations
- To make `logs` and `exec` work, which require a local runner but don't create a job in the CLI - the server creates one and decides to assign it back to the local CLI runner if necessary.

This change preserves both of those properties. That said, I'm not wedded to this refactor, and if there are other reasons that I'm not seeing to keep the runner startup logic cli-global, we don't need to go down this route.

internal/client/job.go
```diff
@@ -55,19 +45,78 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
...
+// - figure out if jobs should be executed locally or remotely.
+// - if job should be executed locally, start a local runner
+// - FUTURE: if a job should be executed remotely, but local VCS is present and dirty, warn.
+// This lives separately from DoJob because the logs and exec commands need to conditionally warm up the
+// local job infrastructure, but don't actually create a job (the server does).
+func (c *Project) setupLocalJobSystem(ctx context.Context) (isLocal bool, newCtx context.Context, err error) {
+	// Automatically determine if we should use a local or a remote runner
+	newCtx = ctx
```
mitchellh, line 56: Since you're using named return values ANYWAYS, I might suggest adding a defer to always log the results of this function:

```go
defer func() {
  c.log.Named("setupLocalJobSystem").Info("result", "isLocal", ..., etc.)
}()
```

internal/client/job.go
```diff
@@ -55,19 +45,78 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
...
+
+	if c.useLocalRunner == nil {
```
mitchellh, line 58: Would love a comment above here reminding me what this tri-state being `nil` implies. i.e. "nil means we should auto-detect" or something.

internal/client/job.go
```diff
@@ -55,19 +45,78 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
...
+
+		// NOTE(izaak): If in the future we need the full project in other places, we should probably cache it on the parent struct.
```
mitchellh, line 60: I would add logging in this block using the logger on `c`

internal/client/job.go
```diff
@@ -55,19 +45,78 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
 	return c.doJobMonitored(ctx, job, ui, nil)
 }
 
+// setupLocalJobSystem does the pre-work required to run jobs locally. This includes:
+// - figure out if jobs should be executed locally or remotely.
+// - if job should be executed locally, start a local runner
+// - FUTURE: if a job should be executed remotely, but local VCS is present and dirty, warn.
+// This lives separately from DoJob because the logs and exec commands need to conditionally warm up the
+// local job infrastructure, but don't actually create a job (the server does).
+func (c *Project) setupLocalJobSystem(ctx context.Context) (isLocal bool, newCtx context.Context, err error) {
```
mitchellh, line 54: I think this function is sufficiently complicated and important that it may be worth unit testing in isolation.

mitchellh (changes requested):
Overall, excellent, and I raelly like this location for this a LOT better.

krantzinator:
Regarding this
>Adds the operation set flags to logs and exec. They both could trigger operations, and without them you have no control over if those operations happen locally or remotely. As a result, up until now, it hasn't been possible to run exec or logs non-locally! I tested them both, and they work with remotely ODR.
-- what platforms did you test this on? I assume k8s? I may test against nomad and/or ECS just to see if any oddities pop up.

izaaklauer:
@krantzinator I tested on docker, k8s, and lambda - I picked lambda because it's the only plugin (I think) that has plugin-custom implementations for `logs` and `exec` - no other platform executes any plugin-specific code to service `logs` or `exec` (because they have a long-running ceb available to handle that). Nomad/ECS specific testing would be awesome though!

krantzinator (approved):
This feels really good. I think the additional platform testing re: my comment on ODR+exec does not have to hold up merging this PR, since we are reasonably confident in our ODR system+tests.
It looks like you addressed @mitchellh's comments well also. 🎉

builtin/aws/lambda/platform_logs.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/builtin/aws/utils"
+	"github.com/pkg/errors"
```
mitchellh, line 15: Nit: this should go into the new line separated section above this because its not a waypoint import.

mitchellh (approved):
Didn’t run it but looks great.

PR #2782, opened 2021-11-29
internal/{cli,server}: Introduce trigger token type

briancain:
This commit introduces a new token type for trigger URL actions and
includes a way to generate it via the user token CLI subcommand.
Currently, as of this pull request, authing with it does nothing and attempting to use it
against a gRPC endpoint will return an error permission denied.


internal/server/proto/server.proto
```diff
@@ -3971,6 +3974,17 @@ message Token {
...
 
+  // The Trigger message is a kind of token type that is only used for
+  // authenticated trigger URL requests. It should not have any other
+  // authorized access to make requests in any other API endpoint.
+  message Trigger {
+    // The user that initiated the trigger token generation
+    string from_user_id = 1;
+
+    // The login token we'd like to create.
+    Login login = 2;
```
mitchellh, line 3985: I don't think you need any fields in the `Trigger` message at all

mitchellh (approved):
One small change please but the rest is good

internal/server/proto/server.proto
briancain: Ah yeah, I don't think we really need the login message. The user that created the token might be useful though?

mitchellh: yeah that sounds good!

evanphx (approved):

briancain:
Force pushed to fixup the merge conflicts with the server protos.

PR #2784, opened 2021-11-30
Various tweaks to the git diff tools

izaaklauer:
No library survives first contact with a production use-case. Turns out we needed a few more things:

### Remote url normalization

Turns out, the `.git` at the end of a remote url is optional in github! You can use `https://github.com/hashicorp/waypoint-examples.git` and `https://github.com/hashicorp/waypoint-examples` interchangeably. This is true for ssh urls too.

I imagine it won't be uncommon for people to omit the trailing `.git` when setting up their project in waypoint. To fix this, i'm now normalizing all remote urls (i.e. removing any trailing `.git`) before doing any comparison.

### Default branch detection

Also turns out, you don't need to give waypoint a remote ref - if you skip it, waypoint will just clone the repo, and you'll get whatever the default HEAD is. For diffing purposes though, we need to know what that default HEAD is.

I'm using `git` directly rather than go-git for this, because we need to actually talk to the remote to discover the default HEAD, and setting up registry auth for go-git was painful. I'm not in love with relying on the output of `git show remote <remote>` either though.

### Recursive file diffs

If you've set a git path on your waypoint project, it would be nice if we only diff against that path. This actually worked previously, but i've added some test cases just to nail it down.


mitchellh (approved):
Note that we do figure out the HEAD in our cloning code if you ever dive back into `go-git`. Its also very cheap cause you don't need to clone anything you can merely request refs (and you can even filter those refs to only HEAD).

krantzinator (approved):

briancain (approved):

PR #2792, opened 2021-12-02
Set the assigned runner ID on the job

izaaklauer:
We had this field, but weren't using it! I'd like to use it to give the CLI more context as to where a job is running.


mitchellh (changes requested):
All changes in `state` (and esp around jobs) must have a unit test attached to them. Please do that but the impl looks fine.

mitchellh (approved):

briancain (approved):
Nice

PR #2793, opened 2021-12-02
ui: timeline component added to details pages

### LINKED ISSUE #2775: ui: add timeline component to details pages
In the current UI, we have a few places where we can see the connections between a build - deployment - release, but these are not prominent. Users may want to find which build was used for this deployment, etc, and having these links clearly displayed in the UI will help.

Thus, we would like to add a timeline component to each artifact's details page to show clearly the succession of events.

(see design documents for actual UI)

sabrinako:
## Notes
Currently does not show destroyed deployments in the timeline, @almonk is still working on designs for the best way to show this

## How to test
1. Pull down the branch `git checkout ui/timeline-component`
2. `yarn start`
3. Check the build, deployment, and release details pages for the timelines
_Optional_
4. Spin up a Waypoint server + make some builds, deployments, and releases
5. `ember serve local`
6. Check the build, deployment, and release details pages for the timelines

## Screenshot
Screen Shot 2021-12-14 at 9 28 35 AM

### DIFF
--- a/ui/app/components/timeline.ts
+++ b/ui/app/components/timeline.ts
@@ -62,6 +62,7 @@ export default class Timeline extends Component<TimelineArgs> {
         isCurrentRoute: this.areWeHere(key),
       } as ArtifactModel);
     }
+    artifactsList.sort((a, b) => (a.type > b.type ? 1 : -1));
     return artifactsList;
   }
 }
--- a/ui/app/components/timeline.ts
+++ b/ui/app/components/timeline.ts
@@ -41,14 +41,8 @@ export default class Timeline extends Component<TimelineArgs> {
   @service router!: RouterService;
 
   areWeHere(currentArtifactKey: string): boolean {
-    let entry = Object.entries(_TYPE_ROUTES).find(([_, value]) =>
-      this.router.currentRouteName.includes(value)
-    );
-
-    if (entry) {
-      return entry[0] === currentArtifactKey;
-    }
-    return false;
+    let routeForKey = _TYPE_ROUTES[currentArtifactKey];
+    return this.router.currentRouteName.includes(routeForKey);
   }
 
   get artifacts(): ArtifactModel[] {
--- a/ui/app/components/timeline.ts
+++ b/ui/app/components/timeline.ts
@@ -1,0 +1,67 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import RouterService from '@ember/routing/router-service';
+import ApiService from 'waypoint/services/api';
+import { Status } from 'waypoint-pb';
+
+const _TYPE_TRANSLATIONS = {
+  build: 'page.artifact.timeline.build',
+  deployment: 'page.artifact.timeline.deployment',
+  release: 'page.artifact.timeline.release',
+};
+
+const _TYPE_ROUTES = {
+  build: 'workspace.projects.project.app.build',
+  deployment: 'workspace.projects.project.app.deployment.deployment-seq',
+  release: 'workspace.projects.project.app.release',
+};
+interface ArtifactModel {
+  sequence: number;
+  type: string;
+  route: string;
+  status?: Status.AsObject;
+  isCurrentRoute: boolean;
+}
+
+export interface TimelineModel {
+  build?: TimelineArtifact;
+  deployment?: TimelineArtifact;
+  release?: TimelineArtifact;
+}
+interface TimelineArtifact {
+  sequence: number;
+  status: Status.AsObject | undefined;
+}
+interface TimelineArgs {
+  model: TimelineModel;
+}
+
+export default class Timeline extends Component<TimelineArgs> {
+  @service api!: ApiService;
+  @service router!: RouterService;
+
+  areWeHere(currentArtifactKey: string): boolean {
+    let entry = Object.entries(_TYPE_ROUTES).find(([_, value]) =>
+      this.router.currentRouteName.includes(value)
+    );
+
+    if (entry) {
+      return entry[0] === currentArtifactKey;
+    }
+    return false;
+  }
+
+  get artifacts(): ArtifactModel[] {
+    let artifactsList: ArtifactModel[] = [];
+    for (let key in this.args.model) {
+      artifactsList.push({
+        sequence: this.args.model[key].sequence,
+        type: _TYPE_TRANSLATIONS[key],
+        route: _TYPE_ROUTES[key],
+        status: this.args.model[key].status,
+        isCurrentRoute: this.areWeHere(key),
+      } as ArtifactModel);
+    }
+    return artifactsList;
+  }
+}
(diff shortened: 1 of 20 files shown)


ui/app/components/timeline.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export default class Timeline extends Component<TimelineArgs> {
+  @service api!: ApiService;
+  @service router!: RouterService;
+
+  get artifacts(): Artifact[] {
+    let timelineObjects = [] as Artifact[];
+    let preload = this.args.currentArtifact.preload;
+    let build = preload?.build ?? null;
+    let deployment = preload?.deployment ?? null;
```
jgwhite, line 39: To get around the TypeScript errors here, I’d take a look at [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). Really powerful/useful feature of TypeScript.

jgwhite:
@sabrinako regarding the wonky timestamps, a couple things to keep in mind:

1. Do we show `status.startTime` or `status.completeTime`? Does it depend on the current state of the operation?
2. We do some monkey business to treat a Build and its PushedArtifact as one thing. You’ll see evidence of this in the template for the build page. Do we need to account for that too?

ui/app/routes/workspace/projects/project/app/deployment/deployment-seq.ts
```diff
@@ -32,16 +37,46 @@ export default class DeploymentDetail extends Route {
...
+      timeline.release = {
+        sequence: release.sequence,
+        status: release.status,
+      };
+    }
+
+    return { ...deployment, release, timeline };
+  }
+
+  resetController(_: Controller, isExiting: boolean, transition: Transition): void {
```
sabrinako, line 74: Fixes a bug if you clicked on the 'deployments' tab when already being there, you get redirected to a blank page.

jgwhite:
This feedback might be a little wacky, but I wonder if the timeline should run in the other direction, just so it matches the nav on the left?

As it is, they run in opposite directions:

CleanShot 2021-12-14 at 16 41 10@2x

sabrinako:
> This feedback might be a little wacky, but I wonder if the timeline should run in the other direction, just so it matches the nav on the left?
> 
> As it is, they run in opposite directions:
> 
> CleanShot 2021-12-14 at 16 41 10@2x

I thought about that as well, but I went with how it's structured in the Figma. @almonk any thoughts?

ui/app/routes/workspace/projects/project/app/build.ts
```diff
@@ -34,8 +39,15 @@ export default class BuildDetail extends Route {
...
 
   async model(params: Params): Promise<Model> {
-    let { builds } = this.modelFor('workspace.projects.project.app') as AppRouteModel;
+    let { builds, deployments, releases } = this.modelFor('workspace.projects.project.app') as AppRouteModel;
     let buildFromAppRoute = builds.find((obj) => obj.sequence === Number(params.sequence));
+    let deploymentFromAppRoute = deployments.find((obj) => {
+      if (obj.preload && obj.preload.artifact) {
+        return obj.preload.artifact.id === buildFromAppRoute?.pushedArtifact?.id;
+      }
+      return obj.artifactId === buildFromAppRoute?.pushedArtifact?.id;
```
jgwhite, line 48: I suspect `preload.artifact.id` and `artifactId` will always concur when working with the real API. Maybe something we can fix in Mirage? What do you think? Worth digging into it?

ui/app/routes/workspace/projects/project/app/build.ts
```diff
@@ -51,6 +63,28 @@ export default class BuildDetail extends Route {
...
+
+    if (releaseFromAppRoute) {
+      timeline.release = {
+        sequence: releaseFromAppRoute.sequence,
+        status: releaseFromAppRoute.status,
+      };
+    }
+
+    result.timeline = timeline;
+
```
jgwhite, line 87: Note for the future: we could potentially do this cross-referencing server-side in [the UI_ methods](https://github.com/hashicorp/waypoint/blob/0ee3f72cee308cd9d3a14b5691f324bb69099571/internal/server/proto/server.proto#L369-L376).

ui/app/routes/workspace/projects/project/app/deployment/deployment-seq.ts
jgwhite: Good catch!

ui/app/services/api.ts
```diff
@@ -46,6 +46,11 @@ const protocolVersions = {
   'client-version': 'ui-0.0.1',
 };
 
+export type BuildExtended = Build.AsObject & {
+  statusReport?: StatusReport.AsObject;
+  pushedArtifact?: PushedArtifact.AsObject;
+};
+
```
jgwhite, line 53: Did we end up using this anywhere?

ui/app/styles/_variables.scss
```diff
@@ -24,6 +24,8 @@ $button-shadow: 0 3px 2px rgba(var(--shadow), 0.2);
   --card: #{dehex(color.$white)};
   --highlighted-list-item: #{dehex(color.$ui-cool-gray-100)};
   --panel: #{dehex(color.$ui-cool-gray-050)};
+  --timeline-dot: #{dehex(color.$ui-gray-400)};
+  --timeline-dash: #{dehex(color.$ui-gray-100)};
```
jgwhite, line 28: Do we need dark mode equivalents of these? (Can wait until the polishing phase, of course 😎)

ui/app/styles/components/image-ref.scss
```diff
@@ -4,6 +4,7 @@
 
   &__tag {
     margin-left: 7px;
+    overflow-wrap: anywhere;
```
jgwhite, line 7: Didn’t realize `anywhere` was a possible value here. TIL!

ui/app/styles/components/timeline.scss
```diff
@@ -0,0 +1,58 @@
+.timeline {
+  li {
+    display: grid;
+    grid-template-columns: scale.$sm-1 auto;
+    grid-template-rows: max-content 20px;
```
jgwhite, line 5: Loving this use of grid (although I must admit I don’t understand it yet!)

jgwhite (approved):
This is looking *wonderful*. I only have superficial comments.

Following this PR, we should discuss as a team how we can make our routes a bit less busy. The `afterModel` hooks are a totally logical place to cross-reference and prepare model data, but they are getting a little large and complex. I’m not sure what the alternative is, but together I’m sure we can dream up some neat ideas.

ui/app/routes/workspace/projects/project/app/build.ts
sabrinako: As discussed on sync, this will be a TODO and handled later since it works as-is

PR #2794, opened 2021-12-02
feature: Introduce trigger URL state store with corresponding service API and state package funcs

briancain:
This pull request introduces the concept of Trigger URL state into the Waypoint server state store. It's the main building blocks for WP-107 for configuration through the grpc API. 

Sub-PR: The CLI Implementation https://github.com/hashicorp/waypoint/pull/2807


internal/server/proto/server.proto
```diff
@@ -3178,6 +3200,98 @@ message ReleaseExpanded {
...
+* Trigger URLs
+********************************************************************/
+
+// The Trigger message is a representation of a URL that can be requested
+// for invoking specific lifecycle operations on a projects applications. These
+// trigger URLs are intended to be used in a CI system for easy configuration.
+// The user is expected to configure and generate the URL ahead of time, and
+// Waypoint will generate a trigger based on the configuration here and return
+// a URL to make requests with.
+message Trigger {
```
briancain, line 3213: Oops, writing the CLI implementation made me realize this message needs an operation reference for what the trigger will be executing. Will update this PR with that.

briancain: Done

internal/server/proto/server.proto
```diff
@@ -849,6 +864,13 @@ message Ref {
     string id = 1;
     string name = 2;
   }
+
+  // Trigger references a Trigger message value to be used for a given operation.
+  // It can be looked up by name or id. Id takes priority if both are set.
+  message Trigger {
+    string id = 1;
+    string name = 2;
```
mitchellh, line 872: We should use a `oneof` here.

internal/server/singleprocess/state/trigger.go
```diff
@@ -0,0 +1,383 @@
...
+
+	err := s.db.Update(func(dbTxn *bolt.Tx) error {
+		if t.Project == nil {
+			return status.Error(codes.FailedPrecondition, "A Project is required to create a trigger")
+		}
+
+		if t.Id != "" {
+			var err error
+			_, err = s.triggerGet(dbTxn, memTxn, &pb.Ref_Trigger{Id: t.Id})
+			if err != nil {
```
mitchellh, line 36: Am I reading this right? This prevents upserts, only updates. So if you specify an ID, you must be updated an existing trigger? If that's true, that doesn't match our other `Put` functions, and you may want to split this into a `Create` and `Update` separately.

internal/server/proto/server.proto
```diff
@@ -344,6 +343,22 @@ service Waypoint {
...
+  rpc UpsertTrigger(UpsertTriggerRequest) returns (UpsertTriggerResponse);
+
+  // GetTrigger returns a requested trigger message. Or an error if it does not exist.
+  rpc GetTrigger(GetTriggerRequest) returns (GetTriggerResponse);
+
+  // DeleteTrigger takes a trigger id and deletes it, if it exists.
+  rpc DeleteTrigger(DeleteTriggerRequest) returns (google.protobuf.Empty);
+
+  // ListTriggers takes a request filter, and returns any matching existing triggers
+  rpc ListTriggers(ListTriggerRequest) returns (ListTriggerResponse);
```
mitchellh, line 360: I recommend in the future when you support triggering the triggers to make an RPC endpoint `RunTrigger` or something and that is what the HTTP side calls too. But, it may be worth doing the RPC alone for simplicity sake and then adding HTTP on top of that.

mitchellh (changes requested):
Looks very good!

I didn't read your state implementation in detail because I want to ask you write tests for it first. The state package is _very well_ tested and the tests are in `internal/server/statetest` or something like that.

internal/server/proto/server.proto
briancain: That's a great idea, I'll plan to tackle that next!

internal/server/proto/server.proto
briancain: Now that I've fleshed out the CLI a bit, I'm thinking of dropping `name` completely and just referring to name. Thoughts? Mostly because we probably don't want to make `name` unique across server installations, just `id`.

internal/server/singleprocess/state/trigger.go
briancain: I think that's a mistake on my part. We should be ok to allow upserts at the state API level here.

internal/server/proto/server.proto
mitchellh: "just referring to ID"* you mean? I think that is better. We have a "description" already for human frindly understanding of why a trigger exists. And ID satisfies our security-through-obscurity for non-authenticated triggers.

briancain: Oops, typo! Yep, meant just referring to *id*.

internal/server/singleprocess/state/trigger.go
briancain: Fixed this up to be a proper Put.

internal/server/proto/server.proto
```diff
@@ -3178,6 +3199,113 @@ message ReleaseExpanded {
...
+  // name can be user defined, or auto generated.
+  string name = 2;
+
+  // description is user defined, describes the purpose
+  // of the trigger.
+  string description = 3;
+
+  // labels that allow users to group triggers together that are
+  // related. There are no default labels.
+  repeated string labels = 4;
```
izaaklauer, line 3227: Curious why this isn't a map<string,string> like other labels.

internal/server/ptypes/status_report.go
```diff
@@ -53,7 +53,7 @@ func ValidateStatusReportRules(v *pb.StatusReport) []*validation.FieldRules {
 	}
 }
 
-// ValidateUpsertArtifactRequest
```
izaaklauer, line 56: nice

izaaklauer (approved):
👍

internal/server/proto/server.proto
briancain: Those labels are more key=>val (literally a map), where as the intention here is a flat list, similar to how you'd label something in GitHub. Just a way to categorize something (like, list me all the trigger URLs for `"test"` or `"production"`, etc).

mitchellh: I wonder if we should call these "tags" or something since "labels" is used so heavily throughout WP otherwise to mean a map of key/value.

mitchellh (changes requested):
Almost there.

internal/server/proto/server.proto
briancain: That change makes sense to me!

briancain:
FYI: Rebased to fix up the server proto gen conflict off `main`.

krantzinator (approved):
Looks sensible to me! I'll try harder to break it with the next round of PRs. 😈

PR #2795, opened 2021-12-03
Display the runner type during CLI operations

izaaklauer:
Waypoint will soon automatically choose if operations will occur locally or remotely. Currently though, users have no visibility into where their operations take place. If there is some kind of error during the operation, users will find it valuable to know where that error occurred. For example, if it's a permissions error, they'll need to either update their local credentials providers, or go modify an ODR profile, depending on where the job ran.

With this change, the runner proto gets a new field, `kind`, which is either local, odr, or remote. Runners will know their own type on startup, and communicate that type back to the server when they first register.

When the CLI creates a job, it watches the job events to see when job's `assigned_runner` ref is set. It then uses that runner ID to query the server for the complete runner proto, and displays the runner location.

## How to verify

Run a local build. Note the line `Performing operation locally`
```
$ waypoint build -local

» Building web...
Performing operation locally
✓ Building image...
 │ Step 1/3 : FROM nginx:stable
 │  ---> c8d03f6b8b91
 │ Step 2/3 : COPY ./public/ /var/www
 │  ---> Using cache

```

Turn on a remote runner (but don't set a default or project runner profile), and run the op again with `-local-false` to force the job to execute on the "static" remote runner.
Note the line `Performing this operation on a remote runner with id "01FP0N64TDN3QN4D0M83RGRPD4"`
```
$ waypoint build -local=false

  Performing this operation on a remote runner with id "01FP0N64TDN3QN4D0M83RGRPD4"

» Cloning data from Git
  Ref: us-east-1

```

Set up an ODR profile.
Note the line `Performing operation on docker with runner profile test`
```
$ ~/dev/waypoint/waypoint build -local=false

  Performing operation on docker with runner profile test

» Cloning data from Git
  Ref: us-east-1
...
```

### Future considerations

If an ODR operation fails, there's a good chance there will be some useful context in the ODR pod's logs, or it's env vars, or platform-specific setup. Even with these changes, there is no good way to figure out which odr instance (i.e. k8s pod) in particular executed the op. The best thing i've found is to compare launch timestamps, or grep the logs of _all_ odr pods for strings I expect to see.

I don't see any obvious path to get that context. 

Hacky idea: We could do something hacky, like have the task launcher plugin set some `WAYPOINT_TASK_INSTANCE_NAME` env var, and teach the runner to discover that and report it back to the server. 

Better (but harder) idea: We could have the task plugin report DeclaredResources via an outparamater (similar to how `status` works), and save those on a new `task` protobuf, and then tie that `task` id back to the `runner` proto somehow. I think this would probably require an RFC.

If anyone has any better ideas, i'm all ears.


internal/server/singleprocess/service_job.go
```diff
@@ -337,7 +337,7 @@ func (s *service) onDemandRunnerStartJob(
 
 	// Arguments for the runner image. Waypoint is ALWAYS assumed to be
 	// the entrypoint for ODR images.
-	args := []string{"runner", "agent", "-vv", "-id", runnerId, "-odr"}
+	args := []string{"runner", "agent", "-vv", "-id", runnerId, "-odr", "-odr-profile-id", od.Id}
```
izaaklauer, line 340: I don't love this way of setting metadata on the runner - it feels very indirect. It would feel much cleaner to me if we called a server api here-ish to create a new "pending" `runner` record of type odr, then just start the runner with the predetermined ID. Then, when the runner registered itself to the server, it would just set its state from "pending" to "active", or something.

It also feels a little weird from a security perspective - fresh runners have the opportunity to lie about their profile id and odr-ness if we rely on them to tell the server. 

If and when it's time to add the next runner flag like this, I think we should revisit this pattern.

internal/server/proto/server.proto
```diff
@@ -2144,16 +2146,33 @@ message Runner {
...
-  // server to change some other behavior:
-  //
-  //   * The server will give ODRs project/app-scoped config if it exists.
-  //   * The server will never assign more than one job to this runner.
-  //     This is also enforced in the runner client-side but the server also
-  //     does this out of caution.
-  //
-  bool odr = 4;
+  // Location indicates what kind of runner this is
+  oneof type {
```
izaaklauer, line 2150: Happy to bikeshed on names here. I don't love `type`.

briancain: We use `kind` for token types, maybe we could use that here too?

izaaklauer: Oh good idea

internal/cli/base.go
```diff
@@ -471,12 +471,7 @@ func remoteOpPreferred(ctx context.Context, client pb.WaypointClient, project *p
 	}
 	hasRemoteRunner := false
 	for _, runner := range runnersResp.Runners {
-		if !runner.Odr {
```
izaaklauer, line 474: This is a great side-effect of this change! We can be more confident now about deciding if a remote op is likely to work.

internal/cli/runner_agent.go
```diff
@@ -272,6 +276,14 @@ func (c *RunnerAgentCommand) Flags() *flag.Sets {
...
 			Target: &c.flagODR,
 			Usage:  "Indicates to the runner it's operating as an on-demand runner.",
+			Hidden: true,
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:   "odr-profile-id",
+			Target: &c.flagOdrProfileId,
+			Usage:  "The ID of the odr profile used to create the task that is running this runner.",
+			Hidden: true,
```
krantzinator, line 286: Should these lines all be indented or is GitHub just being weird about how they're displayed?

internal/client/job.go
```diff
@@ -325,6 +328,40 @@ func (c *Project) queueAndStreamJob(
...
+				event.State.Job.AssignedRunner != nil &&
+				assignedRunner == nil {
+
+				assignedRunner = event.State.Job.AssignedRunner
+
+				runner, err := c.client.GetRunner(ctx, &pb.GetRunnerRequest{RunnerId: assignedRunner.Id})
+				if err != nil {
+					c.UI.Output("Failed to inspect the runner (id %q) assigned for this operation: %s", assignedRunner.Id, err, terminal.WithErrorStyle())
+				}
+				switch runnerType := runner.Kind.(type) {
```
krantzinator, line 343: I love this because it's like self-documenting of the different runners in our runner system.

krantzinator (approved):
Code paths look generally good! I can merge this next week when you're out if you'd like to wait for @evanphx's feedback.

internal/server/singleprocess/service_job.go
mitchellh: We need to add some sort of runner "adoption" or "activation" step for some future work we have planned for exactly the reason you stated. Ideally: runners can relatively easy ATTEMPT to register with the server, but a human (or machine via more privileged API) has to "adopt" the runner to activate it. I think we can do this using some sort of certificate exchange mechanism (so if the runner restarts or something it can prove its adoption for a period of time). I have various thoughts here.

For now, I think this is okay.

internal/cli/base.go
```diff
@@ -471,12 +471,7 @@ func remoteOpPreferred(ctx context.Context, client pb.WaypointClient, project *p
...
-			// NOTE(izaak): There is currently no way to distinguish between a remote runner and a CLI runner.
-			// So if some other waypoint client is performing an operation at this moment, we will interpret
-			// that as a remote runner, and this will return a false positive.
-
-			// Also note that this is designed to run before se start our own CLI runner.
+		if _, ok := runner.Kind.(*pb.Runner_Remote_); ok {
```
mitchellh, line 474: I've always done `v, ok := ... ; ok && v != nil`. What you did is PROBABLY FINE, but talking out loud I've always wondered if this is totally fine haha. You can keep it how it is, just thinking out loud.

internal/client/job.go
```diff
@@ -325,6 +328,40 @@ func (c *Project) queueAndStreamJob(
...
+			if event.State != nil &&
+				event.State.Job != nil &&
+				event.State.Job.AssignedRunner != nil &&
+				assignedRunner == nil {
+
+				assignedRunner = event.State.Job.AssignedRunner
+
+				runner, err := c.client.GetRunner(ctx, &pb.GetRunnerRequest{RunnerId: assignedRunner.Id})
+				if err != nil {
+					c.UI.Output("Failed to inspect the runner (id %q) assigned for this operation: %s", assignedRunner.Id, err, terminal.WithErrorStyle())
```
mitchellh, line 341: Should you be continuing or breaking out of the switch here? It looks like this will fall through and the switch on the next line will panic on error with `runner == nil`.

internal/client/job.go
```diff
@@ -325,6 +328,40 @@ func (c *Project) queueAndStreamJob(
...
+				}
+				switch runnerType := runner.Kind.(type) {
+				case *pb.Runner_Local_:
+					c.UI.Output("Performing operation locally", terminal.WithInfoStyle())
```
mitchellh, line 345: All these should use `ui` not `c.UI`. I don't remember why, but we should be using the passed in UI that is identical to the one we use for other output, shouldn't we?

internal/server/proto/server.proto
```diff
@@ -2134,6 +2134,8 @@ message GetJobStreamResponse {
 ********************************************************************/
 
 message Runner {
+  bool deprecated_is_odr = 4;
```
mitchellh, line 2137: I usually throw these on the end so I can add some comments of what these used to be.

internal/server/proto/server.proto
```diff
@@ -2144,16 +2146,34 @@ message Runner {
...
-  //
-  //   * The server will give ODRs project/app-scoped config if it exists.
-  //   * The server will never assign more than one job to this runner.
-  //     This is also enforced in the runner client-side but the server also
-  //     does this out of caution.
-  //
-  bool odr = 4;
+  // Kind indicates what variety of runner this is
+  oneof kind {
+
```
mitchellh, line 2151: Strike the newline

internal/server/proto/server.proto
```diff
@@ -2144,16 +2146,34 @@ message Runner {
...
+    //
+    //   * The server will give ODRs project/app-scoped config if it exists.
+    //   * The server will never assign more than one job to this runner.
+    //     This is also enforced in the runner client-side but the server also
+    //     does this out of caution.
+    //
+    ODR odr = 5;
+
+    // local indicates this runner was created by a cli instantiation
+    Local local = 6;
```
mitchellh, line 2164: Instead of empty messages you can use `google.protobuf.Empty` here if you want. You can change to a message type later and be fully backwards compatible on the wire. Up to you.

mitchellh (changes requested):
One critical thing, the rest are just things to think about. Looks wonderful. Great job!

internal/cli/base.go
izaaklauer: oh ha interesting - i assume `v == nil` when `runner.Kind` has never been set - I think it's OK here because nil will still bypass this block.

mitchellh: Well, you can do `runner.Kind = (*pb.Runner_Remote_)(nil)` to make `runner.Kind != nil` but once you cast it is `== nil`. I can’t think of any time this would happen without someone acting weird though. (This is the interface vs typed nil)

internal/cli/runner_agent.go
izaaklauer: Github weirdness - it's just one stop.

internal/client/job.go
izaaklauer: Oh that's a very good catch, thank you.

internal/server/proto/server.proto
izaaklauer: If they were intended solely as markers i'd switch, but I imagine we'll want to add more fields to them in the not-too-distant future.

briancain (approved):
Looks good to me! 👍🏻

mitchellh (approved):

krantzinator:
Per discussion with Izaak last week, I'm going to merge this in since it has all the necessary approvals and tests are passing! 🎉

PR #2797, opened 2021-12-03
Add Workspace Create command

catsby:
This PR adds a simple `workspace create <name>` CLI command for creating workspaces with the `UpsertWorkspace` endpoint. 

Workspaces are a bit limited at the moment, so there really isn't much to do here. The `Upsert` API call does not currently indicate if the workspace already existed or not, so we simply display the same generic "registered with server" response.

### DIFF
--- a/internal/cli/workspace_create.go
+++ b/internal/cli/workspace_create.go
@@ -1,0 +1,96 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/clierrors"
+	"github.com/hashicorp/waypoint/internal/pkg/flag"
+	pb "github.com/hashicorp/waypoint/internal/server/gen"
+	"github.com/posener/complete"
+)
+
+type WorkspaceCreateCommand struct {
+	*baseCommand
+
+	flagWorkspaceName string
+}
+
+func (c *WorkspaceCreateCommand) Run(args []string) int {
+	// Initialize. If we fail, we just exit since Init handles the UI.
+	flagSet := c.Flags()
+	if err := c.Init(
+		WithArgs(args),
+		WithFlags(flagSet),
+		WithNoConfig(),
+		WithNoLocalServer(),
+	); err != nil {
+		return 1
+	}
+	args = flagSet.Args()
+
+	if len(args) != 1 {
+		c.ui.Output(c.Help(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	workspaceName := args[0]
+
+	client := c.project.Client()
+	resp, err := client.UpsertWorkspace(c.Ctx, &pb.UpsertWorkspaceRequest{
+		Workspace: &pb.Workspace{
+			Name: workspaceName,
+		},
+	})
+	if err != nil {
+		c.ui.Output(
+			clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	// this is unlikely to happen with a nil error above, but added here to be
+	// defensive.
+	if resp.Workspace == nil {
+		c.ui.Output(
+			fmt.Sprintf("no workspace returned for create command with name %q", workspaceName),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	// the UpsertWorkspace call is idempotent, and does not return any
+	// indication if the workspace was created or if it already existed, so we
+	// report a generic response
+	c.ui.Output("Workspace registered with the server")
+
+	return 0
+}
+
+func (c *WorkspaceCreateCommand) Flags() *flag.Sets {
+	return c.flagSet(0, nil)
+}
+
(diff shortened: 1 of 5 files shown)


internal/cli/workspace_create.go
```diff
@@ -0,0 +1,96 @@
...
+	return "Create a workspace with a given name."
+}
+
+func (c *WorkspaceCreateCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint workspace create <name>
+
+  Create a workspace in Waypoint with the given value. If a workspace with the
+	given name already exists, no error will be returned. This command ignores
+	the -workspace flag.
```
izaaklauer, line 93: I think adding -workspace to our global flag set is starting to be more of a pain than a convenience. It's present on `config sync` too, but isn't respected there. It might be nice to make it something CLI commands can opt into.

internal/cli/workspace_create.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (c *WorkspaceCreateCommand) Run(args []string) int {
+	// Initialize. If we fail, we just exit since Init handles the UI.
+	flagSet := c.Flags()
+	if err := c.Init(
+		WithArgs(args),
+		WithFlags(flagSet),
+		WithNoConfig(),
+		WithNoLocalServer(),
```
izaaklauer, line 26: I can imagine workspaces could be useful to users who only use the local server. Will it not work with a local server?

internal/cli/workspace_create.go
catsby: Agreed! This is a "global" API call so `-workspace` just doesn't mean anything here...

internal/cli/workspace_create.go
catsby: Thanks for bringing this up! I believe I misunderstood what the local server was doing (or supposed to do/offer). After more reading and some discussion, I think this is a mistake. I removed it in 98f0fe07b

internal/cli/workspace_create.go
```diff
@@ -0,0 +1,95 @@
...
+	return "Create a workspace with a given name."
+}
+
+func (c *WorkspaceCreateCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint workspace create <name>
+
+  Create a workspace in Waypoint with the given value. If a workspace with the
+	given name already exists, no error will be returned. This command ignores
+	the -workspace flag.
```
briancain, line 92:
```suggestion
  Create a workspace in Waypoint with the given value. If a workspace with the
  given name already exists, no error will be returned. This command ignores
  the -workspace flag.
```

Just some formatting fixups. This will make the auto-gen docs happy. Also I think you'll need to auto-generate the website docs for this command? I'm surprised the website mdx PR tests didn't fail....you might have to add it to the `commands-nav-data.json` file

internal/cli/workspace_create.go
```diff
@@ -0,0 +1,95 @@
...
+			fmt.Sprintf("no workspace returned for create command with name %q", workspaceName),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	// the UpsertWorkspace call is idempotent, and does not return any
+	// indication if the workspace was created or if it already existed, so we
+	// report a generic response
+	c.ui.Output("Workspace registered with the server")
```
briancain, line 65:
```suggestion
	c.ui.Output("Workspace %q registered with the server", workspaceName, terminal.WithSuccessStyle())
```

internal/cli/workspace_create.go
```diff
@@ -0,0 +1,95 @@
...
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+
+	// this is unlikely to happen with a nil error above, but added here to be
+	// defensive.
+	if resp.Workspace == nil {
+		c.ui.Output(
+			fmt.Sprintf("no workspace returned for create command with name %q", workspaceName),
```
briancain, line 56: The `Output` function can format so so need for `fmt.Sprintf` here

briancain (changes requested):
Looks good Clint! Just a few minor comments from me 👍🏻

internal/cli/workspace_create.go
```diff
@@ -0,0 +1,95 @@
...
+	return c.Flags().Completions()
+}
+
+func (c *WorkspaceCreateCommand) Synopsis() string {
+	return "Create a workspace with a given name."
+}
+
+func (c *WorkspaceCreateCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint workspace create <name>
```
briancain, line 88:
```suggestion
Usage: waypoint workspace create [options] <name>
```

internal/cli/workspace_create.go
catsby: Updated in c169c87b5

internal/cli/workspace_create.go
catsby: This command doesn't not currently offer any options. It's a global command, so `-w`, `-a`, or `-p` don't do anything.

briancain (approved):
Looks great! Thanks for the updates 👍🏻

PR #2799, opened 2021-12-03
CLI: emit dirty vcs warning for remote operations.

izaaklauer:
Merge after https://github.com/hashicorp/waypoint/pull/2779

This change wires in the vcs detection library introduced in https://github.com/hashicorp/waypoint/pull/2772.

## What problem is this solving

Now that waypoint runs operations remotely wherever possible, users who were previously used to local operations will be exposed to remote ops. They may be used to modifying their `waypoint.hcl` file, running something like `waypoint up`, and expecting the runner to take the change. The remote runner will run off a fresh clone of their repo though, and won't see the change.

We can't _just_ look at the waypoint.hcl file though, because they may have template file directives in there, and some plugins read other files (e.x. docker reads the Dockerfile). There's no good way to detect which files we need to be clean to run a safe operation, so we need to warn if anything is dirty.

## What changed

If we're going to execute an operation remotely, and we're in a project directory, and there is any difference detected between that project dir and the vcs state the remote runner will use, we emit a big warning.

## How to verify

Set up a project for remote ops, dirty the project repo by modifying the `waypoint.hcl` file, run an op, and observe the warning:

```
$ waypoint build

       locally with the -local flag.

» Cloning data from Git
  Ref: us-east-1
...
```

### DIFF
--- a/internal/client/job.go
+++ b/internal/client/job.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
@@ -13,6 +15,7 @@ import (
 
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/internal/pkg/finalcontext"
+	"github.com/hashicorp/waypoint/internal/pkg/gitdirty"
 	pb "github.com/hashicorp/waypoint/internal/server/gen"
 	"github.com/hashicorp/waypoint/internal/server/grpcmetadata"
 )
@@ -48,7 +51,7 @@ func (c *Project) doJob(ctx context.Context, job *pb.Job, ui terminal.UI) (*pb.J
 // setupLocalJobSystem does the pre-work required to run jobs locally. This includes:
 // - figure out if jobs should be executed locally or remotely.
 // - if job should be executed locally, start a local runner
-// - FUTURE: if a job should be executed remotely, but local VCS is present and dirty, warn.
+// - if jobs will be executed remotely, but local VCS is present and dirty, warn.
 // This lives separately from DoJob because the logs and exec commands need to conditionally warm up the
 // local job infrastructure, but don't actually create a job (the server does).
 func (c *Project) setupLocalJobSystem(ctx context.Context) (isLocal bool, newCtx context.Context, err error) {
@@ -60,12 +63,15 @@ func (c *Project) setupLocalJobSystem(ctx context.Context) (isLocal bool, newCtx
 	// Automatically determine if we should use a local or a remote runner
 	newCtx = ctx
 
+	// We may use this in multiple places, so we save the result if we obtain it
+	// NOTE(izaak): If in the future we need the full project in other places in this codepath, we should probably cache it early on the parent struct.
+	var project *pb.Project
+
 	// A nil useLocalRunner means the option was not set explicitly when this client was created.
 	// We'll decide a value for it here, and set it for future runs.
 	if c.useLocalRunner == nil {
 		log.Debug("determining if a local or remote runner should be used for this and future jobs")
 
-		// NOTE(izaak): If in the future we need the full project in other places, we should probably cache it on the parent struct.
 		getProjectResp, err := c.client.GetProject(ctx, &pb.GetProjectRequest{Project: c.project})
 		if err != nil {
 			if status.Code(err) == codes.NotFound {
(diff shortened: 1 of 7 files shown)


internal/client/job.go
```diff
@@ -423,3 +498,13 @@ func (c *Project) queueAndStreamJob(
...
 const stateEventPause = 3000 * time.Millisecond
+
+var (
+	warnGitDirty = strings.TrimSpace(`
+       There are local changes that do not match the remote repository. By default, 
+       Waypoint will perform this operation using a remote runner that will use the 
+       remote repository’s git ref and not these local changes. For these changes 
+       to be used for future operations, either commit and push, or run the operation 
+       locally with the -local flag.
+	`)
```
krantzinator, line 509: To fix the indentation here that you see in the CLI, just move everything flush with the line start. If you search for `= strings.TrimSpace(` you'll see that's what we do elsewhere.
```suggestion
There are local changes that do not match the remote repository. By default, 
Waypoint will perform this operation using a remote runner that will use the 
remote repository’s git ref and not these local changes. For these changes 
to be used for future operations, either commit and push, or run the operation 
locally with the -local flag.
`)
```

internal/pkg/gitdirty/gitdirty.go
```diff
@@ -23,6 +23,24 @@ func init() {
 	githubStyleHttpRemoteRegexp = regexp.MustCompile(`http[s]?:\/\/(.*?\..*?)\/(.*)`) // Example: https://git.test/testorg/testrepo.git
 }
 
+// GitInstalled checks if the command-line tool `git` is installed
+func GitInstalled() bool {
```
krantzinator, line 27: Yes! Nice addition.

krantzinator (approved):

internal/config/config.go
```diff
@@ -161,3 +161,8 @@ func Load(path string, opts *LoadOptions) (*Config, error) {
 func (c *Config) HCLContext() *hcl.EvalContext {
 	return c.ctx.NewChild()
 }
+
+// GetConfigPath returns the path to the directory that contains the config file (waypoint.hcl)
+func (c *Config) GetConfigPath() string {
```
mitchellh, line 166: Getters usually omit the `Get`, recommend just `ConfigPath() string` here.

Ref: https://go.dev/doc/effective_go#Getters

internal/client/job.go
```diff
@@ -60,12 +63,15 @@ func (c *Project) setupLocalJobSystem(ctx context.Context) (isLocal bool, newCtx
 	// Automatically determine if we should use a local or a remote runner
 	newCtx = ctx
 
+	// We may use this in multiple places, so we save the result if we obtain it
+	// NOTE(izaak): If in the future we need the full project in other places in this codepath, we should probably cache it early on the parent struct.
```
mitchellh, line 67: Do you mind line-breaking this at anywhere from 70 to 100 columns? (usually I do 70). I know this is a hot topic but the rest of the project follows this style (and you do too in most places)

internal/client/job.go
```diff
@@ -94,6 +101,74 @@ func (c *Project) setupLocalJobSystem(ctx context.Context) (isLocal bool, newCtx
 		// Inject the metadata about the client, such as the runner id if it is running
 		// a local runner.
 		newCtx = grpcmetadata.AddRunner(ctx, c.activeRunner.Id())
+	} else {
+		// We're about to run a remote op. We should check if we have a dirty local vcs,
+		// because the user may expect their local changes to be reflected in the remote
+		// op execution, and they won't.
+		gitDirtyErr := func() error {
```
mitchellh, line 108: Tingling of whether this should just be extracted into a standalone unexported func with arguments.

internal/pkg/gitdirty/gitdirty.go
mitchellh: Can Zen this up:

```
_, err := exec.LookPath("git")
return err == nil
```

(my university called this "boolean zen": https://courses.cs.washington.edu/courses/cse143/16au/style/bad-boolean-zen.html)

internal/pkg/gitdirty/gitdirty.go
```diff
@@ -23,6 +23,24 @@ func init() {
...
+	if _, err := exec.LookPath("git"); err == nil {
+		return true
+	}
+	return false
+}
+
+// GetRepoTopLevel returns the path to the root of the repository that contains pathWithinVcs.
+// Equivalent to git rev-parse --show-toplevel
+func GetRepoTopLevel(log hclog.Logger, pathWithinVcs string) (string, error) {
```
mitchellh, line 36: Getters comment applies here as well. A more descriptive version might be `RepoTopLevelPath`

mitchellh (changes requested):
Core stuff all looks good, some stylistic recommendations.

internal/client/job.go
izaaklauer: oh ha, I don't mind at all. I'll update my editor to move the break to 70. Good to know I can cheat up to 100 and get away with it 😆.

internal/client/job.go
izaaklauer: I have the same tingling - I tried it both ways. It's a very weird collection of arguments though, and only called here, so I went this route, but it feels a bit cute.

internal/client/job.go
izaaklauer: Do you go for 70 columns from where the comment starts, or 70 from the left margin?

internal/client/job.go
briancain: Seems like something like this might be useful else where at some point too? I can see value it in being a func to call.

mitchellh (approved):

krantzinator:
Per discussion with Izaak last week, I'm going to merge this in so it's available for work people may be doing this week while he is out.

PR #2800, opened 2021-12-03
Detecting unstaged files as dirty

izaaklauer:
Turns out we weren't detecting unstaged files previously! This remedies that.

Also contains bit of renaming for clarity.


krantzinator (approved):
😅

izaaklauer:
😅 There are _so_ many little nuances to working with git this way - I think we've nailed at least 90% of the use-cases, but I wouldn't be surprised at all if there are git states out there that will confuse our dirty detector. Luckily it's not in any critical path.

Issue #2391, opened 2021-09-30
0.7 UI Improvements


gregone:
Placeholder task for the upcoming UI work. 

### Subtasks
- [x] https://github.com/hashicorp/waypoint/issues/2613
- [x] https://github.com/hashicorp/waypoint/pull/2632
- [x] https://github.com/hashicorp/waypoint/issues/2614
- [x] https://github.com/hashicorp/waypoint/pull/2802
- [x] https://github.com/hashicorp/waypoint/issues/2392
- [x] https://github.com/hashicorp/waypoint/issues/2775
- [x] https://github.com/hashicorp/waypoint/pull/2852
- [x] https://github.com/hashicorp/waypoint/pull/2879
- [x] https://github.com/hashicorp/waypoint/issues/2883

### Stretch goals
- [ ] https://github.com/hashicorp/waypoint/issues/2740
- [x] https://github.com/hashicorp/waypoint/issues/2881
- [x] https://github.com/hashicorp/waypoint/issues/2855

### Out of scope
- [ ] https://github.com/hashicorp/waypoint/issues/2445

PR #2802, opened 2021-12-06
ui: extract `<Table>`

jgwhite:
## Why the change?

We’re going to start using this new table style in other places for #2391.

## What’s the plan?

- [x] Extract `<Table>` from `<ResourcesTable>`

## What does it look like?

Hopefully identical!

### Before

CleanShot 2021-12-06 at 16 14 24@2x

### After

CleanShot 2021-12-06 at 16 14 11@2x

### How do I test it?

1. `git checkout ui/table`
2. `cd ui && yarn start`
3. [Visit localhost:4200](http://localhost:4200)
4. Browse around
5. Verify tables look decent
6. Try adding some new tables using `<Table>`
7. Verify the API isn’t too terrible 🙃

### DIFF
--- a/ui/app/styles/components/table.scss
+++ b/ui/app/styles/components/table.scss
@@ -1,4 +1,4 @@
-.resources-table {
+.table {
   background: rgb(var(--panel));
   border-radius: 8px;
   padding: 10px;
@@ -55,7 +55,22 @@
     }
   }
 
+  .w-1\/2 {
+    width: 50%;
+  }
+  .w-1\/3 {
+    width: 33.333%;
+  }
+  .w-1\/4 {
+    width: 25%;
+  }
+  .w-1\/5 {
+    width: 20%;
+  }
   .w-2\/5 {
     width: 40%;
+  }
+  .w-3\/5 {
+    width: 60%;
   }
 }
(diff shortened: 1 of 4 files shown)


ui/app/styles/components/table.scss
```diff
@@ -55,7 +55,22 @@
...
+  }
+  .w-1\/5 {
+    width: 20%;
+  }
   .w-2\/5 {
     width: 40%;
   }
+  .w-3\/5 {
+    width: 60%;
+  }
```
jgwhite, line 75: Let me know what y’all think of these helper classes. Can add more documentation, or do them a different way, I’m not precious.

The idea is that you’d use them on the `<col>` element, as per the example.

sabrinako (approved):
Looks good to me! Great job with this, appreciate the `caption` section

ui/app/styles/components/table.scss
```diff
@@ -55,7 +55,22 @@
...
+  .w-1\/4 {
+    width: 25%;
+  }
+  .w-1\/5 {
+    width: 20%;
+  }
   .w-2\/5 {
     width: 40%;
   }
+  .w-3\/5 {
```
gregone, line 73: i would have liked to rename those classnames, I don't think backslashes are very practical. :\

jgwhite: @gregone feel free to change them. I’m not precious :)

Issue #1608, opened 2021-06-07
Bug: `deployment list` command does not line-wrap correctly


izaaklauer:
**Describe the bug**

When running `waypoint deployment list` with a terminal narrower than the longest line, we don't add newlines. Other list commands like `context list` handle this correctly. This doesn't appear to be new behavior in 0.4.0, but is striking because we  added the `health` column.

Example:

Baseline output:
<img width="992" alt="Screen Shot 2021-06-07 at 11 55 49 AM" src="https://user-images.githubusercontent.com/8404559/121051528-87ee8400-c787-11eb-983b-a264ecee56a3.png">

Expected behavior (achieved by shrinking window after running the command):
<img width="801" alt="Screen Shot 2021-06-07 at 11 55 42 AM" src="https://user-images.githubusercontent.com/8404559/121051349-62617a80-c787-11eb-95b2-08ec6e2dce64.png">

Actual behavior:
<img width="800" alt="Screen Shot 2021-06-07 at 11 55 58 AM" src="https://user-images.githubusercontent.com/8404559/121051459-7a38fe80-c787-11eb-9939-367245f39334.png">

It's more striking when there is a longer list of deployments

<img width="716" alt="Screen Shot 2021-06-07 at 11 58 03 AM" src="https://user-images.githubusercontent.com/8404559/121051709-ac4a6080-c787-11eb-99e0-35d313cab7a9.png">

**Waypoint Platform Versions**
Reproducible in 0.4.0 and 0.3.2

izaaklauer:
This isn't a `deployment list` problem in particular - it's all of our commands.

izaaklauer:
![image](https://user-images.githubusercontent.com/8404559/160005374-124b63e1-701e-482f-b51d-ba7a0828c47b.png)

xiaolin-ninja:
Testing around and it looks like this is happening to `runner profile list`, and `release list`... likely all of them?

briancain:
@xiaolin-ninja - absolutely. Basically anywhere we use `UI.Table` which is from the waypoint SDK.

briancain:
@izaaklauer - Hey I think I will pull this out out `0.8.y` only because the solution to fix this is fairly large and non-trivial to render the tables properly for a terminal size. This feels more like a larger task rather than a bug task. If I'm wrong feel free to add it back 👍🏻

izaaklauer:
It's still an issue! This must have reverted.

PR #2807, opened 2021-12-06
feature: Add Trigger URL CLI for managing and creating trigger URL state

briancain:
This pull request builds off of https://github.com/hashicorp/waypoint/pull/2794 and implements the CLI for managing trigger URL state in Waypoint server. Notice that this pull request is targeted to the pull requests branch rather than main. If the other PR gets merged before this one, I'll retarget off `main` instead.

Some of the UX isn't complete, like returning and show the actual HTTP url to trigger requests with when you create a trigger URL. This work will happen next with the http service implementation. Everything else is here in the pull request though for creating/updating, listing, inspection, and deletion.


internal/cli/main.go
```diff
@@ -405,6 +405,33 @@ func Commands(
 			}, nil
 		},
 
+		"trigger": func() (cli.Command, error) {
+			return &helpCommand{
+				SynopsisText: helpText["trigger"][0],
+				HelpText:     helpText["trigger"][1],
+			}, nil
+		},
+		"trigger apply": func() (cli.Command, error) {
```
mitchellh, line 414: "apply" feels a bit weird for triggers, I wonder if we should split into "create" and "update". Triggers don't really feel like some desired state operation... even if internally we support an upsert.

mitchellh (changes requested):
Looks very good! Just a couple questions.

internal/cli/main.go
briancain: I started down the path with `create` and `update` (see git commit history for that) but wasn't really sure how to handle it cleanly without a bunch of duplicated code or a shared lib for handling operations and flags. If you have ideas I'm all ears!

internal/cli/trigger_apply.go
```diff
@@ -0,0 +1,377 @@
...
+	resp, err := c.project.Client().UpsertTrigger(ctx, &pb.UpsertTriggerRequest{
+		Trigger: createTrigger,
+	})
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	// TODO(briancain): update output to show trigger URL with wp server attached once http service is implemented
+	c.ui.Output("Trigger %q (%s) has been %s", resp.Trigger.Name, resp.Trigger.Id,
```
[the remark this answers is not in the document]
briancain: Maybe a `-json` flag that's just a simple map of key values here? Like `{id: "123", url: "http://example.com/123"}`

internal/cli/main.go
mitchellh: The way we've done it before is make it the same command struct, and just put a `Update bool` on it and then switch on that logic in it. So most of the logic is shared, but ie. if `Update == true` then we require an ID. And we can also change some of the help output that way.

internal/cli/trigger_apply.go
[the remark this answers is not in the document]
mitchellh: Possibly, still requires `jq`... speaking as a CI scripter, I'd prefer just a plaintext unix-style output of just what I want for the most common cases.

internal/cli/main.go
briancain: Got it! Good idea. Got that updated now.

internal/cli/trigger_apply.go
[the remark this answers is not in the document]
briancain: Makes sense, I updated it to output the ID (and in the future, the URL) at the end of the creation script. Similar to how `waypoint up` shows deployment and release urls!

briancain:
Frontend tests seem to be stuck on yellow, if you click into circle they're green. Just an FYI.

internal/cli/trigger_list.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+
+			fmt.Println(str)
+		}
+		return 0
+	}
+
+	c.ui.Output("Trigger URL Configs", terminal.WithHeaderStyle())
+
+	tbl := terminal.NewTable("ID", "Name", "Workspace", "Project", "Application", "Operation", "Description", "Labels", "Last Time Active")
```
izaaklauer, line 86: Looks good, but also so wide! I wonder if we should steal an idea from kubectl and have a short version of commands like this, with maybe only the 5 most important columns, and add a `-o wide` option to include all the columns. This is exacerbated by our existing line-wrapping problems: https://github.com/hashicorp/waypoint/issues/1608

Other commands are guilty too though, so it shouldn't block this PR. Just an idea.

briancain: I could see that. I also wouldn't mind punting on it and we could fix it across all commands and use the same flag name 🤔 There's also always the `-json` flag which is nicely formatted and not as wide

mitchellh (approved):

briancain:
Rebased this branch to fix the merge conflict off main from the server proto, and rebased the upstream parent branch into this one to be up to date.

internal/cli/main.go
```diff
@@ -956,6 +990,17 @@ accesses a Waypoint server requires a token.
...
 	},
 
+	"trigger": {
+		"Trigger URL management",
+		`
+Trigger URL management.
+
+Trigger URLs are used execute to lifecycle operations for Waypoint. They are configured
+ahead of the execution request, and Waypoint server will generate a configuration
+URL for clients to send a request to to execute the operation.
```
krantzinator, line 1000:
```suggestion
Trigger URLs are used to execute lifecycle operations for Waypoint. They are configured
ahead of the execution request, and Waypoint server will generate a configuration
URL for clients to send a request to execute the operation.
```

krantzinator:
One question as I'm reviewing: do we specify `on Waypoint server` with other CLI commands' help texts? Is it necessary to specify that? It seems superfluous to me.

```
    create     Create and register a trigger URL and register it to Waypoint server
    delete     Delete a trigger URL from Waypoint server
    inspect    Inspect a trigger URL from Waypoint server
    list       List trigger URL configurations on Waypoint server
    update     Update a trigger URL and register it to Waypoint server
```

internal/cli/trigger_apply.go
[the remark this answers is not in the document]
krantzinator: Heh I came to ask for more info. I guess it'll all be in the `list` output so that's fine.

krantzinator (approved):
Looks good! Excited to do some more in-depth testing once the grpc/http stuff hits. 🎉

PR #2815, opened 2021-12-07
CLI: Warn if project from flags mismatches project from config

izaaklauer:
## What changed

A user can add the `-project` flag when they're in some other project's directory:

Example:

```
$ cd ~/projects/project-a
$ cat waypoint.hcl # it exists
$ waypoint deploy -project=project-b
```

This means one of two things:

1: They added the flag explicitly, and they want to operate against `project-b`.

2: The user wants to operate against the project that matches the directory they're in, but they've copied the `-project` flag from some other command/example.

We could just error here, but it feels onerous to force the user from scenario 1 to `cd` into some safe directory before running their command.

The middle ground here is to continue to allow flags to override config, but issue a warning to help the user from scenario 2. We also force the operation to happen remotely, as the local runner doesn't currently support cloning down some other project's vcs source.

## What this looks like

```
$ waypoint build -project=learn-waypoint-lambda
Warning: Currently in project directory for "example-nginx-ami", but will operate 
against specified project "learn-waypoint-lambda"

» Cloning data from Git
  Ref: us-east-1
...
```

### DIFF
--- a/.changelog/2815.txt
+++ b/.changelog/2815.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-cli: Enhance warning project flag mismatches project parsed from config
+cli: Enhance warning for project flag mismatches when project parsed from config
 ```
--- a/.changelog/2815.txt
+++ b/.changelog/2815.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Enhance warning project flag mismatches project parsed from config
+```
--- a/internal/cli/base.go
+++ b/internal/cli/base.go
@@ -321,6 +321,22 @@ func (c *baseCommand) Init(opts ...Option) error {
 	// config parsing, as precedence order means we take the most specific value
 	// which is the -project flag
 	if c.flagProject != "" {
+		// Warn if the project from config and the project from flags conflict
+		if c.refProject != nil && c.flagProject != c.refProject.Project {
+			c.ui.Output(warnProjectFlagMismatch, c.refProject.Project, c.flagProject, terminal.WithWarningStyle())
+
+			// The config we parsed is for some other project, so we don't want to use it
+			c.cfg = nil
+			c.refProject = nil
+			c.refApps = nil
+
+			// NOTE(izaak): unless we force remoteness, we may spawn a local runner which will operate against
+			// the current config (which isn't relevant)
+			c.Log.Debug("Forcing any future operations to occur remotely because we don't have the relevant waypoint.hcl present.")
+			flagLocal := false
+			c.flagLocal = &flagLocal
+		}
+
 		c.refProject = &pb.Ref_Project{Project: c.flagProject}
 	}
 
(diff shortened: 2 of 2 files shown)


internal/cli/base.go
```diff
@@ -325,6 +325,22 @@ func (c *baseCommand) Init(opts ...Option) error {
...
 	// which is the -project flag
 	if c.flagProject != "" {
+		// Warn if the project from config and the project from flags conflict
+		if c.refProject != nil && c.flagProject != c.refProject.Project {
+			c.ui.Output(warnProjectFlagMismatch, c.refProject.Project, c.flagProject, terminal.WithWarningStyle())
+
+			// The config we parsed is for some other project, so we don't want to use it
+			c.cfg = nil
+			c.refProject = nil
+			c.refApps = nil
```
izaaklauer, line 335: It feels a little bit messy to undo the refs that were set in the above config-parsing block - if someday we set more refs up there, we'll have to remember to roll them back here.

Moving this flag mismatch check into the config parsing block feels messy too though.

.changelog/2815.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli: Enhance warning project flag mismatches project parsed from config
```
krantzinator, line 2: nit - make this easier to parse
```suggestion
cli: Enhance warning for project flag mismatches when project parsed from config
```

krantzinator (approved):
Nice

internal/cli/base.go
mitchellh: Feels like you could move this into the `if !baseCfg.NoConfig` block because thats the only place we populate refProject. From there, it feels cleaner to avoid ever setting these extra fields. That way the place where its set is totally just one place.

mitchellh (changes requested):

internal/cli/base.go
izaaklauer: Hmm, i'm not sure if I entirely follow - we'll still need to set `refProject` outside of `baseCfg.NoConfig`, because there may be commands that need a project ref but don't want config parsed. I took another stab at it though, moving more logic up into the `baseCfg.NoConfig` block to avoid setting the refs in the first place if the config and flag defined project names collide.

Still feels a bit messy, because now we're checking `c.flagProject` twice - but it's not too bad. Let me know if you see a better pathway.

internal/cli/base.go
```diff
@@ -873,5 +885,10 @@ short notation -p and -a.
 The current Waypoint server does not support snapshots. Rerunning the command
 with '-snapshot=false' is required, and there will be no automatic data backups
 for the server.
+`)
+
+	warnProjectFlagMismatch = strings.TrimSpace(`
+Warning: Currently in project directory for %q, but will operate 
+against specified project %q
```
briancain, line 892:
```suggestion
Warning: Currently in project directory for %q, but was requested 
to operate against specified project %q with '-project'
```

I think it'll be useful to be a bit more specific as to why the warning occurs.

briancain (approved):
A minor update request in the warning message, but otherwise looks good to me!

> Still feels a bit messy, because now we're checking c.flagProject twice - but it's not too bad. Let me know if you see a better pathway.

Tried to catch up on this conversation....from what you have now I think that seems ok to me. The project flag check is mostly for the warning to be displayed if no config, then the second time we check immediately after is for setting project refs.

mitchellh (approved):

internal/cli/base.go
krantzinator: I'm going to add this in, as I told Izaak I'd work on merging his PRs while he is out this week.

PR #2830, opened 2021-12-13
Allow hyphens in workspace names

catsby:
The current regex to validate names used in the `UpsertWorkspace` endpoint does not allow hyphens, which I do not believe were meant to be excluded in valid names in https://github.com/hashicorp/waypoint/pull/2758

Note: no changelog entry because at this time this endpoint is not yet released

### DIFF
--- a/internal/server/ptypes/workspace.go
+++ b/internal/server/ptypes/workspace.go
@@ -14,7 +14,7 @@ import (
 
 // WorkspaceNameRegexp is the valid Workspace name regular expression. At this
 // time the only restriction is to not allow spaces.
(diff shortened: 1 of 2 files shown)


mitchellh (commented):
This will allow workspaces to _begin_ with hyphens, that may be something we want to disallow.

catsby:
@mitchellh great point, thanks. I refactored it to allow hyphens and underscore, but not start with either. Please take a look when you get a chance. My regular expressions skills are lacking, but I used this to iterate: https://regex101.com/r/82kJL0/4

mitchellh (approved):

PR #2832, opened 2021-12-14
ui/mirage: fix `UpsertProject` handler

jgwhite:
## Why the change?

Previously this would fail to create a brand new project.

## How do I test it?

1. `git checkout ui/mirage-upsert-project-fix`
2. `cd ui && yarn start`
3. [Try making a new project](http://localhost:4200/default/new)
4. Observe it doesn’t work
5. `git checkout ui/mirage-upsert-project-fix`
6. [Try making a new project](http://localhost:4200/default/new)
7. Verify it now works


gregone (approved):
:shipit:

PR #2834, opened 2021-12-15
ui: fix workspace switcher issues on deployment master-detail pages

jgwhite:
## Why the change?

Originally, the workspace switcher appeared on the `app` page and its immediate child pages (builds, deployments, releases, etc). None of these pages included additional dynamic URL segments beyond `:app_id`. The workspace switcher was implemented with the faulty assumption that this would always be the case.

Why does this matter? Well, the workspace switcher renders links with fully-qualified paths, which need to be fed the correct params (dynamic segments) in order to function correctly. Let’s work through an example to see where this can break down:

Given this path:
```
/:workspace_id/:project_id/app/:app_id/deployments/seq/:deployment_seq
```

And this invocation of `LinkTo`:
```hbs
<LinkTo
  @route="workspace.projects.project.app.deployment.deployment-seq"
  @models={{array "my-workspace" "my-project" "my-app" "1"}}
>
```
Then `@models` will be plumbed into the path as follows:
```
@models={{array "my-workspace" "my-project" "my-app" "1"}}
                └──────┬─────┘ └─────┬────┘ └──┬───┘ └┬┘
       ┌───────────────┘             │         │      │
       │            ┌────────────────┘         │      │
       │            │             ┌────────────┘      └───────┐
 ┌─────┴─────┐ ┌────┴────┐     ┌──┴──┐                 ┌──────┴──────┐
/:workspace_id/:project_id/app/:app_id/deployments/seq/:deployment_seq
```

`@models` are fed in *from the right* (from least significant, if you like). Ember assumes any model you haven’t specified should remain unchanged. So let’s see what happens if a model is missing from the right hand side, let’s remove the sequence param:

```
@models={{array "my-workspace" "my-project" "my-app"}}
                └──────┬─────┘ └─────┬────┘ └──┬───┘
                       │             │         │
                    ┌──┘          ┌──┘         └──────────────┐
                    │             │                           │
               ┌────┴────┐     ┌──┴──┐                 ┌──────┴──────┐
/:workspace_id/:project_id/app/:app_id/deployments/seq/:deployment_seq
```

Which gives us a nonsense path, like this:

```
/previous-workspace/my-workspace/app/my-project/deployments/seq/my-app
```

This is the situation we found ourselves in with the workspace switcher, because we were only ever feeding in `@models` up to the `:app_id`, assuming there would be no further dynamic segments for pages on which the switcher appears.

Phew!

The fix is threefold:

1. Rather than linking to simply `currentRoute`, calculate a suitable “pivot route” which makes sense when switching between workspaces. At present, this is the immediate child route of the `app` route, but this may change in future if the app route gains immediate children with dynamic segments that don’t apply cleanly between workspaces. For example, if `deployment-seq` with it’s `:deployment_seq` dynamic segment were to become a direct child of `app`, then we’d have to tweak the behavior, but at least it’ll be more obvious what’s happening rather than seeing a bizarre nonsense path.
2. Calculate the `@models` array in a generic manner, so it’s more robust to additional dynamic segments.
3. Expand test coverage to catch regressions

## What’s the plan?

- [x] Improve test coverage so it actually catches this issue
- [x] Fix the issue 🦉 
- [x] PR description

### Before

https://user-images.githubusercontent.com/34030/146375921-b072e730-8518-47dc-83e6-7153c8fd8045.mp4

### After

https://user-images.githubusercontent.com/34030/146375949-a1dcc538-80ef-4e62-b5c4-05fef85384ca.mp4

## How do I test it?

1. `git checkout ui/workspace-switcher-fixes`
2. `cd ui && yarn start`
3. `open http://localhost:4200/default/marketing-public/app/wp-matrix`
4. Try switching between workspaces on the deployments, builds, and releases pages
5. Verify everything works as you’d expect

### DIFF
--- a/ui/app/components/workspace-switcher.hbs
+++ b/ui/app/components/workspace-switcher.hbs
@@ -34,6 +34,7 @@
                 @models={{array-concat workspace.name @models}}
                 aria-current={{if (eq workspace.name @current) "page" false}}
                 {{on "click" D.close}}
+                data-test-workspace-link={{workspace.name}}
               >
                 <span class="workspace-switcher__link-contents">
                   {{#if (eq workspace.name @current)}}
--- a/ui/app/controllers/workspace/projects/project/app.ts
+++ b/ui/app/controllers/workspace/projects/project/app.ts
@@ -6,4 +6,39 @@ import { tracked } from '@glimmer/tracking';
 export default class extends Controller {
   @tracked isSwitchingWorkspace = false;
   @service router!: RouterService;
+
+  /**
+   * Returns a suitable “pivot” route when switching between workspaces.
+   * For example, if you’re looking at a list of builds for production
+   * then it makes sense to switch to the list of builds for staging.
+   * However, if you’re looking at an individual deployment in production
+   * then it wouldn’t make sense to switch to that same deployment in staging,
+   * because it won’t be there. Instead, you want to jump up a level to see
+   * the list of all deployments for production.
+   */
+  get workspaceSwitcherTargetRoute(): RouterService['currentRoute'] {
+    let result = this.router.currentRoute;
+
+    while (result.parent && result.parent.localName !== 'app') {
+      result = result.parent;
+    }
+
+    return result;
+  }
+
+  /**
+   * Returns the array of models to go with `workspaceSwitcherTargetRoute`,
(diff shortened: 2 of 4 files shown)


gregone (approved):
LGTM.

PR #2835, opened 2021-12-15
ui: automatic workspace selection

### LINKED ISSUE #2662: ui: Default workspace selection

jgwhite:
## Why the change?

Closes #2662 

## What’s the plan?

- [x] Capture the logic from #2662 in some tests
- [x] Figure out how to do storage properly with ember-simple-auth
- [x] Figure out the slightly surprising way that the presence of `login()` in our default mirage scenario interferes with this behavior
- [x] PR description

### Before

Notice:

1. We land in the `default` workspace, even though it’s not in use (or doesn’t exist at all)
2. If we switch workspace and revisit the UI, we end up back at `default`

https://user-images.githubusercontent.com/34030/150109628-f12ab897-15af-4ea0-9d64-985926f7a015.mp4

### After

Notice:

1. We land in the `production` workspace, because it’s alphabetically first of the workspaces that exist
2. When we switch workspace and revisit the UI, we end up back at the last workspace we saw (`staging` in this case)

https://user-images.githubusercontent.com/34030/150109657-8236a454-f23b-4d8c-9ecc-d90786b45e73.mp4

## How do I test it?

1. `git checkout ui/automatic-workspace-selection`
2. Pre-populate a Waypoint server with a variety of workspaces
3. [Boot the dev server against that Waypoint server](https://github.com/hashicorp/waypoint/blob/main/ui/README.md#running-with-a-local-waypoint-server)
4. Try navigating in various ways
5. Verify workspace selection behaves in a sensible manner

### DIFF
--- a/.changelog/2835.txt
+++ b/.changelog/2835.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Automatically select the appropriate workspace
+```
--- a/ui/app/components/logout/index.hbs
+++ b/ui/app/components/logout/index.hbs
@@ -1,5 +1,5 @@
 {{#if @canLogout}}
-  <button class="auth-link" type="button" {{on "click" this.logout}}>
+  <button data-test-logout-button class="auth-link" type="button" {{on "click" this.logout}}>
     <FlightIcon @name="sign-out" class="nav-icon" />
     Logout
   </button>
--- a/ui/app/instance-initializers/session-workspace-storage.ts
+++ b/ui/app/instance-initializers/session-workspace-storage.ts
@@ -1,0 +1,15 @@
+import ApplicationInstance from '@ember/application/instance';
+
+export function initialize(appInstance: ApplicationInstance): void {
+  let session = appInstance.lookup('service:session');
+
+  session.on('invalidationSucceeded', () => {
+    // Clear workspace on logout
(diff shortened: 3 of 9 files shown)


sabrinako (approved):
Nice work! Looks smooth and working as expected when testing on local as well 🤠👍

PR #2839, opened 2021-12-15
Update to Go 1.17

mitchellh:
This is pretty long in the tooth probably since Go 1.18 now in beta, it just wasn't a priority for us, but I don't see any reason for us to not upgrade to the latest 1.17 version. This updates our CI and Dockerfile. We probably need to update our release repo too.

Don't think we should backport this, and just use it for the next minor.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 
 references:
   images:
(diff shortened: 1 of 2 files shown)


briancain (approved):
Looks good, assuming the tests are happy! 👍🏻

krantzinator (approved):

PR #2840, opened 2021-12-15
feature: RunTrigger gRPC endpoint for Waypoint Server

briancain:
This pull request introduces the service level function `RunTrigger` which will queue a job based on an existing stored trigger URL configuration using the ID requested. It also optionally allows for any configuration variable overrides on the request.

I kept the targeted runner to Any until that work is finished. It will be easier once the targeted config is available based on the project and application.

The next step after this is to work on the HTTP handler. It should register a trigger URL id as a valid route, and if a request to that endpoint is called it will take a run trigger request with a registered trigger URL id and call this grpc func (similar to how the HTTP exec works) to queue the job request and optionally stream back the job outputs to the requester.

I tested this endpoint using the `waypoint-grpc` helper tool in contrib: https://github.com/hashicorp/waypoint/tree/main/contrib/waypoint-grpc


internal/server/singleprocess/service_trigger.go
```diff
@@ -71,3 +76,125 @@ func (s *service) ListTriggers(
...
+	log.Debug("building run trigger job")
+
+	// Build the job(s)
+	job := &pb.Job{
+		Workspace: runTrigger.Workspace,
+		Labels:    map[string]string{"trigger/id": runTrigger.Id},
+	}
+
+	// TODO(briancain): is there an easy way to convert this without the big switch
+	// so that we don't have to add new case statements any time a new operation is supported?
```
mitchellh, line 105: Probably, but depends how ugly you want to get lol. This is probably good for now.

internal/server/singleprocess/service_trigger.go
```diff
@@ -71,3 +76,125 @@ func (s *service) ListTriggers(
...
+	var ids []string
+	if runTrigger.Application == nil {
+		// we're gonna queue multiple jobs for every application in a project
+		project, err := s.state.ProjectGet(runTrigger.Project)
+		if err != nil {
+			return nil, err
+		}
+
+		log.Debug("building multi-jobs for all apps in project", "project", project.Name)
+		for _, app := range project.Applications {
```
mitchellh, line 145: Dunno if Rae's work added a way to queue a job for all applications, but this feels like a function that could live in the state interface instead.

briancain: Good suggestion! I'll take a look, if not I can add it here or in a separate PR. It would be a nice addition.

internal/server/singleprocess/service_trigger.go
```diff
@@ -71,3 +76,115 @@ func (s *service) ListTriggers(
...
+	default:
+		return nil, status.Errorf(codes.Internal,
+			"trigger %q is configured with an unsupported operation %T", runTrigger.Id, op)
+	}
+
+	if len(req.VariableOverrides) > 0 {
+		log.Debug("variable overrides have been requested for trigger job")
+		for i, v := range req.VariableOverrides {
+			switch vType := v.Source.(type) {
+			case *pb.Variable_Cli:
```
briancain, line 128: @krantzinator suggested maybe we could introduce a `RunTrigger` source type here too. Although the CLI type does exactly what we want here, I think.

internal/server/singleprocess/service_trigger.go
briancain: Got that added in this PR!

internal/server/singleprocess/service_trigger_test.go
```diff
@@ -382,3 +382,188 @@ func TestServiceTrigger_DeleteTrigger(t *testing.T) {
...
+				Id: respTrigger.Trigger.Id,
+			},
+		})
+
+		require.Error(err)
+		require.Nil(resp)
+		st, ok := status.FromError(err)
+		require.True(ok)
+		require.Equal(codes.NotFound, st.Code())
+	})
```
catsby, line 447: In the test scenario here we create a trigger (ID generated, URLs supported, etc) for a Project that doesn't exist (yet, or possibly never will at all). So we create a valid Trigger (no error when creating), but it's also an invalid Trigger (error when trying to use). 

This scenario surprised me so I went looking. It seems [UpsertTrigger](https://github.com/hashicorp/waypoint/blob/109efacee9a7d8b4506de4ebf8e537181b03e0d6/internal/server/singleprocess/service_trigger.go#L11) doesn't validate that the `Project` referenced actually exists, just that it's not an empty string. 

This may be beyond your PR, but I feel like we should ensure it exists before we create more things based off of it, especially when those things can manipulate the thing that doesn't yet exist.

internal/server/singleprocess/state/job.go
```diff
@@ -252,6 +252,49 @@ func (s *State) JobCreate(jobs ...*pb.Job) error {
...
+	jobTemplate *pb.Job,
+) ([]*pb.QueueJobRequest, error) {
+	var result []*pb.QueueJobRequest
+	project, err := s.ProjectGet(pRef)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, app := range project.Applications {
+		// We copy by all values because copying the original job includes copying a mutux lock,
```
catsby, line 268: nit: we don't seem to be copying _all_ the `pb.Job` values here. If we're intentionally only copying a few, we should document that we're only copying the ones we really need. Otherwise someone in the future may read this and think that we copied all of them at the time, but now (then?) there are more, and we're now (then) are not copying them all but should be? 

If we do need to copy all without the mutex issues, we're already using [`copystructure`](https://github.com/mitchellh/copystructure) in a few other places, I believe that would avoid the issue.

catsby: Ah, I may have mis-read the code comment here. I read it first as "we copy all" but it's actually "we copy by all values" which threw me. That said, do we need to copy all of them? We seem to be omitting `ExpireTime` and `QueueTime` for instance, which sound relevant?

catsby (commented):
Looks good so far! I had a few nit-picks and some questions about the copy. Probably no blockers but wanted to ask first

internal/server/singleprocess/service_trigger_test.go
briancain: Yep, so this behavior is intentional. I didn't require a project existing on upseting a trigger because the idea is these triggers are going to be configured ahead of time to be used later in CI. Guarding on a project existing to create a trigger would make that much more difficult to achieve. If you try to run a trigger though, we will check if it exists, which I think is the right time to check.

One middle ground would be to display a warning in the CLI if a trigger has been created but no project exists for the requested trigger.

internal/server/singleprocess/state/job.go
briancain: Yeah, I might of missed those. Since this is supposed to be more of a generic job func, I think using something like copystructure is the way to go here 👍🏻 I'll update to use that.

internal/cli/trigger_inspect.go
```diff
@@ -115,6 +117,11 @@ func (c *TriggerInspectCommand) Run(args []string) int {
 		tags = strings.Join(trigger.Tags[:], ", ")
 	}
 
+	var lastActiveTime string
```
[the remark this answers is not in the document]
briancain: I like this idea! I think I'll go with `"n/a"` instead of unknown. Just because if the time hasn't been set, we know it hasn't been active. Where defaulting to unknown makes it seem like we can't know its active time.

internal/server/singleprocess/service_trigger.go
```diff
@@ -71,3 +76,115 @@ func (s *service) ListTriggers(
...
+		job.Application = runTrigger.Application
+		j := &pb.QueueJobRequest{Job: job}
+		jobList = append(jobList, j)
+
+		resp, err := s.QueueJob(ctx, j)
+		if err != nil {
+			return nil, err
+		}
+		ids = append(ids, resp.JobId)
+	}
```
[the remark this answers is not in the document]
briancain: One reason I stayed away from queueJobMulti is it returns both `stop` and `start` job task IDs if the project is backed by on-demand runners. Basically when you go to queue a job, you'll get back 3 job ids, however they aren't labeled and you just have to "know" the order. The reason why this is a bit tricky is the HTTP endpoint is supposed to take this job ID response and stream it back if requested. I think we're safe to assume all run trigger jobs _must_ be backed by on-demand runners, but it'd be nice if maybe the response had more info. Going to figure out a good way to handle this.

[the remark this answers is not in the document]
briancain: I might punt on that for now and address it in the HTTP implementation to be honest, moving to use queueJobMulti in both accounts is an easy fix though and will push that up shortly.

internal/server/singleprocess/service_trigger_test.go
krantzinator: For future work, have we thought about adding a "Test trigger" ability? This could be CLI and a nice fancy UI button. It would cover the "OK I made a trigger; now the dev made a project; now the release engineer needs to test the trigger points to the right project."

krantzinator:
Quick comment -- can we make the error message for a missing op type a bit more informative?
```
! Empty operation type. Operation type must be set with the `-op` flag as one of the following values:
  build, push, deploy, destroy-workspace, destroy-deployment, release, up, init
```

krantzinator:
Hm I'm getting some odd screen wrapping behavior when I have multiple triggers:
Screen Shot 2022-01-05 at 2 42 18 PM

krantzinator:
These cases should all have `operation` removed from the sentence (so just `build` instead of `build operation`). The column heading is `Operation` so it's redundant.
https://github.com/hashicorp/waypoint/blob/f564326568b7eff4496aec3ec9b61151d2fd677d/internal/cli/trigger_inspect.go
```
	case *pb.Trigger_Build:
		opStr = "build operation"
```

krantzinator:
Bug: `trigger inspect` has an `id` flag but also requires an argument that it uses for `<trigger-id>`. It should use one or the other. I hit a panic, as you would expect, when I used the flag but did not provide an argument.

briancain:
@krantzinator - thanks for the feedback!

- I've updated `waypoint trigger list` to only show a few options by default, and included a ` -full` flag to show everything I was showing before.
- I fixed the panic in the CLI if you specify an id by flag but not argument for the `waypoint trigger inspect` command.
- Removed the ` operation` text from inspect and list

Not sure what you are looking for in terms of more information about the operation flag but open to suggestions!

internal/server/singleprocess/service_trigger_test.go
briancain: Maybe like a `waypoint trigger validate <id>`? It could look up the project and optionally the app to make sure they exist?

.changelog/2840.txt
```diff
@@ -0,0 +1,6 @@
+```release-note:feature
+core: Introduce RunTriggers to Waypoint Server and CLI. Triggers can be configured
+ahead of time to execute lifecycle operations on demand through the Waypoint API.
+Currently, the gRPC API is only supported, but in the future an HTTP endpoint
```
krantzinator, line 4:
```suggestion
Currently, only the gRPC API is supported, but in the future an HTTP endpoint
```

internal/server/singleprocess/service_trigger.go
```diff
@@ -71,3 +75,166 @@ func (s *service) ListTriggers(
...
+
+	if len(jobList) > 0 {
+		// Queue the job(s)
+		log.Debug("queueing jobs", "total_jobs", len(jobList))
+
+		// NOTE(briancain): queueJobMulti currently returns 3 Jobs due to On-Demand Runners:
+		// The start task job, the actual queued job, and the stop task job. Users
+		// will generally just care about the middle job for streaming the result
+		// of the RunTrigger
+		respList, err := s.queueJobMulti(ctx, jobList)
```
evanphx, line 215: It's probably ok to leave this out for now, but we should figure out if we should wait for the jobs to be accepted (or at least one job because of ODR) before returning, so that the caller knows that at least the job was valid when the RPC call returns.

evanphx (changes requested):
Minor thing and a note about job validity checking for the future.

internal/cli/trigger_apply.go
```diff
@@ -337,19 +334,13 @@ func (c *TriggerApplyCommand) Flags() *flag.Sets {
...
 			Name:   "deployment-id",
 			Target: &c.flagDeploySeq,
-			Usage:  "The sequence number (short id) for the deployment to use in an operation.",
-		})
-
-		fo.IntVar(&flag.IntVar{
-			Name:   "release-id",
-			Target: &c.flagReleaseSeq,
-			Usage:  "The sequence number (short id) for the release to use in an operation.",
+			Usage:  "The sequence number (short id) for the deployment to use for a release operation.",
```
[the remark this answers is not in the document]
briancain: Hah, nice catch! Oopsie

internal/server/singleprocess/service_trigger.go
briancain: That's a good idea. I ran into some issues while testing where I'd run a bunch of triggers, get ids back, and not see the actual job errors until I looked at the server logs much later. I'll see what I can do to make it better this week before release.

internal/cli/trigger_apply.go
```diff
@@ -94,17 +94,22 @@ func (c *TriggerApplyCommand) Run(args []string) int {
 		}
 
 		// Trigger target
-		if diffTrigger.Workspace != nil {
+		if diffTrigger.Workspace != nil && c.flagWorkspace == "" {
```
xiaolin-ninja, line 97:
```suggestion
		if c.flagWorkspace == "" {
		    if diffTrigger.Workspace != nil {
		       c.flagWorkspace = diffTrigger.Workspace.Workspace
		       } else {
		       c.flagWorkspace = "default"
		   }
```
To make that flow cleaner, and optionally refactor the triggers below to have the flag check as the first conditional for uniformity?

briancain: We'll still have to do this workspace flag check later, because inside this if-statement we're only setting these values if we're updating a trigger URL and one already exists, i.e. the check on line 83. So in the creation case, we'll still need the check below if workspace was not set.

internal/server/singleprocess/service_trigger.go
briancain: Discussed via slack, these jobs are running through on-demand runners, which validate the job already. So no need to do it again here!

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+The Waypoint CLI comes with a subcommand that lets you configure and update
+Trigger URL configurations.
+
+Check out our CLI docs for the `waypoint trigger` command, such as
+[waypoint trigger create](/commands/trigger-create).
+
+## Pre-configuring a Trigger URL
+
+As an example, we can create a Trigger URL config that deploys the latest artifact
+to the produciton workspace. We'll need to figure the Trigger to use the
```
krantzinator, line 41:
```suggestion
to the produciton workspace. We'll need to configure the Trigger to use the
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+right project, application, and workspace. Then we can specify which [lifecycle operation](docs/lifecycle)
+we want the trigger to execute when run. We can also include a `description` or
+`tags` for extra information for the user.
```
krantzinator, line 44:
```suggestion
As an example, you can create a Trigger URL config that deploys the latest artifact
to the production workspace. First, configure the Trigger to use the
right project, application, and workspace. Then, specify which [lifecycle operation](docs/lifecycle)
you want the trigger to execute when run. You can also include a `description` or
`tags` for extra information for the user.
```

krantzinator: You go back and forth in the docs between `you` and `we` and I think we should be consistent. I've noted some suggested changes for this, but not everything.

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+command will return a full HTTP URL that Waypoint server will be able to accept requests
+with for executing this operation.
+
+For now, we can use this ID to make a request to Waypoint server to execute the
+trigger.
+
+## Modifying an existing Trigger URL Configuration
+
+If you wish to see all of your configured triggers inside a Waypoint server,
+we can run the `waypoint trigger list` command:
```
krantzinator, line 71:
```suggestion
run the `waypoint trigger list` command:
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+  01FS57YH4TQMSP6HZ1QZ03QE2R | release-latest       | default   | go-gitops-0 | go          | release
+  01FS582B3X8VW0251C1XDQ56RR | deploy-statusreport  | default   | go-gitops-0 | go          | status report deployment
+  01FS58DAXTE832BM5JMG1Z2XMA | release-statusreport | default   | go-gitops-0 | go          | status report release
+
+```
+
+In this example, we are configuring Trigger URLs for the `go-gitops-0` project.
+
+Now, let's say you have a _status report_ trigger that _only_ runs on your app
+called `go` in the default workspace, but wish for that trigger to refresh the
```
krantzinator, line 90:
```suggestion
called `go` in the `default` workspace, but wish for that trigger to refresh the
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+status in the `production` workspace.
+
+First, we can inspect the Trigger URL configuration to see how it's configured:
```
krantzinator, line 93:
```suggestion
First, inspect the Trigger URL configuration to see how it's configured:
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+     Authenticated: true
+         Operation: status report deployment
+         Workspace: default
+           Project: go-gitops-0
+       Application: go
+
+```
+
+For this trigger to run in the `production` workspace, we can update the
+triggers configuration:
```
krantzinator, line 111:
```suggestion
For this trigger to run in the `production` workspace, update the
trigger's configuration:
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+                ID: 01FS582B3X8VW0251C1XDQ56RR
+  Last Time Active: n/a
+     Authenticated: true
+         Operation: status report deployment
+         Workspace: production
+           Project: go-gitops-0
+       Application: go
+```
+
+For now, we can use this ID to make a request to Waypoint server to execute the
```
krantzinator, line 131:
```suggestion
For now, you can use this ID to make a request to Waypoint server to execute the
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+URL job for each job queued.
+
+In the future, the HTTP implementation will handle these jobs for you, and
+give you the ability to stream the output of the job easily in CI.
+
+## Running a Trigger URL Configuration via HTTP
+
+In an upcoming version of Waypoint, we will provide an HTTP endpoint for easy
+access to request running a configured Trigger. When you create a trigger,
+you'll also now see an HTTP url:
```
krantzinator, line 157:
```suggestion
access to request running a configured Trigger. When you create a trigger,
you'll then see an HTTP url:
```

website/content/docs/triggers.mdx
```diff
@@ -0,0 +1,179 @@
...
+      -app=backend-svc \
+      -workspace=production \
+      -name=deploy-latest \
+      -op=deploy
+Trigger ID: 123ABC456DEF
+Trigger URL:
+https://waypoint.server.address/api/v1/trigger/123ABC456DEF
+```
+
+Then you will be able to simply add a curl request to your CI configuration
```
krantzinator, line 171: Really emphasize this is future, not now.
```suggestion
When this upcoming feature enhancement is released, you will be able to add a curl request to your CI configuration
```

krantzinator (commented):
Could you add a note on trigger-specific variables? I want to test it but it's not covered in the docs :)

evanphx (approved):

website/content/docs/triggers.mdx
briancain: Thanks for the catch here! I merged your other suggestions but github doesn't like this multi-line suggestion so I'll make the changes locally

catsby (approved):

krantzinator (approved):
I have more testing I want to do when HTTP hits, but this looks good!

briancain:
Thank you all who took the time to review and test! Super helpful, appreciate you all ❤️ 🙏🏻

briancain:
Hey there @paladin-devops - soooon! I hope to have a draft PR up this week 😄 Once it's up, it'll likely go out in the next minor Waypoint release. I can give you a ping if you're interested in checking it out in the PR form.

briancain:
Awesome @paladin-devops !! That's great to hear. Appreciate you trying that PR early. Very pleased to hear it worked like a charm 😄

PR #2865, opened 2021-12-23
Pin the build version of the kaniko container

izaaklauer:
To avoid introducing unexpected changes on each build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ RUN touch /tmp/.keep
 #--------------------------------------------------------------------
 # This target is explicitly invoked from the command line, it's not used
 # by the non-odr stages.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN touch /tmp/.keep
 #--------------------------------------------------------------------
 # This target is explicitly invoked from the command line, it's not used
 # by the non-odr stages.
-FROM gcr.io/kaniko-project/executor:latest as odr
+FROM gcr.io/kaniko-project/executor:v1.6.0 as odr
```
briancain, line 35: Maybe we should use `1.7.0`? I think that's latest at the moment.

briancain (approved):
Makes sense to me!

mitchellh (approved):

PR #2878, opened 2022-01-05
Upsert Workspace with remote init operation is called

catsby:
This PR adds a call to `UpsertWorkspace` when executing a remote init operation (for example, from the UI). 

Note that the UI does not currently offer any option to specify the Workspace

### DIFF
--- a/internal/runner/operation_init.go
+++ b/internal/runner/operation_init.go
@@ -4,6 +4,8 @@ import (
 	"context"
 
 	"github.com/hashicorp/go-hclog"
+	"google.golang.org/grpc/codes"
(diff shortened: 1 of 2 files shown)


krantzinator (approved):

evanphx (approved):

evanphx:
This is fine to do, but I'll also mention that the state APIs also create the workspace whenever needed.

catsby:
> the state APIs also create the workspace whenever needed.

Indeed! Internally we have discussed changing this to require the explicit creation first (with the exception of the `init` command). This is part of some first-steps toward requiring explicit creation

PR #2879, opened 2022-01-06
ui: updated the ui of list items on deployments tab

### LINKED ISSUE #2796: ui: convert deployment list info to new design
## Context
With the new redesign, we wanted to streamline the look and information shown on each deployment list item. @almonk has designed a new version of the deployment list item to go along with the new UI. See the Figma design document for specific implementation details

sabrinako:
## Notes
- The URL shown on each list item is the **_deployment URL_**, the designs state it should be the _**release URL**_, but in consideration of the context and after discussion, we chose to show the deployment URL here
- Currently the frontend does not receive the commit message with git commit information at all, so that needs to be implemented later
- As noted in some other PRs, currently there are no animated icons in `flight`, so the `Deploying...` icon is the `clock` icon until the animated loading one is available

## Screenshots
**_Note the left sidebar_**
### Before
Screen Shot 2022-01-06 at 3 06 24 PM

### After
Screen Shot 2022-01-12 at 12 33 14 PM

## How to test
1. Pull down the branch `git checkout ui/deployment-card-state`
2. `yarn start`
3. Check the deployments tab of an application
4. Optional: test UI against a real Waypoint server


ui/app/components/app-item/deployment.hbs
```diff
@@ -7,29 +7,66 @@
...
+            (clean-up-url @deployment.preload.deployUrl)
+            (t (concat "status_report_indicator.label."
+              (downcase @deployment.statusReport.health.healthStatus))
+            )
+          )
+        ) as |vars|}}
+          <span
+            data-test-health-status={{downcase vars.deploymentHealth}}
+            class="deploy-url__status deploy-url__status--{{downcase vars.deploymentHealth}}"
+          />
```
jgwhite, line 30: CleanShot 2022-01-11 at 17 27 02@2x

When a URL is present, it looks like we’re only using the color of the indicator to convey the health status. We should add an additional means to convey this information.

Per [WCAG SC 1.4.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html):

> Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element

ui/app/components/app-item/deployment.hbs
```diff
@@ -7,29 +7,66 @@
...
+      {{!-- ROW: git commit information --}}
+      {{#if @deployment.preload.jobDataSourceRef.git.commit}}
+        {{#let @deployment.preload.jobDataSourceRef.git.commit as |commit|}}
+          {{!-- TODO: change to if statement- if user profile picture available, use that. else use flight icon --}}
+          <FlightIcon @name="git-commit"/>
+          <span data-test-git-commit class="git-commit" title="Triggered by commit: {{commit}}">
+            <CopyButton @clipboardText={{commit}}>
+            <span class="git-commit__sha">{{truncate-commit commit}}</span>
+            </CopyButton>
+            {{!-- TODO: when commit message available, add to this --}}
```
jgwhite, line 66: Following on from our discussion of TODOs the other day, I think it could be just a little clearer that we’re saying “*when* Waypoint gains the ability to store commit author and message data *then* add…”

ui/app/components/app-item/deployment.hbs
```diff
@@ -7,29 +7,66 @@
...
+          data-test-time-icon
+          @name={{or
+            (and (eq status.state 0) "help")
+            (and (eq status.state 1) "clock")
+            (and (eq status.state 2) "check")
+            (and (eq status.state 3) "alert-triangle")
+          }}
+          @size="16"
+        />
+        <span data-test-operation-status>
```
jgwhite, line 48: Are we intending to add the tooltips back to these at some point? Or are we reserving that information for the detail page?

jgwhite (changes requested):
Looking good. I have one significant comment regarding accessibility of the health status indicator but all great apart from that.

ui/app/components/app-item/deployment.hbs
jgwhite: Just to hammer the point home, here’s how a simulation of how this would look to someone with deuteranopia (via [Sim Daltonism](https://michelf.ca/projects/sim-daltonism/)):

CleanShot 2022-01-11 at 17 41 46@2x

ui/app/components/app-item/deployment.hbs
sabrinako: I assumed it'd be too busy with the tooltips, but @almonk can weigh in as well

ui/app/components/app-item/deployment.hbs
jgwhite: @sabrinako How about we show the URL if and only if the health status is `available`? I think that would adhere pretty closely to the WCAG guidelines, on the basis that the presence of a URL implies availability.

ui/app/components/app-item/deployment.hbs
jgwhite: I’m fine with that. Probably also better from an a11y perspective, so we’re not nesting semi-interactive stuff within a link.

ui/app/components/app-item/deployment.hbs
jgwhite: @gregone pointed out that we might want to also show the URL in the `unknown` state, for deployments that don’t/can’t provide health status. I think this is worth digging into, but perhaps in a subsequent PR.

almonk:
I didn't make this clear enough so my bad – but my idea was to _only_ show the release URL if the state was healthy. So in the case of any state OTHER than healthy, you'd see the status text...

CleanShot 2022-01-12 at 16 04 34@2x

jgwhite (approved):
🎉

PR #2888, opened 2022-01-07
builtin/docker: fix crash during stop task with ODR

mitchellh:
A crash could occur if a container is already removing, since we set err
to nil and then later check `err.Error()`. This crash causes no actual
problems cause the container is already auto-removing, but it makes the
logs noisier than they need to be.


briancain (approved):
🐋

PR #2891, opened 2022-01-10
website: document dynamic variable values

mitchellh:
This adds docs for #2889.

This uses the new `dynamic` function name which I'm opening a PR for separately now.


evanphx (approved):

izaaklauer (approved):
Nice!

The other docs path I can think of is that we link to `app-config/dynamic.mdx` from each plugin (e.g. [here](https://www.waypointproject.io/plugins/terraform-cloud#required-parameters)), but that page only describes app config. We might want to add a link to `variables/dynamic.mdx` there too.

mitchellh:
Good catch @izaaklauer! I made sure the docs were clear of configdynamic.

PR #2892, opened 2022-01-10
internal/config/dynamic: rename configdynamic() to dynamic()

mitchellh:
This was discussed heavily in Slack and Google Docs.

`configdynamic` remains functionality indefinitely, its harmless to keep it working for now as long as we note its deprecated and do not document that form.


evanphx (approved):

izaaklauer (approved):
Nice!

If we want to keep going down this vein, we might want to also rename `configsourcer` plugins to `dynamicsourcer`s.

mitchellh:
Yeah I agree with that. That's a lot more surgical, deep change so we can do that another time. Its less user facing anyways (besides the docs).

PR #2893, opened 2022-01-11
ui: 0.7 cleanup (fix dark mode styling, spacing on timeline comp, etc.)

sabrinako:
## Included in this PR
- [x] Clean up styling in dark mode (ensure all coloring looks correct)
- [x] Add additional space after the check mark icons of each line in timeline component
- [x] Fix breadcrumbs (seems like most of them except for deployments are off)
- [x] Page headers with icon are missing (build and release details pages)

## How to test
1. Pull down branch `git checkout ui/cleanup-0.7`
2. `yarn start`
3. Click around and check that the things state above look correct

### Before (Dark Mode)
Screen Shot 2022-01-11 at 12 50 36 PM
^ Also note the workspace switcher coloring

Screen Shot 2022-01-11 at 12 52 11 PM
^ Note breadcrumbs

### After (Dark Mode)
Screen Shot 2022-01-11 at 12 45 23 PM
Screen Shot 2022-01-11 at 12 46 37 PM
^ Note breadcrumbs


ui/app/templates/workspace/projects/project/app/logs.hbs
```diff
@@ -1,4 +1,4 @@
 {{page-title (concat @model.application.application "Logs")}}
 
 <h3>Application logs</h3>
-<LogStream @req={{@model}}></LogStream>
+<LogStream @req={{@model.request}}></LogStream>
```
jgwhite, line 4: What‘s the background on this change? It seems different from the others.

gregone, line 4: is this intentional? Doesn't seem related to the PR, and I see no other changes on the logs route

jgwhite (commented):
This is looking fabulous. One question before I approve.

ui/app/templates/workspace/projects/project/app/logs.hbs
sabrinako: It's to handle the breadcrumbs, I added the application name to the model so we could access it for that. If I remembered correctly, you can't use `modelFor` outside of the `model` function, otherwise I would have just grabbed it that way

jgwhite: Ah gotcha, because the model for the route is now an object *containing* the request, where before it was just the request 👍

jgwhite: With that clarification, I’m going to approve this.

jgwhite (approved):

jgwhite:
- [ ] Logs and Exec pages have headings but Deployments, Builds, and Releases pages do not

Perhaps none of these pages needing a heading anymore?

CleanShot 2022-01-12 at 15 31 14@2x

CleanShot 2022-01-12 at 15 31 02@2x

PR #2895, opened 2022-01-12
Feat/docker auth

### LINKED ISSUE #2577: Waypoint Docker builder FROM private registry fails
**Describe the bug**
While using a Dockerfile in the Waypoint build stage with a base image from a private registry, waypoint build fails. This fails even while logged into the private registry on the same CLI running Waypoint (no remote operations enabled).

**Steps to Reproduce**
1. Write Dockerfile with a FROM layer, pulling from a private registry (requiring authentication)
2. Log in to the registry with docker login
3. Run waypoint build

I set up a local private registry to replicate this issue (following this post https://docker.github.io/get-involved/docs/communityleaders/eventhandbooks/docker101/registry/) which I've seen in other scenarios with a private registry on Dockerhub itself.

waypoint build
```
» Building docker-lab...
❌ Building image...
 │ Step 1/1 : FROM localhost:5000/alpine:3.6
! unable to stream build logs to the terminal: Head
  http://localhost:5000/v2/alpine/manifests/3.6: no basic auth credentials
```

as opposed to "docker build ."
```
docker build .
Sending build context to Docker daemon  8.704kB
Step 1/1 : FROM localhost:5000/alpine:3.6
3.6: Pulling from alpine
Digest: sha256:36c3a913e62f77a82582eb7ce30d255f805c3d1e11d58e1f805e14d33c2bc5a5
Status: Downloaded newer image for localhost:5000/alpine:3.6
 ---> 43773d1dba76
Successfully built 43773d1dba76
```

**Expected behavior**
Waypoint successfully pulls down base image from private registry and continues with build.

Waypoint CLI & server version 0.6.1
Docker version 20.10.7, build f0df350

**Additional context**
The current workaround to this issue is to first pull down the image that the Dockerfile (to be used by waypoint build) will use. It seems that the Docker builder plugin IS capable of recognizing and using the image when it is cached locally, but cannot reach out to the remote private registry at build time to use it as a base image.

This workaround is doable for local execution of Waypoint, but this breaks in a pipeline environment. For example, in a Jenkins pipeline running Waypoint commands, Jenkins would have to be told which image to pull down from the private registry (after logging in) using a docker pull command, prior to the execution of waypoint build. This seems redundant and I think Waypoint should be capable of pulling down the necessary base image at build time.

I have not tested out this behavior with Waypoint runners, but I'd imagine the situation would grow more complex there, and possibly could involve the use of hooks prior to the build to login to Docker and pull down the image.

paladin-devops:
Fixes #2577.

Uses [AuthConfig](https://pkg.go.dev/github.com/docker/docker/api/types?utm_source=godoc#AuthConfig) type from Docker's Go API in order to authenticate to a private registry when an image is built. This is required if a base image is from a private repo.

Example build stanza with username/password auth configured:
```hcl
  build {
    use "docker" {
      auth {
        hostname       = "registry.hub.docker.com"
        username       = "dockeruser"
        password       = "dockerpass"
      }
    }
  }
```

All of the following parameters are supported (map of string):
```
        hostname      = ""
        username      = ""
        password      = ""
        email         = ""
        auth          = ""
        serverAddress = ""
        identityToken = ""
        registryToken = ""
```


builtin/docker/builder.go
```diff
@@ -424,14 +436,28 @@ func (b *Builder) buildWithDocker(
...
+	var authMap = make(map[string]types.AuthConfig)
+	if b.config.Auth != nil {
+		authMap[authConfig["host_name"]] = types.AuthConfig{
+			Username:      authConfig["user_name"],
+			Password:      authConfig["password"],
+			Email:         authConfig["email"],
+			Auth:          authConfig["auth"],
+			ServerAddress: authConfig["server_address"],
+			IdentityToken: authConfig["identity_token"],
+			RegistryToken: authConfig["registry_token"],
```
evanphx, line 448: `authConfig` should be a struct instead of a map to avoid folks setting fields that are ignored.

evanphx (changes requested):
Looks fine other than using a struct for `Auth`. Could you also rebase the PR?

builtin/docker/builder.go
```diff
@@ -39,6 +39,17 @@ type Builder struct {
 	config BuilderConfig
 }
 
+type Auth struct {
+	Hostname      string `hcl:"hostname,optional"`
```
evanphx, line 43: This field doesn't appear to be used?

evanphx (changes requested):

evanphx:
@paladin-devops Super! Yes, please take a look at the referenced plugin components and see about making the have a common auth mechanism. Thanks!

evanphx:
@paladin-devops So I think I realized one reason I was confused about the hostname: In the registry build case, it's the hostname embedded in the image name that governs where the auth info goes. So maybe error out if they specify a Hostname to the auth block in registry.

builtin/docker/builder.go
```diff
@@ -424,14 +452,28 @@ func (b *Builder) buildWithDocker(
 		return status.Errorf(codes.InvalidArgument, "buildkit is required to use platform option")
 	}
 
+	var authMap = make(map[string]types.AuthConfig)
+	if (Auth{} != *b.config.Auth) {
```
evanphx, line 456: This is a really strange conditional, I guess just checking if the value is empty. I've never seen this before, maybe leave a comment (and personally I'd put the `Auth{}` on the right hand side).

builtin/docker/pull/builder.go
```diff
@@ -47,8 +48,8 @@ type BuilderConfig struct {
 	// Control whether or not to inject the entrypoint binary into the resulting image
 	DisableCEB bool `hcl:"disable_entrypoint,optional"`
 
-	// The docker specific encoded authentication string to use to talk to the registry.
-	EncodedAuth string `hcl:"encoded_auth,optional"`
```
evanphx, line 51: So we can't delete this because we'll break everyone that is using it, so the new Auth mechanism has to work along side EncodedAuth (probably just use EncodedAuth if set, otherwise the new Auth).

evanphx (changes requested):
So we need to still support EncodedAuth and probably at a minimum, move the `Auth{}` to the right hand side of those empty checks.

builtin/docker/builder.go
paladin-devops: Agreed, I couldn't find a simpler way to check if a struct is empty, though! Will add a comment and move to right side.

builtin/docker/pull/builder.go
paladin-devops: I'll work that back in there for backwards-compatibility.

paladin-devops: Still working on the registry piece, but for the puller, all of these test cases now work as expected:

```hcl
    //auth used for private repo, expected build success
    workspace "pull" {
      use "docker-pull" {
        image = "paladindevops/alpine"
        tag   = "very-latest"
        auth {
          username = "username"
          password = "password"
        }
      }
    }

    //encoded_auth used for private repo, expected build success
    workspace "pull-encoded" {
      use "docker-pull" {
        image = "paladindevops/alpine"
        tag = "very-latest"
        encoded_auth = "encoded_docker_auth"
      }
    }

    //no auth used for private repo, expected build failure
    workspace "pull-noauth" {
      use "docker-pull" {
        image = "paladindevops/alpine"
        tag = "very-latest"
      }
    }

    //no auth used for public repo, expected build success
    workspace "pull-public" {
      use "docker-pull" {
        image = "alpine"
        tag = "latest"
      }
    }
```

evanphx (approved):
Looks good! Could you resolve the merge conflicts?

evanphx:
@paladin-devops Ok, we're in the home stretch! You'll need to include a changelog entry and then regenerate the website docs by doing: `go run ./cmd/waypoint docs -website-mdx`

Issue #2791, opened 2021-12-01
nomad-jobspec deployment with Nomad ODR deploys to ODR's namespace


paladin-devops:
**Describe the bug**
In my Nomad jobspec file, I specify the job's namespace to be the Waypoint workspace of the deploy. However when I try to deploy with remote operations by using a Nomad ODR, which runs in a namespace that is different from my application job's namespace, the application job deploys into the Nomad ODR job's namespace (which is specified in the runner profile).

**Steps to Reproduce**
1. Set a Nomad runner profile with a given namespace (waypoint).

Command:
```
waypoint runner profile set -plugin-type=nomad -name=nomad-large -plugin-config=./nomad-large-profile.json
```
Config:
```
{
  "datacenter": "dc1",
  "namespace": "waypoint",
  "nomad_host": "http://waypoint01.node.dc1.consul:4646",
  "region": "global",
  "resources_cpu": "2000",
  "resources_memory": "4000"
}
```
2. Use project apply to set the runner profile for an existing project with a Nomad jobspec plugin
```
waypoint project apply -runner-profile=nomad-large my-project
```
3. Deploy the project w/remote operations (after it has already been built and initialized)

Command:
```
waypoint deploy -workspace=prod -var-file=./vars.wpvars my-project/my-app
```
Deploy stanza:
```
deploy {
    use "nomad-jobspec" {
      jobspec = templatefile("${path.app}/app.waypoint.nomad.hcl", {
        namespace         = workspace.name
        region            = var.region
      })
    }
  }
```
Job file excerpt:
```
job "app" {
  name  = "app"
  datacenters  = local.dcs //from elsewhere in the jobspec
  type = "$${var.job_type}" //from elsewhere in the jobspec
  namespace = "${namespace}"
  region = "${nomad_region}"
```

The result of these steps is that the Nomad job of "app" is deployed to the namespace "waypoint", and NOT the namespace "prod" which would be templated into the Nomad job file by "workspace.name".

**Expected behavior**
The Waypoint runner is spawned in the namespace defined in its runner profile configuration (this part is currently happening and does not appear to be broken), and the application's Nomad job is deployed to the namespace specified in the jobspec.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: 0.6.2.
* Waypoint Server Platform and Version: Nomad 1.2.2
* Waypoint Plugin: nomad-jobspec

**Additional context**
[...]

PR #2896, opened 2022-01-12
Feat/nomad odr namespace fix

paladin-devops:
Addresses #2791.

The Nomad API client's namespace configuration will be set equal to the namespace in the jobspec, for Nomad jobspec deployments. The image below showcases the desired result of a jobspec deployment with `namespace = "dev"` set in its jobspec, where the Waypoint server & static runner are deployed to a namespace (in this case, default), the ODR is deployed to a namespace (in this case, default), and the actual application (in this case, nomad-test) is deployed to a separate namespace, dev, by the ODR.

image

My ODR profile, for reference:
```json
{
	"datacenter": "dc1",
	"namespace": "default",
	"nomad_host": "http://my-nomad-host:4646",
	"region": "global",
	"resources_cpu": "200",
	"resources_memory": "600"
}
```
Excerpt from my jobspec, for reference:
```hcl
job "nomad-test" {
  type = "service"
  namespace = "dev"
```

### DIFF
--- a/builtin/nomad/jobspec/platform.go
+++ b/builtin/nomad/jobspec/platform.go
@@ -83,6 +83,9 @@ func (p *Platform) Deploy(
 	// Set our deployment ID on the meta.
 	job.SetMeta(metaId, result.Id)
 
+	// Update our client to use the Namespace set in the jobspec
+	client.SetNamespace(*job.Namespace)
+
 	// Register job
 	s.Update("Registering job %q...", *job.Name)
 	regResult, _, err := jobclient.Register(job, nil)
--- a/.changelog/2896.txt
+++ b/.changelog/2896.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+plugin/nomad: Fix Nomad job namespace when using ODRs
+```


briancain (commented):
This change looks good to me, pretty easy fix! If you could, please rebase the commit history down to just the change (i.e. remove the merge commits) and add a `bug` fix [changelog .txt file](https://github.com/hashicorp/waypoint/blob/main/.github/CHANGELOG_GUIDE.md)! Thank you!

builtin/nomad/jobspec/platform.go
```diff
@@ -83,6 +83,9 @@ func (p *Platform) Deploy(
 	// Set our deployment ID on the meta.
 	job.SetMeta(metaId, result.Id)
 
+	// Update our client to use the Namespace set in the jobspec
+	client.SetNamespace(*job.Namespace)
```
briancain, line 87: Does this need to be a pointer here, or can we use the `job.Namepsace` string value directly? 🤔

briancain (approved):
Looks good to me! Just had a minor question. Thanks for taking the time to open this pull request and fixup the issue you reported! 🙏🏻

builtin/nomad/jobspec/platform.go
evanphx: We checked and it is a pointer already, @briancain was mistaken.

PR #2902, opened 2022-01-13
website: upgrade downloads page

BRKalow:
Upgrading the downloads page package, which includes some retry logic when fetching release data from releases.hashicorp.com. This should hopefully reduce the likelihood of a failed site deployment immediately after a product release.


thiskevinwang (approved):
✅

PR #2922, opened 2022-01-19
ui: add ember-testing-sprite-embed

jgwhite:
See https://github.com/hashicorp/flight/blob/%40hashicorp/flight-icons%2F2.0.2/ember-flight-icons/index.js#L9-L13 for background.


jgwhite:
Looks like an a11y validation is failing due to the addition of the spritesheet. Looking into it.

jgwhite:
Icons!

CleanShot 2022-01-19 at 17 43 33@2x

sabrinako (approved):
Thanks for figuring this out, it'll make the Percy testing even better 🎉  LGTM!

PR #2926, opened 2022-01-20
Allow more time for grpc connection

catsby:
Extend the dial timeout for our gRPC connection to be 10 seconds, up from 5. I found with Waypoint on ECS the `grpc.DialContext()` call (around line 107 at time of writing) was taking around 6 seconds. Doubled the timeout to be safe. I'm not entirely sure why this method takes so long _sometimes_. For now though this change will greatly reduce the number of `context exceeded` errors from the CLI when talking to Waypoint on ECS


mitchellh (approved):

briancain (approved):
Sounds good 👍🏻

PR #2930, opened 2022-01-21
ui: limit number of deployments requested

### LINKED ISSUE #2919: ui: Limit number of deployments requested
@briancain and I discussed this and we think as a temporary measure we could limit the number of deployments requested to ~50~ 10. We should still improve the performance of these methods, and pursue true pagination in the API, but the ~50~ 10 limit will hopefully make Waypoint UI usable in cases like this.

_Originally posted by @jgwhite in https://github.com/hashicorp/waypoint/issues/2304#issuecomment-1015858019_

jgwhite:
## Why the change?

Closes #2919

## What’s the plan?

- [x] Set `limit` on `OperationOrder` in `UI_ListDeployments` request
- [x] Document the results in this PR

### Before

~10s for ~50 deployments

CleanShot 2022-01-21 at 13 17 21@2x

### After

~2s for the latest 10 of ~50 deployments (still pretty slow, but tolerable)

CleanShot 2022-01-21 at 13 17 42@2x

## How do I test it?

1. Generate lots of deployments on a Waypoint server of your choosing
2. `git checkout ui/deployments-limit`
3. `cd ui && yarn start local`
4. [Visit http://localhost:4200](http://localhost:4200)
5. Navigate to the app in question
6. Verify that the loading time is tolerable

### DIFF
--- a/.changelog/2930.txt
+++ b/.changelog/2930.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: Limit number of deployments requested
+```
--- a/ui/app/services/api.ts
+++ b/ui/app/services/api.ts
@@ -104,6 +104,8 @@ export default class ApiService extends Service {
 
     let order = new OperationOrder();
     order.setDesc(true);
+    // See https://github.com/hashicorp/waypoint/issues/2919 for more context on this limit
+    order.setLimit(10);
     req.setOrder(order);
 
     let resp = await this.client.uI_ListDeployments(req, this.WithMeta());


briancain (approved):
I am not frontend, but this change makes sense to me as a temporary workaround until we can fix that slow query 🙏🏻

sabrinako (approved):
Code looked good but just wanted to test on local first, definitely a shorter load time but do agree we need to do some more improvements to this endpoint's performance in the future

PR #2935, opened 2022-01-24
Runner Adoption

mitchellh:
This adds runner adoption, a mechanism where runners can enter a "pending" state and must be explicitly "adopted" by a human (or API) before they are accepted and given any work. There are many new concepts introduced in this PR:

To do this, this introduces a new API that runners call on startup `RunnerToken` which blocks until they are adopted. This is fully backwards compatible: prior runners that use a login token continue to work, and don't need to call the `RunnerToken` API. 

This introduces a new token specifically for runners called a "runner token". Operators can create a runner token manually using `waypoint runner token` and avoid the entire adoption process -- useful for autoscaling scenarios.

A new concept called a "server cookie" is also introduced in this PR. The server cookie is similar to an Erlang cookie. It is generated on startup, available via API or `waypoint server cookie`, and can be specified as a header/metadata in ANY API call as an additional check that you are communicating to the proper server. The server cookie is _required_ for the runner adoption API as a basic check that the calling runner is talking to the right cluster. This is not a security mechanism.

## Backwards Compatibility

This PR is fully backwards compatible. If a Waypoint user/operator did nothing, everything behaves identically. 

## Demo

Here is a demo from an earlier implementation, but the basics are still the same:

https://user-images.githubusercontent.com/1299/150837533-7fe3959a-18be-4a9d-8793-ce5b85e55a01.mp4

## RFC

This fully implements WP-109.


internal/runner/runner.go
```diff
@@ -192,15 +232,76 @@ func (r *Runner) Id() string {
...
+		log.Warn("cookie not set for runner, will skip adoption process")
+	}
+
+	if adopt {
+		// Register and initialize the adoption flow (if necessary) by requesting
+		// our token.
+		log.Debug("requesting token with RunnerToken (initiates adoption)")
+		tokenResp, err := r.client.RunnerToken(tokenCtx, &pb.RunnerTokenRequest{
+			Runner: r.runner,
+		})
```
evanphx, line 278: Just so that I'm clear, this becomes a sort of "door knock" protocol if the runner already has a token. Ie, the server will say "hey, yup, that's a good token!" and just return the same token.

internal/server/singleprocess/service_runner.go
```diff
@@ -93,16 +245,64 @@ func (s *service) RunnerConfig(
...
+		if err := s.state.RunnerOffline(record.Id); err != nil {
+			log.Error("failed to mark runner as offline. This should not happen.", "err", err)
 		}
 	}()
 
+	// We'll set preadopt to true if we're in a state where we can implicitly
+	// adopt the runner if we don't have explicit adoption. This is allowed
+	// in two cases: (1) legacy login tokens for pre WP 0.8 and (2) manually
+	// created runner tokens that match this runner.
+	preadopt := false
```
evanphx, line 260: By my reading of the logic, it's not possible to get to line 290 without setting `preadopt` so do we need it? We're rejecting all paths that don't set `preadopt` it appears.

evanphx (changes requested):
One question I answered myself, but another question about the `preadopt` logic.

internal/server/singleprocess/service_runner.go
mitchellh: I thought about this overnight and agree.

evanphx (approved):

internal/cli/runner_list.go
```diff
@@ -0,0 +1,166 @@
...
+		f.BoolVar(&flag.BoolVar{
+			Name:   "pending",
+			Target: &c.flagPending,
+			Usage:  "List only runners pending adoption.",
+		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:   "json",
+			Target: &c.flagJson,
+			Usage: "Output trigger URL configuration list information as JSON. This includes " +
```
briancain, line 122:
```suggestion
			Usage: "Output runner configuration list information as JSON. This includes " +
```

internal/server/singleprocess/service_runner.go
```diff
@@ -67,6 +69,156 @@ func (s *service) RunnerGetDeploymentConfig(
 	}, nil
 }
 
+func (s *service) AdoptRunner(
+	ctx context.Context,
+	req *pb.AdoptRunnerRequest,
```
briancain, line 74: It'd be good to get these service functions in the runner with the `serverptype` Request validations like our other service funcs have.

briancain (changes requested):
Looks pretty good to me! Just a few minor comments ✨ 🍪

internal/server/singleprocess/service_runner.go
mitchellh: Done!

briancain (approved):
✨ 🍪 🏃🏻

krantzinator (approved):

Issue #2817, opened 2021-12-08
Nomad Blue-Green Deployment Support


paladin-devops:
**Is your feature request related to a problem? Please describe.**
I'm working on implementing  blue-green deployments using Waypoint with Nomad in conjunction with Consul. The approach I've taken loosely follows [this guide](https://learn.hashicorp.com/tutorials/consul/service-mesh-traffic-splitting) (loosely because this tutorial uses docker-compose, not Nomad).

**Describe the solution you'd like**
A mechanism to control blue-green promotions/switches in Consul/Nomad.

**Describe alternatives you've considered**
I've implemented a solution which is driven by Waypoint and other Hashicorp tools, whose major components are described below.

Waypoint w/Nomad jobspec:
Project w/Docker build with Nomad jobspec deploy plugin, and some input metadata. The other interesting part here is the hook. This hook will update a value in Consul KV which contains vars for a Consul-Terraform-Sync module to consume.
```hcl
project = "website"

app "nginx" {
  build {
    use "docker" {
      dockerfile = templatefile("${path.app}/Dockerfile", {
        tag          = var.nginx_base_image_tag
      })
      context            = "${path.app}/../.."
      buildkit = true
    }
    registry {
      use "docker" {
        image        = "docker.io/myrepo/${app.name}"
        tag          = var.version
        encoded_auth = var.docker_auth
      }
    }
  }
  deploy {
    hook {
      when    = "after"
      command = [ "/usr/local/bin/curl", "--request", "PUT", "--data", "@contents", "http://consul.service.consul:8500/v1/kv/cts" ]
    }
    use "nomad-jobspec" {
      jobspec = templatefile("${path.app}/nginx.waypoint.nomad.hcl", {
        env       = workspace.name
        region    = var.nomad_region
        version   = var.version
      })
    }
  }
}
```

Contents of payload to Consul KV
```hcl
blue_tag="v1" //old version
green_tag="v2" //new version
```

Nomad canary:
The excerpts from my Nomad jobspec below show how I'm using a canary deployment with metadata on the instances of my service to set the version for CTS to consume. One way I'm limited here with Waypoint is that Waypoint does not have support for promoting Nomad canaries.
```hcl
//both the update and service stanzas are at the group level of the Nomad job spec
  update {
      max_parallel = 1
      health_check = "checks"
      stagger      = "10s"
      auto_revert  = true
[...]

Issue #1536, opened 2021-05-25
internal/stepgroup: StepGroup.Status wipes text and doesn't update text color


briancain:
**Describe the bug**

When using StepGroup.Status to update the color of some output, it ends up instead printing the words `ok`, or `warn`, or `error` instead of updating the requested text color for a given status.

**Steps to Reproduce**

```golang
s := sg.Add("The start of the stepgroup")
...
...
// A few lines later
s.Update("some text to be printed!!")
s.Status(terminal.StatusOk)
s.Done()
```

The output ends up just looking like (in green):

```shell
# surrounding output
✓ ok
```

```golang
s := sg.Add("The start of the stepgroup")
...
...
// A few lines later
s.Update("some text to be printed!!")
s.Status(terminal.StatusWarn)
s.Done()
```

Additionally, using `StatusWarn` prints `warn` in the green StatusOK color text:

```shell
# surrounding output
✓ warn
```

**Expected behavior**

Using `StepGroup.Status` after a `StepGroup.Update` should update the color of the output to the given status level, and it shouldn't replace the text with `ok`, `warn`, or `error`.

**Waypoint Platform Versions**

Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: `main` 09dba88a3a918a8216eb39ac858c2660ba3cc37b
* Waypoint Server Platform and Version: N/A
* Waypoint Plugin: N/A

briancain:
Potentially fixed by https://github.com/hashicorp/waypoint/pull/1480

PR #2938, opened 2022-01-25
Feat/nomad-jobspec canary promote releaser

paladin-devops:
This PR introduces a releaser for Nomad, which promotes a canary deployment. Supporting canaries in releases for Nomad bring canary/blue-green deployment strategies into Waypoint's lifecycle (#2817).

This is still very much a work in progress. Although I've successfully promoted a Nomad canary deployment with this plugin, I still intend to fully implement the release manager for the releaser, the destroyer (though this will likely be redundant with the Nomad platform destroyer), and there are other configuration options/changes I would like to include:

- [x] Specify particular Nomad task groups to promote (by default this promotes all)
- [x] A Consul URL, or IP/port combo if Consul isn't used
- [x] Add canary option to the Nomad platform plugin
- [x] Support generation ID changes in releaser with nomad-jobspec platform deployment
- [x] Evaluate if scaling should be supported
- [x] Evaluate if failing a canary deployment should be supported
- [x] Evaluate if reverting to a previous deployment should be supported

This is my first attempt at a plugin beyond the tutorial, and I look forward to any feedback while I continue with this draft!


briancain:
Thanks for all the details and keeping us in the loop @paladin-devops - Let us know when you're ready for a team review! For now, we'll give you the space to continue to work on this PR while it's still in draft form :) but wanted to say feel free to ping us when you are feeling like this is ready.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,547 @@
...
+	} else if deploy == nil {
+		st.Update("No active deployment for Nomad job")
+		return err
+	}
+
+	canaryDeployment := false
+	groupsToPromote := make([]string, len(deploy.TaskGroups))
+	for taskGroupName, taskGroup := range deploy.TaskGroups {
+		if r.config.Groups != nil {
+			if isElementExist(r.config.Groups, taskGroupName) && taskGroup.DesiredCanaries != 0 {
```
krantzinator, line 123: You use `> 0` when updating the task group's canary, so for consistency, can we check `>` here too? Or is there a specific reason for checking `!= 0`?
```suggestion
			if isElementExist(r.config.Groups, taskGroupName) && taskGroup.DesiredCanaries != 0 {
```

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,547 @@
...
+				canaryDeployment = true
+				groupsToPromote = append(groupsToPromote, taskGroupName)
+				continue
+			}
+		} else if taskGroup.DesiredCanaries != 0 {
+			canaryDeployment = true
+			// if no groups to promote are specified in the config, promote all groups
+			// that have canaries
+			groupsToPromote = append(groupsToPromote, taskGroupName)
+		}
```
krantzinator, line 133: Is this check supposed to be here, or just L123-L127 (I'm not sure what it's checking that wasn't already checked above)? Does this have the potential of a `nil` reference since we aren't checking for the existence of the task group here?

krantzinator (commented):
Just a few small comments and questions for now! Let me know if you need help on this docs generation as well. Looks great!

builtin/nomad/jobspec/platform.go
```diff
@@ -46,86 +53,246 @@ func (p *Platform) GenerationFunc() interface{} {
...
+	} else if *job.Status == "completed" {
+		jobResource.Health = sdk.StatusReport_PARTIAL
+		jobResource.HealthMessage = fmt.Sprintf("Job %q is reporting partially available!", state.Name)
+	} else if *job.Status == "failed" || *job.Status == "lost" {
+		jobResource.Health = sdk.StatusReport_DOWN
+		jobResource.HealthMessage = fmt.Sprintf("Job %q is reporting down!", state.Name)
+	} else {
+		jobResource.Health = sdk.StatusReport_UNKNOWN
+		jobResource.HealthMessage = fmt.Sprintf("Job %q is reporting unknown!", state.Name)
+	}
```
tgross, line 202: I think we might be co-mingling several different status values here. The status of a job is only one of [running, dead, or pending](https://github.com/hashicorp/nomad/blob/v1.2.5/nomad/structs/structs.go#L4021-L4025) (see also the [API docs](https://www.nomadproject.io/api-docs/jobs#status)). 

Whereas:
* An allocation's client status is one of [pending, running, complete, failed, or lost](https://github.com/hashicorp/nomad/blob/v1.2.5/nomad/structs/structs.go#L9322-L9328). 
* A deployment can be [running, paused, failed, successful, cancelled, pending, blocked, or unblocking](https://github.com/hashicorp/nomad/blob/v1.2.5/nomad/structs/structs.go#L8911-L8919)
* An eval can be [blocked, pending, complete, failed, canceled](https://github.com/hashicorp/nomad/blob/v1.2.5/nomad/structs/structs.go#L10370-L10376) (yes, those two objects spell canceled differently 😊 ). 
* I wasn't able to figure out which object has `queued` or `started`

The reason this granularity matters to a Waypoint user is that a Nomad job can be `running` but have a bunch of its allocations in various pending or failing states! So I suspect we want to aggregate information about the job, deployment, and its allocations and then present that as a top-level Waypoint job status.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,547 @@
...
+	jobClient := client.NomadClient.Jobs()
+	deploymentClient := client.NomadClient.Deployments()
+
+	st.Update("Getting job...")
+	jobs, _, err := jobClient.PrefixList(target.Name)
+	if err != nil {
+		return status.Errorf(codes.Aborted, "Unable to fetch Nomad jobs: %s", err.Error())
+	} else if target.Name != jobs[0].ID {
+		return status.Errorf(codes.Aborted, "Job not found: %s", err.Error())
+	}
```
tgross, line 98: In the Nomad CLI if the prefix isn't an exact match, we present a list of options to the user in the error message so they can refine the prefix match. Also, because `PrefixList` is a list API, Nomad doesn't return an error on empty results, so I believe this would panic with an index error if there's no results?

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,547 @@
...
+
+	st.Update("Getting latest deployments for job")
+	deploy, _, err := jobClient.LatestDeployment(*job.ID, q)
+	if err != nil {
+		return status.Errorf(codes.Aborted, "Unable to fetch latest deployment for Nomad job: %s", err.Error())
+	} else if deploy == nil {
+		st.Update("No active deployment for Nomad job")
+		return err
+	}
```
tgross, line 117: Because we have the deployment at this point, we might want to consider checking the initial deployment status before we go rummaging through the task groups. The deployment might have already failed, completed, been marked successful, etc.

Also, "latest deployment" means literally that and not necessarily the one you're expecting (this is admittedly a bit of a gotcha in the Nomad API!). When you submit a job:

* (1) The job is forwarded to the leader, and stored in the Nomad state store. At the same time, an evaluation is created.
* (2) A scheduler worker (not necessarily on the leader) dequeues the evaluation and creates a plan with the deployment. The plan is sent to the leader to be "applied".

So imagine if we submit a job via the API and then immediately hit `LatestDeployment` before the new deployment is created. We could end up with the deployment for the _previous_ version of the job, which would be confusing because we'd try to promote a possibly already-promoted deployment. So we'll want to match the version for this deployment against the job version we've just submitted.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,547 @@
...
+	// TODO: Force timeout if exceeds healthy deadline or progress deadline of job
+	var currentTaskGroupState *api.DeploymentState
+	var groupHealthy bool
+	for _, group := range groupsToPromote {
+		if group != "" {
+			currentTaskGroupState = deploy.TaskGroups[group]
+			groupHealthy = false
+			for !groupHealthy {
+				if currentTaskGroupState.HealthyAllocs < len(currentTaskGroupState.PlacedCanaries) {
+					time.Sleep(5 * time.Second)
```
tgross, line 151: The top-level function this is in `resourceJobCreate` has a `context.Context`. Should we be using that here rather than having a fixed sleep without a way of breaking it? Is there any kind of backoff we should be using?

tgross (changes requested):
Hi @paladin-devops! Thanks for opening this PR! I'm a developer on the Nomad team so I'm going to admit I'm not as up-to-speed on Waypoint's code base. But I did take a pass through this looking at the Nomad-specific logic and I've left some comments for you.

builtin/nomad/jobspec/releaser.go
paladin-devops: @krantzinator `> 0` makes more sense here, thanks!

builtin/nomad/jobspec/releaser.go
paladin-devops: @krantzinator this check is here in case the user did not provide any task groups in the releaser's config.

Before implementing the ability to specify particular task groups to be promoted in the releaser config, I was using the `PromoteAll` Nomad API method. So when hitting this else block, to avoid adding a task group which has no canary allocs (`DesiredCanaries == 0`) to the `groupsToPromote` slice, we verify that it has canaries. If not, we continue to the next iteration, and if none of the task groups have canaries, we exit early (L135-L138).

So in effect, the logic of this for loop & hitting this else block is "if there are no groups specified in the releaser, promote every group that has canaries".

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,547 @@
...
+        %{ endfor ~}
+
+        // Ensure we set PORT for the URL service. This is only necessary
+        // if we want the URL service to function.
+        PORT = 80
+      }
+    }
+  }
+
+	group "app-gateway" {
```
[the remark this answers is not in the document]
paladin-devops: I'm not sure what's going on with my Go formatting 😅 in my IDE (Atom) each of these appear to be aligned. I opened the file in Vim and the spacing being off was apparent so I just fixed it up there. Guess I'll do my dev work in Vim now 🤷🏻‍♂️

builtin/nomad/jobspec/platform.go
paladin-devops: @tgross thank you for advising on this! It looks like "queued" and "started" may have initially been sourced from the [TaskGroupSummary struct](https://github.com/hashicorp/nomad/blob/35c22bcb6c31d3605d9fe9f2a02cefee04e6c4b1/api/jobs.go#L1003-L1010), but I agree that it would make sense to do some kind of top-level status. Here's what I'm thinking...

If a deployment's status is "completed", skip the eval check and do alloc health check.

If a deployment's status is running, check eval ID returned. Since the eval may return a "blocked", "failed" or "canceled" status, that would allow us to skip the alloc evaluation and jump straight to a final unhealthy status.

If a deployment's status is "paused", "failed", or "blocked", early exit (skip alloc health check) with a final status.

If a deployment's status is "pending" or "unblocking", I'm a bit unsure how to proceed here... I'm thinking either provide an UNKNOWN status, or wait for some period of time and check again for a different status until we hit a timeout on the status check. 

For the alloc health check, if all allocs are "running", we go with "READY". If some are "running" and some are "pending" or "failed", we report "PARTIAL". If all are "lost" or "failed", we report "DOWN". I feel like if we have allocs in the "complete" state, then technically the job is "DOWN", but that could possibly be perceived as unhealthy to Waypoint users, even if the job ran to completion as intended (like a batch job).

Final status options for Waypoint, for reference:
- UNKNOWN: We cannot determine the status of the resource
- ALIVE: The resource exists but might not be ready to handle requests
- READY: The resource exists and is ready to handle requests
- DOWN: The resource is down and not responding
- MISSING: We're expecting it to exist, but it does not
- PARTIAL: Some resources in deployment are OK, others are not OK

builtin/nomad/jobspec/releaser.go
paladin-devops: @tgross I added a check to make sure the list of jobs that the `PrefixList` actually has an element before checking the exactness of the job ID. Thanks!

& regarding presenting a list of options to the user, since the job ID we're checking against is actually the job ID from the Waypoint deploy phase, I think we should still fail here, but maybe output the name of the first one or two jobs returned in the list, saying something like "maybe you meant to promote the canary deployment for X job". Is this what you had in mind?

builtin/nomad/jobspec/releaser.go
paladin-devops: @tgross I caught up to this comment after my response to your remarks on the status check, and I agree a similar "early exit" strategy could be applied here if the deployment fails/is unhealthy.

And thank you for the tip on the gotcha! I'll add in a check to verify that the job version returned by the job `Info` call matches up with the job version returned by the `LatestDeployment`.

builtin/nomad/jobspec/releaser.go
tgross: Yeah, that's the sort of thing I had in mind.

builtin/nomad/jobspec/platform.go
tgross: > If a deployment's status is "completed", skip the eval check and do alloc health check.

A deployment could be "complete" with failed-but-replaced allocations. Or you could complete the deploy and then immediately have an unrelated alloc failure. I'm not sure of the semantics the Waypoint user would want there in terms of the output of this command?

> If a deployment's status is running, check eval ID returned. Since the eval may return a "blocked", "failed" or "canceled" status, that would allow us to skip the alloc evaluation and jump straight to a final unhealthy status.

If the deployment is "running", the eval that led to that deployment can only be "complete". The scheduler may have _also_ spawned a new blocked eval if the plan was only partially applied. Ex. if you submit a job that needs 20 allocs but can only fit 18, we start the deployment for those 18 and then create a new "blocked" eval to re-plan the remaining 2 allocations once more resources become available. In that case it sounds like that'd be `PARTIAL`?

> If a deployment's status is "paused", "failed", or "blocked", early exit (skip alloc health check) with a final status.
>
> If a deployment's status is "pending" or "unblocking", I'm a bit unsure how to proceed here... I'm thinking either provide an UNKNOWN status, or wait for some period of time and check again for a different status until we hit a timeout on the status check.

A failed deployment is definitely a failure (`DOWN`?). 

A paused deployment has been manually paused by the operator with [`nomad deployment pause`](https://www.nomadproject.io/docs/commands/deployment/pause). Not sure what's the best Waypoint state to reflect that.

The "blocked", "pending", "unblocking" states are kind of exotic. Those are for [multi-region deployments](https://learn.hashicorp.com/tutorials/nomad/multiregion-deployments). I'm not sure if Waypoint supports federated Nomad regions at all yet, but if you got those you'd want to poll on them until they become terminal.

> For the alloc health check, if all allocs are "running", we go with "READY". If some are "running" and some are "pending" or "failed", we report "PARTIAL". If all are "lost" or "failed", we report "DOWN". I feel like if we have allocs in the "complete" state, then technically the job is "DOWN", but that could possibly be perceived as unhealthy to Waypoint users, even if the job ran to completion as intended (like a batch job).

That all seems reasonable to me.

Speaking of non-service jobs, you might also want to special case system jobs because they don't use deployments.

builtin/nomad/jobspec/releaser.go
paladin-devops: @tgross took some time, but I think I've got this covered! 

https://github.com/hashicorp/waypoint/pull/2938/files#diff-8465d427b10de27ab3337b90735cb0728c5f62a8ce0612476219fbb6aacc2574R148-R176

I have currently arbitrarily set the context deadline to 5 minutes here ([since it's the default](https://www.nomadproject.io/docs/job-specification/update#healthy_deadline)). Either in this PR or as an enhancement later I want to update this to pull the `HealthyDeadline` from the task group currently being assessed, and possibly also use child Go routines to concurrently check the task groups. Then if one of the task groups doesn't meet its healthy deadline before others don't, we may not have to wait extra time.

builtin/nomad/jobspec/platform.go
paladin-devops: @tgross after further evaluation, I took a different direction here. It occurred to me that since Waypoint continuously polls for app status via the `resourceJobStatus` function (see: #1612), it may not be best for us to rely on the deployment or evaluation from the deployment event, since the information they provide will become stale after the deployment event. Because of that, in the update I made, we now check for the 3 job statuses, running/pending/dead, and only if the job is running do we dig deeper into the status of the job's allocs:

```go
		if running == currentJobVersionAllocs {
			jobResource.Health = sdk.StatusReport_READY
			jobResource.HealthMessage = fmt.Sprintf("Job %q is reporting ready!", state.Name)
		} else if running > 0 && (complete > 0 || failed > 0 || pending > 0 || lost > 0) {
			jobResource.Health = sdk.StatusReport_PARTIAL
			jobResource.HealthMessage = fmt.Sprintf("Some allocations are running for Job %q!", state.Name)
		} else if (complete + pending + failed + lost) == currentJobVersionAllocs {
			jobResource.Health = sdk.StatusReport_DOWN
			jobResource.HealthMessage = fmt.Sprintf("No allocations for Job %q are running!", state.Name)
		}
```
https://github.com/hashicorp/waypoint/pull/2938/commits/608e787bb24121939dc603c72261ba7a6c0d15d8#diff-15311425528bec1ed88b93dfce5b4fa4ef9e23779e2d1f2229a55e976ccbb913R203-R230

paladin-devops: @tgross I've updated the check the list of evals for an alloc, and if there's any blocked evals, but all allocs are running then the status becomes PARTIAL.

Eval check:
```go
	if *job.Status == "running" {
		// Get list of evaluations for job
		evals, _, err := jobClient.Evaluations(*job.ID, q)
		evalStatus := "unblocked"
		for _, eval := range evals {
			switch eval.Status {
			case structs.EvalStatusBlocked:
				evalStatus = "blocked"
				break
			case structs.EvalStatusPending:
				break
			case structs.EvalStatusComplete:
				break
			case structs.EvalStatusFailed:
				break
			case structs.EvalStatusCancelled:
				break
			default:
				evalStatus = "unknown"
			}
		}
```

Alloc check & eval check:
```go
		if running == currentJobVersionAllocs && evalStatus == "unblocked" {
			jobResource.Health = sdk.StatusReport_READY
			jobResource.HealthMessage = fmt.Sprintf("Job %q is reporting ready!", state.Name)
		} else if running == currentJobVersionAllocs && evalStatus == "blocked" {
			jobResource.Health = sdk.StatusReport_PARTIAL
			jobResource.HealthMessage = fmt.Sprintf("Allocs for job %q are running, but there is at least one blocked evaluation!", state.Name)
		}
```

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,576 @@
...
+			d := time.Now().Add(time.Minute * time.Duration(5))
+			log.Debug(fmt.Sprintf("Healthy deadline: %s", d.String()))
+			ctx, cancel := context.WithDeadline(ctx, d)
+			defer cancel()
+			groupHealthy := false
+			for !groupHealthy {
+				currentTaskGroupState = deploy.TaskGroups[group]
+				if currentTaskGroupState.HealthyAllocs < len(currentTaskGroupState.PlacedCanaries) {
+					select {
+					case <-time.After(5 * time.Second): // forces sleep for 5 seconds
```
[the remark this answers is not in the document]
paladin-devops: @tgross thank you for the suggestion! I've updated the 5 second delay to instead use ticker to avoid the leak.

Also, I did not include another check for the deployment by calling the API at the top of the loop because earlier in the function [we get the deployment](https://github.com/hashicorp/waypoint/pull/2938/files/c80b5aca1fd787fa6f80e4a02715a82576e75c5a#diff-8465d427b10de27ab3337b90735cb0728c5f62a8ce0612476219fbb6aacc2574R111) for the first time. So unless the deployment status changed between that initial check and when we hit this loop, we still have a pretty current status of the deployment. I mostly did it this way to avoid a potentially likely unnecessary second call to the Nomad API, since we have a current deployment status already. However if you think making that second call is worth it to avoid hitting the select, I can definitely update it to get the deployment status again prior to hitting the select.

builtin/nomad/jobspec/platform.go
```diff
@@ -2,23 +2,31 @@ package jobspec
...
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"strings"
+	"time"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad/api"
+	"github.com/hashicorp/nomad/nomad/structs"
```
tgross, line 13: This package isn't a stable API. We import it into the Nomad Terraform provider and some of the plugins, and it's hugely painful and something we'd love to undo. Let's avoid pulling it into Waypoint and inflicting that pain on the Waypoint developers too 😀 

(That being said, although the alloc status consts are available in [`api/allocations.go`](https://github.com/hashicorp/nomad/blob/main/api/allocations.go#L17-L29) the eval and job status consts are not and absolutely should be. We'll need to fix this in the Nomad repo.)

go.mod
```diff
@@ -98,14 +99,14 @@ require (
...
 	go.opencensus.io v0.23.0
-	go.uber.org/goleak v1.1.10
-	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
-	golang.org/x/oauth2 v0.0.0-20210402161424-2e8d93401602
-	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
-	google.golang.org/api v0.44.0
-	google.golang.org/genproto v0.0.0-20210602131652-f16073e35f0c
-	google.golang.org/grpc v1.39.1
-	google.golang.org/protobuf v1.26.0
+	go.uber.org/goleak v1.1.12
```
tgross, line 102: This is kicking up some linter issues with the `go.sum`. I think the only reason `go.mod` is being updated here at all is because we pulled in `github.com/hashicorp/nomad/nomad/structs`? If we don't do that then we don't need to update this either.

builtin/nomad/jobspec/platform.go
```diff
@@ -46,86 +54,314 @@ func (p *Platform) GenerationFunc() interface{} {
...
+			case structs.EvalStatusComplete:
+				break
+			case structs.EvalStatusFailed:
+				break
+			case structs.EvalStatusCancelled:
+				break
+			default:
+				evalStatus = "unknown"
+			}
+		}
```
tgross, line 218: Could this be squashed down to a `hasSquashedEvals` bool? That'd make it more clear what the purpose of this block is and let the check below just be on that bool.

tgross (commented):
This is looking pretty good from a Nomad perspective @paladin-devops.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,577 @@
...
+			defer cancel()
+			ticker := time.NewTicker(5 * time.Second)
+			groupHealthy := false
+			for !groupHealthy {
+				currentTaskGroupState = deploy.TaskGroups[group]
+				if currentTaskGroupState.HealthyAllocs < len(currentTaskGroupState.PlacedCanaries) {
+					select {
+					case <-ticker.C:
+					case <-ctx.Done(): // cancelled
+						return status.Errorf(codes.Aborted, "Healthy deadline reached: %s", ctx.Err())
```
briancain, line 172:
```suggestion
						return status.Errorf(codes.Aborted, "Context cancelled from timeout checking health of task group %q: %s", group, ctx.Err())
```

I think this is more accurate here?

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,577 @@
...
+	if r.config.FailDeployment {
+		u, _, err = deploymentClient.Fail(deploy.ID, wq)
+	} else {
+		u, _, err = deploymentClient.PromoteGroups(deploy.ID, groupsToPromote, wq)
+	}
+	if err != nil {
+		return err
+	}
+
+	st.Update(fmt.Sprintf("Monitoring evaluation %s", string(u.EvalID)))
```
briancain, line 198: StepGroup I believe supports string formatting through its parameters, so you shouldn't need `fmt` here, should be able to do `"Monitoring evaluation %q", string(u.EvalID)` without the `fmt.Sprintf`!

briancain: I might of missed this else where, so if there are other places, it would be good to get those updated too!

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,577 @@
...
+	result.Url = job.Meta["waypoint.hashicorp.com/release_url"]
+	return nil
+}
+
+func (r *Releaser) resourceJobDestroy(
+	client *nomadClient,
+	state *Resource_Job,
+	sg terminal.StepGroup,
+) error {
+	// Do nothing because the platform will destroy the job for us
```
briancain, line 216: Maybe a `log.Trace` here saying no action happened would be good for debugging.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,577 @@
...
+		s.Update("No job was found")
+		s.Status(terminal.StatusError)
+		s.Done()
+		s = sg.Add("")
+
+		jobResource.Name = state.Name
+		jobResource.Health = sdk.StatusReport_MISSING
+		jobResource.HealthMessage = sdk.StatusReport_MISSING.String()
+	} else if job == nil {
+		return status.Errorf(codes.FailedPrecondition, "Nomad job response cannot be empty")
```
briancain, line 260: Instead of failing the entire status job, we could make this case an UNKNOWN health with this as the health message 🤔

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,577 @@
...
+		}
+	}
+
+	return &result, nil
+}
+
+// ReleaserConfig is the configuration structure for the Releaser.
+type ReleaserConfig struct {
+	Groups         []string `hcl:"groups,optional"`
+	FailDeployment bool     `hcl:"fail_deployment,optional"`
```
briancain, line 417: Variable comments here would be excellent! Could be a copy of the doc comments defined below.

builtin/nomad/jobspec/platform.go
```diff
@@ -46,86 +54,314 @@ func (p *Platform) GenerationFunc() interface{} {
...
 	// Set our deployment ID on the meta.
 	job.SetMeta(metaId, result.Id)
 
 	// Update our client to use the Namespace set in the jobspec
-	client.SetNamespace(*job.Namespace)
+	client.NomadClient.SetNamespace(*job.Namespace)
 
 	// Register job
-	s.Update("Registering job %q...", *job.Name)
+	st.Update(fmt.Sprintf("Registering job %q...", *job.Name))
```
briancain, line 121: Same feedback here, no need for `fmt`, the StepGroup func should support string formatting!

briancain (changes requested):
Hey there @paladin-devops !! Thanks for taking on this work and doing such a thorough job of it. I've done a first pass of the code to give you some additional review on the Waypoint side. It's a lot of comments, but this is a big PR! Overall I think this is in the right direction.

I haven't tested it yet, mostly just code review. I will do that kind of review next once all the Nomad/Waypoint feedback has been addressed 😄 Thanks again, well done!

builtin/nomad/jobspec/platform.go
paladin-devops: @tgross will do! I originally wrote this with a bool, but I changed to using a string to account for the unknown case, so we could report an unknown status if there were a new Nomad eval status introduced but not handled here.

builtin/nomad/jobspec/platform.go
paladin-devops: Updated!

builtin/nomad/jobspec/releaser.go
paladin-devops: Thanks @briancain! Since `st` here is of type `terminal.Status`, did you mean to convert this (and other usages of Status.Update) to use `StepGroup`?

Also just for clarification, I used `terminal.Status` here instead of `StepGroup` as I saw other comments mentioning to not use `StepGroup` until #1536 is resolved.

briancain: Yeah, so this is a bit confusing. This `st.Update` is a StepGroup, and `Status` is apart of that implementation I believe. The issue linked basically says we should avoid setting `Status` after a `StepGroup` output due to that bug, but I don't think we're doing that here! Just some extra info for you right now as to what that issue is about 😄 

But what I was suggesting is this should be possible, no need to use `fmt.Sprintf`:

```golang
st.Update("Monitoring evaluation %s", string(u.EvalID))
```

briancain:
Looks great @paladin-devops ! You might need to regenerate the website mdx, otherwise you can ignore the Vercel failure right now! I will re-request review for myself so I can give this a proper test 🙏🏻

builtin/nomad/jobspec/releaser.go
paladin-devops: @briancain I actually gave that a try, but it wouldn't compile 🤔 resulting in this warning:
```
too many arguments in call to st.Update
	have (string, string)
	want (string)
```

Dug through the core code, and _think_ the root cause is that `runnerUIStatus` (which is what I'm using in `resourceJobCreate`) does not do the formatting for us already...

```go
func (u *runnerUIStatus) Step(status string, msg string) {
	u.b.mu.Lock()
	defer u.b.mu.Unlock()

	if u.b.evc == nil {
		return
	}
```

https://github.com/hashicorp/waypoint/blob/main/internal/runner/ui.go#L255-L262

...but `runnerUISGStep` does!

```go
func (u *runnerUISGStep) Update(str string, args ...interface{}) {
	msg := fmt.Sprintf(str, args...)

	u.sg.ui.mu.Lock()
	defer u.sg.ui.mu.Unlock()
```

https://github.com/hashicorp/waypoint/blob/main/internal/runner/ui.go#L416-L421

I saw examples in other plugins which use basic concatenation with `status.Update` instead of string formatting, could do that here as well! Let me know what you think. 😄 

```go
	st.Update("Deleting Lambda function version " + deployment.Version)
```

briancain: Oops - ok that makes sense. I guess it was originally a `Status` and not a `StepGroup`! I think your example below with the string concatenation makes sense 👍🏻 Sorry for the mixup!

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,598 @@
...
+				continue
+			}
+		} else if taskGroup.DesiredCanaries > 0 {
+			canaryDeployment = true
+			// If no groups to promote are specified in the config, promote all groups
+			//   that have canaries
+			groupsToPromote = append(groupsToPromote, taskGroupName)
+		}
+	}
+	if !canaryDeployment {
```
briancain, line 153: Is this technically a release failure? If there's no canaries defined in the jobspec file, should we abort the release? Right now we say the release was successful in this case.

```
... deploy output ....
✓ Job "web" is reporting ready!

» Releasing example-nodejs...
⚠️ Nomad canary allocations not detected in job task groups.
✓ Release successfully rolled out!

✓ Finished building report for Nomad platform
✓ Finished building report for Nomad job resource
```

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,598 @@
...
+			health_check = "task_states"
+		}
+
+		service {
+			name = "app"
+			port = 80
+			connect {
+				sidecar_service {}
+			}
+		}
```
briancain, line 512: This job block needs some formatting help. I'm guessing it's just a mix of tabs and whitespace! `network {` on line 490 looks a bit off too.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,598 @@
...
+`)
+
+	doc.Example(`
+// The waypoint.hcl file
+release {
+	use "nomad-jobspec-canary" {
+		groups = [
+			"app"
+		]
+		fail_deployment = false
```
briancain, line 481: I'm thinking since this is optional, we don't include it in the docs. If it's unset, it should default to false.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,598 @@
...
+	if !canaryDeployment {
+		st.Step(terminal.StatusWarn, "Nomad canary allocations not detected in job task groups.")
+		log.Info("Canaries not detected")
+		return nil
+	}
+
+	var currentTaskGroupState *api.DeploymentState
+	for _, group := range groupsToPromote {
+		if group != "" {
+			st.Update(fmt.Sprintf("Checking task group: %s", group))
```
briancain, line 162: I'd love it if we could also print some health information here from Nomad allocation while we wait. Similar to what we do with the Kubernetes plugin:

```
» Deploying tetris...
✓ Kubernetes client connected to https://127.0.0.1:36367 with namespace default
✓ Creating deployment...
✓ Expected "http" port "3000" for app "tetris-v1"
⠏ Waiting on deployment to become available: requested=1 running=0 ready=0
```

Otherwise for unhealthy deployments, it is unclear what is being checked or why it hasn't progressed. I pasted some extra output from the Nomad CLI that I was looking at while waiting for the task group to come up. I think maybe we could show some of this info from `currentTaskGroupState` if possible.

```
Latest Deployment
ID          = 89389c34
Status      = running
Description = Deployment is running but requires manual promotion

Deployed
Task Group  Auto Revert  Promoted  Desired  Canaries  Placed  Healthy  Unhealthy  Progress Deadline
app         true         false     1        1         3       0        3          2022-02-24T14:18:15-08:00

Allocations
ID        Node ID   Task Group  Version  Desired  Status   Created     Modified
a4a534b0  51dfa101  app         1        run      failed   1m23s ago   43s ago
5da6f961  51dfa101  app         1        stop     failed   3m2s ago    1m23s ago
bfb30a7e  51dfa101  app         1        stop     failed   4m12s ago   3m2s ago
55861890  51dfa101  app         0        run      running  13m21s ago  13m11s ago

Time                       Type             Description
2022-02-24T14:10:03-08:00  Alloc Unhealthy  Unhealthy because of failed task
2022-02-24T14:10:03-08:00  Not Restarting   Exceeded allowed attempts 2 in interval 30m0s and mode is "fail"
2022-02-24T14:10:03-08:00  Terminated       Exit Code: 1, Exit Message: "Docker container exited with non-zero exit code: 1"
2022-02-24T14:10:02-08:00  Started          Task started by client
2022-02-24T14:09:45-08:00  Restarting       Task restarting in 16.755013793s
2022-02-24T14:09:45-08:00  Terminated       Exit Code: 1, Exit Message: "Docker container exited with non-zero exit code: 1"
2022-02-24T14:09:44-08:00  Started          Task started by client
2022-02-24T14:09:26-08:00  Restarting       Task restarting in 18.214921411s
2022-02-24T14:09:26-08:00  Terminated       Exit Code: 1, Exit Message: "Docker container exited with non-zero exit code: 1"
2022-02-24T14:09:24-08:00  Started          Task started by client
```

I haven't figured out why my Nomad deployment was unhealthy and the container kept exiting, it might be an issue on my end! But some UX I noticed could be improved :)

briancain (commented):
OK! I was able to give this a test run. I haven't got a successful Canary promotion yet, but I got through enough of it to give a better review of the pull request.

And major props adding status to jobspec, this is quite a large pull request with a lot of major features, so kudos! ✨

builtin/nomad/jobspec/releaser.go
briancain: I noticed this happens on the very _first_ run. No canary allocations are detected in the task group, and the second run, they exist 🤔

builtin/nomad/jobspec/releaser.go
briancain: If you have an idea of why it might be exiting let me know. I don't see any failures in docker logs or Nomad 🤔 I can provide Nomad jobspec and Waypoint HCL files of course if needed.

builtin/nomad/jobspec/releaser.go
paladin-devops: Yep, canaries are introduced in the `update` stanza, and since the very first deployment of a job isn't considered an "update", all allocations are deployed right away. I think it'd make sense to add something to check if it's `jobVersion` 0, and then not report an error here!

briancain: That sounds good to me!

builtin/nomad/jobspec/releaser.go
paladin-devops: @briancain the jobsepc and HCL would definitely help! I can take a look at that for you, & will look into how to improve the UX here.

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,598 @@
...
+`)
+
+	doc.SetField(
+		"groups",
+		"List of task group names which are to be promoted.",
+	)
+
+	doc.SetField(
+		"fail_deployment",
+		"If true, fails the canary deployment.",
```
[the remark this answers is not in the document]
paladin-devops: > Is it that if it's true, instead of promoting the deployment for release, it fails the deployment by marking it as bad in Nomad? 

^ correct! But agreed that it could be articulated better! I intend to update it to "If true, marks the deployment as failed." - stolen from the Nomad CLI docs! 😄 

https://www.nomadproject.io/docs/commands/deployment/fail#command-deployment-fail

builtin/nomad/jobspec/releaser.go
paladin-devops: You're absolutely right, it's possibly something with the linter in one of the linter plugins I'm using in my IDE (Atom) confusing Go with markdown 😅 should be fixed up now, with the help of good old vim.

builtin/nomad/jobspec/releaser.go
paladin-devops: Since the pre-requisite for promoting a Nomad canary deployment is that all canary allocations are healthy, I added a line to display the # of healthy allocs vs. the number of deployed canaries. Once they're equal, that's when the promotion can happen!

```go
st.Update("Waiting on allocations to become healthy: healthy allocs=" + strconv.Itoa(currentTaskGroupState.HealthyAllocs) + " placed canaries=" + strconv.Itoa(len(currentTaskGroupState.PlacedCanaries)))
```

And the output looks like (couldn't copy it live from terminal so took screenshot):
image

briancain: Looks amazing! Love it - thanks for adding that in 👍🏻

builtin/nomad/jobspec/releaser.go
```diff
@@ -0,0 +1,605 @@
...
+
+	// We'll update the user in real time
+	// TODO: Replace ui.Status with StepGroups once this bug
+	// has been fixed: https://github.com/hashicorp/waypoint/issues/1536
+	st := ui.Status()
+	defer st.Close()
+
+	rm := r.resourceManager(log, dcr)
+	if err := rm.CreateAll(
+		ctx, log, st, &result, target,
```
briancain, line 326: This is a pretty subtle error, but we need to include a `StepGroup` parameter here for `CreateAll`. The reason being, for when a Release encounters a failure, Waypoint automatically "rolls back" to a previous version. We do this through the plugin SDK, and end up calling `DestroyAll` directly rather than the top level `Destroy` here. `DestroyAll` expects a stepgroup, and without it, we get a bad error:

```
! 2 errors occurred:
        * rpc error: code = Aborted desc = Context cancelled from timeout checking
  health of task group "app": context deadline exceeded
        * Error during rollback: 1 error occurred:
        * argument cannot be satisfied: type: terminal.StepGroup. This is a bug in the
  go-argmapper library since this shouldn't happen at this point.
```

This is basically saying when our internal SDK attempted to call `DestroyAll` during a Release failure, the arguments it were given could not satisfy any defined Destroy functions that expected a StepGroup.

For the fix, it is pretty simple:

```diff

diff --git a/builtin/nomad/jobspec/releaser.go b/builtin/nomad/jobspec/releaser.go
index 237218fe8..7cb514b10 100644
--- a/builtin/nomad/jobspec/releaser.go
+++ b/builtin/nomad/jobspec/releaser.go
@@ -319,11 +319,12 @@ func (r *Releaser) Release(
        // TODO: Replace ui.Status with StepGroups once this bug
        // has been fixed: https://github.com/hashicorp/waypoint/issues/1536
        st := ui.Status()
+       sg := ui.StepGroup()
        defer st.Close()
+       defer st.Wait()

        rm := r.resourceManager(log, dcr)
        if err := rm.CreateAll(
-               ctx, log, st, &result, target,
+               ctx, log, st, sg, &result, target,
        ); err != nil {
                return nil, err
        }

```

briancain (commented):
Yoo, thanks for all the updates @paladin-devops - I still can't get my Canary to successfully work, but loving all the updates. I found a bug in the failure scenario for releases, otherwise this looks great to me.

I've asked some others on the team to help test this PR too and see if they can get it to work. If you have any other advice for getting the Canary successful I'm all ears 😄

briancain:
I'm pretty good with the code changes here 👍🏻 I'll give the team a few days to see if others can get it to work on their systems.

I assume a Canary release works on yours @paladin-devops ? Is there something specific from the jobspec that I might need? I really only used the `update` stanza:

```hcl
// Nomad jobspec file app.nomad.tpl
job "web" {
  datacenters = ["dc1"]
  group "app" {
    update {
      max_parallel = 1
      canary       = 1
      auto_revert  = true
      auto_promote = false
      health_check = "task_states"
    }

    task "app" {
      driver = "docker"
      config {
        image = "${artifact.image}:${artifact.tag}"

        // For local Nomad, you prob don't need this on a real deploy
        network_mode = "host"
      }

      env {
        %{ for k,v in entrypoint.env ~}
        ${k} = "${v}"
        %{ endfor ~}

        // For URL service
        PORT = "3000"
      }
    }
  }
}
```

```hcl
project = "example-nodejs"

config {
  env = {
    WP_TEST = "test-string"
  }
}

app "example-nodejs" {
  build {
    use "pack" {}
    registry {
      use "docker" {
        image = "nodejs-example"
        tag   = "1"
        local = true
      }
    }
  }

  deploy {
    use "nomad-jobspec" {
      // Templated to perhaps bring in the artifact from a previous
      // build/registry, entrypoint env vars, etc.
      jobspec = templatefile("${path.app}/app.nomad.tpl")
    }
  }

  release {
    use "nomad-jobspec-canary" {
      groups = [
        "app"
      ]
    }
  }
}
```

builtin/nomad/jobspec/releaser.go
paladin-devops: Thanks Brian! Just pushed the update, good catch!

briancain (approved):
Thanks again for this! Looks great, appreciate all of the time spent on feedback and review from us.

tgross (approved):
LGTM 👍

briancain:
Let's merge this thing! 🎉

PR #2943, opened 2022-01-25
Add `.vscode/launch_template.json`

jgwhite:
Just a few debug profiles to get us started

### DIFF
--- a/.vscode/launch_template.json
+++ b/.vscode/launch_template.json
@@ -1,0 +1,40 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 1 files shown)


gregone:
thank you!

gregone (approved):

krantzinator (approved):
Wooooo 🎉 🙌🏻

krantzinator:
@jgwhite Actually, idea on this -- should we name this `launch_template.json` or some such, so that we can alter our local launch.json with various things and have that in our local `gitignore` and not worry about overwriting all the time?

jgwhite:
> @jgwhite Actually, idea on this -- should we name this `launch_template.json` or some such, so that we can alter our local launch.json with various things and have that in our local `gitignore` and not worry about overwriting all the time?

Great idea @krantzinator. I’ve applied this change.

PR #2950, opened 2022-01-26
internal/server: Cache Horizon URL Name and Host for Deploy Bundle

### LINKED ISSUE #2304: Improve performance of UI_ListDeployments and UI_ListReleases
These new endpoints are not especially performant. They have been observed to take >10s response time with ~50 deployments with a wp server on a t3-micro, and ~1s with a small project on a developer laptop:

image

There is some low-hanging fruit here, but will require adding support for more performant queries to our state package: https://github.com/hashicorp/waypoint/blob/f75353c7ca810885c594d37b51bdef9ef6d77c2b/internal/server/singleprocess/service_ui_deploy.go#L27-L28

briancain:
Prior to this commit, when crafting a Deploy bundle, we would include
the applications deployment URL for each deployment in the bundle.
Looking up this hostname requires making an HTTP request to Horizon. The
way this was originally written, we would look up that URL for all
requested deployments. If a user had 80+ deployments in the bolt db,
this request would take quite some time.

Given that each applications deployment vanity name and hostname should
be the same, and only differs by the deployment sequence, we don't need
to get the hostname for each requested deployment. Instead we can grab
it once, and use it for the remaining deployments in the bundle.

Fixes #2304

In my testing, I had around 180 deployments and the request took about 5 seconds to process.
Now with this fix, the same number of deployments only takes ~30-50 ms. For reference, it only
takes about 20 ms to list a single deployment.

### DIFF
--- a/internal/server/singleprocess/service_ui_deploy.go
+++ b/internal/server/singleprocess/service_ui_deploy.go
@@ -39,7 +39,12 @@ func (s *service) UI_ListDeployments(
 		return nil, err
 	}
 
-	var deployBundles []*pb.UI_DeploymentBundle
+	var (
+		deployBundles []*pb.UI_DeploymentBundle
+
+		deployURLName string
+		deployURLHost string
+	)
 
 	for _, deploy := range deployList {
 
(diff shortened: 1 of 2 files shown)


internal/server/singleprocess/service_ui_deploy.go
```diff
@@ -106,7 +119,16 @@ func (s *service) UI_ListDeployments(
...
+
+				deployURLName = hostname.Hostname
+				deployURLHost = strings.TrimPrefix(hostname.Fqdn, hostname.Hostname)
 			}
+		} else {
+			bundle.DeployUrl = fmt.Sprintf(
+				"%s--%s%s",
+				deployURLName,
+				(&ptypes.Deployment{Deployment: deploy}).URLFragment(), // Deployment Sequence Number
+				deployURLHost)
```
jgwhite, line 131: I might be reading this wrong, but I think if `deploy.HasEntrypointConfig == false` then we’d end up executing this `else` clause and trying to set `bundle.DeployUrl` even though there’s no `deployURLHost` to work with.

briancain: Done!

briancain:
General FYI - CircleCI is currently having an outage, so it looks like our `Integration` tests are gonna be stuck for a while.

krantzinator (approved):

PR #2953, opened 2022-01-27
plugin/docker: add no-cache option

radriaanse:
Allow to pass through the no-cache flag to disable the build cache.

This is useful in cases where you want to make sure you're doing a clean build or to mitigate issues with the build cache that arise from time to time.

### DIFF
--- a/.changelog/2953.txt
+++ b/.changelog/2953.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/docker: Add parameter to disable the build cache
+```
--- a/builtin/docker/builder.go
+++ b/builtin/docker/builder.go
@@ -71,6 +71,9 @@ type BuilderConfig struct {
 
 	// Controls the passing of the target stage
 	Target string `hcl:"target,optional"`
+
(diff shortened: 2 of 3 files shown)


radriaanse:
Verified it's working, default behavior stays the same (cache is used):
```
» Building nocache...
-> Initializing Docker client...
-> Building image...
Step 1/2 : FROM scratch
 --->
Step 2/2 : COPY . .
 ---> Using cache
 ---> c124aba5ef06
```

And after setting to true the build log shows no cached layers are used:
```
» Building nocache...
-> Initializing Docker client...
-> Building image...
Step 1/2 : FROM scratch
 --->
Step 2/2 : COPY . .
 ---> c124aba5ef06
Successfully built c124aba5ef06
```

Also tested with buildkit, seems to work just fine as well but can't verify using logs since those are broken.

krantzinator (approved):
Look fine to me, thank you!
A nit that doesn't pertain to this PR specifically is that we should probably update the `buildWithDocker` func to take in options; it's getting a bit unwieldy.

PR #2963, opened 2022-01-28
Hash labels as part of runner tokens and adoption

mitchellh:
This improves #2935 by making adoption label-aware. If a label set changes, the adoption state of a runner is reset. Additionally, runner tokens include a hash of the labels they're minted for, so if the label changes, tokens are invalidated.

This prevents the following insecure scenario: a runner registers and is adopted with a set of labels for development targeting. After adoption, the runner restarts with a set of labels for production targeting. With this PR, this scenario would result in the runner reseting to the un-adopted state and the previously distributed tokens becoming invalid.

For pre-adoption, users can specify the `-label` flag to the `waypoint runner token` command to create a token for a specific set of labels.

### DIFF
--- a/pkg/server/ptypes/runner.go
+++ b/pkg/server/ptypes/runner.go
@@ -13,11 +13,17 @@ import (
 )
 
 // RunnerLabelHash calculates a unique hash for the set of labels on the
-// runner. If there are no labels, the hash value is always zero.
+// runner. This generates a consistent hash value for an empty set of labels.
+// The result is never 0.
 func RunnerLabelHash(v map[string]string) (uint64, error) {
-	if len(v) == 0 {
-		return 0, nil
+	if v == nil {
+		v = map[string]string{}
 	}
+
+	// We always set this special key so that even empty label sets have
+	// a non-zero hash value. This MUST NEVER BE CHANGED otherwise the
+	// hash values for all previously issued tokens will be invalidated.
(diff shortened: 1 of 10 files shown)


evanphx (approved):

briancain (approved):
👍🏻 🏷️

pkg/server/ptypes/runner.go
```diff
@@ -4,13 +4,30 @@ import (
...
+// The result is never 0.
+func RunnerLabelHash(v map[string]string) (uint64, error) {
+	if v == nil {
+		v = map[string]string{}
+	}
+
+	// We always set this special key so that even empty label sets have
+	// a non-zero hash value. This MUST NEVER BE CHANGED otherwise the
+	// hash values for all previously issued tokens will be invalidated.
+	v["waypoint.hashicorp.com/runner-hash"] = "1"
```
catsby, line 26: At the risk of sounding pedantic, but considering the imporatnance stressed in this code comment, is it worth adding a test to ensure `RunnerLabelHash(nil)` always returns the expected value? (`9642272480295632742` according to running it locally).

catsby (approved):

pkg/server/ptypes/runner.go
mitchellh: Good idea.

PR #2967, opened 2022-02-01
statetest: Add missing tests and make timeouts more flexible

evanphx:
Most timeouts were adjust to 3 seconds when the operation is expected to
block, allowing the state implementation more freedom about how quickly
it can detect and signal WatchSets.

This also brings out a config var sorted implementation that was embedded in singleprocess to be more general so other state implementations can use it.

### DIFF
--- a/pkg/server/sort/config.go
+++ b/pkg/server/sort/config.go
@@ -1,10 +1,16 @@
 package sort
 
 import (
+	"errors"
+	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
+	"github.com/hashicorp/waypoint/pkg/config/funcs"
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"github.com/mitchellh/pointerstructure"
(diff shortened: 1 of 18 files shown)


briancain (approved):
🧪 ✨ 
edit: I see the tests are failing, but overall this PR makes sense to me 👍🏻

pkg/server/sort/config.go
```diff
@@ -79,3 +85,188 @@ var (
...
+
+type OrderRequest struct {
+	Set       [][]*pb.ConfigVar
+	Merge     bool
+	Runner    *pb.Ref_RunnerId
+	Workspace *pb.Ref_Workspace
+	Labels    map[string]string
+}
+
+func OrderVariables(req *OrderRequest) ([]*pb.ConfigVar, error) {
```
mitchellh, line 140: Docs would be nice.

mitchellh (changes requested):
Looks great. One doc request and then CI tests are failing.

PR #2970, opened 2022-02-01
core: HTTP Trigger URLs

briancain:
This pull request introduces a new endpoint into Waypoint server for
running Trigger URLs. By request, a user can concurrently stream back
job event stream output through HTTP. It does not allow un-authenticated
triggers to stream back the job event stream, and returns an HTTP 204 No Content.

Fixes https://github.com/hashicorp/waypoint/issues/2987


internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,380 @@
...
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"github.com/hashicorp/waypoint/pkg/serverconfig"
+)
+
+// Message is the message we return to the requester when streaming output
+type Message struct {
+	JobId    string `json:"jobId,omitempty"`
+	Message  string `json:"message,omitempty"`
+	ExitCode string `json:"exitCode,omitempty"`
+}
```
briancain, line 29: Going to update this `Message` to have a `Value interface{}` and a `ValueType string`

internal/server/httpapi/trigger_test.go
```diff
@@ -0,0 +1,46 @@
...
+		//triggerHandler.ServeHTTP(responseRecorder, request)
+
+		//if responseRecorder.Code != 404 {
+		//	t.Errorf("Want status '%d', got '%d'", 404, responseRecorder.Code)
+		//}
+
+		//if strings.TrimSpace(responseRecorder.Body.String()) != "404" {
+		//	t.Errorf("Want '%s', got '%s'", "404", responseRecorder.Body)
+		//}
+	})
```
briancain, line 40: I'm going to update this to use a mocked func to send specific event stream data so we can get some proper tests that the HandleTrigger func handles stream events properly.

internal/cli/trigger_apply.go
```diff
@@ -311,12 +313,17 @@ func (c *TriggerApplyCommand) Run(args []string) int {
...
-	c.ui.Output("Trigger ID: %s", resp.Trigger.Id, terminal.WithSuccessStyle())
-	// TODO(briancain): update output to show trigger URL with wp server attached once http service is implemented
-	//c.ui.Output("Trigger URL: %s", resp.TriggerURL, terminal.WithSuccessStyle())
+	triggerID := resp.Trigger.Id
+	addr := strings.Split(c.clientContext.Server.Address, ":")[0]
+	port := serverconfig.DefaultHTTPPort
+	serverAddr := fmt.Sprintf("%s:%s", addr, port)
+	serverTriggerURL := fmt.Sprintf("https://%s/v1/trigger/%s", serverAddr, triggerID)
+
+	c.ui.Output(" Trigger ID: %s", triggerID, terminal.WithSuccessStyle())
```
catsby, line 325:
```suggestion
	c.ui.Output("Trigger ID: %s", triggerID, terminal.WithSuccessStyle())
```

Not sure if the leading space is intentional here

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,378 @@
...
+
+		// Our API client
+		client := pb.NewWaypointClient(grpcConn)
+
+		requestVars := mux.Vars(r)
+		runTriggerId := requestVars["id"]
+
+		overrideVarJSONRaw := r.URL.Query().Get("variables")
+		var (
+			vo                map[string]string
```
catsby, line 80: Nit pick on names, suggest renaming:

- `overrideVarJSONRaw` -> `variableJSONRaw`
- `vo` -> `variables`

I feel like that will line up with the `Get("variables")` part above better. At first read it wasn't clear what `vo` represented, or would represent, especially sitting just above the `variableOverrides` variable.

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,378 @@
...
+			resp, err = client.NoAuthRunTrigger(ctx, runTriggerReq)
+		}
+		if err != nil {
+			log.Error("server failed to run trigger", "id", runTriggerId, "err", err)
+
+			if status.Code(err) == codes.PermissionDenied {
+				http.Error(w, fmt.Sprintf("request not authorized to run trigger: %s", err), 401)
+			} else {
+				// improve http error code, which is more applicable for general queue failures?
+				http.Error(w, fmt.Sprintf("server failed to run trigger: %s", err), 412)
```
catsby, line 133: nit: we could use http status code constants like `http. StatusUnauthorized` and `http.StatusPreconditionFailed` here so people like myself don't have to go look up what `401` and `412` stand for because I can't remember :P

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,378 @@
...
+			}
+
+			return
+		}
+		if resp == nil {
+			http.Error(w, fmt.Sprintf("server returned no job ids from run trigger %q", runTriggerId), 500)
```
catsby, line 139: `RunTrigger()` (and I assume `NoAuthRunTrigger()`) [can return a gRPC error code of `NotFound`](https://github.com/hashicorp/waypoint/blob/133491b031efb585a34151ed26b29958d952b27f/internal/server/singleprocess/state/db.go#L107), which I don't think should be considered a `500` internal server error. We should either check for that not found error and pass it on, or otherwise just return a `404` here either way

catsby (commented):
Looks good so far, just minor nits I've found.

internal/cli/trigger_apply.go
briancain: It is! I tried aligning the two on the “:” here, similar to how waypoint up aligns the ui and server host name. It looks less weird on the command line, but I’m ok to delete the space if it’s not worth it

internal/server/httpapi/trigger.go
briancain: Oh yeah, very good call here!

internal/cli/trigger_apply.go
catsby: 👍 if it's intentional then let's just leave it

internal/server/httpapi/trigger.go
briancain: Oh, I just checked out the code here...this is happening right above here!

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+							http.StatusInternalServerError)
+						return
+					}
+
+					var (
+						jobComplete bool
+						exitCode    string
+					)
+
+					log.Trace("reading job stream for job", "job_id", jId)
```
catsby, line 246: Might be helpful to move this line to above the `Recv()` call, so if that fails we have record of what job id we tried to use, in the event it's not contained in the error message(s)

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+
+					// read and send the stream
+					for {
+						resp, err := stream.Recv()
+						if err != nil {
```
catsby, line 251: We call `stream.Recv()` on line 225 above but we don't seem to do anything with the values other than error check. Is there a reason those checks can't be done in this `for` loop? If we need to get the `GetJobStreamResponse_Open_` response first, we could check for that first and then `continue` to restart the `for` loop. Would that work?

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+							// This shouldn't happen, but if it does, just ignore it.
+							log.Warn("nil response received, ignoring")
+							continue
+						}
+
+						select {
+						case <-cn.CloseNotify():
+							log.Trace("client closed connection to stream")
+							return
+						default:
```
catsby, line 267: nit: since the bulk of the logic is in this `default:` block, we could move this `select` to the top like so:

```go
for {
  select {
    case <-cn.CloseNotify():
      log.Trace("client closed connection to stream")
      return
    default:
      // Get jobstream output and return string chunk message back
      time.Sleep(time.Second)
  }
  [ .. rest of the code .. ]
}
```

This format reduces the stack by 1 and I feel is just easier to reason about. It also moves the listening for the `CloseNotify()` to the first thing we do in the loop, instead of `Recv()`

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+									log.Info("job completed successfully")
+									exitCode = "0"
+								} else {
+									exitCode = "1"
+									st := status.FromProto(event.Complete.Error)
+									log.Warn("job failed", "code", st.Code(), "message", st.Message())
+									http.Error(w,
+										fmt.Sprintf("job failed to complete: job code %s: %s", st.Code(), st.Message()),
+										http.StatusInternalServerError)
+								}
```
catsby, line 290: Question, do we need to return here? Either with completed or the error block?

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+
+								// Lock to ensure multiple routines don't send back a message at the same
+								// time to the receiver and mess up the incoming message
+								mu.Lock()
+								// send the message back
+								err := enc.Encode(msg)
+								if err != nil {
+									log.Error("failed to encode job stream output to send back", "err", err)
+									http.Error(w, fmt.Sprintf("server failed to encode job stream output: %s", err), 500)
+									return
```
catsby, line 379: Do we need to call `mu.Unlock()` here before we return?

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+
+			enc := json.NewEncoder(w)
+
+			// NOTE(briancain): We skip every two jobs here because when we call RunTrigger
+			// via gRPC, it eventually queues the trigger jobs through on-demand runners, and
+			// queueJobMulti returns three jobs: StartTask, the job to be queued, and StopTask. People
+			// really only expect output from the job to be queued, so we only stream that back.
+
+			// For example, a trigger that queues 2 Waypoint operation jobs returns 6 total job ids:
+			// Job List: [ {0: StartTask, 1: WP Operation 1, 2: StopTask}, {3: StartTask, 4: WP Operation 2, 5: StopTask}, ... ]
```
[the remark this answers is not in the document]
briancain: We're selecting Waypoint operations from this list. In reality, the list is flat and not grouped like this, but I choose to show them grouped this way because the surrounding job ids (i.e. Start and Stop task) are directly related to the job they wrap.

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,378 @@
...
+		jobIds := resp.JobIds
+
+		streamOutput := r.URL.Query().Get("stream")
+
+		// Attempt to stream output back, on request.
+		if streamOutput != "" {
+			if !requireAuth {
+				// We do not allow streaming job stream info if a no-auth token trigger was requested
+				log.Trace("server does not allow for streaming job stream output for no-token trigger URLs")
+				w.WriteHeader(http.StatusNoContent)
```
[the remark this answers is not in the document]
briancain: This specific status code was a request/feedback from @mitchellh. I think it matches how Netlify handles no-auth triggers. An unauthorized status could make sense here too 🤔

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,378 @@
...
+		if resp == nil {
+			http.Error(w, fmt.Sprintf("server returned no job ids from run trigger %q", runTriggerId), 500)
+			return
+		}
+		jobIds := resp.JobIds
+
+		streamOutput := r.URL.Query().Get("stream")
+
+		// Attempt to stream output back, on request.
+		if streamOutput != "" {
```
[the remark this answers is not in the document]
briancain: Makes sense to me!

internal/server/httpapi/trigger.go
briancain: Yep, this is because the way our job stream events work, we need to make sure the first event is always the open message. If not, we should exit right away. This is similar to how we stream job output on the CLI too. I like it separate from the `for` only because it's a required pre-condition that we only care about before reading the rest of the stream and beyond that we don't need to check for it again.

internal/server/httpapi/trigger.go
briancain: One sec, I have a commit that will fix this up 😄

briancain: I've updated this func locally to instead return the error directly in Message rather than an http error.

internal/server/httpapi/trigger.go
briancain: Hm, yes I think so!

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,398 @@
...
+				log.Error("failed to stream job output, could not create http.CloseNotifier")
+				http.NotFound(w, r)
+				return
+			}
+			flusher, ok := w.(http.Flusher)
+			if !ok {
+				log.Error("failed to stream job output, could not create http.Flusher")
+				http.NotFound(w, r)
+				return
+			}
```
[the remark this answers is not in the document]
briancain: Yep, if the requester isn't capable I believe. And yeah, I don't know if `NotFound` is the right response here, probably more of an InternalServerError?

internal/server/httpapi/trigger.go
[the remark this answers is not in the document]
briancain: I think based on your other comment feedback, we can return `http.StatusNoContent` if no streaming was requested, and otherwise if a stream was requested for a non-authenticated token, we return `http.StatusUnauthorized`. 🤔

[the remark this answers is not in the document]
catsby: I suppose I didn't or don't quite understand the flow here. If the scenario is that you can trigger this without a token but require a token to receive the stream, then `StatusNoContent` makes sense if there is no token present. I think the name and purpose of `requireAuth` has thrown me a bit here.

internal/server/httpapi/trigger.go
catsby: 👍

internal/server/httpapi/trigger.go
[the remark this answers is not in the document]
briancain: Updated to be an InternalServerError instead, this makes more sense to me 👍🏻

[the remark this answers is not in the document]
catsby: I would think a `400`  `StatusBadRequest`? It's not the servers fault I think? That's based on my assumption that the client made a request that it failed to indicate it can actually handle the response to. 

Not a big deal to worry about though, we can leave as-is if you would like

internal/server/httpapi/trigger.go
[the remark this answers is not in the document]
briancain: Perhaps removing the brackets in the comment will be less confusing to the reader? I can see now how it might make them think the job ids come in some kind of grouped set rather than a flat slice.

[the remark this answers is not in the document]
catsby: Yeah I would say remove the brackets IMO. I read that as a slice of structs, oddly enough.

internal/server/httpapi/trigger.go
catsby: Makes sense, thanks 👍

internal/server/httpapi/trigger.go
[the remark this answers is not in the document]
briancain: Good idea, got that updated now!

catsby (approved):
👍

internal/server/httpapi/trigger.go
```diff
@@ -0,0 +1,413 @@
...
+		runTriggerId := requestVars["id"]
+
+		variablesJSONRaw := r.URL.Query().Get("variables")
+		var (
+			variables         map[string]string
+			variableOverrides []*pb.Variable
+		)
+
+		if variablesJSONRaw != "" {
+			if err := json.Unmarshal([]byte(variablesJSONRaw), &variables); err != nil {
```
mitchellh, line 92: I would say this feels a little weird to just pass URL-encoded raw JSON as a GET parameter. I'm trying to think if I've seen this done before... in some ways a `variables[key]=value` approach feels like it'd make more sense and something I've seen in other frameworks (Rails for example). Or, base64-encoding the JSON, but that makes it slightly more user hostile.

This could work fine... I'm just thinking...

pkg/server/proto/server.proto
```diff
@@ -32,6 +32,11 @@ service Waypoint {
   // OIDC provider.
   rpc CompleteOIDCAuth(CompleteOIDCAuthRequest) returns (CompleteOIDCAuthResponse);
 
+  // Attempts to run a trigger given a trigger ID reference. If the trigger does
+  // not exist, we return not found. If the trigger exists but requires authentication
+  // we return an error.
+  rpc NoAuthRunTrigger(RunTriggerRequest) returns (RunTriggerResponse);
```
mitchellh, line 38: Is there a reason you did this versus overloading this on `RunTrigger`?

mitchellh (commented):
Looks amazing! A couple questions.

internal/server/httpapi/trigger.go
briancain: Yep, it's a bit strange. I know people are going to want to override their input vars in some way through this flow, I wasn't sure of a better solution over http 😅 I'm happy to try something better for sure

pkg/server/proto/server.proto
mitchellh: I think so... we have other APIs now that behave similarly I believe (esp with the Runner stuff).

internal/server/httpapi/trigger.go
mitchellh: No problem, let's go with this!

pkg/server/proto/server.proto
briancain: Sounds good, I'll combine them!

briancain: We chatted about this again in slack, and turns out we can do this a better way. I've opened an issue to track the work for now: https://github.com/hashicorp/waypoint/issues/2990. And for this PR I added some code comments on the two funcs with a link to the issue!

mitchellh (approved):

PR #2971, opened 2022-02-02
Handle auth error in workspaces route

gregone:
Because of the `redirect` hook added in https://github.com/hashicorp/waypoint/pull/2835, a user with an invalid auth token would hit the workspaces route, trigger an error from the api, and be stuck in a weird state (not logged out, not logged in), with a cryptic button in the UI.
Screen Shot 2022-01-31 at 19 14 44

### DIFF
--- a/ui/app/routes/workspaces.ts
+++ b/ui/app/routes/workspaces.ts
@@ -1,14 +1,21 @@
-import Route from '@ember/routing/route';
-import { inject as service } from '@ember/service';
 import ApiService from 'waypoint/services/api';
+import Route from '@ember/routing/route';
 import SessionService from 'ember-simple-auth/services/session';
+import { Workspace } from 'waypoint-pb';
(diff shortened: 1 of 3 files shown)


jgwhite (approved):
Good catch 👍

ui/app/routes/workspaces.ts
```diff
@@ -1,14 +1,21 @@
...
   @service session!: SessionService;
 
   async redirect(): Promise<void> {
-    let workspaces = await this.api.listWorkspaces();
+    let workspaces: Workspace.AsObject[] = [];
+    try {
+      workspaces = await this.api.listWorkspaces();
+    } catch (error) {
+      // Send Authentication or other error if it exists
+      this.send('error', error);
```
jgwhite, line 17: This makes me realize my mental model of how `redirect` behaves is incomplete with regard to errors.

PR #2988, opened 2022-02-04
Generate swagger spec from the server protobuf schema

izaaklauer:
This updates the `make gen/server` command to generate a swagger spec off the waypoint server protobuf schema. It also adds annotations to the ListProjects rpc to set the url path. Eventually we'll need to add these annotations to all the RPCs we intend to support with via swagger/grpc-gateway.


pkg/server/proto/server.proto
```diff
@@ -86,7 +87,9 @@ service Waypoint {
   // ListProjects returns a list of all the projects. There is no equivalent
   // ListApplications because applications are a part of projects and you
   // can use GetProject to get more information about the project.
-  rpc ListProjects(google.protobuf.Empty) returns (ListProjectsResponse);
+  rpc ListProjects(google.protobuf.Empty) returns (ListProjectsResponse) {
+    option (google.api.http) = { get: "/api/v1/projects" };
```
gregone, line 91: I think we may want to change this to `/waypoint/....` to match patterns in other cloud services

gregone: Additionally, if we want to use those URLs with a prefix in one (or both place), it may be better to handle this at the router/mux level?

gregone (changes requested):

pkg/server/proto/server.proto
evanphx: yeah, the prefix (be it product name and/or version) will be there, it's a question of if the annotations include it or if the router handles it.

pkg/server/server.go
```diff
@@ -1,5 +1,6 @@
 package server
 
-//go:generate sh -c "protoc -I../../thirdparty/proto/api-common-protos -I ../.. ../../pkg/server/proto/server.proto --go_out=plugins=grpc:../.. --go-json_out=../.."
+//go:generate sh -c "protoc -I../../thirdparty/proto/api-common-protos -I ../.. ../../pkg/server/proto/server.proto --go_out=plugins=grpc:../.. --go-json_out=../.. --swagger_out=../.."
+//go:generate mv ./proto/server.swagger.json ./gen/
```
briancain, line 4: Are there any extra tools that should go in `make tools` for generating swagger?

izaaklauer: Yes! Done

gregone (approved):

PR #2989, opened 2022-02-04
internal/server: RunnerConfig: create runner after token validation

mitchellh:
Functionally, it doesn't make a difference, but practically there's no
reason to create the runner before validating the token. If the token is
invalid, it just creates more work that was unnecessary. If the token is
valid, we would've done it anyways.


mitchellh:
If tests pass I'm going to just merge this since it shouldn't affect anything that isn't already tested.

evanphx (approved):

PR #2994, opened 2022-02-07
Fix minor typo

martinb3:
Sorry, this typo of allotted was bothering me in some output. Let me know if I should submit differently. Thanks.

### DIFF
--- a/builtin/aws/lambda/platform.go
+++ b/builtin/aws/lambda/platform.go
@@ -367,7 +367,7 @@ func (p *Platform) Deploy(
 	}
 
 	if ver == nil {
-		return nil, fmt.Errorf("Lambda was unable to prepare the function in the aloted time")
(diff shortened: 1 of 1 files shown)


builtin/aws/lambda/platform.go
```diff
@@ -367,7 +367,7 @@ func (p *Platform) Deploy(
 	}
 
 	if ver == nil {
-		return nil, fmt.Errorf("Lambda was unable to prepare the function in the aloted time")
+		return nil, fmt.Errorf("Lambda was unable to prepare the function in the alotted time")
```
briancain, line 370: Good catch! I think it's `allotted`?

```suggestion
		return nil, fmt.Errorf("Lambda was unable to prepare the function in the allotted time")
```

martinb3: Indeed it is!

briancain (approved):

PR #2995, opened 2022-02-07
Adding -from-kubernetes-config and -from-kubernetes-kubeconfig to login

izaaklauer:
It's not uncommon for me to have many contexts live, and always set it when invoking kubectl
rather than rely on the default. It's also not uncommon to have multiple kubeconfig files, and
specify that path on the command line rather than move it to the default location.

Example of using these flags:
```
$ waypoint login -from-kubernetes -from-kubernetes-namespace=waypoint -from-kubernetes-context=arn:aws:eks:us-east-2:999999999999:cluster/prod -from-kubernetes-kubeconfig=~/kube/prod-config
```

### DIFF
--- a/website/content/commands/login.mdx
+++ b/website/content/commands/login.mdx
@@ -56,5 +56,7 @@ or file.
 - `-from-kubernetes-service=<string>` - The name of the Kubernetes service to get the server address from when using the -from-kubernetes flag.
 - `-from-kubernetes-secret=<string>` - The name of the Kubernetes secret that has the Waypoint token when using the -from-kubernetes flag.
 - `-from-kubernetes-namespace=<string>` - The name of the Kubernetes namespace that has the Waypoint token when using the -from-kubernetes flag.
+- `-from-kubernetes-kubeconfig=<string>` - Path to the kubeconfig file for the cluster Waypoint was installed into when using the -from-kubernetes flag.
+- `-from-kubernetes-context=<string>` - The name of the Kubernetes context that Waypoint was installed into when using the -from-kubernetes flag.
 
 @include "commands/login_more.mdx"
--- a/.changelog/2995.txt
+++ b/.changelog/2995.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+cli: Add -from-kubernetes-context and -from-kubernetes-kubeconfig flags to login command
+```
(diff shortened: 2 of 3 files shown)


mitchellh:
The reason I didn't originally expose these is because the K8S client library respects the standard Kubernetes environment variables. `KUBECONFIG` and `KUBECONTEXT`. I don't mind exposing these as flags, just sharing my original reasoning.

izaaklauer:
oh! I honestly didn't know about those env vars.

website/content/commands/login.mdx
```diff
@@ -56,5 +56,7 @@ or file.
 - `-from-kubernetes-service=<string>` - The name of the Kubernetes service to get the server address from when using the -from-kubernetes flag.
 - `-from-kubernetes-secret=<string>` - The name of the Kubernetes secret that has the Waypoint token when using the -from-kubernetes flag.
 - `-from-kubernetes-namespace=<string>` - The name of the Kubernetes namespace that has the Waypoint token when using the -from-kubernetes flag.
+- `-from-kubernetes-kubeconfig=<string>` - Path to the kubeconfig file for the cluster Waypoint was installed into when using the -from-kubernetes flag.
+- `-from-kubernetes-context=<string>` - The name of the Kubernetes context that Waypoint was installed into when using the -from-kubernetes flag.
```
catsby, line 60: Considering Mitchell's comment, it may be worth mentioning here that if omitted we'll try to use the default env var values

catsby (approved):

website/content/commands/login.mdx
mitchellh: I would actually just specify the `EnvVar` setting and the help will auto-include this.

briancain:
Should we close this and point people at setting the right k8s env vars instead?

website/content/commands/login.mdx
izaaklauer: Hmm, looks like we only have an EnvVar setting for plugin docs, and even that seems to not work: https://github.com/hashicorp/waypoint/issues/3135

xiaolin-ninja:
Update from discussion: we should update the doc gen if we want this to work.

briancain:
I vote we close this one. It's been a full year open and we haven't needed it.

briancain:
Since there were no 👎🏻 I'm gonna go ahead and close this one so it doesn't get stale!

PR #3004, opened 2022-02-08
Runner inspect

xiaolin-ninja:
This PR adds `waypoint runner inspect <name>` UI.
Screen Shot 2022-02-07 at 10 55 02 PM

Switches the "ID" and "state" columns in `runner list`.

& some sneaky cosmetic code adjustments.

### DIFF
--- a/internal/cli/main.go
+++ b/internal/cli/main.go
@@ -180,30 +180,27 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"init": func() (cli.Command, error) {
 			return &InitCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"up": func() (cli.Command, error) {
 			return &UpCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"destroy": func() (cli.Command, error) {
 			return &DestroyCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"exec": func() (cli.Command, error) {
 			return &ExecCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
+
 		"config": func() (cli.Command, error) {
 			return &helpCommand{
 				SynopsisText: helpText["config"][0],
@@ -235,6 +232,7 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
+
 		"logs": func() (cli.Command, error) {
 			return &LogsCommand{
 				baseCommand: baseCommand,
@@ -252,19 +250,16 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"artifact list": func() (cli.Command, error) {
 			return &ArtifactListCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"artifact list-builds": func() (cli.Command, error) {
 			return &BuildListCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"artifact push": func() (cli.Command, error) {
 			return &ArtifactPushCommand{
 				baseCommand: baseCommand,
@@ -277,19 +272,16 @@ func Commands(
 				HelpText:     helpText["deployment"][1],
 			}, nil
 		},
-
 		"deployment deploy": func() (cli.Command, error) {
 			return &DeploymentCreateCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"deployment destroy": func() (cli.Command, error) {
 			return &DeploymentDestroyCommand{
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"deployment list": func() (cli.Command, error) {
 			return &DeploymentListCommand{
 				baseCommand: baseCommand,
@@ -301,7 +293,6 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
-
 		"release list": func() (cli.Command, error) {
 			return &ReleaseListCommand{
 				baseCommand: baseCommand,
@@ -359,6 +350,7 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
+
 		"status": func() (cli.Command, error) {
 			return &StatusCommand{
 				baseCommand: baseCommand,
@@ -382,7 +374,6 @@ func Commands(
 				HelpText:     helpText["hostname"][1],
 			}, nil
 		},
-
 		"hostname register": func() (cli.Command, error) {
 			return &HostnameRegisterCommand{
 				baseCommand: baseCommand,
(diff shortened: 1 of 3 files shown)


internal/cli/main.go
```diff
@@ -180,30 +180,27 @@ func Commands(
 				baseCommand: baseCommand,
 			}, nil
 		},
-
```
xiaolin-ninja, line 183: I noticed an inconsistency in the spacing between these commands, but also an apparent pattern of a space before a chunk of commands relating to the same operation that got broken at some point. So I made the code prettier to follow that pattern.

internal/cli/runner_inspect.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	if len(c.args) == 0 {
+		c.ui.Output("Runner name required.", terminal.WithErrorStyle())
+		return 1
+	}
+	id := c.args[0]
+
+	resp, err := c.project.Client().GetRunner(c.Ctx, &pb.GetRunnerRequest{RunnerId: id})
+	if err != nil && status.Code(err) != codes.NotFound {
```
briancain, line 45: Do we need another error case for when a requested runner profile is not found? I think it'll ignore that case here.

internal/cli/runner_inspect.go
```diff
@@ -0,0 +1,136 @@
...
+	var kindStr = "unknown"
+	if v, ok := kindMap[reflect.TypeOf(resp.Kind)]; ok {
+		kindStr = v
+	}
+	var lastSeenStr string
+	if v, err := ptypes.Timestamp(resp.LastSeen); err == nil {
+		lastSeenStr = humanize.Time(v)
+	}
+	var stateStr = "unknown"
+	stateStr = strings.ToLower(resp.AdoptionState.String())
```
briancain, line 77: Hm, I think we can simplify this with a `switch`?

```
switch kind := resp.Kind.(type) {
case *pb.Runner_Odr:
case ....
...
}
```

internal/cli/runner_inspect.go
```diff
@@ -0,0 +1,136 @@
...
+	return c.Flags().Completions()
+}
+
+func (c *RunnerInspectCommand) Synopsis() string {
+	return "Show detailed information about a runner."
+}
+
+func (c *RunnerInspectCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint runner inspect ID
```
briancain, line 131:
```suggestion
Usage: waypoint runner inspect [options] ID
```

Since we got the json flag

briancain (changes requested):
Looks pretty good to me! I think you need to generate the website for the new CLI docs too? The test didn't fail because it hasn't existed yet, I think.

internal/cli/runner_inspect.go
catsby: It panics:

```
[go][main](4)$ waypoint runner inspect 01FSW91WAMHBQZ542NKKR2ST1Q

» Runner:
              ID: 01FSW91WAMHBQZ542NKKR2ST1Q
  Adoption State: new
            Kind: remote
          Labels: map[]
          Online: false
          
[go][main](4)$ waypoint runner inspect puppy

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x34dac60]

goroutine 1 [running]:
github.com/hashicorp/waypoint/internal/cli.(*RunnerInspectCommand).Run(0xc000efb2e0, {0xc000eaa570, 0xc5, 0x3e0eeb0})
        /Users/clint/go-src/github.com/hashicorp/waypoint/internal/cli/runner_inspect.go:69 +0x580
github.com/mitchellh/cli.(*CLI).Run(0xc000dd0000)
        /Users/clint/Go/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x5f8
github.com/hashicorp/waypoint/internal/cli.Main({0xc00006e080, 0x3717960, 0xc0000001a0})
        /Users/clint/go-src/github.com/hashicorp/waypoint/internal/cli/main.go:127 +0x545
main.main()
        /Users/clint/go-src/github.com/hashicorp/waypoint/cmd/waypoint/main.go:14 +0x7f
```

internal/cli/runner_inspect.go
```diff
@@ -0,0 +1,136 @@
...
+	resp, err := c.project.Client().GetRunner(c.Ctx, &pb.GetRunnerRequest{RunnerId: id})
+	if err != nil && status.Code(err) != codes.NotFound {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	if c.flagJson {
+		var m jsonpb.Marshaler
+		m.Indent = "\t"
+		str, err := m.MarshalToString(resp)
```
catsby, line 53: I think we should build up some other things in a map of sorts first and then render them to json, similar to what we do with [deployments](https://github.com/hashicorp/waypoint/blob/b49074ec5c8e86a1ce37dc9e7358b22f71b6f050/internal/cli/deployment_list.go#L150-L152). As is, this leaves some of the information behind that you normally get without using `-json`. 

Examples:

```
[waypoint][runner-inspect](4)$ waypoint runner inspect 01FSW91WAMHBQZ542NKKR2ST1Q

» Runner:
              ID: 01FSW91WAMHBQZ542NKKR2ST1Q
  Adoption State: new
            Kind: remote
          Labels: map[]
          Online: false

[waypoint][runner-inspect](4)$ waypoint runner inspect -json 01FSW91WAMHBQZ542NKKR2ST1Q
{
        "id": "01FSW91WAMHBQZ542NKKR2ST1Q",
        "remote": {

        }
}
```

internal/cli/runner_inspect.go
catsby: seems this was based off of https://github.com/hashicorp/waypoint/blob/357e3d3f9f4189e5abfce09253f3e14f8f362ed4/internal/cli/runner_list.go#L62-L102 from `runner_list`, so it's at least consistent with that command

briancain: Ah yeah, I remember seeing that go by now. I think maybe switch is easier to read if it can be done? Not a huge deal

internal/cli/runner_inspect.go
briancain: Heh, interestingly enough, I believe most of the `inspect` commands format the raw proto response as json directly, so this would match the other inspect commands.

xiaolin-ninja: yeah, I vote leave it and make it a to-do to redo item the other inspects with good parsing

internal/cli/runner_inspect.go
xiaolin-ninja: ^ Noticed inconsistency in the codebase with adding the [options]. Made an issue here: https://github.com/hashicorp/waypoint/issues/3009

.changelog/3004.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+cli: Add runner list command
```
briancain, line 2:
```suggestion
cli: Add `waypoint runner inspect` command
```

briancain (approved):
🎉 Looks great! I think we deferred the work around `json` and `[options]` for another PR today in our GH sync.

PR #3005, opened 2022-02-08
Add tooling to generate protos with docker

izaaklauer:
This introduces a new method to generate protos. You can run `make docker/tools` go create a new local docker container that contains all the tools in `tools/tools.go` and the correct version of protoc, then run `make docker/gen/server` to regenerate protobufs using that container.

This is an alternative to:

- Building protos locally by downloading the correct version of `protoc` locally and installing it (and its includes), and running `make tools; make gen/server`. 

- Building protos in nix with `nix-shell; make gen/server`

The downside here is that we're pinning the protoc version in the dockerfile, so if we upgrade protoc we'll need to update both here and in nix. I don't know of a great way to extract the `v3.15.8` from the nix definitions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,3 +151,13 @@ tools: # install dependencies and tools required to build
 	$(GO_CMD) generate -tags tools tools/tools.go
 	@echo
 	@echo "Done!"
+
+.PHONY: docker/tools
+docker/tools:
+	@ echo "Building docker tools image"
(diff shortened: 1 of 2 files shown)


mitchellh:
On the flip side, if you turn the protoc version into a VAR in the Dockerfile, its pretty easy for me to pull that out using Nix.

gregone (approved):

briancain (approved):
🐋 🚀

PR #3007, opened 2022-02-08
Make runner profile optional in waypoint HCL

xiaolin-ninja:
^ yup!

Original idea was to have `profile = ""` even if the user doesn't want to define any profile and have the server choose a default, but after considerations this is redundant and will bug out if the user omits this field.


briancain (approved):
👍🏻

catsby (approved):

PR #3019, opened 2022-02-10
ui: ember-a11y-refocus@^2.3.0

### LINKED ISSUE #2882: ui: polling causes ember-a11y-refocus to refire
**Describe the bug**
With the new package `ember-a11y-refocus`, screen readers now announce when the user has navigated to a new page and refocuses them to the top of the page, since this is a SPA. Unfortunately, with how our data is collected/stored with polling and not ember data, this announcement of the user navigating to a new page happens every time the frontend successfully polls the backend.

**Steps to Reproduce**
1. Start up the UI `yarn start`
2. Start up the screen reader
3. Check that `ember-a11y-refocus` runs when navigating to a new page
4. Wait on an application overview page for a little bit until a polling event occurs

**Expected behavior**
Should not fire whenever polling occurs

jgwhite:
## Why the change?

Closes #2882

Thanks to ember-a11y-refocus#334

## How do I test it?

1. `git checkout ui/ember-a11y-refocus@2.3.0`
2. `cd ui && yarn start`
3. [Visit a deployment](http://localhost:4200/default/marketing-public/app/wp-matrix/deployments/4)
4. Move focus to some element other than the skip link
5. Wait 15 seconds for the automatic refresh
6. Verify focus does not jump back to the skip link
7. Try navigating to a different page
8. Verify focus has jumped back to the skip link

### DIFF
--- a/.changelog/3019.txt
+++ b/.changelog/3019.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: fixed issue with focus jumping back to the skip link on automatic refresh
+```
--- a/ui/package.json
+++ b/ui/package.json
@@ -53,7 +53,7 @@
     "codemirror": "^5.62.2",
     "date-fns": "^2.15.0",
     "docker-parse-image": "^3.0.1",
-    "ember-a11y-refocus": "^2.1.0",
+    "ember-a11y-refocus": "^2.3.0",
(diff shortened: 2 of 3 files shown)


gregone (approved):
Perfect!

Issue #3026, opened 2022-02-14
`arm64` function support for `aws-lambda` deployments


thiskevinwang:
**Is your feature request related to a problem? Please describe.**

(This feels like half feature request and half problem.)

I'm on an M1 Max macbook pro and running `waypoint up`

With 0 docker config, Waypoint appears to build an `arm64` image

```console
Image built: waypoint.local/waypoint-apollo-lambda:latest (arm64)
```

But Waypoint also appears to deploy a lambda function with a fixed `x86_64` arch, regardless of the Docker image arch.

This is my `waypoint.hcl`

```hcl
project = "waypoint-apollo-lambda"

app "waypoint-apollo-lambda" {
  build {
    use "docker" {
      // buildkit = true
      // platform = "amd64"
    }

    registry {
      use "aws-ecr" {
        region     = "us-east-1"
        repository = "waypoint-apollo-lambda"
        tag        = gitrefpretty()
      }
    }
  }

  deploy {
    use "aws-lambda" {
      region = "us-east-1"
      memory = 256
    }
  }

  release {
    use "aws-alb" {
    }
  }
}
```

After deployment (and working around https://github.com/hashicorp/waypoint/issues/2066), visiting the loadbalancer URL returns a **502 Bad Gateway**.

**Describe the solution you'd like**

An option in the `aws-lambda` config, like `platform`

```hcl
  deploy {
    use "aws-lambda" {
      region = "us-east-1"
      memory = 256
      platform = "arm64"
    }
  }
```

... and potentially a warning prompt if there is any arch mismatch between `docker {}` and `aws-lambda {}`

**Describe alternatives you've considered**

Alternative no. 1 — use `amd64` Docker image

(This works in the `build` stanza, but is a compromise because the end result is an `amd64` image on a `x86_64` function)

```hcl
  build {
    use "docker" {
      buildkit = true
      platform = "amd64"
    }
```

Alternative no. 2 — manually update lambda arch and additional follow up steps. 

(This is closer to ideal — `arm64` image on `arm64` function — but there is manual clicking involved.)

1. [Lambda] Update arch to `arm64` and find the proper `arm64` ECR image
   ![image](https://user-images.githubusercontent.com/26389321/153871684-4b97cc0c-3d57-4f18-b16b-eaa8147d6871.png)
2. [Lambda] Publish new function version
3. [EC2] Find my loadbalancer; Find the listener's Target Group
4. [EC2] Deregister current Target 
   This causes a temporary `503`
[...]

PR #3032, opened 2022-02-17
feat(aws-lambda): support `architecture`

thiskevinwang:
# Description

This adds an optional `architecture` field (`x86_64 | arm64`) to aws-lambda to support setting architecture on lambda creation and update.

--- 

Short term option for https://github.com/hashicorp/waypoint/issues/3026

### DIFF
--- a/builtin/aws/lambda/platform.go
+++ b/builtin/aws/lambda/platform.go
@@ -337,7 +337,7 @@ func (p *Platform) Deploy(
 					case "ValidationException":
 						// likely here if Architectures was an invalid
 						if architecture != lambda.ArchitectureX8664 && architecture != lambda.ArchitectureArm64 {
-							return nil, fmt.Errorf("architecture must be either x86_64 or arm64")
+							return nil, fmt.Errorf("architecture must be either \"x86_64\" or \"arm64\"")
 						}
 						return nil, err
 					}


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/DataDog/opencensus-go-exporter-datadog v0.0.0-20210527074920-9baf37265e83
 	github.com/adrg/xdg v0.2.1
 	github.com/armon/circbuf v0.0.0-20190214190532-5111143e8da2
-	github.com/aws/aws-sdk-go v1.36.31
+	github.com/aws/aws-sdk-go v1.43.0
```
thiskevinwang, line 14: Lambda `Architectures` seemed to have been introduced in `1.41.0`

builtin/aws/lambda/platform.go
```diff
@@ -271,11 +279,22 @@ func (p *Platform) Deploy(
...
+			Architectures: aws.StringSlice([]string{architecture}),
 		})
 
 		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case "ValidationException":
+					// likely here if Architectures was invalid
+					if architecture != lambda.ArchitectureX8664 && architecture != lambda.ArchitectureArm64 {
+						return nil, fmt.Errorf("architecture must be either x86_64 or arm64")
```
briancain, line 293:
```suggestion
						return nil, fmt.Errorf("architecture must be either \"x86_64\" or \"arm64\"")
```

Small nit, but suggested values here are easier to read I think if we quote them.

builtin/aws/lambda/platform.go
```diff
@@ -314,6 +334,12 @@ func (p *Platform) Deploy(
 					switch aerr.Code() {
 					case "ResourceConflictException":
 						return nil, err
+					case "ValidationException":
+						// likely here if Architectures was an invalid
+						if architecture != lambda.ArchitectureX8664 && architecture != lambda.ArchitectureArm64 {
+							return nil, fmt.Errorf("architecture must be either x86_64 or arm64")
```
briancain, line 340:
```suggestion
							return nil, fmt.Errorf("architecture must be either \"x86_64\" or \"arm64\"")
```

briancain (approved):
I haven't tried this, but the code makes sense to me providing an option to set the architecture for a lambda 👍🏻

PR #3037, opened 2022-02-22
feature: add label-based runner targeting for operations

xiaolin-ninja:
Now users can also target runners by runner labels. In the Waypoint configs, they can specify:

```
runner {
  target_labels = {
    "env" = {
        "default" = "test"
        "prod" = "prod"
    }[workspace.name]
  }
}
```

or

```
app {
  runner {
    target_labels = {
      "env" = {
        "default" = "test"
        "prod" = "prod"
      }[workspace.name]
    }
  }
}
```

It also adds a "labels" column in `runner list`.
```
❯ waypoint-dev runner list
      ID      |   STATE    |  KIND  |  LABELS   | LAST REGISTERED  
--------------+------------+--------+-----------+------------------
  test        | preadopted | remote | env:test  | 1 second ago     
```

Some notes on implementation choices:

Profile takes precedence over labels, so if someone has both a runner profile and label defined, we ignore the labels. I also chose to ignore the project-level runner stanza completely if app-level runner configs are defined, instead of parsing and piecing together labels and profiles. We could do that... but the current product doesn't support targeting distinct runners for individual apps anyway, so we could cross that bridge when we get there? It also felt a little too "nice". heh.

I also opted to put the runner look-up and targeting logic on the client-side instead of server-side (differing from the [RFC](https://docs.google.com/document/d/1J9EURm5wMczeXlqmdvnehqJQl1z4GEg12yCoRcSULiY)); i.e. looking up a list of all runners and finding the one with the right labels, as opposed to adding a "by label" lookup in the db. Because.... the infrastructure is already there and it's easy. _But_ listing all the runners might bite us in the butt a few years down the line with database load... but also how many runners can a project have anyway?! I was also debating whether labels are something we want to store in the db, since they feel ephemeral (as in the English word)... but we are storing adoption state, which might have the same degree of permanence. So anyway I'm happy to take the server-side approach if we feel it's better.


internal/client/job.go
```diff
@@ -235,18 +237,45 @@ func (c *Project) doJobMonitored(ctx context.Context, job *pb.Job, ui terminal.U
...
+					Name: configRunner.Profile,
+				}
+			}
+			if configRunner.TargetLabels != nil {
+				if job.OndemandRunner != nil {
+					log.Warn("both runner profile and target labels are set, profile takes precedence.")
+				} else {
+					runners, err := c.client.ListRunners(ctx, &pb.ListRunnersRequest{})
+					if err != nil {
+						errors.Wrapf(err, "no runners found.")
```
briancain, line 265: Since we're using the `status`/`codes` errors else where in this file, this would probably be better to return a `status.NotFound` error

internal/server/singleprocess/service_job.go
```diff
@@ -165,6 +165,8 @@ func (s *service) queueJobReqToJob(
 		}
 	}
 
+	// THIS IS THE PART WE NEED TO CHANGE.
```
briancain, line 168: Is this a left over comment now?

briancain (commented):
Looks pretty good overall! Only had a few comments before trying this out 😄

internal/client/job.go
briancain: We could also just return the error the `ListRunners` endpoint returns, although if we want to wrap it, we could mention the job client failed to list runners or something more specific to this file?

xiaolin-ninja: ^ realized that it would not error and simply return an empty list, so modified this code behavior.

internal/client/runner_test.go
```diff
@@ -117,7 +117,26 @@ func Test_remoteOpPreferred(t *testing.T) {
...
+		}
+		_, err := client.UpsertProject(ctx, &pb.UpsertProjectRequest{Project: project})
+		require.Nil(err)
+
+		runnerCfgs := []*configpkg.Runner{{TargetLabels: map[string]string{"env": "test"}}}
+
+		remote, err := remoteOpPreferred(ctx, client, project, runnerCfgs, log)
+		require.Nil(err)
+		require.True(remote)
+	})
```
briancain, line 137: I think we should have a good test around checking that `TargetLabels` is a proper subset of the runner labels

internal/client/job.go
```diff
@@ -235,21 +236,63 @@ func (c *Project) doJobMonitored(ctx context.Context, job *pb.Job, ui terminal.U
...
+					targeted := false
+					for _, r := range runners.Runners {
+						n := len(configRunner.TargetLabels)
+						// Check whether a runner matches every target label
+						// If multiple runners match labels, this will take the first that matches
+						for k, v := range configRunner.TargetLabels {
+							if val, ok := r.Labels[k]; ok && v == val {
+								n -= 1
+							}
+						}
```
briancain, line 278: @xx and I paired on this - We can simplify this check for looking at when `TargetLabels` has to be a subset of `Runners` with a bool

internal/client/runner_test.go
briancain: We can do the positive case with a few key,vals in each, and then a bad case for when it's not a subset. That should cover it I think!

mitchellh (changes requested):
Commentary was done in Zoom.

xiaolin-ninja:
After discussion, client-side targeting is not the right approach, especially because it wouldn't work with polling. Closing this PR and making 3 separate incremental PRs instead for a server-side implementation.

PR #3039, opened 2022-02-23
Fix panic in config get

xiaolin-ninja:
Screen Shot 2022-02-23 at 12 53 03 PM

When running `waypoint config get -app=<app>` outside the project directory, Waypoint panics instead of returning a proper error.

Now it returns:
```
❯ waypoint-dev config get -app=example-java
please also define project with -project
```


briancain (approved):
👍🏻 Backport too? edit: nevermind, already fixed!

I do think this could use a changelog though since it fixed a bug

PR #3044, opened 2022-02-25
Fix the -set-default flag for `context create`.

izaaklauer:
The flag was unused before this point. This also sets the default state to "true", which is what I think most users would expect.

## How to verify
```
$ waypoint context create is-default
$ waypoint context list
    |         NAME         |  PLATFORM  |                                    SERVER ADDRESS
----+----------------------+------------+----------------------------------------------------------------------------------------
  * | is-default           | n/a        |

$ waypoint context create -set-default=false not-default

Context "not-default" created.
$ waypoint context list
    |         NAME         |  PLATFORM  |                                    SERVER ADDRESS
----+----------------------+------------+----------------------------------------------------------------------------------------
  * | is-default           | n/a        |
    | not-default          | n/a        |
```

### DIFF
--- a/internal/cli/context_create.go
+++ b/internal/cli/context_create.go
@@ -61,7 +61,7 @@ func (c *ContextCreateCommand) Flags() *flag.Sets {
 			Name:    "set-default",
 			Target:  &c.flagSetDefault,
 			Default: true,
-			Usage:   "Set this context as the new default for the CLI.",
+			Usage:   "Set this context as the new default for the CLI. Defaults to true.",
 		})
 		f.StringVar(&flag.StringVar{
 			Name:   "server-addr",
--- a/internal/cli/context_create.go
+++ b/internal/cli/context_create.go
@@ -43,6 +43,13 @@ func (c *ContextCreateCommand) Run(args []string) int {
 		return 1
 	}
 
+	if c.flagSetDefault {
+		if err := c.contextStorage.SetDefault(name); err != nil {
+			c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		}
+	}
+
 	c.ui.Output("Context %q created.", name, terminal.WithSuccessStyle())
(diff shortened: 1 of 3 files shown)


catsby (approved):
We should update the documentation as well https://github.com/hashicorp/waypoint/blob/7d04fc110757f4808f5bb5394804eeba5639175f/website/content/commands/context-create.mdx

internal/cli/context_create.go
```diff
@@ -51,9 +58,10 @@ func (c *ContextCreateCommand) Flags() *flag.Sets {
 	return c.flagSet(0, func(set *flag.Sets) {
 		f := set.NewSet("Command Options")
 		f.BoolVar(&flag.BoolVar{
-			Name:   "set-default",
-			Target: &c.flagSetDefault,
-			Usage:  "Set this context as the new default for the CLI.",
+			Name:    "set-default",
+			Target:  &c.flagSetDefault,
+			Default: true,
+			Usage:   "Set this context as the new default for the CLI. Defaults to true.",
```
briancain, line 64: BTW we don't need the `Defaults to true` in the doc string since the CLI adds this automatically:

```
brian@localghost:waypoint λ waypoint context create -help                                                                                                                                                                                                                                                             
Command Options:
...
...

  -set-default
      Set this context as the new default for the CLI. Defaults to true. The
      default is true.
```

briancain: I can make a PR to fix this real quick

briancain: https://github.com/hashicorp/waypoint/pull/3048

izaaklauer: Oh ha, I was fixing the website docs. Looks like the CLI prints the default, but the website docs don't?

izaaklauer: Agree that the cli help text is more important than the website docs, so I support reverting this for now.

PR #3046, opened 2022-02-27
feat: pass docker arch to ecr to lambda

thiskevinwang:
# Description

This adds `docker` **architecture**, and passes it to the `ecr` component and then `lambda` component (via the protobufs) for a _0-config_ option to automatically set a lambda architecture that is compatible with the ECR image architecture.

The HCL value from #3032 takes precedence.

This closes https://github.com/hashicorp/waypoint/issues/3026


builtin/docker/plugin.proto
```diff
@@ -22,16 +23,16 @@ message Image {
...
     // docker is set if the image is in a local Docker daemon registry.
-    google.protobuf.Empty docker = 4;
+    google.protobuf.Empty docker = 5;
 
     // Waypoint 0.6 and earlier supported "img" as a Dockerless builder.
     // We removed this in 0.7, hence we prefix this now with "unused". We
     // keep the field around so we can detect old clients sending it and
     // show an error. Do not set this.
-    google.protobuf.Empty unused_img = 5;
+    google.protobuf.Empty unused_img = 6;
```
briancain, line 35: if I remember right, these numbers should _never_ be changed. Otherwise old proto clients will use the wrong variables during communication. It's totally fine to make `architecture` the _next_ number here, no need to bump the other ones.

builtin/aws/lambda/mapper.go
```diff
@@ -0,0 +1,18 @@
...
+// DockerArchitectureMapper maps a docker image architecture to a valid lambda architecture.
+func DockerArchitectureMapper(src string, log hclog.Logger) string {
+	switch src {
+	case "amd64", "x86_64":
+		return lambda.ArchitectureX8664
+	case "arm64", "aarch64":
+		return lambda.ArchitectureArm64
+	}
+	log.Warn("unsupported docker architecture", "arch", src, "defaulting to:", lambda.ArchitectureX8664)
+	return lambda.ArchitectureX8664
```
briancain, line 17: This could just be the `default` case in the switch

briancain (changes requested):
I know this is still in draft, but figured I'd give a lil' review!

builtin/docker/plugin.proto
thiskevinwang: Ah TIL! and that makes sense

builtin/aws/lambda/mapper.go
```diff
@@ -0,0 +1,19 @@
+package lambda
```
evanphx, line 1: Could you move this to `utils.go` instead? Mappers are a specific argmapper concept that are automatically called and have to be registered. This appears to just be a helper function to convert one string to another.

evanphx (changes requested):
If you change that file so it's not confused as an argmapper mapper, we'll be good to go!

builtin/aws/lambda/mapper.go
thiskevinwang: TIL! Ok, renamed this `mapper.go` file `utils.go`

PR #3048, opened 2022-02-28
fixup: Don't include default value

briancain:
The CLI package will append the default value of a flag automatically so
no need to include it ourselves.

### DIFF
--- a/internal/cli/context_create.go
+++ b/internal/cli/context_create.go
@@ -61,7 +61,7 @@ func (c *ContextCreateCommand) Flags() *flag.Sets {
 			Name:    "set-default",
 			Target:  &c.flagSetDefault,
 			Default: true,
-			Usage:   "Set this context as the new default for the CLI. Defaults to true.",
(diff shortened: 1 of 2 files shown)


briancain:
Not really sure what to do about this other than fix our CLI docs auto-gen.

Currently our CLI package auto-appends a default value string to all CLI help docs, but our CLI docs autogen tool does not do that, so the website does not include defaults.

With the default in the doc string though, it reads bad on the CLI:

```
  -set-default
      Set this context as the new default for the CLI. Defaults to true. The
      default is true.
```

briancain:
Sounds good @mitchellh - We can fix the CLI docs in this PR here, and then open an issue to track fixing up the CLI doc autogen tool to include defaults for the website docs.

izaaklauer:
Ha crazy coincidence, also just wrote this loosely-related issue after some feedback from @koesbong : https://github.com/hashicorp/waypoint/issues/3049

izaaklauer (approved):

PR #3052, opened 2022-03-01
Start runner liveness check _before_ starting the runner.

izaaklauer:
Without this, with the helm runner-only install using the adoption flow, the runner pauses
waiting to be adopted before starting it's health check. It sits in that state for 10 seconds,
then k8s gives up on it and terminates it becasue it's failed too many health checks.

Verified through local testing

### DIFF
--- a/.changelog/3052.txt
+++ b/.changelog/3052.txt
@@ -1,3 +1,0 @@
-```release-note:bug
-runner: start health check before blocking on adoption
-```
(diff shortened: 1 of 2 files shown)


.changelog/3052.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+runner: start health check before blocking on adoption
```
mitchellh, line 2: Don't need to changelog this, cause its only a bug with the runner adoption which was never released.

mitchellh (changes requested):
Good catch! One note.

briancain (approved):
🏃🏻

PR #3053, opened 2022-03-01
docs: Fix typo for config source-set

backspace:
Maybe this was a problem introduced in #868?

### DIFF
--- a/internal/cli/app_docs.go
+++ b/internal/cli/app_docs.go
@@ -384,7 +384,7 @@ func (c *AppDocsCommand) mdxFormatConfigSourcer(name, ct string, doc *docs.Docum
 
 	if len(doc.Fields()) > 0 {
 		fmt.Fprintf(w, "\n\n### Source Parameters\n\n"+
-			"The parameters below are used with `waypoint config set-source` to configure\n"+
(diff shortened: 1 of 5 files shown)


backspace:
ah I did see that the string existed in multiple places but I didn’t realise it needed to be run locally. I can try that later today 🤓

briancain:
Sounds great @backspace ! The trick to making it work is you'll just need a little Waypoint server locally, it does some plugin doc generation there when you run `make docs/website-mdx`. You should be able to just install it to docker to make it work right.

backspace:
Okay, I ran that locally and it changed three files. But there’s [another instance](https://github.com/hashicorp/waypoint/blob/main/website/content/partials/components/configsourcer-null.mdx) that didn’t change… should I change it manually? 😯

Also I clearly didn’t read `CONTRIBUTING.md` and used the Github UI to propose this change with the default branch name. I can close and reopen if that’s helpful!

briancain:
@backspace - no worries, the branch name is fine!

I'm not sure why that one didn't update, strange. I'd say update it manually and that should be good. Thank you!

backspace:
👍🏻 done now, thank you!

PR #3059, opened 2022-03-02
Fixing two bugs in the adopt runner flow

izaaklauer:
Bug 1: we were ignoring an error

But 2: we weren't propagating the token from the runner adoption flow into all the contexts we were using to make client api calls.


mitchellh (approved):
Wait for tests to pass but this is good, I was on Zoom with Izaak to verify this fix.

briancain (approved):

PR #3061, opened 2022-03-02
change runner adoption protobuf

xiaolin-ninja:
Some [cleverness in modifying the `runner list` led to a change in output. ](https://hashicorp.slack.com/archives/C013QT1KG9W/p1646232325910549)This changes the protobuf to make the cleverness work.


briancain (approved):
Makes sense to me 👍🏻

izaaklauer (approved):
it'll be non-ideal that we'll print "preadopted" instead of "pre-adopted" in the CLI, but it's good enough for me for now I think.

PR #3062, opened 2022-03-03
Do not add ODR profile to job if profile is empty

izaaklauer:
Currently, if you have a default runner profile but no project-specific target, the waypoint server fails the job with the following error: `no runner profile name or id specified`. This is because the client's `waypointHCL.ConfigRunner()` is returning non-nil, but the profile field is an empty string, so the job is getting a runner profile assigned (but with an empty profile string).

The server is expecting a nil odr profile on a job to mean "no preference" [here](https://github.com/hashicorp/waypoint/blob/ff4406443d64d5f16812bef32d8b4428656aae6e/internal/server/singleprocess/service_job.go#L171), not an assigned profile with an empty string.

This fixes the problem by not assigning an odr profile to the job unless the profile name is actually set.

### DIFF
--- a/internal/client/job.go
+++ b/internal/client/job.go
@@ -245,8 +245,10 @@ func (c *Project) doJobMonitored(ctx context.Context, job *pb.Job, ui terminal.U
 		}
 		// If runner config is found, assign to job
 		if configRunner != nil {
-			job.OndemandRunner = &pb.Ref_OnDemandRunnerConfig{
-				Name: configRunner.Profile,
+			if configRunner.Profile != "" {
+				job.OndemandRunner = &pb.Ref_OnDemandRunnerConfig{
+					Name: configRunner.Profile,
+				}
 			}
 		}
 	}


catsby (approved):
LGTM, though I did not test.

I think we should add a changelog for this if this is a "bug" that has been released. If this kind of profile targeting is not yet released then probably ok without one

xiaolin-ninja (changes requested):
I can't find that error message you're referring to could you drop a link or how to reproduce?

Right now there is a check in `remoteOpPreferred` for whether the profile is an empty string: https://github.com/hashicorp/waypoint/blob/main/internal/client/runner.go#L116. The job should run locally if `runnerCfg.Profile` is empty and there is no default, otherwise it should grab the default. So this fix on the job isn't necessary, unless we want to refactor this logic.

xiaolin-ninja:
I ran some tests with a runner stanza but no profile set in the configs, and traced it through debugger... when it [hits the server ](https://github.com/hashicorp/waypoint/blob/ff4406443d64d5f16812bef32d8b4428656aae6e/internal/server/singleprocess/service_job.go#L171), `job.OndemandRunner` is correctly nil and it goes through the flow to pull a default ODR. In the case where there's no default profile, it goes local. So I wasn't able to reproduce this.

xiaolin-ninja (approved):
After discussion, we realized this is fixed in targetable runners 2, but this PR is separate because it needs to be good for Alpha.

Also before anyone says anything, I requested Izaak to nest the ifs for label-based targeting. :P

PR #3068, opened 2022-03-04
builtin/aws: add `cpu_architecture` aws-ecs parameter

smaeda-ks:
This is to support deploying Docker images built by the Apple M1 chip to ECS:
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-arm64.html

Adding new `cpu_architecture` parameter (optional, no default value):

```hcl
app "example-nodejs" {
  labels = {
    "service" = "example-nodejs",
    "env"     = "dev"
  }

  build {
    use "docker" {}
    registry {
      use "aws-ecr" {
        region     = "ap-northeast-1"
        tag        = "latest"
      }
    }
  }

  deploy {
    use "aws-ecs" {
      region = "ap-northeast-1"
      memory = "512"
      cpu_architecture = "ARM64"
    }
  }
}
```

NOTE: Currently, this only works with the `docker` build plugin, and not with the `pack` plugin. But I believe that's due to the upstream buildpacks issue with arm64, not coming from Waypoint itself.

### DIFF
--- a/.changelog/3068.txt
+++ b/.changelog/3068.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+builtin/aws/ecs: Add `cpu_architecture` aws-ecs parameter
+to support deploying Docker images built by the Apple M1 chip on ECS
+```
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -96,13 +96,14 @@ func (p *Platform) ConfigSet(config interface{}) error {
 		return err
 	}
 
-	if c.CpuArchitecture != "" {
+	if c.Architecture != "" {
+		c.Architecture = strings.ToUpper(c.Architecture)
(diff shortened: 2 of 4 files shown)


briancain (commented):
Hey @smaeda-ks ! Thank you for adding this. Overall the code make sense to me. Could you please add a changelog and generate the website docs?

smaeda-ks:
@briancain Thank you! Just pushed those additional commits :)

briancain (approved):
Thanks again!

evanphx:
This looks great, but can you update it to be `architecture` to be in line with a recent change over in lambda? https://github.com/hashicorp/waypoint/blob/main/builtin/aws/lambda/platform.go#L801

From PR: https://github.com/hashicorp/waypoint/pull/3046

smaeda-ks:
@evanphx Thanks! Does this look good?
https://github.com/hashicorp/waypoint/pull/3068/commits/20be53ccd1ddb591abb5a854457d8a86623deab8

PR #3069, opened 2022-03-04
gRPC interceptor to nil any "any" fields for gRPC-gateway

mitchellh:
The protojson encoder tries to decode "any" fields, but we use any fields as opaque blobs and don't want them decoded (except as raw bytes). Therefore, we get encoding errors. This solves the issue by simply nullifying any "any" fields that are intended for gRPC gateway because the HTTP API doesn't need these fields anyways.


izaaklauer (approved):
tested and modified and it works!

PR #3070, opened 2022-03-04
K8s: Don't error build/deploy/release because a previous deployment isn't found

catsby:
If for some reason a previous deployment is not found and on `waypoint up` we try to remove it, the entire operation will be reported as a failure even if we successfully deployed a new version.

Here we check for not found, and simply continue


briancain (approved):
Ah yeah, makes sense! We're already destroying it anyway 👍🏻

briancain:
Backport?

catsby:
> Backport?

@briancain agreed, backporting to 0.7.x

PR #3084, opened 2022-03-10
HDS hack spike 

amyrlam:
IS:

/auth
image

/default
image

For production, some of the mass copying of design-system-components styling should be sorted in a more sustainable way, but wanted to power through to get something for the look & feel of the hack day. 

***

Questions:

(1)
I still had to run Docker and the Waypoint CLI to use `ember serve` beyond the /auth page, is that expected? Without ` waypoint user token` in terminal via the CLI, I couldn't log in.

(2)
I noticed in https://github.com/hashicorp/waypoint/pull/2933 the background was white, is the color mode an option somewhere in the UI?

(3)
Button https://design-system-components-hashicorp.vercel.app/components/button might be the next best value add, or Badge https://design-system-components-hashicorp.vercel.app/components/badge, I just haven't seen as many of those personally in the Waypoint UI yet


amyrlam:
There are some less than desirable hacks in here, like copying over the majority of component styling form design-system-components, but just sharing for your reference! 

If you have DS things in mind that may be high value, let me know

And if you have any guidance on the questions above ^^ please lmk

sabrinako:
Responding to questions:
1. You shouldn't have to use Docker/Waypoint CLI to run locally, should only have to do that if you run `ember serve local`.
2. Not sure if I fully understand the question, but Waypoint should be responsive to the computer's setting light/dark mode. We don't have anywhere on the UI to toggle it manually.
3. I think you're right, those two would be a good addition. I'll try to work on those today!

amyrlam:
(1) Thanks a lot - appreciate your help on the dev env! Sorted in https://github.com/hashicorp/waypoint/pull/3085

(2) I see "light mode" now, that's really cool. I'll have to take a closer look at the code, didn't realize it was a built-in feature

(3) The tricky thing is that some of the global CSS will make implementing the DS challenging. I was originally copying in the CSS from the components on https://github.com/hashicorp/design-system-components line-by-line, but then copied in almost everything. Not an ideal practice, of course. Cool, I'll also try to find an example of `Hds::Badge` and `Hds::Button` today, probably later in the day though

ui/yarn.lock
```diff
@@ -2302,11 +2357,25 @@
     ember-cli-babel "^7.26.6"
     ember-cli-htmlbars "^5.7.1"
 
+"@hashicorp/ember-flight-icons@^2.0.1":
```
sabrinako, line 2360: Looks like flight icons is in here twice

ui/yarn.lock
```diff
@@ -2302,11 +2357,25 @@
...
+    "@hashicorp/flight-icons" "^2.1.1"
+    ember-cli-babel "^7.26.11"
+    ember-cli-htmlbars "^6.0.1"
+
 "@hashicorp/flight-icons@^2.0.0":
   version "2.0.0"
   resolved "https://registry.yarnpkg.com/@hashicorp/flight-icons/-/flight-icons-2.0.0.tgz#84910c97f2bc643a6ca039ca7ed9be0bf14f8ee2"
   integrity sha512-jNRHHCFvptOHLXPIlnZ4A1LIFtS6m+Bv72NsE3Pl7KHMwcqkLHDN8gNa6AHf80cvFE+pR3sEx6Yjxu79jWJczQ==
 
+"@hashicorp/flight-icons@^2.1.1":
```
sabrinako, line 2374: Second instance, different version

ui/yarn.lock
amyrlam: Yeah, it's kind of confusing:
- The package is brought in already for using `FlightIcon`
- Then it's also a dependency of @hashicorp/design-system-components (use `FlightIcon` in some of the components like `Hds::IconTile`)

I ran this in terminal:

```
waypoint (hack-hds) 🌓 ~ pwd
/Users/amyrlam/src/hashicorp/waypoint/ui
waypoint (hack-hds) 🌓 ~ yarn list @hashicorp/flight-icons
yarn list v1.22.17
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ @hashicorp/design-system-components@0.1.0
│  └─ @hashicorp/flight-icons@2.1.1
└─ @hashicorp/flight-icons@2.0.0
✨  Done in 1.30s.
```

But...we've seen issues before with mismatched versions, ref: https://github.com/hashicorp/design-system-components/issues/34

For now I could bump the existing `@hashicorp/ember-flight-icons` in a separate PR, should just be adding new icons, no breaking changes for the codebase?

sabrinako: Should be fine to make a PR for upgrading since we're doing it to a feature branch and not to main

amyrlam: made change on https://github.com/hashicorp/waypoint/tree/amy/hack-hds-2 which will PR later

amyrlam:
close in favor of https://github.com/hashicorp/waypoint/pull/3094 (thanks @sabrinako for the fixes!)

PR #3085, opened 2022-03-10
Add /ui README update

amyrlam:
Could definitely be better fleshed out, but wanted to PR it now so is not forgotten / is a step in the right direction.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -53,7 +53,7 @@ that are static and are re-loaded on page refresh.
 
 Troubleshooting:
 
-- If you run into issues with `ember serve`, try deleting the directory `ui/node_modules/` and rerunning an `ember serve`.
+- If you run into issues with `ember serve`, try deleting the directory `ui/node_modules/`, rerunning `yarn install`, and rerunning an `ember serve`.
 - You will need a Waypoint CLI token to log in, e.g. the output of `waypoint user token`. To achieve this, run Docker and the Waypoint CLI.
 
 ### Running with a local Waypoint Server
--- a/ui/README.md
+++ b/ui/README.md
@@ -50,11 +50,11 @@ that are static and are re-loaded on page refresh.
 
 - `ember serve`
 - The app will be available at [http://localhost:4200](http://localhost:4200).
+- When prompted for a token, you can use any non-empty string (i.e. `my-cool-token`)
 
 Troubleshooting:
 
 - If you run into issues with `ember serve`, try deleting the directory `ui/node_modules/`, rerunning `yarn install`, and rerunning an `ember serve`.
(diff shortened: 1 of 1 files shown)


ui/README.md
```diff
@@ -51,6 +51,11 @@ that are static and are re-loaded on page refresh.
 - `ember serve`
 - The app will be available at [http://localhost:4200](http://localhost:4200).
 
+Troubleshooting:
+
+- If you run into issues with `ember serve`, try deleting the directory `ui/node_modules/` and rerunning an `ember serve`.
+- You will need a Waypoint CLI token to log in, e.g. the output of `waypoint user token`. To achieve this, run Docker and the Waypoint CLI.
```
jgwhite, line 57: In mirage-mode, this can actually be any string at all

ui/README.md
```diff
@@ -51,6 +51,11 @@ that are static and are re-loaded on page refresh.
 - `ember serve`
 - The app will be available at [http://localhost:4200](http://localhost:4200).
 
+Troubleshooting:
+
+- If you run into issues with `ember serve`, try deleting the directory `ui/node_modules/` and rerunning an `ember serve`.
```
jgwhite, line 56:
```suggestion
- If you run into issues with `ember serve`, try deleting the directory `ui/node_modules/`, rerunning `yarn install`, and rerunning an `ember serve`.
```

ui/README.md
jgwhite: Maybe this could read:

> You can use literally any non-empty string as the login token, i.e. `my-cool-token`.

jgwhite: Thinking about this some more, we should probably move this instruction up into the list above, i.e.:

```md
This returns data in-browser with Mirage.js (a mocking framework)
active. This means that the network requests will be intercepted
and return [mocked objects](https://github.com/hashicorp/waypoint/tree/master/ui/mirage/services)
that are static and are re-loaded on page refresh.

- `ember serve`
- The app will be available at [http://localhost:4200](http://localhost:4200).
- When prompted for a token, you can use any non-empty string (i.e. `my-cool-token`)
```

amyrlam: Just tried it on my machine - thanks both! Updated the README with ^^, LMK what you think

sabrinako (approved):

amyrlam:
@sabrinako thanks for the review, what's the merge process? Do I merge in as the author, or does the Waypoint team?

sabrinako:
> @sabrinako thanks for the review, what's the merge process? Do I merge in as the author, or does the Waypoint team?

If you have the ability to merge you're allowed to as long as you get 1 approving review! The tests are all passing so you should be good to merge. If you can't do it on your end just message me

PR #3087, opened 2022-03-10
Runner resilience part 1/N: startup resilience to server being down or going down

mitchellh:
This is the first body of work towards making runners more resilient to servers being down (and ultimately vice versa). This part focuses on the server going down or already being down during the runner startup process. 

This PR makes the runner able to tolerate the following conditions:

1. Server down prior to the runner ever starting - the runner blocks and attempts to reconnect
2. Server goes down while waiting on adoption - the runner reconnects and restarts the adoption (which is idempotent)
3. Server goes down after adoption but before the initial config stream - reconnect and re-establish initial config stream
4. Server goes down after start (adoption and config established) - reconnect to the config stream in the background

Prior to this PR, all four scenarios above would crash or exit the runner. 

All four scenarios are unit tested.

This PR _does not_ improve runner resilience with regards to job execution. That is a future PR.

### DIFF
--- a/.changelog/3087.txt
+++ b/.changelog/3087.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+core: runners automatically reconnect on startup if the server is unavailable
+or becomes unavailable during the startup process
+```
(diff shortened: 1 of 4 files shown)


izaaklauer (approved):

briancain (approved):
Love the tests! 👨🏻‍🍳 💋

PR #3089, opened 2022-03-10
UI: Setup for design-system-components

sabrinako:
Currently no visible changes to the end user with this PR, just setup work

### What changed
The main change is in the `ember-cli-build.js` file, the `sassOptions` is what allows us to load the design system's styling. The other changes you see are about installing the necessary packages.

### How to test
1. Pull down branch
2. `yarn` to install needed dependencies
3. `yarn start`
4. Make sure nothing broke
5. Optional: edit/add an `<Hds::___>` component somewhere in the UI to see that the design system's CSS actually flows through

### DIFF
--- a/ui/app/styles/_layout.scss
+++ b/ui/app/styles/_layout.scss
@@ -74,8 +74,8 @@ ul {
 ul.sidebar {
   grid-area: menu;
   overflow: scroll;
-  height:0;
-  min-height:100%;
+  height: 0;
+  min-height: 100%;
 
   li:first-child {
     hr {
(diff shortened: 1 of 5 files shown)


amyrlam (approved):
LGTM, added in a `Hds::Button` locally for good measure

image

- Seems like we should tweak the design-system-components README for the `ember-cli-build` step, even if you already have ember-cli-sass
- Also, I should bring this change to the spike in https://github.com/hashicorp/waypoint/pull/3084

amyrlam:
made https://github.com/hashicorp/design-system-components/pull/214

PR #3092, opened 2022-03-10
builtin/vault(docs) clarify `key` value format difference

smaeda-ks:
This PR attempts to improve the current ambiguous description of the `key` parameter in the `vault` configsourcer doc:
https://www.waypointproject.io/plugins/vault#key

Basically, keys stored in the "KV2" (KV version 2) secret engine must be prefixed with `/data`.

```hcl
    PASSWORD_KV1 = dynamic("vault", {
      path = "kv1/my-secret"
      key = "password"
    })
    PASSWORD_KV2 = dynamic("vault", {
      path = "kv2/data/my-secret"
      key = "/data/password"  # super confusing
    })
```

ref. https://github.com/hashicorp/waypoint/blob/v0.7.2/builtin/vault/config_sourcer.go#L188

This is how `cachedSecretVal.Secret.Data` looks different between KV and KV2:

KV:
> map[string]interface {}{“password”:“foobar”}

KV2:
> map[string]interface {}{“data”:map[string]interface {}{“password”:“foobar”}, “metadata”:map[string]interface {}{“created_time”:“2022-03-10T18:14:10.883066333Z”, “custom_metadata”:interface {}(nil), “deletion_time”:“”, “destroyed”:false, “version”:“1"}}

### DIFF
--- a/builtin/vault/config_sourcer.go
+++ b/builtin/vault/config_sourcer.go
@@ -333,8 +333,16 @@ config {
       key = "password"
     })
 
-    "DATABASE_HOST" = dynamic("vault", {
-      path = "kv/database-host"
+    # KV Version 2
+    "PASSWORD_FOO" = dynamic("vault", {
+      path = "secret/data/my-secret
+      key = "/data/password"  # key must be prefixed with "/data" (see below)
+    })
+
+    # KV Version 1
+    "PASSWORD_BAR" = dynamic("vault", {
+      path = "kv1/my-secret
+      key = "password"
     })
   }
 }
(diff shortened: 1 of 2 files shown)


izaaklauer (commented):
I think we might want to lead with instructions on Vault KV Version 2. I think the majority of our users are going to be on KV V2, and as it stands they might read the begining of the description and then just go for it.

smaeda-ks:
@izaaklauer Thank you! Applied that suggestion and also updated the example snippet as well

briancain (approved):
Looks good to me, thanks for updating the documentation for this! 🙏🏻 🎉

PR #3094, opened 2022-03-11
More HDS hack spike: swap out components and bump ember-flight-icons

amyrlam:
Bump ember-flight-icons
  - Have same versions in the yarn.lock, one comes from design-system-components (re: https://github.com/hashicorp/waypoint/pull/3084#discussion_r824113529)
  
Swap `Hds::LinkTo::Standalone` for `Pds::CtaLink`
  - The `Hds::LinkTo::Standalone` is intentionally not styled like a button
image

Swap `Hds::Button` for button
  - Left some notes about the LinkTo's on the same screen. I did not swap them out here since the `Hds::Link (Standalone)` / `HDS::LinkTo (Standalone)` component requires an icon (is why looks off center with the below link)

image

Swap `Hds::Badge` for class="badge badge--version"
  - Went with @color="neutral-dark-mode" because of the existing grey background row select

image

Replaces: https://github.com/hashicorp/waypoint/pull/3084

Note: Since the HDS components do not have dark mode by default, that would be a blocker for production

### DIFF
--- a/ui/app/components/app-item/deployment.hbs
+++ b/ui/app/components/app-item/deployment.hbs
@@ -11,7 +11,7 @@
     @route="workspace.projects.project.app.deployments.deployment"
     @models={{array @deployment.sequence}}
   >
-    <b class="badge badge--version">v{{@deployment.sequence}}</b>
+    <Hds::Badge @text="v{{@deployment.sequence}}" @size="small" @color="neutral-dark-mode" />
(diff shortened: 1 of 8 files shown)


amyrlam:
@gregone I'm going to merge this into `ui/hack-hds` and open that as a Draft PR, for discussion tomorrow at our meeting

PR #3097, opened 2022-03-14
internal/runner: make Accept resilient to the server going down

mitchellh:
This starts by addressing only the initial job stream opening. This still does not add any resilience after the job stream is open (i.e. when a job is running).

I'm opening this PR early since this required some changes to how we track the state of the underlying runner config and it is complex enough on the surface that I didn't want to complicate a future PR. Additionally, this PR on its own still incrementally makes the runner more resilient to failure modes.

This PR builds on #3087 and adds the following scenarios the runner can now withstand:

* The server goes down after runner start but prior to `Accept`.
* The server goes down during `Accept` (its a stream, so while the initial handshake is blocking)
* The runner registration is disconnected at some point prior to or during `Accept`. This is usually due to the server going down, but the behavior that the runner sees is the same. 
  - This last one in particular is nice to resolve since the error was not obvious. The error returned to users was "runner not found". This is no longer possible.

## Runner State from Booleans to Monotonic Ints

A core change made here is the change from a set of booleans to represent state to monotonic unsigned ints. This lets us still retain the boolean-like behavior (anything greater than 0 is "true") while also giving us some "happens after" semantics we can check. This is important because on disconnect, we need to detect when we're reconnected _after_ the disconnect. Before, we only had a true/false of whether we were connected, and it was a race to determine if the "true" was still "true" because we hadn't yet detected its "false." Numbers solve this.

As a bonus, changing to this method let us remove a few cases that required special handling :)

### DIFF
--- a/internal/runner/accept.go
+++ b/internal/runner/accept.go
@@ -132,7 +132,7 @@ func (r *Runner) accept(ctx context.Context, id string) error {
 				// RunnerConfig stream to re-establish. We wait for the config
 				// generation to increment.
 				if r.waitStateGreater(&r.stateConfig, stateGen) {
-					return status.Errorf(codes.Internal, "early exit while waiting for reconnect")
+					return status.Error(codes.Internal, "early exit while waiting for reconnect")
 				}
 
 				continue
--- a/internal/runner/accept.go
+++ b/internal/runner/accept.go
@@ -108,9 +108,11 @@ func (r *Runner) accept(ctx context.Context, id string) error {
 		ctx = serverclient.TokenWithContext(ctx, tok)
 	}
 
-	// Open a new job stream. NOTE: we purposely do NOT use ctx above
-	// since if the context is cancelled we want to continue reporting
-	// errors.
+	// Open a new job stream. This retries on connection errors. Note that
+	// this loop doesn't respect the accept timeout because gRPC has no way
+	// to time out of a "WaitForReady" RPC call (it ignores context cancellation,
+	// too). TODO: do a manual backoff with WaitForReady(false) so we can
+	// weave in accept timeout.
 	retry := false
 	var client pb.Waypoint_RunnerJobStreamClient
 	for {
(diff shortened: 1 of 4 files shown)


internal/runner/accept.go
```diff
@@ -114,10 +111,37 @@ func (r *Runner) accept(ctx context.Context, id string) error {
...
+			retry = true
+
+			if status.Code(err) == codes.Unavailable || status.Code(err) == codes.NotFound {
+				log.Warn("server down during job stream open, will attempt reconnect")
+
+				// Since this is a disconnect, we have to wait for our
+				// RunnerConfig stream to re-establish. We wait for the config
+				// generation to increment.
+				if r.waitStateGreater(&r.stateConfig, stateGen) {
+					return status.Errorf(codes.Internal, "early exit while waiting for reconnect")
```
briancain, line 135:
```suggestion
					return status.Error(codes.Internal, "early exit while waiting for reconnect")
```

I figure this is good as `Error` since there's no message formatting.

internal/runner/accept.go
```diff
@@ -114,10 +111,37 @@ func (r *Runner) accept(ctx context.Context, id string) error {
 	// Open a new job stream. NOTE: we purposely do NOT use ctx above
 	// since if the context is cancelled we want to continue reporting
 	// errors.
-	log.Debug("opening job stream")
-	client, err := r.client.RunnerJobStream(streamCtx)
-	if err != nil {
-		return err
+	retry := false
+	var client pb.Waypoint_RunnerJobStreamClient
+	for {
```
briancain, line 116: Does this need any kind of timeout awareness to prevent infinite loop retries? Maybe infinite loop retries is by design!

briancain, line 116: Does this need any kind of timeout awareness to prevent infinite loop retries? Maybe infinite loop retries is by design!

internal/runner/runner.go
```diff
@@ -82,10 +81,10 @@ type Runner struct {
...
 	// when the root context cancels.
 	stateCond       *sync.Cond
-	stateConfig     bool // config stream is connected
-	stateConfigOnce bool // true once we process config once, success or error
-	stateJobReady   bool // ready to start accepting jobs
-	stateExit       bool // true when exiting
+	stateConfig     uint64 // config stream is connected
+	stateConfigOnce uint64 // true once we process config once, success or error
+	stateJobReady   uint64 // ready to start accepting jobs
+	stateExit       uint64 // true when exiting
```
briancain, line 87: These doc strings should be updated to the new way they're being handled with increments!

briancain, line 87: Do these struct doc strings need to be updated to what `uint64` values mean in this case? Perhaps it's just 0 true 1 false, N "something else"?

briancain (approved):
Looks pretty good to me, just a few minor requests for fixups!

internal/runner/accept.go
mitchellh: Due to reasons I put in the comment, this is very difficult to solve right now. I'm proposing just keeping this as a TODO and doing this another time. For now, it would retry forever past the `AcceptTimeout` setting but I think that's okay, that's only true so long as the server remains down.

PR #3099, opened 2022-03-14
Server: Introduce Task for tracking on-demand runner spawned tasks

briancain:
This pull request is 1/Many I'll be working on to introduce more history into the tracking for on-demand runner spawned tasks. Initially, this PR focuses on the state model for tracking spawned task history. It currently only keeps track of job id refs that will be associated with the spawned tasks once the jobs have been queued. Task refs can be looked up either directly by a TrackTask id or the main Job Id the task was responsible for spawning.

Reviewer note: There might be more data `Task` should hold onto, but be aware _a lot_ of the data about the running operation already exists inside the referenced job id which can be looked up. If there's something else `Task` should have, let's add it!

No changelog yet until the last PR is ready - coming soon! Targeted for a major release which is at the moment version 0.8.0.


internal/server/singleprocess/state/track_task.go
```diff
@@ -0,0 +1,317 @@
...
+	schemas = append(schemas, tracktaskIndexSchema)
+}
+
+// TrackTaskPut creates or updates the given TrackTask.
+func (s *State) TrackTaskPut(t *pb.TrackTask) error {
+	memTxn := s.inmem.Txn(true)
+	defer memTxn.Abort()
+
+	err := s.db.Update(func(dbTxn *bolt.Tx) error {
+		if t.TaskJob == nil {
```
briancain, line 29: Actually thinking maybe this doesn't need to be required, depending on at what point we have the server upsert a TrackTask. I _think_ the first place will be where we wrap a requested job with Start/Stop tasks and I think by that point we might know the job ids for everything - So maybe this is ok. The downside if there's other task data to update later (say in the runner) the requester will need to make sure to include the original TaskJob Id along with the original TrackTask Id.

Maybe something to leave for now - and fix once the server actually uses this state?

internal/server/singleprocess/state/track_task.go
```diff
@@ -0,0 +1,317 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/hashicorp/go-memdb"
+	bolt "go.etcd.io/bbolt"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+var tracktaskBucket = []byte("tracktask")
```
mitchellh, line 15: Bikeshed for the whole thing: I'd name this `task`. State by definition is tracking something, so i.e. we don't do `trackjob`, just `job`. Similar vibe here.

internal/server/singleprocess/state/track_task.go
mitchellh: I'd keep it as an assertion. I think you're right: we'll always have the job.

pkg/server/proto/server.proto
```diff
@@ -902,6 +902,22 @@ message Ref {
...
+  }
+
+  // TrackTask references a TrackTask message by its id or the job id it queued
+  message TrackTask {
+    oneof ref {
+      // the id of the tracktask record
+      string id = 1;
+
+      // The main job ID that the task initiated
+      string job_id = 2;
```
mitchellh, line 918: Neat! There are 3 tasks revolving around a task: start, run, stop. Its unclear to me as a first time reader what this is referencing, it sounds like "run". Could you clarify in doc comments?

pkg/server/proto/server.proto
```diff
@@ -1087,6 +1105,32 @@ message TaskLaunchInfo {
   repeated string arguments = 3;
 }
 
+// TrackTask tracks the life of an on-demand runner task that spawns Start
+// and Stop tasks for any kind of job/operation in Waypoint. Automatic jobs
+// such as project polling or status report generation spawn on-demand
+// runner tasks, and this message can be used to track the life of those
+// automated jobs. Note that every on-demand runner task is wrapped with a
+// Start and Stop task, which we track here as well.
+message TrackTask {
```
mitchellh, line 1114: Would just rename to `Task` (along with everything else, see first review comment which explains why)

pkg/server/proto/server.proto
```diff
@@ -1087,6 +1105,32 @@ message TaskLaunchInfo {
...
+  // Currently the only required field when Upserting a TrackTask.
+  Ref.Job task_job = 2;
+
+  // Both start_job and stop_job relate to task_job as these jobs were
+  // responsibile for starting and stopping the task job.
+  Ref.Job start_job = 3;
+  Ref.Job stop_job = 4;
+
+  // This task info message represented as json format
+  string task_json = 5;
```
mitchellh, line 1131: What is this?

pkg/server/proto/server.proto
```diff
@@ -1087,6 +1105,32 @@ message TaskLaunchInfo {
...
+  // assumes the TrackTask message already exists in the database.
+  string id = 1;
+
+  // The referred job stores data like ODR config, the target runner used,
+  // timestamp for when job started or completed, operation state for task, etc.
+  // Currently the only required field when Upserting a TrackTask.
+  Ref.Job task_job = 2;
+
+  // Both start_job and stop_job relate to task_job as these jobs were
+  // responsibile for starting and stopping the task job.
```
mitchellh, line 1126: Nit: starting and stopping the _runner_ which executed the task job.

mitchellh (changes requested):
One big bikeshed, a few other comments. Mostly 🔥

internal/server/singleprocess/state/track_task.go
briancain: Totally - I initially reached for calling it `Task`, and wondered since it's a top level message maybe `Task` is better reserved for something else, but this is probably the thing it should be!

briancain: Done!

pkg/server/proto/server.proto
briancain: Done!

pkg/server/proto/server.proto
briancain: This was the metadata json suggestion for the frontend to consume raw task info like `artifact_json`. Not hooked up to anything yet.

pkg/server/proto/server.proto
briancain: Oh that's right! It's the run job. Will clarify in the comments 👍🏻

pkg/serverstate/statetest/test_task.go
```diff
@@ -12,42 +12,42 @@ import (
 
 func init() {
 	tests["tracktask"] = []testFunc{
```
mitchellh, line 14: Missed one here

mitchellh:
One note for @evanphx, its not necessary for HCP to [yet] implement this and the implementation may still change. I think this PR will introduce state test failures cause of missing entries, I wonder if you can just skip/ignore them for now, make them noops.

internal/server/singleprocess/state/task.go
```diff
@@ -0,0 +1,317 @@
...
+		taskId = task.Id
+	default:
+		return nil, status.Error(codes.FailedPrecondition, "No valid ref id provided in Task ref to taskIdByRef")
+	}
+
+	return []byte(strings.ToLower(taskId)), nil
+}
+
+func (s *State) taskByJobId(jobId string) (*pb.Task, error) {
+	trackedTasks, err := s.TaskList()
```
mitchellh, line 272: Ah hah, I would add a `jobId` field to the index and use an index lookup for this. I think we'll regularly have a lot of tasks and this is going to get nasty.

briancain: Ah yeah, good call!

pkg/server/proto/server.proto
```diff
@@ -1087,6 +1105,35 @@ message TaskLaunchInfo {
   repeated string arguments = 3;
 }
 
+// Task tracks the life of an on-demand runner task that spawns Start
+// and Stop tasks for any kind of job/operation in Waypoint. Automatic jobs
+// such as project polling or status report generation spawn on-demand
+// runner tasks, and this message can be used to track the life of those
+// automated jobs. Note that every on-demand runner task is wrapped with a
+// Start and Stop task, which we track here as well.
+message Task {
```
mitchellh, line 1114: I also just thought, we're probably going to need some form of task expiry. You can get a lot of the timestamps from the jobs, but there is probably a state API we'll eventually need. You don't need to do anything in this PR I just wanted to put it in your head so you're aware of it and think about it as the implementation progresses.

evanphx: I think my concern with this is likely in the upper layer, mostly figuring out when TaskDelete is going to be called. The Task api used doesn't mandate an explicit place for Tasks to be finished, so I'd like to understand how we'll prevent Task values from building up over time.

mitchellh: I think they'll be part of the pruning similar to jobs (not sure how you did that in the sql-side). Otherwise, I think we set a timeout currently on ODR tasks so we can just double or triple that (for a really safe buffer) prior to deleting the task. But yes, this is a really important thing to look closely at @briancain when you do the implementation.

mitchellh: I wouldn't just add pruning, I'd think about how the lifecycle relates to it since tasks are tied to jobs, unlike jobs which stand on their own. Pruning _may_ be the answer, but I think we can potentially have a more solid solution.

briancain: Makes sense...I should of hit refresh before sending that comment! I missed the first one after Evans.

I can see us cleaning up tasks on ODR timeouts for sure, that use-case makes sense to me 👍🏻 Perhaps I can write up a little gist idea to chat about in slack for a plan going forward. I can foresee this being addressed in Phase 3 of the plan, where we implement actually storing these Task messages on ODR job queues in the server and later in the runner on execution.

pkg/server/proto/server.proto
briancain: Done!

pkg/server/proto/server.proto
briancain: I added a TODO above the Task message docs for me for phase 3! I think that should cover it until I get there later after the API/CLI phase 2.

mitchellh (approved):
Looks good. If we decide to change course we can always do so since this doesn't currently actually get used anywhere.

PR #3102, opened 2022-03-15
internal/{server,cli}: Attempt to forcefully cancel a job

briancain:
This commit exposes the `force` option for forcefully canceling a job in
Waypoint. The `force` option already existed at the state level,
including unit tests, so this commit simply exposes the `force` option
at the API level and introduces a flag in the CLI.


internal/cli/job_cancel.go
```diff
@@ -59,14 +60,26 @@ func (c *JobCancelCommand) Run(args []string) int {
...
 }
 
 func (c *JobCancelCommand) Flags() *flag.Sets {
 	return c.flagSet(flagSetOperation, func(set *flag.Sets) {
+		f := set.NewSet("Command Options")
+		f.BoolVar(&flag.BoolVar{
+			Name:    "force",
+			Target:  &c.flagForce,
+			Default: false,
+			Usage:   "Will attempt to forcefully cancel a job.",
```
mitchellh, line 81: I would recommend adding a lot more warning help text here. Because **force cancel is very dangerous.** I didn't expose it on purpose because people tend to just give it a shot if cancel doesn't work right away, but it is _very dangerous_. I think it makes sense to expose it but the help text should reflect the danger.

A forcefully canceled job will attempt to send a cancellation signal but it is not guaranteed. If the job is actually still running, the result will never be read. The status of it will never be followed up. If it is stuck, it will remain stuck forever. If another job was queued for the same app/workspace, it will begin running immediately which can cause concurrency issues depending on the plugin. 

A force cancel will _never fail_ to update the server state, but _may fail_ to clean up resources. A graceful cancel will _never fail_ to clean up resources, but _may fail_ to cancel the server state (since it gracefully waits).

My recommendation would perhaps be: "Will forcefully cancel the job. This will immediately mark the job as complete in the server, regardless of the real job status. This may leave dangling resources or cause concurrency issues if the underlying job doesn't gracefully cancel."

Note I specifically removed "attempt" from the help text. Force cancel cannot fail.

mitchellh (changes requested):

mitchellh (approved):

internal/cli/job_cancel.go
briancain: Roger that. Good call! I updated this PR to include some warning text, additional CLI UX warnings on if the flag is used, and the CLI will sleep for a few seconds to give users a chance to ctrl-c if they aren't sure about force cancel.

PR #3103, opened 2022-03-15
chore: `make install` conditionally remove existing waypoint file

jfreeland:
If you don't have a `waypoint` file at `$(GOPATH)/bin` then `make install` fails here.

But you could just as easily `touch $GOPATH/bin/waypoint`.


briancain (approved):
👍🏻 worked for me!

PR #2771, opened 2021-11-23
Replace -remote flag with -local

izaaklauer:
Merge after https://github.com/hashicorp/waypoint/pull/2760/files

Replaces the -remote flag with the -local flag. Nice deprecation warning for -remote.

The impetus for this change is that shortly, waypoint will automatically default operations to happen remotely if they're likely to succeed remotely, and locally if otherwise. With that change in place, it will be much more likely that users will want to override an operation to happen locally, and adding `-local` is more friendly than adding `-remote=false`.

### How to verify

Flag usage
```
$ waypoint artifact build -local

» Building web...
✓ Initializing Docker client...
✓ Building image...
 │ Step 1/3 : FROM nginx:stable
 │  ---> c8d03f6b8b91
 │ Step 2/3 : COPY ./public/ /var/www
 │  ---> Using cache
 │  ---> 61409f8fbdaf
 │ Step 3/3 : COPY ./nginx.conf /etc/nginx/conf.d/default.conf
 │  ---> Using cache
 │  ---> 5e11c4551189
 │ Successfully built 5e11c4551189
 │ Successfully tagged waypoint.local/web:latest
✓ Injecting Waypoint Entrypoint...
Image built: waypoint.local/web:latest (amd64)

$ waypoint artifact build -local=false

» Building web...
! Project nginx-project does not have a data source configured. Remote jobs
  require a data source such as Git to be configured with the project. Data
  sources can be configured via the CLI or UI. For help, see :
  https://www.waypointproject.io/docs/projects/git#configuring-the-project
```

-remote deprecation
```
$ waypoint artifact build -remote
! The -remote flag has been deprecated. Use -local=false instead

$ waypoint artifact build -remote=false
! The -remote flag has been deprecated. Use -local=true instead

```

PR #3107, opened 2022-03-16
doc: deprecate `-remote` flag

smaeda-ks:
ref. https://github.com/hashicorp/waypoint/pull/2771

replace `-remote` flag with `-local`

### DIFF
--- a/website/content/commands/index.mdx
+++ b/website/content/commands/index.mdx
@@ -55,26 +55,58 @@ For example, to see help about the up subcommand:
 $ waypoint up -h
 Usage: waypoint up [options]
 
-  Perform the build, deploy, and release steps for the project.
+  Perform the build, deploy, and release steps.
 
 Global Options:
 
+  -app=<string>
+      App to target. Certain commands require a single app target for Waypoint
+      configurations with multiple apps. If you have a single app, then this
+      can be ignored. This is aliased as "-a".
+
+  -plain
+      Plain output: no colors, no animation. The default is false.
+
+  -project=<string>
+      Project to target. This is aliased as "-p".
+
   -workspace=<string>
-      Workspace to operate in. Defaults to 'default'. The default is default.
+      Workspace to operate in. This is aliased as "-w".
 
 Operation Options:
 
   -label=<key=value>
       Labels to set for this operation. Can be specified multiple times.
 
-  -remote
-      True to use a remote runner to execute. This defaults to false unless
-      'runner.default' is set in your configuration. The default is false.
+  -local
+      True to use a local runner to execute the operation, false to use a
+      remote runner. If unset, Waypoint will automatically determine where the
+      operation will occur, defaulting to remote if possible.
 
   -remote-source=<key=value>
       Override configurations for how remote runners source data. This is
       specified to the data source type being used in your configuration. This
       is used for example to set a specific Git ref to run against.
+
+  -var=<key=value>
+      Variable value to set for this operation. Can be specified multiple
+      times.
+
+  -var-file=<string>
+      HCL or JSON file containing variable values to set for this operation.
+      If any "*.auto.wpvars" or "*.auto.wpvars.json" files are present, they
+      will be automatically loaded.
+
+Command Options:
+
+  -prune
+      Prune old unreleased deployments. The default is true.
+
+  -prune-retain=<int>
+      The number of unreleased deployments to keep. If this isn't set or is
+      set to any negative number, then this will default to 1 on the server.
+      If you want to prune all unreleased deployments, set this to 0. The
+      default is -1.
 ```
 
 Running `waypoint -autocomplete-install` will add the waypoint autocomplete capability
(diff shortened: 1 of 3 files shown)


website/content/commands/index.mdx
```diff
@@ -55,26 +55,58 @@ For example, to see help about the up subcommand:
 $ waypoint up -h
 Usage: waypoint up [options]
 
-  Perform the build, deploy, and release steps for the project.
```
izaaklauer, line 58: I'd vote for leaving `for the project` here - it answers the question "what will waypoint build, deploy, and release?"

website/content/commands/index.mdx
```diff
@@ -55,26 +55,58 @@ For example, to see help about the up subcommand:
...
+      can be ignored. This is aliased as "-a".
+
+  -plain
+      Plain output: no colors, no animation. The default is false.
+
+  -project=<string>
+      Project to target. This is aliased as "-p".
+
   -workspace=<string>
-      Workspace to operate in. Defaults to 'default'. The default is default.
```
izaaklauer, line 63: I'd vote for us to leave the default value documented here (but not say it twice haha)

website/content/docs/kubernetes/container-build.mdx
```diff
@@ -58,14 +58,14 @@ file. You can test this using the Waypoint CLI by running `waypoint build`. This
 command will perform the build only, so it can be used for testing. The
 `waypoint build` command should be run in the same directory as your project
 (with the `waypoint.hcl`). If you want to force a
-[remote run](#local-vs-remote-builds), use the `-remote` flag.
+[remote run](#local-vs-remote-builds), use the `-local=false` option.
```
izaaklauer, line 61: I might go with something like:

```
Waypoint will automatically run your build remotely if you have a git repo configured for your project and a runner and runner profile configured. If you want to force a build to happen locally, use the `-local` flag.
```

website/content/docs/kubernetes/container-build.mdx
```diff
@@ -58,14 +58,14 @@ file. You can test this using the Waypoint CLI by running `waypoint build`. This
...
 
 ```shell-session
 $ waypoint build
 ...
 
 # or, to run it remotely in your Kubernetes cluster
-$ waypoint build -remote
+$ waypoint build -local=false
```
izaaklauer, line 68: I vote for 

~~~


```shell-session
# Let waypoint choose to run your build remotely if possible:
$ waypoint build
...

# or, force it to run locally on your development machine:
$ waypoint build -local
~~~

File: website/content/docs/projects/remote.mdx [hunk 5]
```diff
@@ -44,10 +44,10 @@ $ waypoint up my-project/web
 
 If you have the local source checked out and your terminal working directory
 is in your project source code, you can omit the project name and use the
-`-remote` flag:
+`-local=false` option:
```
izaaklauer, line 47: We've also removed the old `app/project` syntax.

I'd go for something like:
~~~
Remote operations require:
- At least one [runner is installed and running](/docs/runner). Waypoint
  automatically installs a runner with `waypoint install`.
- Your project is configured with a [data source](/docs/waypoint-hcl/runner)
  so that the runners can clone the project source code.
- Your project has [remote operations enabled](/docs/waypoint-hcl/runner).

Waypoint will automatically run your operations remotely if it detects these are true, and that you
have a runner profile configured.

## Usage
### CLI


If waypoint is running your operations remotely but you'd prefer them to run locally, add the 
`-local` flag:

```shell-session
$ waypoint up -local
```
~~~

website/content/docs/projects/remote.mdx
```diff
@@ -60,7 +60,7 @@ logic is executing remotely.
 
 Waypoint only allows a single operation per distinct `(workspace, project, app)`
 combination. This applies to both local and remote operations. Therefore,
-if one user runs `waypoint up` locally and another runs `waypoint up -remote`,
+if one user runs `waypoint up` locally and another runs `waypoint up -local=false`,
```
izaaklauer, line 63: This is actually true even for two local builds! I'd vote we document that case instead: 

```
Therefore, if two users attempt run `waypoint up` at the same time, even if they add the `-local` flag, one of the operations will wait for the other to complete before continuing
```

izaaklauer (changes requested):
Thanks for spotting these problems with the docs! I think they actually need a more complex update - i've added some comments and context, but let me know if you'd like to hand it over to us to make the changes.


Context on the `-local` flag - previously, all waypoint operations were performed locally, unless you added the `-remote` flag, because remote operations were advanced and wouldn't work in all scenarios.

In 0.6, we updated waypoint to first verify if a project was configured to run successfully remotely - it checks that the project has a git source configured, has a static runner checked in, and has an ODR profile set. If all of those are true, it'll automatically run the op remotely, and otherwise run it locally.

The `-local=false` option is an escape hatch for users with some exotic setup that our heuristics don't handle properly. For nearly all users, if waypoint doesn't automatically run something remotely, adding `-local=false` will result in a failed operation, likely because they don't have git ref or an ODR config. So I'd like to avoid pointing users to adding `-local=false` anywhere in the general docs.

website/content/commands/index.mdx
smaeda-ks: @izaaklauer Sorry for the confusion! This is actually taken from the actual CLI output of `$ waypoint up -h` with the latest version, so I leave this section as-is...!

smaeda-ks:
@izaaklauer Thank you so much for the cotext! I've updated my changes as per your suggestions (with some modifications).
If this is going to require a lot more fundamental documentation updates I'm totally fine with handing this over your way. Thanks!

PR #3109, opened 2022-03-17
builtin/nomad/jobspec: fix dropped errors

alrs:
This fixes three dropped errors in `builtin/nomad/jobspec`. No changelog entry is warranted.


briancain (approved):
Thanks!

PR #3111, opened 2022-03-17
K8s: Don't error/hang if deleting a service that isn't found

catsby:
If a K8s deployment has multiple active deployments and a user runs `waypoint destroy -auto-approve`, Waypoint will hang trying to destroy the service multiple times. 

Here we check to see if the service is not found, and if so simply continue.

### DIFF
--- a/builtin/k8s/releaser.go
+++ b/builtin/k8s/releaser.go
@@ -338,6 +338,7 @@ func (r *Releaser) resourceServiceCreate(
 
 func (r *Releaser) resourceServiceDestroy(
 	ctx context.Context,
+	log hclog.Logger,
 	state *Resource_Service,
 	sg terminal.StepGroup,
 	csinfo *clientsetInfo,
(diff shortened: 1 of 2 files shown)


builtin/k8s/releaser.go
```diff
@@ -358,7 +358,9 @@ func (r *Releaser) resourceServiceDestroy(
 
 	step = sg.Add("Deleting service...")
 	if err := serviceclient.Delete(ctx, state.Name, metav1.DeleteOptions{}); err != nil {
-		return err
+		if !errors.IsNotFound(err) {
+			return err
+		}
```
briancain, line 363: Maybe we should log the error otherwise? Just for debug purposes.

briancain (approved):

builtin/k8s/releaser.go
catsby: Good idea, added in de591fb38

PR #3116, opened 2022-03-19
Switch away from google.protobuf.Any

evanphx:
Follows on from https://github.com/hashicorp/waypoint-plugin-sdk/pull/68, removes usage of google.protobuf.Any to avoid the deref behavior when converting to json.

Also finishes up the conversion to the new protobuf API.


izaaklauer (approved):

catsby (approved):

mitchellh:
I assume `opaqueany` kept the same tag numbers and structure as an `Any`, since otherwise this will be a major breaking change. But otherwise, it should be totally fine.

evanphx:
@mitchellh Yup, I was extremely careful to be sure opaqueany/Any used the same tag numbers an protobuf.any/Any. In fact, I straight up copied the definition and did not alter it to be sure.

PR #3119, opened 2022-03-21
support runner job stream reattachment, make runners resilient to server outage mid-execution

mitchellh:
This makes runners fully resilient to server outages even during the middle of an active job execution.

Job execution uses a gRPC stream, and there is no way to "continue" an existing gRPC stream after a disconnect; the client must re-establish a new stream. To make this work, I made two changes. For the runner, the most significant change was to wrap the existing stream interface and do the reconnect automatically so it _feels_ like a single continuous stream but in fact it is reconnecting.

The server needed a change as well so that it was aware that a stream opening was a reconnection rather than a new job stream. To do this, this introduces a change to the server API: the `RunnerJobStream` endpoint supports job reattachment. 

Tests were added to verify behavior.


internal/runner/accept.go
```diff
@@ -89,63 +88,131 @@ func (r *Runner) AcceptExact(ctx context.Context, id string) error {
...
+			client.CloseSend()
+		}
 
-				retry = true
-				continue
-			}
+		streamCtxLock.Lock()
+		defer streamCtxLock.Unlock()
+		if streamCancel != nil {
+			// Wrap in a func() so that if we retry, we don't stack defers.
```
evanphx, line 129: Not sure what this comment is referring to....

internal/runner/accept.go
```diff
@@ -89,63 +88,131 @@ func (r *Runner) AcceptExact(ctx context.Context, id string) error {
...
+			// but it is used when the streamCancel is being reset so that
+			// we don't set it up and race with cancelation.
+			streamCtxLock.Lock()
+			defer streamCtxLock.Unlock()
+
+			// Mark that we canceled
+			atomic.StoreInt32(&canceled, 1)
+
+			// Cancel the context
+			streamCancel()
```
evanphx, line 153: Do we need to check that `streamCancel` is not nil? Most of the other places do...

internal/runner/accept.go
```diff
@@ -89,63 +88,131 @@ func (r *Runner) AcceptExact(ctx context.Context, id string) error {
...
+	stateGen = r.readState(&r.stateConfig)
 
-		break
+	// Open a new job stream. This retries on connection errors. Note that
+	// this retry loop doesn't respect the accept timeout because gRPC has no way
+	// to time out of a "WaitForReady" RPC call (it ignores context cancellation,
+	// too). TODO: do a manual backoff with WaitForReady(false) so we can
+	// weave in accept timeout.
+	// NOTE: we purposely do NOT use ctx above since if the context is
+	// cancelled we want to continue reporting errors.
```
evanphx, line 203: I think this comment should come around line 181 instead? Is that the situation it's referring to?

internal/runner/reattach.go
```diff
@@ -0,0 +1,262 @@
...
+		if r.waitStateGreater(&r.stateConfig, stateGen) {
+			return status.Error(codes.Internal, "early exit while waiting for reconnect")
+		}
+
+		// Get the new state generation in case we get another disconnect
+		// we can continue retrying and detecting new connections.
+		stateGen = r.readState(&r.stateConfig)
+
+		log.Info("opening job stream")
+		client, err = r.client.RunnerJobStream(context.Background(), grpc.WaitForReady(true))
```
evanphx, line 181: How long with this block for if the server is down? Indefinitely? Since we're using a new context, can you add a comment about this lines "blocking mode"?

internal/runner/reattach.go
```diff
@@ -0,0 +1,262 @@
...
+			Event: &pb.RunnerJobStreamRequest_Request_{
+				Request: &pb.RunnerJobStreamRequest_Request{
+					RunnerId:      r.id,
+					ReattachJobId: c.jobId,
+				},
+			},
+		}); err != nil {
+			if status.Code(err) == codes.Unavailable ||
+				status.Code(err) == codes.NotFound {
+				continue
```
evanphx, line 204: We're abandoning `client` and the top of the loop will use `c.client`, the later I presume is totally dead. Should we be assign `c.client = client` as soon as we get one?

evanphx (changes requested):
Looks great! This is a hair problem and this comes out pretty clean. I've made a few comments about clarifications and perhaps needing to move/add some comments for later readers.

internal/runner/accept.go
mitchellh: Yep, good find.

internal/runner/reattach.go
mitchellh: That's a good idea. This is a bug due to restrucuring from goto to a loop here.

mitchellh: Actually, just used a `goto` to avoid having to execute `f` again, and commented that.

izaaklauer (approved):
This is some intense computer code. 

I've read through the logic and the tests, and it looks good to me.

Thanks for tackling this!

internal/runner/reattach.go
mitchellh: This was a mistake, now fixed to use the context given to `accept` so the lifecycle matches that.

evanphx (approved):

PR #3121, opened 2022-03-22
ui: Regenerate libs from protos

jgwhite:
## Why the change?

The main motivation is #3116, but also it’s been a while since we regenerated the libs and there are plenty of changes to test against.

## What’s the plan?

- [x] Account for `opaqueany` in the build process and UI dependency graph
- [x] Rebuild the dependencies
- [x] Improve documentation around this process

## What versions of the tools did you use?

| Tool | Version |
| --- | --- |
| protoc | 3.17.3 |
| ts-protoc-gen | 0.15.0 |
| protoc-gen-grpc-web | 1.3.1 |

## What does it look like?

Hopefully no changes!

## How do I test it?

1. Check out the branch
2. Install dependencies
3. Trying booting the app locally as you normally would (against Mirage and a real server)
4. Try out the new guidance in the README


ui/README.md
```diff
@@ -85,12 +85,14 @@ if you've made API changes in `/internal/server` and want to use those on the fr
 
 - MacOS only: `brew install gnu-sed` then follow the instructions to replace the default `sed`
 - Download [the 1.1.2 release of `mockery`](https://github.com/vektra/mockery/releases/tag/v1.1.2) and install in your `/go/bin` directory
-- install `ts-protoc-gen`: `yarn global add ts-protoc-gen` or `npm i -g ts-protoc-gen`
-- install `protoc-gen-grpc-web`: `brew install protoc-gen-grpc-web`
+- Install [`protoc` 3.17.3](https://github.com/protocolbuffers/protobuf/releases/tag/v3.17.3)
```
jgwhite, line 88: It still feels a little weird that we’re using a different version of protoc for the UI libs, but let’s not rock that particular boat right now. Also: protoc is now at 3.19, so we’re behind the times for both core and UI.

ui/README.md
```diff
@@ -85,12 +85,14 @@ if you've made API changes in `/internal/server` and want to use those on the fr
...
+- Install `ts-protoc-gen`: `yarn global add ts-protoc-gen` or `npm i -g ts-protoc-gen`
+- Install `protoc-gen-grpc-web`: `brew install protoc-gen-grpc-web`
 
 #### Generate the API definitions
 
-- `go generate ./internal/server`
+- `make docker/tools`
+- `make docker/gen/server`
```
jgwhite, line 95: The core generation logic uses protoc 3.15, which is captured in the docker version of these commands. Note that currently these docker images don’t work on Apple silicon.

ui/yarn.lock
```diff
@@ -3947,7 +3947,8 @@ anymatch@~3.1.1:
     picomatch "^2.0.4"
 
 "api-common-protos@link:./lib/api-common-protos":
-  version "1.0.0"
+  version "0.0.0"
+  uid ""
```
jgwhite, line 3951: I’m still ensure why Yarn does this. It seems to me that `version "1.0.0"` is more correct. 🤷‍♀️

izaaklauer (approved):

PR #3122, opened 2022-03-22
[RelAPI Onboarding] Add release API metadata file

mdeggies:
👋  This PR adds a `.release/release-metadata.hcl` file to the repo. This contains static metadata that will be processed and sent as part of the payload in RelAPI POST requests, which will be sent when staging and production releases are triggered.  

This can be merged now, but will not have any effect until after the RelAPI launch. Similar additions are being added across all projects that publish to releases.hashicorp.com.

### DIFF
--- a/.release/release-metadata.hcl
+++ b/.release/release-metadata.hcl
@@ -1,0 +1,5 @@
+url_docker_registry_dockerhub = "https://hub.docker.com/r/hashicorp/waypoint"
+url_docker_registry_ecr = "https://gallery.ecr.aws/hashicorp/waypoint"
+url_license = "https://github.com/hashicorp/waypoint/blob/main/LICENSE"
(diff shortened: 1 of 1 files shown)


briancain (approved):
Thanks @mdeggies ! Should we backport this to any other release branches, or are we good to keep this on `main` going forward? We're currently on the 0.7.x release cycle, so maybe just that one?

mdeggies:
Thanks for the review, Brian! Yep, we'll need to backport to all active release branches. I just added the 0.7.x label, but lmk if there are others to add.

briancain:
Sounds great to me! If we're all good I can go ahead and merge @mdeggies 👍🏻

mdeggies:
Sounds good.Thanks again!

PR #3138, opened 2022-03-24
Output variable values used and obfuscate sensitive values

krantzinator:
This PR adds support for obfuscating variables set to `sensitive` in the waypoint.hcl, and also supports displaying the final variable values in CLI output and logs, as well as the UI (though this part is only supported and not yet implemented).

This will allow users to see the used variable values upon the completion of a job, as well as enable us to add support for actually looking up which values were used on a given build/deploy/release.
This also enables future improvements such as a "noop" dry-run that returns the values that would be used if an operation was completed.

`sensitive` values are called `sensitive` to match what Terraform and Packer call them, to give consistency among our HashiCorp products. Unique to Waypoint though is the obfuscation of the values by using SHA256 encoding. This is not meant to be a security mechanism -- the values are stored on the server, which must be secured by the user -- but it does serve as a way to both obfuscate the values in logging and other outputs, while still enabling a user to verify that the value they expected the job to run with was the final value used.

The final values are stored as Variable_Refs so as to separate the values set on the server (when a user saves a value in the UI) and the final set of values used. This also enables an easier format for easily translating to logs and outputs.

The `Sensitive` item is added to the `Variable` message to support further communication between the UI and the variables set in the waypoint.hcl as part of future improvements.

To see it in action, you can use the below waypoint.hcl on our waypoint-examples/docker/go project:
```
project = "example-go"

app "example-go" {
  labels = {
    "service" = "example-go",
    "env"     = "dev"
  }

  build {
    use "pack" {}

    registry {
      use "docker" {
        image = "rae/go"
        tag   = var.tag
        local = var.local
      }
    }
  }

  deploy {
    use "docker" {}
  }
}

variable "tag" {
  type = string
  sensitive = true
  default = "hello"
}

variable "local" {
  default = true
}
```

To see how complex HCL types come through, replace the variable definition and usage of `tag` with:
```
...
        tag   = var.tag[0]
...
variable "tag" {
        type       = list(string)
        default   = ["hello", "othertag", "moretag", "whatifi", "havelotsof", "tagsand things"]
}
```

A couple of screenshots showing some example outputs:
Screen Shot 2022-03-23 at 9 51 38 PM

An example of the line breaks I was going for; without the manual line break intervention, the use of variable values longer than the terminal width severely messed up the column title widths:
Screen Shot 2022-03-23 at 9 53 05 PM


internal/cli/artifact_build.go
```diff
@@ -41,6 +42,46 @@ func (c *ArtifactBuildCommand) Run(args []string) int {
...
+				buildResult.Build.VariableRefs[iv].Type,
+				buildResult.Build.VariableRefs[iv].Source,
+			}
+			tbl.Rich(
+				columns,
+				[]string{
+					terminal.Green,
+				},
+			)
+		}
```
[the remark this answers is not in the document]
briancain: Additionally, I'd love for the `Type` column here to call out when a variable is sensitive. The SHA is a good indicator, but being explicit will be very helpful too:

```
» Variables used:
     VARIABLE    |                     VALUE                      |  TYPE  | SOURCE
-----------------+------------------------------------------------+--------+----------
  registry_local | 7c9863a97af3acabf62da25c6e6f1b82483d635afad14b | sensitive,bool   | cli
                 | 0e5e445266420fa0d6                             |                  |

```

My formatting is off but you get the idea 😄

briancain:
Oops, I hit submit too fast! I also ran into a weird thing where this string input var was right aligned? Very strange, not sure what's going on there 🤔 

```
» Variables used:
     VARIABLE    |         VALUE         |  TYPE  | SOURCE
-----------------+-----------------------+--------+----------
  image          | localhost:5000/tetris | string | cli
  namespace      |                       | string | default
  registry_local | false                 | bool   | cli
  release_port   |                  3000 | string | default
  tag            | latest                | string | default
```

The input var is defined as:

```
variable "release_port" {
  default     = "3000"
  type        = string
  description = "Port to open for the releaser."
}
```

internal/config/variables/variables.go
```diff
@@ -93,6 +98,8 @@ type Variable struct {
 	// declaration, the type of the default variable will be used.
 	Type cty.Type
 
+	Sensitive bool
```
briancain, line 101: This could use a doc comment!

internal/cli/artifact_build.go
```diff
@@ -41,6 +42,46 @@ func (c *ArtifactBuildCommand) Run(args []string) int {
...
+		sort.Strings(inputVars)
+		for _, iv := range inputVars {
+			// We add a line break in the value here because the Table word wrap
+			// alone can't accomodate the column headers to a long value
+			val := buildResult.Build.VariableRefs[iv].Value
+			if len(val) > 45 {
+				for i := range val {
+					// line break every 45 characters
+					if i%46 == 0 && i != 0 {
+						val = val[:i] + "\n" + val[i:]
```
xiaolin-ninja, line 66: As is, the first chunk would be 46 chars because of indexing.

```suggestion
						val = val[:i-1] + "\n" + val[i:]
```

internal/cli/artifact_build.go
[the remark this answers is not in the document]
xiaolin-ninja: iirc from when Rae & I talked about this, we don't want to explicitly say that it's a sensitive value to reduce risk.

internal/cli/deployment_create.go
```diff
@@ -118,6 +119,51 @@ func (c *DeploymentCreateCommand) Run(args []string) int {
...
+		sort.Strings(inputVars)
+		for _, iv := range inputVars {
+			// We add a line break in the value here because the Table word wrap
+			// alone can't accomodate the column headers to a long value
+			val := deployment.VariableRefs[iv].Value
+			if len(val) > 45 {
+				for i := range val {
+					// line break every 45 characters
+					if i%46 == 0 && i != 0 {
+						val = val[:i] + "\n" + val[i:]
```
xiaolin-ninja, line 147:
```suggestion
						val = val[:i-1] + "\n" + val[i:]
```

internal/cli/release_create.go
```diff
@@ -156,6 +157,49 @@ func (c *ReleaseCreateCommand) Run(args []string) int {
...
+		sort.Strings(inputVars)
+		for _, iv := range inputVars {
+			// We add a line break in the value here because the Table word wrap
+			// alone can't accomodate the column headers to a long value
+			val := result.Release.VariableRefs[iv].Value
+			if len(val) > 45 {
+				for i := range val {
+					// line break every 45 characters
+					if i%46 == 0 && i != 0 {
+						val = val[:i] + "\n" + val[i:]
```
xiaolin-ninja, line 183:
```suggestion
						val = val[:i-1] + "\n" + val[i:]
```

+1 on Brian's comment, would be great to extract this into a helper.

internal/cli/up.go
```diff
@@ -50,6 +51,51 @@ func (c *UpCommand) Run(args []string) int {
...
+		sort.Strings(inputVars)
+		for _, iv := range inputVars {
+			// We add a line break in the value here because the Table word wrap
+			// alone can't accomodate the column headers to a long value
+			val := result.Release.Release.VariableRefs[iv].Value
+			if len(val) > 45 {
+				for i := range val {
+					// line break every 45 characters
+					if i%46 == 0 && i != 0 {
+						val = val[:i] + "\n" + val[i:]
```
xiaolin-ninja, line 79:
```suggestion
						val = val[:i-1] + "\n" + val[i:]
```

internal/config/variables/variables.go
```diff
@@ -666,8 +680,71 @@ func EvaluateVariables(
...
+
+	jobVals, diags := getJobValues(vs, iv, salt)
+	if diags.HasErrors() {
+		return nil, nil, diags
+	}
+	return iv, jobVals, diags
+}
+
+// getJobValues combines the Variable and Value into a VariableRef,
+// encoding any 'sensitive' values as SHA256 values with the given salt.
```
xiaolin-ninja, line 696:
```suggestion
// hashing any 'sensitive' values as SHA256 values with the given salt.
```
Sorry super nit, "encoding" & "hashing" aren't the same. Encoding is reversible and not a security mechanism, hashing is a security tool and irreversible.

internal/config/variables/variables.go
```diff
@@ -666,8 +680,71 @@ func EvaluateVariables(
...
+			val = buf.String()
+			t = "complex"
+		}
+
+		if vs[v].Sensitive {
+			// salt shaker
+			saltedVal := salt + val
+			h := sha256.Sum256([]byte(saltedVal))
+			hVal := h[:]
+			val = hex.EncodeToString(hVal)
```
xiaolin-ninja, line 737:
```suggestion
			val = hex.EncodeToString(h[:])
```
I'd skip defining a single-use value

website/content/docs/waypoint-hcl/variables/input.mdx
```diff
@@ -172,6 +176,18 @@ variable "image_tag" {
...
+
+[inpage-sensitive]: #sensitive-values
+
+If set to true, the variable value will be obfuscated in outputs. This should not
+be relied on as a security measure on its own, but can be used as a way to hide
+literal variable values from logging, CLI output, and the UI. Variable values
+shared with the user via these methods will generate a SHA256 encoded value using
+the server cookie as the salt along with the set value. Therefore the user running an
+operation can verify if the value they expected to be used was used, with the output
+SHA256 encoded value.
```
xiaolin-ninja, line 189:
```suggestion
shared with the user via these methods will generate a SHA256 hashed value using
the server cookie as the salt along with the set value. Therefore, the user running an
operation can verify whether the expected value was used by verifying the outputted
SHA256 hashed value.
```

internal/cli/artifact_build.go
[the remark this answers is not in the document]
xiaolin-ninja: iirc from when Rae & I talked about this, we don't want to explicitly say that it's a sensitive value to reduce risk.

pkg/server/proto/server.proto
```diff
@@ -619,6 +623,12 @@ message Variable {
     HclPos start = 2;
     HclPos end = 3;
   }
+
+  message Ref {
```
mitchellh, line 627: Nitpick on name: `Ref` to me usually means a database-style foreign key. That's how I've thought of it when we used refs in Waypoint. I feel like this should be something else, like `Value`.

pkg/server/proto/server.proto
```diff
@@ -3165,6 +3194,12 @@ message Deployment {
   // Resources that this deployment has created or manages.
   repeated DeclaredResource declared_resources = 19;
 
+  // Variable refs store the final value used on the operation for each variable
+  // defined in the waypoint.hcl. Any variables with `sensitive` set in the
+  // waypoint.hcl will have a value encoded as SHA256 so the user can verify
+  // the value used.
+  map<string, Variable.Ref> variable_refs = 21;
```
mitchellh, line 3201: I don't think we need these on operations, because the operation contains a reference to the job (`job_id`), and you can get the variables from the job. For APIs, we can eagerly do that and put it on the response perhaps. But I think no reason to have it here?

pkg/server/proto/server.proto
mitchellh: Oh, I think I see, you named this this because of `DataSource.Ref`. I named that `Ref` because of Git refs. Thats what git calls the format used to point to a commit. I think variable should still be `Value` or something like that.

pkg/server/proto/server.proto
```diff
@@ -1541,6 +1552,12 @@ message Job {
   // exposed via the data_source parameter itself.
   DataSource.Ref data_source_ref = 110;
 
+  // Variable refs store the final value used on the operation for each variable
+  // defined in the waypoint.hcl. Any variables with `sensitive` set in the
+  // waypoint.hcl will have a value encoded as SHA256 so the user can verify
+  // the value used.
+  map<string, Variable.Ref> variable_refs = 113;
```
mitchellh, line 1559: Nitpick on field name, maybe `variable_final_values`

pkg/server/proto/server.proto
```diff
@@ -619,6 +623,12 @@ message Variable {
     HclPos start = 2;
     HclPos end = 3;
   }
+
+  message Ref {
+    string value = 2;
+    string type = 3;
+    string source = 4;
```
mitchellh, line 630: String types spook me out. Maybe they're right, but here to just challenge the idea.

* Can `source` be an enum?
* I think you can encode the type in the type system, the same way it is for `Variable`. Example in the previous comment.

pkg/server/proto/server.proto
```diff
@@ -619,6 +623,12 @@ message Variable {
     HclPos start = 2;
     HclPos end = 3;
   }
+
+  message Ref {
+    string value = 2;
```
mitchellh, line 628: It feels like this message should be a oneof...

I feel like this is slightly more analogous to `message Variable` too.

```proto
message Ref {
  Source source = 1;

  enum Source {
    // TODO: comments
    UNKNOWN = 0;
    FILE = 1;
    // etc...
  }

  oneof value {
    // sensitive values just have a salted hash of the value. If you have the
    // secret value, you can SHA-256 hash it with the server cookie as a
    // salt to verify this value matched.
    string sensitive = 1;

    // The raw value of this variable. 
    string str = 1;
    bool bool = 1;
    int64 num = 1;
    string hcl = 1;
  }
}
```

mitchellh: Also, it seems like `sensitive` is duplicated here from `Variable` but I think they serve two diff purposes:

* The `sensitive` on `Variable` notes that that variable is _currently_ sensitive. This can change at any point from false to true or true to false.
* The `sensitive` on `Ref` denotes the variable was sensitive _at the time_ the value was rendered, and also uses the protobuf oneof and type system to make it clear the value is the hash (versus the `string str` field). 
  - If at a future point, a variable is updating to no longer be sensitive, past rendered values are still sensitive. So you can't rely on the `Variable.sensitive` field to let you know about this. You have to encode it in the `Ref`.

internal/runner/operation.go
```diff
@@ -248,6 +260,7 @@ func (r *Runner) executeJob(
 		core.WithDataDir(projDir),
 		core.WithLabels(job.Labels),
 		core.WithVariables(inputVars),
+		core.WithVariableRefs(jobVars),
```
mitchellh, line 263: If you remove variable refs from operations as I suggested, then you can nix this whole code path plus everything in `core`.

internal/runner/operation.go
```diff
@@ -225,8 +226,19 @@ func (r *Runner) executeJob(
...
+	clientResp, _ := r.client.GetServerConfig(context.Background(), &empty.Empty{})
+	var serverCookie string
+	if clientResp != nil && clientResp.Config != nil {
+		serverCookie = clientResp.Config.Cookie
+	}
+	// We set both inputVars and jobVars on the project.
+	// inputVars is the set of cty.Values to use in our hcl evaluation
+	// and jobVars is the matching set of variable refs to store on the job that
+	// has sensitive values obfuscated and is used for user-facing feedback/output.
+	inputVars, jobVars, diags := variables.EvaluateVariables(log, pbVars, cfg.InputVariables, serverCookie)
```
mitchellh, line 241: Pairing with the next comment, if you remove refs from operations, I'd update the job here with the rendered variables. This includes updating the server that the variables are now available... this might require a new event.

mitchellh (changes requested):
I didn't do a _full_ line by line review because I think there is enough in my feedback. 

The core logic here feels really solid. I have some important feedback on the structure of the data and where it lives. But the actual core encoding of the values and storage generally is great.

internal/cli/artifact_build.go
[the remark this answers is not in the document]
mitchellh: > iirc from when Rae & I talked about this, we don't want to explicitly say that it's a sensitive value to reduce risk.

I think we should encode this information in the API (see my proto proposed changes). Terraform shows when a variable is sensitive or not. 

And I can't think of any threat model reason to not share this information. I know we don't have a threat model right now, but I can't think of anything hiding this information protects against. The protection itself is based on SHA-256 being safe, knowing it is a SHA-256 doesn't give you any head starts. In my opinion, it can also be trivially assumed to be sensitive when the value is obviously a SHA-256 hash and the value is non-sensical for the variable name.

internal/config/variables/variables.go
krantzinator: Doh, missed adding the TODO :D

internal/cli/artifact_build.go
krantzinator: I'm just going to make this `i%45`; the indexing 😵 got me but that's what I want -- up to the first 45 chars inclusive (index 0-44), and then the next set exclusive ([45:].

xiaolin-ninja: Ah, I tested this and if you do i%45 it doesn't crop right! The first chunk ends up as 45 chars but every subsequent chunk is only 44. @krantzinator

xiaolin-ninja:
Just in case this comment got lost in the rebasing, for the manual line break logic, it needs to be
```
if i%46 == 0 && i != 0 {
  val = val[:i-1] + "\n" + val[i:]
  ...
}
```
and not 
``````
if i%45 == 0 && i != 0 {
  val = val[:i] + "\n" + val[i:]
  ...
}
```
Because otherwise the first chunk will be alright at 45 chars, but subsequent chunks will be 44 chars.

briancain (approved):

briancain:
The failing test is weird, but otherwise this looks good with my feedback! 👍🏻

internal/cli/up.go
```diff
@@ -247,3 +264,38 @@ Usage: waypoint up [options]
...
+		// We add a line break in the value here because the Table word wrap
+		// alone can't accomodate the column headers to a long value
+		if len(v.Value) > 45 {
+			for i := range v.Value {
+				// line break every 45 characters
+				if i%46 == 0 && i != 0 {
+					v.Value = v.Value[:i] + "\n" + v.Value[i:]
+				}
+			}
+		}
```
xiaolin-ninja, line 286: [See my other comment](https://github.com/hashicorp/waypoint/pull/3138#issuecomment-1079226152)

This code crops the first chunk as 46 chars, and every subsequent chunk as 45 chars. This code below should work to crop every chunk as 45 chars:

```suggestion
	if len(v.Value) > 45 {
		for i := 45; i < len(v.Value); i += 46 {
			v.Value = v.Value[:i] + "\n" + v.Value[i:]
		}
	}
```

pkg/server/gen/server.swagger.json
```diff
@@ -5967,6 +5967,13 @@
           "$ref": "#/definitions/hashicorp.waypoint.Job.DataSource.Ref",
           "description": "Ref of the data was fetched for this job. This is available after\nthe Ref event is sent down by GetJobStream. This is NOT used to specify\nthe ref that should be downloaded. That level of configuration should be\nexposed via the data_source parameter itself."
         },
+        "variable_final_values": {
+          "type": "object",
+          "additionalProperties": {
+            "$ref": "#/definitions/hashicorp.waypoint.Variable.FinalValue"
+          },
+          "description": "Variable refs store the final value used on the operation for each variable\ndefined in the waypoint.hcl. Any variables with `sensitive` set in the\nwaypoint.hcl will have a value encoded as SHA256 so the user can verify\nthe value used."
```
xiaolin-ninja, line 5975:
```suggestion
          "description": "Variable refs store the final value used on the operation for each variable\ndefined in the waypoint.hcl. Any variables with `sensitive` set in the\nwaypoint.hcl will have a value hashed with SHA256 so the user can verify\nthe value used."
```

pkg/server/proto/server.proto
```diff
@@ -1541,6 +1569,12 @@ message Job {
   // exposed via the data_source parameter itself.
   DataSource.Ref data_source_ref = 110;
 
+  // Variable refs store the final value used on the operation for each variable
+  // defined in the waypoint.hcl. Any variables with `sensitive` set in the
+  // waypoint.hcl will have a value encoded as SHA256 so the user can verify
```
xiaolin-ninja, line 1574:
```suggestion
  // waypoint.hcl will have a value hashed with SHA256 so the user can verify
```

xiaolin-ninja (changes requested):
Looks great! I added a suggestion for the manual new-line algorithm, & some more nits about using "hash" vs "encode" :)

pkg/server/proto/server.proto
```diff
@@ -619,6 +623,32 @@ message Variable {
...
+      string str = 2;
+      bool bool = 3;
+      int64 num = 4;
+      // hcl stores values of any complex type in a raw string format
+      string hcl = 5;
+    }
+
+    Source source = 6;
+    enum Source {
+      // TODO: comments
```
mitchellh, line 641: There is a TODO here

pkg/server/proto/server.proto
```diff
@@ -619,6 +623,32 @@ message Variable {
     HclPos start = 2;
     HclPos end = 3;
   }
+
+  message FinalValue {
+    // value stores one of three simple types (string, number, bool) as their
+    // matching protobuf format, or any complex hcl type as a raw string.
+    oneof value {
+      string sensitive = 1;
```
mitchellh, line 631: Please comment on the format of this

pkg/server/proto/server.proto
xiaolin-ninja: @krantzinator missed one!

mitchellh (commented):
Everything looks really good, I'm looking at that test failure now.

xiaolin-ninja (approved):

mitchellh (approved):

PR #3140, opened 2022-03-24
update actual version constraints

krantzinator:


gregone (approved):
Looks like [this is the way forward](https://github.com/ember-cli/ember-new-output/blob/master/package.json#L65). Thanks @krantzinator !

PR #3141, opened 2022-03-24
pkg/server: fix some copy-pasta in `server.proto`

jgwhite:


briancain (approved):
👍🏻 We should generate the protos here too so the generated Go docs are fixed too!

briancain (approved):
Looks good!

PR #3149, opened 2022-03-25
Consumable service handlers

izaaklauer:
## What does this solve?

This is an architectural change to allow other projects to import and embed the singleprocess waypoint server. 

## What changed?
More than I would have liked for one PR, but here we are.

Broadly: 
- Moves the singleprocess package from `internal` to `pkg`, allowing it to be imported
- Adds a number of new singleprocess server start options, allowing it to be customized more heavily for use by other projects. 
- Moves the `state` package out of `singleprocess` and into its own internal packate, `boltdbstate`. Other projects shouldn't need this, so I didn't move it out of `internal`.

This PR should have no functional changes - it only moves existing logic.

## Future work
The singleprocess server still needs more configurability around authentication flows and job streaming. 

Closes https://github.com/hashicorp/waypoint/pull/3105, which a different approach to solve the same issue


pkg/server/singleprocess/auth.go
```diff
@@ -133,20 +133,26 @@ func (s *service) cookieFromRequest(ctx context.Context) string {
 	return ""
 }
 
+func (s *Service) AuthenticateCookie(
+	ctx context.Context, token, endpoint string, effects []string,
+) {
+
+}
```
evanphx, line 140: This doesn't appear to be used anywhere, perhaps left over?

evanphx (approved):
Just a comment about a method that appears to not be used, otherwise looks great!

pkg/server/singleprocess/auth.go
izaaklauer: Good catch, thank you!

pkg/server/singleprocess/auth.go
```diff
@@ -123,7 +123,7 @@ func (s *service) tokenFromContext(ctx context.Context) *pb.Token {
 
 // cookieFromRequest returns the server cookie value provided during the request,
 // or blank if none (or a blank cookie) is provided.
-func (s *service) cookieFromRequest(ctx context.Context) string {
+func CookieFromRequest(ctx context.Context) string {
```
briancain, line 126: What's the reason for exporting this instead of keeping it a func pointer? 🤔

pkg/server/singleprocess/service.go
```diff
@@ -0,0 +1,359 @@
...
+		conf, err := state.ServerConfigGet()
+		if err != nil {
+			return nil, err
+		}
+		if conf.Cookie == "" {
+			if err := state.ServerConfigSet(conf); err != nil {
+				return nil, err
+			}
+		}
+	}
```
briancain, line 143: Not sure if this code was copied in the move but we should probably handle the not `ok` case here? Return an `codes.FailedPrecondition` error?

pkg/server/singleprocess/service.go
```diff
@@ -0,0 +1,359 @@
...
+			return nil, err
+		}
+	}
+
+	// Setup background polling if not disabled
+	if !cfg.pollingDisabled {
+		// Setup the background context that is used for internal tasks
+		s.bgCtx, s.bgCtxCancel = context.WithCancel(context.Background())
+
+		// TODO: When more items are added, move this else where
```
briancain, line 196: We could remove this todo, seems ok to keep this here in the startup for now 🤔

briancain (approved):
Looks good, lots of changes! 😬 🤞🏻 I only had a couple of minor questions....

pkg/server/singleprocess/auth.go
izaaklauer: Other potential waypoint server implementations might not want to reinvent the process of cookie extraction! This function was functionally static anyway (it didn't rely on that pointer receiver at all), so removing that and exporting it makes it easier to consume.

pkg/server/singleprocess/service.go
izaaklauer: This was a bit of a hack - inside this `if` is where I was hiding the logic that was only relevant to the singleprocess server, but didn't have good configuration flags for.

After thinking about it a bit, I managed to find better homes for this logic.

PR #3150, opened 2022-03-25
cli: Filter job list on workspace, project, app, job state, or jobs target runners.

briancain:
No changelog because `waypoint job list` is not released yet!

edit: Update to this PR...job filtering happens now by request on the server. This means the client won't have to ingest the entire job list and filter it on its own, and the server won't have to send over the entire job list if the client only is requesting a subset of them.


internal/cli/job_list.go
```diff
@@ -42,6 +45,37 @@ func (c *JobListCommand) Run(args []string) int {
...
+				if pb.Job_State_name[int32(job.State)] == strings.ToUpper(state) {
+					found = true
+					break
+				}
+			}
+			if !found {
+				continue
+			}
+		}
+		tmpj = append(tmpj, job)
```
catsby, line 75: [server.proto](https://github.com/hashicorp/waypoint/blob/4ef4746c00a0531163c9f55bc60a5dd531b4d93f/pkg/server/proto/server.proto#L193-L195) mentions List Jobs is internal and not for public use yet. I assume that's more of a stability thing rather than just trying to keep it internal, so based on that would it be more appropriate here to modify [ListJobsRequest](https://github.com/hashicorp/waypoint/blob/4ef4746c00a0531163c9f55bc60a5dd531b4d93f/pkg/server/proto/server.proto#L2070) to contain ref fields for filtering by workspace, project, or application while we're here?

catsby (commented):
Looks fine but I wonder if we should expand ListJobsRequest to enable us to have this kind of filtering done for us on the server

internal/cli/job_list.go
briancain: Hmm, yes, I believe that was a stability choice early on in Waypoint! Maybe we should flip it to just be `ListJobs` instead of `XListJobs` then?

That makes sense to update this endpoint to accept filter requests 👍🏻  If we do filtering on the server, we should implement the filtering in the db state package and update `ListJobRequest` API to take workspace, project, app, job operation, and state `Ref`s to pass on through to the job list db query. That'll be a lot more work than this but that's probably the right way to do it 🤔

izaaklauer: Yup, server side filtering is the right way to implement this. That's where we're doing other filtering too, so I don't think it would be too much more work!

briancain: Done.

internal/cli/job_list.go
```diff
@@ -173,9 +235,32 @@ func (c *JobListCommand) Run(args []string) int {
...
 		f := set.NewSet("Command Options")
+		f.EnumVar(&flag.EnumVar{
+			Name:   "state",
+			Target: &c.flagState,
+			Values: jobStateValues,
+			Usage:  "List jobs that only match the requested state. Can be repeated multiple times.",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "target-runner",
```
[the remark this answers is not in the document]
briancain: Very good call! especially if that's how the other commands reference it 👍🏻

izaaklauer (approved):
This looks great!! With this filtering in place, I support making JobList not internal.

xiaolin-ninja (approved):

Issue #3157, opened 2022-03-29
UI/Server: UI does not surface latest project job failures


briancain:
**Describe the bug**

If a new project has an initial error, such as https://github.com/hashicorp/waypoint/issues/2454 where the `Init` operation fails to find git on the path, then the following operation `QueueProject` also fails, the UI never updates to say the latest failure in the project.

![Screenshot from 2022-03-29 11-16-30](https://user-images.githubusercontent.com/810277/160678598-07acfd54-efa7-47ea-83b4-5b470910be02.png)

```
brian@localghost:k8s λ waypoint job list                                                                                                                                                                                                                                                          ±[main]

» Waypoint Jobs
              ID             |  OPERATION   |  STATE  | TIME COMPLETED |       TARGET RUNNER        | WORKSPACE |      PROJECT       | APPLICATION
-----------------------------+--------------+---------+----------------+----------------------------+-----------+--------------------+--------------
  01FZBE7Q04J7ZRGCWKV3D9BC0D | StopTask     | Success | 16 seconds ago | *                          | default   | go-gitops-0        |
  01FZBE7PZY9FXYZ1VCN9J470BH | Poll         | Success | 16 seconds ago | 01FZBE7PZZJSC3CY8M74WYB9BA | default   | go-gitops-0        |
  01FZBE7Q04SKMXMREBSDV7K8CS | StartTask    | Success | 19 seconds ago | *                          | default   | go-gitops-0        |
  01FZBE6WMW5KMJTJGDHM9S5FQR | StopTask     | Success | 42 seconds ago | *                          | default   | go-gitops-0        |
  01FZBE6WMPRSBANGRFQ899B0HY | QueueProject | Error   | 42 seconds ago | 01FZBE6WMPMMXNACFGGG32YJ8P | default   | go-gitops-0        |
  01FZBE6SPAZDS7VP8YZ64Y7ZGE | StopTask     | Success | 46 seconds ago | *                          | default   | go-gitops-0        |
  01FZBE6WMWMZXE60S7JHAFVTC0 | StartTask    | Success | 46 seconds ago | *                          | default   | go-gitops-0        |
  01FZBE6SP2K27EB2SKBR3GR8K8 | Poll         | Success | 46 seconds ago | 01FZBE6SP2MX082NWJEBRZ2MEZ | default   | go-gitops-0        |
[...]

PR #3158, opened 2022-03-29
internal: More Descriptive SingletonIds for Poll Jobs

briancain:
Prior to this commit, our poll operation would set the poll jobs
singleton Id to Project/App/Workspace. However not all project
poll operations have an application, such as the initial QueueProject
operation. In this case, we only set Workspace and Project. This means
the singletonId for the Job included a double slash, such as:

```
poll-trigger/go-gitops-0//default
```

Where the application was not set for the job.

This commit fixes this by rearranging the singleton to start at
Workspace, Project, and then optionally Application if exists on the
job.

Discovered via https://github.com/hashicorp/waypoint/issues/3157

edit: This pull request also enhances the application status report poll singletonId to include the workspace and project it belongs to. Otherwise this would of been a job collision issue with multiple separate app names across projects in a single Waypoint server.

### DIFF
--- a/internal/runner/operation_poll.go
+++ b/internal/runner/operation_poll.go
@@ -103,6 +103,13 @@ func (r *Runner) executePollOp(
 		}
 	}
 
+	// NOTE(briancain): We set a singleton ID for a poll project operation to ensure that the
+	// poll handler does not fire off many operations of the same kind more than once,
+	// clogging up the job system. By setting a singleton ID that is unique to this
+	// application or project, we can ensure only 1 operation will be active at once rather than
+	// many operations (such as in the case where a poll interval is shorter than it
+	// takes to run the operation)
+
 	// We assume a project and workspace is set given this is Project polling
 	singletonId := strings.ToLower(fmt.Sprintf(
 		"poll-trigger/%s/%s",
--- a/internal/runner/operation_poll.go
+++ b/internal/runner/operation_poll.go
@@ -111,7 +111,7 @@ func (r *Runner) executePollOp(
 	))
 	// Not all jobs set an application
 	if job.Application.Application != "" {
-		singletonId = "/" + job.Application.Application
+		singletonId = "/" + strings.ToLower(job.Application.Application)
 	}
 
 	log.Debug("queueing job")
--- a/internal/runner/operation_poll.go
+++ b/internal/runner/operation_poll.go
@@ -103,18 +103,24 @@ func (r *Runner) executePollOp(
 		}
 	}
 
+	// We assume a project and workspace is set given this is Project polling
+	singletonId := strings.ToLower(fmt.Sprintf(
+		"poll-trigger/%s/%s",
+		job.Workspace.Workspace,
+		job.Application.Project,
+	))
+	// Not all jobs set an application
+	if job.Application.Application != "" {
+		singletonId = "/" + job.Application.Application
+	}
+
 	log.Debug("queueing job")
 	queueResp, err := r.client.QueueJob(ctx, &pb.QueueJobRequest{
 		Job: &pb.Job{
 			// We set a singleton ID to verify that we only setup
 			// a queue operation once (in case it is taking longer to
 			// process than the poll interval).
-			SingletonId: strings.ToLower(fmt.Sprintf(
-				"poll-trigger/%s/%s/%s",
-				job.Application.Project,
-				job.Application.Application,
-				job.Workspace.Workspace,
-			)),
+			SingletonId: singletonId,
 
 			// Target only our project, we don't need an app for this.
 			Application: &pb.Ref_Application{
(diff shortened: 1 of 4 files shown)


izaaklauer:
Is there a backwards-compat angle here where if someone upgrades their waypoint server with a poll job pending, the new server will fail that poll job because the format changed? Even if so, it's probably OK as long as it doesn't get stuck in some weird state when it does.

pkg/server/singleprocess/poll_application.go
```diff
@@ -240,7 +240,12 @@ const (
 )
 
 // appStatusPollSingletonId generates an application status polling job singleton ID
-// for the given app and operation type.
-func appStatusPollSingletonId(appName string, operationType appStatusPollOperationType) string {
-	return fmt.Sprintf("app-status-poll/%s/%s", appName, operationType)
+// for the given workspace, project, app and operation type.
+func appStatusPollSingletonId(
```
izaaklauer, line 244: At this point I kind of wish we had some had a structured message here - this is a whole lot of info to encode positionally into a string. Understand if that's a bigger harder refactor for now though.

briancain: Yeah, the main idea here for why I originally had a structured singleton ID is that the same poll handler per app shouldn't fire off more than 1 job at a time for status reporting. I think this is about as much info as we'd want in a singleton ID to keep poll jobs unique. Otherwise there's a bug here in Waypoint server where any app in separate projects would block on a separate job running given they'd have the same singleton id.

briancain:
> Is there a backwards-compat angle here where if someone upgrades their waypoint server with a poll job pending, the new server will fail that poll job because the format changed? Even if so, it's probably OK as long as it doesn't get stuck in some weird state when it does.

Hmm, nope, there shouldn't be any failures here around the singleton ID format. There's nothing currently relying on the singleton id format. The main purpose to introduce a singleton id for these jobs was to ensure a poller doesn't fire off more than 1 operation at a time per-application when building a status report poll or project poll. The job system will enforce a wait if it detects there's already a job running with the same singleton ID. So in the case of an upgrade mid poll operation, it should be safe for the ID to change here. Worse case is you get 2 Tasks that refresh a status 🙏🏻

pkg/server/singleprocess/poll_application.go
briancain: Added a note! The reasoning here might of been only in my head, so calling it out explicitly hopefully will clear things up for why we're doing this.

evanphx (approved):
This makes sense, though I agree with izaak that the encoding of these is getting weird. Maybe we should be using a `key=val,...` format instead of just presuming that each segment has a specific meaning.

briancain:
@evanphx - If we really want to make sure no one uses the singleton id for anything we could take the `workspace/project/app` string and SHA it to create `poll/4b1aaab2ab96f5d81f8481ade0cea7d8bb4194ac134a8fc1c610bed77270a388` from `workspace/project/app`. The main reason for the singleton id is just making sure these poll jobs are unique for their operating scope so the same poller doesn't fire off lots of duplicate jobs due to a short poll interval without the previous operations completing. Workspace, project, and app are already data on a job that people can access too.

evanphx:
I like having it be a human readable string because it makes debugging easier.

internal/runner/operation_poll.go
```diff
@@ -103,18 +103,31 @@ func (r *Runner) executePollOp(
...
+
+	// We assume a project and workspace is set given this is Project polling
+	singletonId := strings.ToLower(fmt.Sprintf(
+		"poll-trigger/%s/%s",
+		job.Workspace.Workspace,
+		job.Application.Project,
+	))
+	// Not all jobs set an application
+	if job.Application.Application != "" {
+		singletonId = "/" + strings.ToLower(job.Application.Application)
```
mitchellh, line 121: Should this be `+=`?

mitchellh:
> Is there a backwards-compat angle here where if someone upgrades their waypoint server with a poll job pending, the new server will fail that poll job because the format changed? Even if so, it's probably OK as long as it doesn't get stuck in some weird state when it does.

I think there is a scenario where one poll job could overlap with another, but its unlikely I think and would just be a momentary hiccup.

internal/runner/operation_poll.go
briancain: Omg yes, good catch....

mitchellh:
> This makes sense, though I agree with izaak that the encoding of these is getting weird. Maybe we should be using a `key=val,...` format instead of just presuming that each segment has a specific meaning.

We use the singleton ID as an easy way to look up conflicting jobs. 

We could make this a `map[string]string` to support a more label-like approach and then use `hashstructure` to create a unique hash of the map and use that for indexing purposes. This doesn't have to be user-visible. Right now we're overloading what is user-visible (the singleton ID) with its purpose (a unique lookup key). 

That's a deeper change and we should merge this regardless to start, but just noting for the future.

PR #3159, opened 2022-03-29
Test getting and setting all odr config fields

izaaklauer:
`reflect.DeepEqual` doesn't serve us here - the proto messages have weird internal fields that differ that we don't care about. If anyone knows a more efficient way to compare these structs i'm all ears.


evanphx (approved):

briancain:
> If anyone knows a more efficient way to compare these structs i'm all ears

This seems ok to me for now. Anything too fancy might hide other failures in our tests?

PR #3160, opened 2022-03-29
cli: Detail exactly why command failed for source-get

briancain:
Also, show the full help including command usage, rather than just flag
help.


briancain:
> assuming that using c.Help() instead of c.Flags().Help() is intentional?

Yep yep! So `c.Help()` will print the command usage and command docs, and then it also prints the flag output right after! `c.Flags().Help()` only prints the command flags.

PR #3166, opened 2022-03-31
ui: updated breadcrumbs to include current page

### LINKED ISSUE #2670: ui: Include current page in breadcrumbs
Currently the current page is not included in the breadcrumbs:

This is presumably because including the title in both the breadcrumbs *and* the `<h1>` would lead to duplication, and therefore visual noise. The project settings page, however, *does* include the current page in the breadcrumbs and it’s not so bad:

Personally, in practice I find the fact that the current page is included in the breadcrumbs to be confusing and often leads me to jump up on page higher in the hierarchy than I indented. I suspect this is due to expectations about breadcrumbs from other sites that use the breadcrumb pattern. When I canvased the rest of the frontend team, we all agreed that the current implementation can be confusing.

I’d like the current page to consistently appear in the breadcrumbs.

We could keep things as they are and make the project settings page fall in line with the others instead.

sabrinako:
## Other updates in this PR
- Removed icons to make more streamlined
- Included breadcrumbs for list views (applications, etc) and operation versions

## Additional notes
- Originally, there was no differentiation between `Applications` page and a project's page in breadcrumbs. Apps list is the first and only thing you see when selecting a project right now. But an `Applications` breadcrumb was added in the future case that there will most likely be other screens that are top-level to a project and not about specific applications. (The `Settings` page did have its own breadcrumb already, also adding to the argument to add the `Applications` breadcrumb)
- With removal of icons, the breadcrumb stating the list before a specific page (ie `Projects / marketing-public / Applications / web` we now know what object type `marketing-public` is and `web` is without relying on the icons as clues (removes some mental overhead for users to memorize what each icon correlates with).

## Screenshots
Screen Shot 2022-03-31 at 4 36 54 PM
Screen Shot 2022-03-31 at 4 37 04 PM
Screen Shot 2022-03-31 at 4 37 40 PM

### DIFF
--- a/.changelog/3166.txt
+++ b/.changelog/3166.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+ui: Updated UI of breadcrumbs and UX to include current page
+```
--- a/ui/app/routes/workspace/projects/project/app.ts
+++ b/ui/app/routes/workspace/projects/project/app.ts
@@ -39,7 +39,10 @@ export default class App extends Route {
     return [
       {
         label: model.application.project,
-        icon: 'folder',
+        route: 'workspace.projects.project.apps',
+      },
+      {
+        label: 'Applications',
         route: 'workspace.projects.project.apps',
       },
     ];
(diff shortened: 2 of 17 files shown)


jgwhite (approved):
Sorry it took me so long to review this!

I think it’s looking great and I definitely prefer it this way.

There’s one tiny remaining discrepancy:

image

What do you think? Should we address this now too?

sabrinako:
@jgwhite I addressed the change you mentioned above, I know you technically approved this PR already but just wanted to comment to note that it's been fixed 😄

PR #3169, opened 2022-03-31
Consumable authentication handlers

izaaklauer:
In much the same vein as https://github.com/hashicorp/waypoint/pull/3149, this refactoring allows other waypoint server implementations to consume the authentication logic here. This should not change any behavior or functionality.


pkg/server/singleprocess/auth.go
```diff
@@ -459,8 +473,15 @@ func (s *Service) GenerateLoginToken(
...
+	encodedId, err := s.encodeId(ctx, currentUser.Id)
+	if err != nil {
+		msg := "failed to encode id when generating a login token"
+		log.Error(msg, "currentUser.Id", currentUser.Id, "err", err)
+		return nil, status.Error(codes.Internal, msg)
+	}
+
 	login := &pb.Token_Login{
-		UserId: currentUser.Id,
+		UserId: encodedId,
```
izaaklauer, line 484: I'll note that I don't love this flow. It's not at all intuitive when IDs need to be encoded/decoded, and I don't have any great heuristics to offer. It will be easy for us to implement new logic in the future that skips doing this.

pkg/server/singleprocess/service.go
```diff
@@ -141,16 +141,18 @@ func New(opts ...Option) (pb.WaypointServer, error) {
...
 
-	// If we haven't initialized our server config before, do that once.
-	conf, err := state.ServerConfigGet()
-	if err != nil {
-		return nil, err
-	}
-	if conf.Cookie == "" {
-		err := state.ServerConfigSet(conf)
-		if err != nil && status.Convert(err).Code() != codes.Unimplemented {
+	if !cfg.skipServerConfigInit {
```
izaaklauer, line 144: This is a mostly unrelated change - it allows the server to be started without initializing a global cookie.

pkg/server/singleprocess/auth.go
```diff
@@ -653,7 +707,12 @@ func (s *Service) ConvertInviteToken(ctx context.Context, req *pb.ConvertInviteT
 		}
 
 		// Setup the login information for the new user
-		invite.Login.UserId = user.Id
+		invite.Login.UserId, err = s.encodeId(ctx, user.Id)
+		if err != nil {
+			msg := "failed to the current user's id when converting an invite token"
```
xiaolin-ninja, line 712:
```suggestion
			msg := "failed to encode the current user's id when converting an invite token"
```

evanphx (approved):

PR #3173, opened 2022-04-04
Fixing `make tools`

izaaklauer:
Turns out, leading whitespace is not allowed for go generate commands.


briancain (approved):
Ooooo, this must be why I couldn't get `protoc-gen-grpc-gateway` to install last time I tried after the refactor 😄

PR #3174, opened 2022-04-04
Updating to grpc-gateway v2

izaaklauer:
Note: OSS waypoint doesn't currently depend on any of the gateway generated code, so this change is not risky.


tools/tools.go
```diff
@@ -25,8 +25,8 @@ import (
 //go:generate go install github.com/evanphx/grpc-gateway/protoc-gen-swagger
 	_ "github.com/evanphx/grpc-gateway/protoc-gen-swagger"
 
-//go:generate go install github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
-	_ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway"
+//go:generate go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+	_ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway"
```
izaaklauer, line 29: This is the only thing that actually changed - everything else is the result of `go mod tidy` and `make gen/server`

evanphx (approved):

briancain (approved):

PR #3182, opened 2022-04-06
ci: Only notify PR check failures if `website/content` has changes

briancain:
Prior to this commit, any changes at all from running `make
gen/website-mdx` would notify a failure. This is extra tricky because
npm and lock files might change which causes `git diff` to notify the PR
that a failure occured. In reality, this check was only to tell the PR
author that they needed to run `make gen/website-mdx` to generate CLI
docs automatically. This commit updates that behavior to only check a
`git diff` in the `website/content` folder, where the auto generated
docs live.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,7 +106,7 @@ jobs:
       - run:
           command: cd ./website; npx --no-install next-hashicorp format
       - run: |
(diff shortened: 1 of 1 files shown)


BRKalow (approved):
Thanks!

PR #3185, opened 2022-04-06
cli: Sort uncomplete jobs as "latest"

briancain:
Prior to this commit, when the job list CLI would sort on Complete time,
jobs with a nil complete time would be sorted as "oldest". Really as a
user you would expect these to show up as "latest" with no complete time
field set. This commit fixes the original behavior to make these jobs
show up at the beginning when there is no complete time set.


briancain:
This should go out with the 0.8.0 release so I put it in that milestone!

evanphx (approved):

Issue #3184, opened 2022-04-06
Feature Request: Support for AWS Lambda Function URLs


jason-riddle:
**Is your feature request related to a problem? Please describe.**
As of today, deploying an application to AWS Lambda via Waypoint requires using an AWS ALB. This is great, but can be costly depending on your traffic patterns.

**Describe the solution you'd like**
AWS just announced support for [AWS Lambda Function URLs](https://aws.amazon.com/blogs/aws/announcing-aws-lambda-function-urls-built-in-https-endpoints-for-single-function-microservices/). Note that this is different from API Gateway (which I've opened a separate ticket for here: https://github.com/hashicorp/waypoint/issues/3183).

As far as a price comparison, from the blog post it mentions that

> Function URLs are included in Lambda’s request and duration pricing.

So I interpreted that as there is no additional cost to use this feature. This can be huge if you don't need the added functionality of an ALB or an API Gateway and you are just looking to keep costs as low as possible.

**Describe alternatives you've considered**
Right now, only ALB is supported as an endpoint. So there is no other alternative.

**Explain any additional use-cases**
This would mainly be for proof of concepts. It's not clear from the docs how production ready this solution is.

**Additional context**
None.

briancain:
Hey @jason-riddle ! Looks like @thiskevinwang is currently working on this in a pull request. Thanks for the feature request!

PR #3187, opened 2022-04-07
feat(aws-lambda): add URL releaser component

thiskevinwang:
# Description

This is adds a `lambda-function-url` plugin (with Releaser component only) that leverages the new Lambda Function URLs.

This approach only releases a function URL for the unpublished Lambda function, and does not leverage URLs for aliases of specific function versions

> Note, from AWS:
> 
> You can add a function URL only to the unpublished function version, or to an alias. To configure a function URL, go to the [unpublished function page](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/go-gin?tab=configure), or choose an alias from the [Aliases tab](https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/go-gin?tab=aliases).

This leaves the CORs configuration as a TODO for the near future.

Closes https://github.com/hashicorp/waypoint/issues/3184

### Usage

```hcl
#...
  release {
    use "lambda-function-url" {
    }
  }
}
```

### Screenshots

image

Waypoint UI

image


builtin/aws/lambda/lambda.go
```diff
@@ -5,10 +5,11 @@ import (
 	sdk "github.com/hashicorp/waypoint-plugin-sdk"
 )
 
-//go:generate protoc -I ../../../.. --go_opt=plugins=grpc --go_out=../../../.. waypoint/builtin/aws/lambda/plugin.proto
+//go:generate protoc -I ../../../.. -I ../../../thirdparty/proto/opaqueany --go_out=../../../.. waypoint/builtin/aws/lambda/plugin.proto
```
thiskevinwang, line 8: I didn't touch the `waypoint` repo for sometime, and this `go:generate` directive stopped working for me.

Running this...
```
//go:generate protoc -I ../../../.. --go_opt=plugins=grpc --go_out=../../../.. waypoint/builtin/aws/lambda/plugin.proto
```
...outputted this
```console
[Info  - 9:42:00 AM] 2022/04/07 09:42:00 protoc -I ../../../.. --go_opt=plugins=grpc --go_out=../../../.. waypoint/builtin/aws/lambda/plugin.proto

	operation="generate"

[Info  - 9:42:00 AM] 2022/04/07 09:42:00 --go_out: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC

See https://grpc.io/docs/languages/go/quickstart/#regenerate-grpc-code for more information.

	operation="generate"

[Info  - 9:42:00 AM] 2022/04/07 09:42:00 lambda.go:10: running "protoc": exit status 1

	operation="generate"

[Error - 9:42:00 AM] Request workspace/executeCommand failed.
  Message: exit status 1
  Code: 0 
[Error - 9:42:00 AM] 2022/04/07 09:42:00 command error: exit status 1

[Error - 9:42:20 AM] Request textDocument/definition failed.
  Message: no identifier found
  Code: 0 
[Error - 9:42:20 AM] Request textDocument/definition failed.
  Message: no identifier found
  Code: 0 
```

builtin/aws/lambda/plugin.proto
```diff
@@ -4,6 +4,8 @@ package aws.lambda;
 
 option go_package = "waypoint/builtin/aws/lambda";
 
+// import "opaqueany/any.proto";
```
thiskevinwang, line 7: Todo: figure out:

image

builtin/aws/lambda/platform.go
```diff
@@ -677,6 +677,32 @@ func (p *Platform) Destroy(
...
+					}
+					st.Step(terminal.StatusOK, "Deleted alias: "+"Alias_"+deployment.Version)
+					st.Update("Attempting to delete again...")
+					_, err = lamSvc.DeleteFunction(&lambda.DeleteFunctionInput{
+						FunctionName: aws.String(deployment.FuncArn),
+						Qualifier:    aws.String(deployment.Version),
+					})
+				}
+			}
+		}
```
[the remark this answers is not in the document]
briancain: Agreed, let me know if you have any questions on this! Steps can be a little tricky at the moment to look right.

builtin/aws/lambda/releaser.go
```diff
@@ -0,0 +1,323 @@
...
+	ctx context.Context,
+	log hclog.Logger,
+	release *Release,
+	ui terminal.UI,
+) (*sdk.StatusReport, error) {
+	var report sdk.StatusReport
+	report.External = true
+	defer func() {
+		report.GeneratedTime = timestamppb.Now()
+	}()
```
briancain, line 229: This would be a good thing to double check on - but I believe the SDK will generate the time for you. But I'm not sure!

builtin/aws/lambda/releaser.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (r *Releaser) resourceFunctionUrlCreate(
+	ctx context.Context,
+	log hclog.Logger,
+	sess *session.Session,
+	sg terminal.StepGroup,
+	ui terminal.UI,
+	dep *Deployment,
+) error {
```
briancain, line 77: We should include StepGroup output here too. If you look at the other plugins, you'll get a good sense of the kind of wording you'd want to use.

briancain (commented):
Looking good Kevin. I know this is still in draft, and it looks like there's some todos, but figured I'd drop a few comments as you work on it. Excited to see this land! 🎉

builtin/aws/lambda/releaser.go
thiskevinwang: I copied the `Status` implementation from [`aws-alb`](https://github.com/hashicorp/waypoint/blob/6dd2723d0dd4790eafb27facd6e94ed2f2a18cfa/builtin/aws/alb/releaser.go#L782) 😅 , and also hadn't yet read through the [docs](https://www.waypointproject.io/docs/extending-waypoint/plugin-interfaces/status).

It looks like several other built-in plugins explicitly pass `timestamppb.Now(),` to the status report
- [`nomad/platform`](https://github.com/hashicorp/waypoint/blob/6dd2723d0dd4790eafb27facd6e94ed2f2a18cfa/builtin/nomad/platform.go#L371)
- [`k8s/releaser`](https://github.com/hashicorp/waypoint/blob/0119dccca2a19a0b242d8d2db9b7aa70d2e38069/builtin/k8s/releaser.go#L856)
- [`nomad/jobspec/releaser`](https://github.com/hashicorp/waypoint/blob/b6e78045a94d2ca7ab5703b1ee35405fd0c25be3/builtin/nomad/jobspec/releaser.go#L418)

I'm also looking at [`waypoint-plug-sdk`](https://github.com/hashicorp/waypoint-plugin-sdk/blob/main/framework/resource/manager.go#L446) code for the first time, and not entirely sure if this line means that there is a default value for `GeneratedTime ` 🤔

briancain: Yep totally.... I believe we added automatic timestamp generation in the SDK after those plugins were written so I think it's not required anymore? https://github.com/hashicorp/waypoint-plugin-sdk/blob/3c2ebd53709006279ffd89272b6e4cbb76e71228/framework/resource/manager.go#L444-L451

Those other plugins might need to be updated. I'd give it a shot and check the reported generated time to confirm it works. If so we could make a PR to fix up those other plugins!

thiskevinwang: Yea according the UI, `GeneratedTime` looks like it doesn't need to be set. I'll go ahead and remove it here
image

thiskevinwang: removed: https://github.com/hashicorp/waypoint/pull/3187/commits/0bda64cd1d5f9e95756d12a1fb013efd337b7552

builtin/aws/lambda/releaser.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	// TODO(thiskevinwang): source cors from HCL config
+	cors := lambda.Cors{}
+
+	addPermissionInput := lambda.AddPermissionInput{
+		Action:              aws.String("lambda:InvokeFunctionUrl"),
+		FunctionUrlAuthType: aws.String(functionUrlAuthType),
+		FunctionName:        aws.String(dep.FuncArn),
+		Principal:           aws.String("*"),
```
briancain, line 97: Do you think we should make this configurable? We can have `Principal` default to `"*"` with the option to override in a config if necessary.

builtin/aws/lambda/releaser.go
```diff
@@ -0,0 +1,323 @@
...
+	s := sg.Add("Gathering health report for lambda url deployment: %q", release.Url)
+	defer s.Done()
+
+	report := sdk.StatusReport{
+		External: true,
+	}
+
+	// check the function url status
+	log.Info("Checking function url status...", "url", release.Url)
+	if resp, err := http.Get(release.Url); err != nil {
```
briancain, line 253: Is there an AWS SDK function here we could use instead to validate the creation/existence of a lamba url instead of http.Get?

builtin/aws/lambda/releaser.go
```diff
@@ -0,0 +1,323 @@
...
+		case http.StatusOK:
+			log.Info("Lambda URL returned a 200")
+			report.Health = sdk.StatusReport_READY
+			report.HealthMessage = "Lambda URL appears to be healthy"
+			break
+		default:
+			log.Error("Lambda URL returned a non-200 response: %d", resp.StatusCode)
+			report.Health = sdk.StatusReport_DOWN
+			report.HealthMessage = fmt.Sprintf("Lambda URL returned a non-200 response: %d", resp.StatusCode)
+			break
```
briancain, line 268: I don't thinkkkkk `break` is required in switch statements here 🤔

briancain (changes requested):
Overall looks pretty great! Just a few comments and requests from me 👍🏻

builtin/aws/lambda/releaser.go
thiskevinwang: Yea I think you're right — it should be configurable.

I think a `principal` config should only be applicable only when AuthType is `AWS_IAM` though (this looks like how the dashboard option behaves)

thiskevinwang: I also realized I need to rework the permission creation logic.

Lambda unfortunately doesn't have an `edit-permission` API, only an [`add-permission`](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/add-permission.html) and [`remove-permission`](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/remove-permission.html).

...The AWS console performs a _remove-then-add_ when you **edit** a permission.

I think the simplest approach is to use a fixed `statementId` and **always**...
- remove-permission
  - no-op if not-found [^1]
- add-permission
... on every `waypoint up`

[^1]: This accounts for the very first `waypoint up` where a permission doesn't exist yet. It also accounts for a state where a permission was removed, and the following add-operation fails, leave you with no permission.

Alternatively, we could [`get-policy`](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/get-policy.html) first to find & determine if permissions need to be created or updated at all... but this looks like it gets unnecessarily complex very quick

builtin/aws/lambda/releaser.go
thiskevinwang: Ah, TIL: `switch` `break` semantics differ slightly between [Go](https://go.dev/play/p/OkbAtjsgGWw) and [TypeScript](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAWwIYzACgJSIN4BQixiECAzlIpalCOYgLyIBMADG0SeQO4xQQAFokw065XIRLTSqcgFNWHAFxcZMsmHJwANvIB0OuAHNMAIgASMM9jUyAJvOCoQOqKvXrN2vYZPmAITg4GzsAXwIIgjQMHCA)

builtin/aws/lambda/releaser.go
thiskevinwang: I don't _think_ there's a _healthcheck-like_ API for Lambda 🤔. (Makes sense the its ephemeral nature)
- [lambda - available commands](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/index.html#available-commands)
- (looking for something like [elbv2::describe-target-health](https://docs.aws.amazon.com/cli/latest/reference/elbv2/describe-target-health.html))

1. I'm now realizing that `AWS_IAM` auth type would cause this unsigned http.Get to be return a `503`
Maybe 
2. Maybe `wait` or `invoke` are viable options...

builtin/aws/lambda/releaser.go
```diff
@@ -0,0 +1,323 @@
...
+
+	// Destroy All
+	return rm.DestroyAll(ctx, log, sg, ui, sess)
+}
+
+// ReleaserConfig is the configuration structure for the Releaser.
+type ReleaserConfig struct {
+	// "AWS_IAM" or "NONE"
+	AuthType string `hcl:"auth_type,optional"`
+	// todo(kevinwang): CORS
```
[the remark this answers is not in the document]
thiskevinwang: Yea, I was hoping to add `CORS` config as an _improvement_ in a follow up PR. Is that reasonable?

builtin/aws/lambda/releaser.go
thiskevinwang: Ok, I took a stab at the above logic (the "Alternatively..." approach) - https://github.com/hashicorp/waypoint/pull/3187/commits/5f9f3fdd5cc6ad6b4082fe1f2eb6cde6740c8841

At least with my local testing, it works quite well with the various permutations of `auth_type` and `principal`:D

```hcl
# full public access
auth_type = "NONE" // can be omitted
// principal       // must be omitted, otherwise validation error 

# any authenticated AWS user, from any account
auth_type = "AWS_IAM"
principal = "*" // can be omitted

# any principal under <AWS_ACCOUNT_ID>... that is also permitted to perform `lambda:InvokeFunctionUrl`
auth_type = "AWS_IAM"
principal = <AWS_ACCOUNT_ID>

```

briancain: This is a great question! I will bring this up in our next GitHub triage sync and let you know this week 👍🏻 Appreciate you taking the time to look into this and laying out the options. Makes it easier for us!

builtin/aws/lambda/releaser.go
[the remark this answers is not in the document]
briancain: Very reasonable! Sounds good. I'd say if you plan on getting to it immediately, we can leave this be and wait for the follow-up PR. If it might be a few weeks or so, I'd open a GitHub issue about it so we don't lose track of the work.

builtin/aws/lambda/releaser.go
briancain: Does the `lambda` `get-function` help at all? I don't know if it has any kind of health or status in the return response. Otherwise yeah maybe the way to go is an `http.Get`. Might be another good question to ask the team this week and see what they say.

thiskevinwang: `get-function` doesn't return any useful status information. I'll shoot a message to the Waypoint team slack for some opinions!

builtin/aws/lambda/releaser.go
briancain: Ok, we think the best course of action here is to make the default value for auth type to be `"NONE"`, and then let folks configure it and principal by request.

evanphx:
Hi @thiskevinwang, so we discussed how this plugin should be exposed and available. The issue with putting it into the main lambda plugin is that it will become a default releaser, meaning all users of this plugin will suddenly use it. Because there are like 20+ ways to invoke a lambda, it doesn't feel like like we should be setting up this one specific invoke functionality as the default releaser.

Ergo, I think the best solution is to actually move this to a new plugin that just has the release functionality on it. I'd probably call it `lambda-function-url` and have it live under `builtin/aws/lambda/function_url`. It's a bit more work to setup the new plugin, but not a bunch. The upside is that in a user's waypoint.hcl, they'll be able to do: 

```
app "blah" {
   release {
      use "lambda-function-url" {
           ....
      }
  }
}
```

This aids heavily in understanding how the lambda is being released and helps folks understand that they can built these specific plugins.

builtin/aws/lambda/releaser.go
thiskevinwang: @briancain if I'm understanding this correctly, then that means there are no further code changes required.. (?)

briancain: I think you're right! Looks good, the config validator will do the check for us based on what you mentioned above for the right use-cases for auth, so looks good! Resolved. Thanks :)

thiskevinwang:
@evanphx ah got it! I can move this to `builtin/aws/lambda/function_url` and rename it

evanphx (changes requested):
A couple minor things!

builtin/aws/lambda/function_url/releaser.go
```diff
@@ -0,0 +1,529 @@
...
+
+	// todo(thiskevinwang): CORS
+	// doc.SetField(
+	// 	"cors",
+	// 	"the CORS configuration for the Lambda function URL",
+	// 	docs.Summary(
+	// 		"Use CORS to allow access to your function URL from any domain. You can also use CORS to control access for specific HTTP headers and methods in requests to your function URL",
+	// 	),
+	// 	docs.Default("{}"),
+	// )
```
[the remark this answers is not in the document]
thiskevinwang: I'll remove this for now!

builtin/aws/lambda/function_url/releaser.go
```diff
@@ -0,0 +1,529 @@
...
+				pType := reflect.TypeOf(st.Principal)
+
+				switch pType {
+				case reflect.TypeOf(""):
+					if st.Principal != principal {
+						// principal has changed, we should update permissions
+						reset = true
+					}
+				case reflect.TypeOf(map[string]interface{}{}):
+					fmt.Println("3")
```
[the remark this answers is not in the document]
thiskevinwang: 😅 Removed!

evanphx (approved):

briancain (approved):
Thank you Kevin!! This looks great ✨

thiskevinwang:
Hey @evanphx / @briancain belated question:

Is this PR's changes expected to be included in `v0.8.x`, or will it be available in the next `v0.9.x` minor line?

I spent an hour or so trying to debug why my remote runner (k8s) was responding with...
```console
! 1 error occurred:
        * plugin "lambda-function-url" not found
```
...only to release that this PR's changes weren't in [release/0.8.x](https://github.com/hashicorp/waypoint/tree/release/0.8.x) 🤦‍♂️

---

```console
$ waypoint -v  (Runner Pod)
CLI: v0.8.2 (5b8885480+CHANGES)
Server: v0.8.2

$ waypoint -v (Local CLI)
CLI: v0.8.2 (5b8885480)
Server: v0.8.2
```

briancain:
@thiskevinwang - looks like this pull request wasn't backported, so it didn't make it into the 0.8.2 release. We can post backport this PR now and it will go out in the next release. Sorry about that!

PR #3188, opened 2022-04-07
Consumable job service handlers

izaaklauer:
This is the last in a series of PRs meant to make our service handlers consumable by other projects.

Previous PRs in this series:
https://github.com/hashicorp/waypoint/pull/3169
https://github.com/hashicorp/waypoint/pull/3149

This PR focuses on service_job.go. This adds a new `Reader` method to the `LogStreamProvider`, which allows other projects to inject a log stream reader and not depend on the built-in singleprocess reader.

This should not change any behavior or functionality.

### DIFF
--- a/pkg/server/logstream/logstream.go
+++ b/pkg/server/logstream/logstream.go
@@ -11,11 +11,30 @@ import (
 
 // Provider provides a log stream tracker
 type Provider interface {
-	StartWriter(ctx context.Context, log hclog.Logger, state serverstate.Interface, job *serverstate.Job) (Tracker, error)
+
+	// StartWriter starts a new log writer. Requires state to persist logs.
+	StartWriter(ctx context.Context, log hclog.Logger, state serverstate.Interface, job *serverstate.Job) (Writer, error)
+
+	// StartReader starts a new log reader.
+	StartReader(ctx context.Context, log hclog.Logger, job *serverstate.Job) (Reader, error)
+
+	// ReadCompleted returns all the log entries for a job that has been completed,
+	// by reading them out of persistent storage.
+	ReadCompleted(ctx context.Context, log hclog.Logger, state serverstate.Interface, job *serverstate.Job) ([]*pb.GetJobStreamResponse_Terminal_Event, error)
 }
 
-// Tracker collects and tracks loggable events
-type Tracker interface {
+// Writer collects and tracks loggable events
+type Writer interface {
 	Flush(ctx context.Context)
 	NewEvent(ctx context.Context, event *pb.RunnerJobStreamRequest_Terminal) error
+}
+
+// Reader reads terminal events for a given
+type Reader interface {
+
+	// ReadStream returns a batch of log entries for a job that's currenly active.
+	// If zero exist and block is true, this will block waiting for
+	// available entries. If block is false and no more log entries exist,
+	// this will return nil.
+	ReadStream(ctx context.Context, block bool) ([]*pb.GetJobStreamResponse_Terminal_Event, error)
 }
--- a/pkg/server/singleprocess/logstream.go
+++ b/pkg/server/singleprocess/logstream.go
@@ -60,7 +60,7 @@ func (l *singleProcessLogStreamProvider) ReadCompleted(ctx context.Context, log 
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to read log batch for completed job")
 		}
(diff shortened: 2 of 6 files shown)


pkg/server/logstream/logstream.go
```diff
@@ -11,11 +11,30 @@ import (
...
+
+	// ReadCompleted returns all the log entries for a job that has been completed,
+	// by reading them out of persistent storage.
+	ReadCompleted(ctx context.Context, log hclog.Logger, state serverstate.Interface, job *serverstate.Job) ([]*pb.GetJobStreamResponse_Terminal_Event, error)
 }
 
-// Tracker collects and tracks loggable events
-type Tracker interface {
+// Writer collects and tracks loggable events
+type Writer interface {
```
izaaklauer, line 27: Unrelated rename - we were mixing `Tracker` and `Writer` before.

pkg/server/singleprocess/logstream.go
```diff
@@ -17,29 +18,72 @@ import (
...
 	}, nil
 }
 
-// singleProcessLogStreamTracker implements logstream.Tracker
-type singleProcessLogStreamTracker struct {
+func (s *singleProcessLogStreamProvider) StartReader(ctx context.Context, log hclog.Logger, job *serverstate.Job) (logstream.Reader, error) {
+	var outputR *logbuffer.Reader
+
+	// NOTE(izaak): Not having an output buffer on the job isn't an error condition.
+	// Not sure when it would happen though.
```
izaaklauer, line 32: I'm preserving this behavior because I don't fully understand it, but if anyone does feel free to chime in.

pkg/server/singleprocess/service_job_test.go
```diff
@@ -672,6 +674,158 @@ func TestServiceGetJobStream_bufferedData(t *testing.T) {
...
+	}
+}
+
+// Tests that a client can connect to a completed job and still read
+// its output.
+// NOTE: this does not test that a client can read job logs for a
+// long-completed job. Some server state implementation (namely this one)
+// do not persist job logs, so streaming completed logs only works
+// if the server hasn't restarted or pruned them from memory.
+func TestServiceGetJobStream_completedBufferedData(t *testing.T) {
```
izaaklauer, line 757: Prior to this, we were only testing reading from a completed job. I renamed that test, and added a new test to read from a job that's still running but has buffered output.

pkg/server/singleprocess/service_job.go
```diff
@@ -603,17 +605,53 @@ func (s *Service) GetJobStream(
...
+						return status.Error(codes.Internal, msg)
+					}
+
+				// NOTE: at present (2022-02-25) there is no exposed CLI or UI API that
+				// causes GetJobStream to be called on a completed job. Thusly this code below
+				// is entirely optimistic about future API usage. But it is tested!
+				case pb.Job_SUCCESS, pb.Job_ERROR:
+					// This means that the requested stream finished before GetJobStream was
+					// called. As such, we'll replay the output events.
+					events, err := s.logStreamProvider.ReadCompleted(ctx, log, s.state(ctx), job)
```
izaaklauer, line 630: This is the biggest shift. We didn't have a concept of explicitly reading completed jobs before - we would always read any buffered logs, then read streamed logs, and this case was handled by the buffered log reading.

Other server impls, though, benefit from getting an explicit trigger to read completed job logs - they can use it to pull logs out of persistent storage. Our implementation of ReadComplete here just reads buffered logs from memory.

evanphx: Totally makes sense

pkg/server/singleprocess/service_job.go
```diff
@@ -683,11 +716,47 @@ func (s *Service) getJobStreamOutputInit(
...
+						Terminal: &pb.GetJobStreamResponse_Terminal{
+							Events: []*pb.GetJobStreamResponse_Terminal_Event{{
+								Timestamp: timestamppb.Now(),
+
+								// NOTE(izaak): really not sure if this is the right kind of terminal event
+								Event: &pb.GetJobStreamResponse_Terminal_Event_Line_{
+									Line: &pb.GetJobStreamResponse_Terminal_Event_Line{
+										Style: terminal.ErrorStyle,
+										Msg:   msg,
+									},
```
evanphx, line 745: This is fine, this will just output a single line of text.

evanphx (approved):
All looks good! Nice job abstracting those bits.

PR #3193, opened 2022-04-08
chore: add lambda platform config validation - add tests

thiskevinwang:
# Description

- This adds `aws-lambda` platform config validation for `timeout` and `architecture`
- This updates the platform test

### DIFF
--- a/builtin/aws/lambda/platform.go
+++ b/builtin/aws/lambda/platform.go
@@ -19,6 +19,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/lambda"
+	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/waypoint/builtin/aws/ecr"
 	"github.com/hashicorp/waypoint/builtin/aws/utils"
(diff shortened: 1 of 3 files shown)


builtin/aws/lambda/platform.go
```diff
@@ -38,12 +39,46 @@ type Platform struct {
...
+		}
+
+		var validArchitectures []string
+		for _, arch := range lambda.Architecture_Values() {
+			validArchitectures = append(validArchitectures, fmt.Sprintf("\"%s\"", arch))
+		}
+
+		if err := utils.Error(validation.ValidateStruct(c,
+			validation.Field(&c.Architecture,
+				validation.In(architectures...).Error(fmt.Sprintf("Unsupported function architecture \"%s\". Must be one of [%s], or left blank", c.Architecture, strings.Join(validArchitectures, ", "))),
```
briancain, line 63:
```suggestion
				validation.In(architectures...).Error(fmt.Sprintf("Unsupported function architecture %q. Must be one of [%s], or left blank", c.Architecture, strings.Join(validArchitectures, ", "))),
```

builtin/aws/lambda/platform.go
```diff
@@ -38,12 +39,46 @@ type Platform struct {
...
+	if c.Architecture != "" {
+		architectures := make([]interface{}, len(lambda.Architecture_Values()))
+
+		for i, ca := range lambda.Architecture_Values() {
+			architectures[i] = ca
+		}
+
+		var validArchitectures []string
+		for _, arch := range lambda.Architecture_Values() {
+			validArchitectures = append(validArchitectures, fmt.Sprintf("\"%s\"", arch))
```
briancain, line 58:
```suggestion
			validArchitectures = append(validArchitectures, fmt.Sprintf("%q", arch))
```

builtin/aws/lambda/platform.go
```diff
@@ -38,12 +39,46 @@ type Platform struct {
 // ConfigSet is called after a configuration has been decoded
 // we can use this to validate the config
 func (p *Platform) ConfigSet(config interface{}) error {
-	_, ok := config.(*Config)
+	c, ok := config.(*Config)
 	if !ok {
 		// this should never happen
 		return fmt.Errorf("Invalid configuration, expected *lambda.Config, got %s", reflect.TypeOf(config))
```
briancain, line 45:
```suggestion
		return fmt.Errorf("Invalid configuration, expected *lambda.Config, got %q", reflect.TypeOf(config))
```

builtin/aws/lambda/platform_test.go
```diff
@@ -1,40 +1,40 @@
...
+		cfg := &Config{}
+		require.NoError(t, p.ConfigSet(cfg))
+	})
+
+	t.Run("disallows unsupported architecture", func(t *testing.T) {
+		var p Platform
+		cfg := &Config{
+			Architecture: "foobar",
+		}
+		require.Error(t, p.ConfigSet(cfg))
```
briancain, line 21: Small test nit - some extra validation here might be checking the kind of error as well if there's a specific invalid config format error returned.

briancain (changes requested):
Hey there! Looks pretty good to me. My main recommendation is in Go when we format strings, you can use `%q` for quoting. That will save you from having to escape quotes inside of a string. We also prefer quoting `%q` any variable values in messages to let users know in the case where that value is nil. Then it'll show up like `This strings value is "" this!` rather than `This strings value is  this!` with a double space.

Otherwise looks great, love some extra config validation and tests 😄

builtin/aws/lambda/platform_test.go
thiskevinwang: @briancain is something like this (https://github.com/hashicorp/waypoint/pull/3193/commits/6ced6c2e8ae6ca0a34c3b88d137adda89348f79a) what you meant?

briancain: Yes! Perfect 👍🏻

PR #3196, opened 2022-04-08
cli: Pass context to `Bootstrapped` in `server_run`

jgwhite:
## Why the change?

Without the context argument, the casting always failed and we’d never show the prompt with the bootstrap command.

### Before

CleanShot 2022-04-08 at 11 42 06@2x

### After

CleanShot 2022-04-08 at 11 41 14@2x

## How do I test it?

1. `git checkout cli/run-bootstrap-fix`
2. `make bin`
3. `rm data.db`
4. `./waypoint server run -accept-tos`
5. Verify you see the bootstrap message

### DIFF
--- a/.changelog/3196.txt
+++ b/.changelog/3196.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fix missing bootstrap hint with server run command
+```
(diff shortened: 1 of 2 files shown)


briancain (approved):
Good find! Thanks 👍🏻 👢

PR #3202, opened 2022-04-08
Makefile: Add make targets for running full state and service tests

briancain:
This commit adds two makefile target helpers for running the full test
suite for both boltdbstate and the service API functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,5 +191,6 @@ test/boltdbstate:
 
 .PHONY: test/service
 test/service:
+	$(warning "Running the full service suite requires `docker-compose up`! Some Tests rely on a local Horizon instance to be running.")
 	@echo "Running service API server tests..."
 	go test -test.v ./pkg/server/singleprocess/ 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -182,3 +182,14 @@ tools: # install dependencies and tools required to build
...
+
+.PHONY: test/boltdbstate
+test/boltdbstate:
+	@echo "Running state tests..."
+	go test -test.v ./internal/server/boltdbstate 
+
+
+.PHONY: test/service
+test/service:
+	@echo "Running service API server tests..."
```
catsby, line 194: Maybe add a comment here that mentions users should run `docker-compose up` first to run this? I wouldn't necessarily expect this target to do that for users, but we should mention it if this these tests will fail otherwise.

catsby (approved):

Makefile
briancain: Good idea!

briancain: Pushed that up 👍🏻

PR #3203, opened 2022-04-08
Feature: Introduce `Task` to Waypoint core to track On-Demand Runner operations

briancain:
This pull request introduces a new concept in Waypoint core known as `Task`. Tasks track the lifecycle state of an on-demand runner through its Start, Run, and Stop jobs that get spawned for running operations remotely. By tracking this state specifically, we can get more insight into what Waypoint is doing when it runs jobs/operations remotely. It also includes the API and CLI functionality to inspect the state of a given Task, as well as list existing tasks.

Note that this pull request first few commits are taken from: https://github.com/hashicorp/waypoint/pull/3108. This branch includes the suggested PR feedback, plus the work to get the Task state machine tracked as Start, Run, and Stop jobs progress on the runner.


mitchellh:
Going to start reviewing this... in the future please make PRs of incremental changes on new branches _against_ the feature branch. It makes it way easier to separate out what needs review, what changed, etc.

internal/runner/operation.go
```diff
@@ -42,6 +42,47 @@ func (r *Runner) executeJob(
 ) (*pb.Job_Result, error) {
 	job := assignment.Job
 
+	var task *pb.Task
+	if job.Task != nil {
```
mitchellh, line 46: I was thinking we'd do all of this on the server side in the state system or the API system. I think that introduces less edge cases. When a runner acks a job, it is running, so at that point we can look up any tasks associated with the job, and update their state accordingly. We should be able to represent all state transitions cleanly this way, and much more easily test it (since state/API tests are easy unit tests to write).

In addition, I'd prevent modifying task state with `UpsertTask`, so there is no API path to break that internal state. It is purely event driven rather than trusting API clients.

Thoughts?

internal/server/boltdbstate/task.go
```diff
@@ -79,8 +79,51 @@ func (s *State) TaskDelete(ref *pb.Ref_Task) error {
...
+// memdb transaction. This is often used via the API for building out
+// a complete picture of a task beyond the job ID refs.
+func (s *State) GetJobsByTaskRef(
+	task *pb.Task,
+) (startJob *pb.Job, taskJob *pb.Job, stopJob *pb.Job, err error) {
+	memTxn := s.inmem.Txn(true)
+	defer memTxn.Abort()
+
+	err = s.db.View(func(dbTxn *bolt.Tx) error {
+		resp, err := s.JobById(task.StartJob.Id, nil)
```
mitchellh, line 92: Nesting read locks is _fine_ but weird. This is why we usually have the unexported versions that take the transaction. If `jobById` doesn't exist, I'd introduce that, and pass in `dbTxn` instead.

pkg/server/proto/server.proto
```diff
@@ -1179,6 +1186,76 @@ message Task {
...
+
+  // The overall state of the Task triple. See the enum def for details for
+  // each possible Task state. This will be computed for each GetTaskResponse
+  // based on the Task job ids receieved from the database.
+  State job_state = 7;
+
+  // The resource that gets created to run the task job for this Task
+  // TODO(briancain): This field has not been implemented yet. See
+  // "internal/runner/operation.go" for more info.
+  string resource_name = 8;
```
mitchellh, line 1198: What is this? I didn't find anything in operation.go.

pkg/server/proto/server.proto
```diff
@@ -372,6 +372,17 @@ service Waypoint {
   // UpsertStatusReport updates or inserts a statusreport.
   rpc UpsertStatusReport(UpsertStatusReportRequest) returns (UpsertStatusReportResponse);
 
+  // Internal Task functions
+
+  // UpsertTask updates or inserts a Task
+  rpc UpsertTask(UpsertTaskRequest) returns (UpsertTaskResponse);
```
mitchellh, line 378: Not fully convinced we need a write/update API for tasks at all. I think this goes back to my original feedback in the part 1 or part 2 PR. I'd manage all task state changes internally, and only expose _operations_ on tasks like `CancelTask` or something like that, rather than letting API callers modify state.

pkg/serverstate/serverstate.go
```diff
@@ -177,7 +177,8 @@ type Interface interface {
 	TaskPut(*pb.Task) error
 	TaskGet(*pb.Ref_Task) (*pb.Task, error)
 	TaskDelete(*pb.Ref_Task) error
-	TaskList() ([]*pb.Task, error)
+	TaskList(*pb.ListTaskRequest) ([]*pb.Task, error)
+	GetJobsByTaskRef(*pb.Task) (*pb.Job, *pb.Job, *pb.Job, error)
```
mitchellh, line 181: Do we normally have a `Get` prefix? I don't recall. I thought idiomatic Go was `JobsByTaskRef` instead.

mitchellh (changes requested):

pkg/server/proto/server.proto
briancain: Yep, this makes sense if we move the task driven state into the server!

internal/server/boltdbstate/task.go
briancain: That's true. Yeah it makes a lot more sense! Good call. `jobById` does exist so I'll use that instead

briancain: Addressed in ab18de5de2757fd3a519b5277b080a6201376a0a

pkg/serverstate/serverstate.go
briancain: Addressed in 3270d63103b7197eef42d6d88928cd04bd927d75

pkg/server/proto/server.proto
briancain: Fixed the comment in 73a9c4c437b3e574722e41c45e4fac357c6f90f5

internal/runner/operation.go
briancain: This was made in a follow up PR, intended to be merged into this one! https://github.com/hashicorp/waypoint/pull/3278

briancain:
Tests passed! 😎

mitchellh (approved):

mitchellh (approved):

PR #3212, opened 2022-04-11
Test reading/writing ODR target labels

izaaklauer:
Was a relatively new feature that could stand a test!

### DIFF
--- a/pkg/serverstate/statetest/test_runner_ondemand.go
+++ b/pkg/serverstate/statetest/test_runner_ondemand.go
@@ -15,6 +15,7 @@ import (
 func init() {
 	tests["runner_ondemand"] = []testFunc{
 		TestOnDemandRunnerConfig,
+		TestOnDemandRunnerConfigLabelTargeting,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


briancain (approved):
👍🏻 Makes sense! I'm always down for more tests. I know we originally added some at the service level, so seems good to include more state level tests around this feature ✨ ❤️

pkg/serverstate/statetest/test_runner_ondemand.go
```diff
@@ -142,3 +143,37 @@ func TestOnDemandRunnerConfig(t *testing.T, factory Factory, restartF RestartFac
...
+	require.NoError(err)
+
+	resp, err := s.OnDemandRunnerConfigGet(&pb.Ref_OnDemandRunnerConfig{
+		Id: rec.Id,
+	})
+	require.NoError(err)
+	require.NotNil(resp)
+
+	// Ensure the target saved properly
+	require.Equal(rec.TargetRunner.Target, resp.TargetRunner.Target)
```
briancain, line 178: I think this go-test failure is happening here? We might have to compare the actual map string string, since I don't think go likes comparing two protos of `pb.Ref_Runner_Labels { ... }`

mitchellh: Yeah, be very careful when comparing proto structures, cause they often contain other state. In general, I avoid it.

mitchellh (approved):
One comment then 👍

Issue #3194, opened 2022-04-08
aws-lambda: Make storage configurable


thiskevinwang:
**Is your feature request related to a problem? Please describe.**
Feature request: lambda ephemeral storage is now configurable, from `512 MB` to `10240 MB` (10 GB)
-  24 MAR 2022 
- https://aws.amazon.com/blogs/aws/aws-lambda-now-supports-up-to-10-gb-ephemeral-storage/

![CleanShot 2022-04-08 at 01 42 05@2x](https://user-images.githubusercontent.com/26389321/162372037-f0a5528c-06fd-4347-9600-14866b578d56.png)

**Describe the solution you'd like**
An optional `storage` field should be made available on the `aws-lambda` `deploy` stanza

```hcl
  deploy {
    use "aws-lambda" {
      region  = var.region
      memory  = 512
      storage = 9000
    }
  }
```

**Describe alternatives you've considered**
None

**Explain any additional use-cases**
None

**Additional context**

The AWS SDK API for this was introduced in https://github.com/aws/aws-sdk-go/releases/tag/v1.43.25

PR #3213, opened 2022-04-12
feat(aws-lambda): storage

thiskevinwang:
- feat(aws-lambda): storage
- chore: docs

Closes https://github.com/hashicorp/waypoint/issues/3194

### DIFF
--- a/.changelog/3213.txt
+++ b/.changelog/3213.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+plugin/aws-lambda: add support for lambda storage size
+```
--- a/.changelog/3213.txt
+++ b/.changelog/3213.txt
@@ -1,3 +1,3 @@
-```release-note:feature
+```release-note:improvement
 plugin/aws-lambda: add support for lambda storage size
 ```
--- a/builtin/aws/lambda/platform.go
+++ b/builtin/aws/lambda/platform.go
@@ -12,6 +12,8 @@ import (
 	"time"
 
 	"golang.org/x/crypto/ssh"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -358,7 +360,7 @@ func (p *Platform) Deploy(
 					case "ValidationException":
 						// likely here if Architectures was invalid
 						if architecture != lambda.ArchitectureX8664 && architecture != lambda.ArchitectureArm64 {
-							return nil, fmt.Errorf("architecture must be either \"x86_64\" or \"arm64\"")
+							return nil, status.Error(codes.FailedPrecondition, "architecture must be either \"x86_64\" or \"arm64\"")
 						}
 						// other validation error
 						return nil, err
--- a/builtin/aws/lambda/platform.go
+++ b/builtin/aws/lambda/platform.go
@@ -145,6 +145,9 @@ const (
 
 	// The instruction set architecture that the function supports.
 	DefaultArchitecture = lambda.ArchitectureX8664
+
+	// The storage size (in MB) for the function's `/tmp` directory
+	DefaultStorageSize = 512
 )
 
 const lambdaRolePolicy = `{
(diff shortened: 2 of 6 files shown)


.changelog/3213.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
```
briancain, line 1: This should be `improvement` instead of feature. In Waypoint, we reserve the `feature` release note for brand new components rather than something that is improving on an existing thing.

builtin/aws/lambda/platform.go
```diff
@@ -829,6 +869,11 @@ type Config struct {
 	// Valid values are: "x86_64", "arm64"
 	// Defaults to "x86_64".
 	Architecture string `hcl:"architecture,optional"`
+
+	// The storage size (in MB) of the Lambda function's `/tmp` directory.
+	// Must be a value between 512 and 10240.
+	// Defaults to 512 MB.
+	Storage int `hcl:"storage,optional"`
```
briancain, line 876: I recommend calling this `storagemb` so that it's obvious what unit Storage is without having to look it up in the Waypoint docs. We do something similar in Nomad for Memory:

https://tip.waypointproject.io/plugins/nomad#resources-memorymb

builtin/aws/lambda/platform.go
```diff
@@ -269,39 +277,65 @@ func (p *Platform) Deploy(
...
+						// likely here if Architectures was invalid
+						if architecture != lambda.ArchitectureX8664 && architecture != lambda.ArchitectureArm64 {
+							return nil, fmt.Errorf("architecture must be either \"x86_64\" or \"arm64\"")
+						}
+						// other validation error
+						return nil, err
+					// Function update in progress error
+					case "ResourceConflictException":
+						log.Warn(fmt.Sprintf("Function update in progress. Waiting 2 seconds before retrying... (%d/30)", i+1))
+						time.Sleep(2 * time.Second)
```
briancain, line 333: This timeout makes sense, but let's leverage `context` instead of a sleep and for iterator. You can see an example of how this might work in the nomad jobspec releaser:

https://github.com/hashicorp/waypoint/blob/8d9e228fb8506ab7bfb169ca778b9b6dc206ec36/builtin/nomad/jobspec/releaser.go#L176-L178

https://github.com/hashicorp/waypoint/blob/8d9e228fb8506ab7bfb169ca778b9b6dc206ec36/builtin/nomad/jobspec/releaser.go#L184-L188

Let me know if you have any questions or need help figuring this out!

builtin/aws/lambda/platform.go
```diff
@@ -269,39 +277,65 @@ func (p *Platform) Deploy(
...
+
+			if err != nil {
+
+				if aerr, ok := err.(awserr.Error); ok {
+					switch aerr.Code() {
+					// unrecoverable error
+					case "ValidationException":
+						// likely here if Architectures was invalid
+						if architecture != lambda.ArchitectureX8664 && architecture != lambda.ArchitectureArm64 {
+							return nil, fmt.Errorf("architecture must be either \"x86_64\" or \"arm64\"")
```
briancain, line 326: It doesn't look like we're formatting anything, so I think we could return error messages using the grpc status and codes:

```suggestion
							return nil, status.Error(codes.FailedPrecondition, "architecture must be either \"x86_64\" or \"arm64\"")
```

briancain (changes requested):
Awesome! Looking really good. Let me know if you need any help working on the suggestions here. But otherwise this is a great feature to add to the lambda plugin.

.changelog/3213.txt
thiskevinwang: fixed!

builtin/aws/lambda/platform.go
thiskevinwang: Done! - https://github.com/hashicorp/waypoint/pull/3213/commits/dcc917edcaaeb9fae31db8db2320e9f1aaaba61c

builtin/aws/lambda/platform.go
thiskevinwang: Done! - https://github.com/hashicorp/waypoint/pull/3213/commits/f3c758f20a5647096e35fad636cd890379b87bb6

builtin/aws/lambda/platform.go
thiskevinwang: Neat and TIL! Addressed - https://github.com/hashicorp/waypoint/pull/3213/commits/d046cbccab1ef7217e12fcdabc671ec4400f9fde

briancain (approved):

PR #3216, opened 2022-04-13
Adds $TESTARGS option to test/boltdbstate step, with doc/example

catsby:
Adds ability to run a specific state test to the `test/boltdbstate` make target and some minimal docs explaining


briancain (approved):
Love it! Good idea with the code doc example too 👍🏻

izaaklauer (approved):
woo nice!

PR #3220, opened 2022-04-13
builtin/nomad/jobspec: fix dropped error

alrs:
This fixes a dropped `err` variable in `builtin/nomad/jobspec`.  No changelog entry is warranted.


briancain (approved):
Thanks!

Issue #3175, opened 2022-04-05
“Add variable” icon is misaligned


jgwhite:
**Describe the bug**
![CleanShot 2022-04-05 at 12 45 20@2x](https://user-images.githubusercontent.com/34030/161737704-a27380e1-2155-4606-ba7e-26400cfa6dcc.png)

**Expected behavior**
The button should be correctly aligned.

**Waypoint Version**
db48e2d6c15b235676109481e095ce06e58171af

(Ignore the version number in the screenshot; it’s simulated and we need to update it)

PR #3231, opened 2022-04-15
UI: Fix Add variable icon is misaligned

andrew-hashicorp:
The `Add variable` button in the `true` block was missing the `"button--add-variable"` CSS class, causing the icon to become misaligned when the component was displaying existing input variables / add new input variable form.

Before:
Screen Shot 2022-04-15 at 1 50 58 PM

After:
Screen Shot 2022-04-15 at 1 50 19 PM

This matches how it's implemented in [`waypoint/ui/app/components/project-config-variables/list.hbs`](https://github.com/hashicorp/waypoint/blob/main/ui/app/components/project-config-variables/list.hbs)

Closes https://github.com/hashicorp/waypoint/issues/3175

### DIFF
--- a/ui/app/components/project-input-variables/list.hbs
+++ b/ui/app/components/project-input-variables/list.hbs
@@ -13,7 +13,7 @@
     {{/each}}
   </ul>
   <Pds::ButtonSet>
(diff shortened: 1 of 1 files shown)


sdav9375 (approved):
This looks great!

gregone (approved):

PR #3233, opened 2022-04-18
internal/{service,cli}: Introduce CancelTask API and CLI

briancain:
This commit will be going into the feature branch for pull request #3203. Made separately for easier review.

This pull request introduces a `CancelTask` API which will gracefully attempt to cancel each job inside the Task request. Based off of the feedback from the feature PR to include a task cancel endpoint.


internal/cli/task_cancel.go
```diff
@@ -0,0 +1,103 @@
...
+			Name:    "run-job-id",
+			Target:  &c.flagRunJobId,
+			Default: "",
+			Usage:   "Cancel a Task by Run Job Id.",
+		})
+	})
+}
+
+func (c *TaskCancelCommand) Synopsis() string {
+	return "Cancel an On-Demand Runner Task running in Waypoint"
```
catsby, line 93: > an On-Demand
Does this only work with ODR?

internal/server/boltdbstate/task.go
```diff
@@ -79,6 +79,40 @@ func (s *State) TaskDelete(ref *pb.Ref_Task) error {
...
+// when we cancel a job, we also call a `db.Update` on each job when we update its
+// state in the database. This caused a deadlock. For now we cancel each job
+// separately.
+func (s *State) TaskCancel(ref *pb.Ref_Task) error {
+	task, err := s.TaskGet(ref)
+	if err != nil {
+		return err
+	}
+
+	s.log.Debug("canceling start job for task by id", "task id", task.Id, "start job id", task.StartJob.Id)
```
catsby, line 95: Super-nit: "for task by id" seems superfluous for these debug messages.

internal/server/boltdbstate/task.go
```diff
@@ -274,6 +308,67 @@ func (s *State) taskDelete(
 	return nil
 }
 
+/*
+func (s *State) taskCancel(
```
catsby, line 312: do we have plans for this code or was it left in by accident?

pkg/server/singleprocess/service_task_test.go
```diff
@@ -354,3 +354,85 @@ func TestServiceTask_ListTaskFilters(t *testing.T) {
...
+
+		_, err = client.CancelTask(ctx, &pb.CancelTaskRequest{
+			Ref: &pb.Ref_Task{
+				Ref: &pb.Ref_Task_Id{
+					Id: resp.Task.Id,
+				},
+			},
+		})
+		require.NoError(err)
+	})
```
catsby, line 437: This test only verifies that we can call CancelTask and not receive an error. It doesn't seem to actually test if the task gets canceled. I think we should also query or list the task and verify it is cancelled.

pkg/server/singleprocess/service_task_test.go
```diff
@@ -354,3 +354,85 @@ func TestServiceTask_ListTaskFilters(t *testing.T) {
...
+		Task: serverptypes.TestValidTask(t, &pb.Task{
+			TaskJob:  &pb.Ref_Job{Id: runJobId},
+			StartJob: &pb.Ref_Job{Id: startJobId},
+			StopJob:  &pb.Ref_Job{Id: stopJobId},
+		}),
+	})
+	require.NoError(t, err)
+	taskId := resp.Task.Id
+
+	// Create, should get an ID back
```
catsby, line 410: This seems like mistaken copy-paste

internal/cli/task_cancel.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	} else if c.flagRunJobId != "" {
+		taskReq = &pb.CancelTaskRequest{
+			Ref: &pb.Ref_Task{
+				Ref: &pb.Ref_Task_JobId{
+					JobId: c.flagRunJobId,
+				},
+			},
+		}
+	}
```
catsby, line 61:
```suggestion
taskReq = &pb.CancelTaskRequest{
  Ref: &pb.Ref_Task{},
}

if taskId != "" {
  taskReq.Ref.Ref = &pb.Ref_Task_Id{
    Id: taskId,
  }
} else if c.flagRunJobId != "" {
  taskReq.Ref.Ref = &pb.Ref_Task_JobId{
    JobId: c.flagRunJobId,
  }
}
```

I feel like refactoring this to only show what is different between the two branches improves readability. Feel free to discard

internal/cli/task_cancel.go
```diff
@@ -0,0 +1,103 @@
...
+	ctx := c.Ctx
+
+	var taskId string
+	if len(c.args) == 0 && c.flagRunJobId == "" {
+		c.ui.Output("Task ID or Run Job Id required.\n\n%s", c.Help(), terminal.WithErrorStyle())
+		return 1
+	} else {
+		taskId = c.args[0]
+	}
```
catsby, line 35:
```suggestion
	if len(c.args) == 0 && c.flagRunJobId == "" {
		c.ui.Output("Task ID or Run Job Id required.\n\n%s", c.Help(), terminal.WithErrorStyle())
		return 1
	}
	taskId := c.args[0]
```

Nit: remove the unnecessary `else` block

internal/cli/task_cancel.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (c *TaskCancelCommand) Run(args []string) int {
+	// Initialize. If we fail, we just exit since Init handles the UI.
+	if err := c.Init(
+		WithArgs(args),
+		WithFlags(c.Flags()),
+	); err != nil {
+		return 1
+	}
+	ctx := c.Ctx
```
catsby, line 27: nit: we don't seem to get any value out of this statement as we don't appear to do anything with `ctx` beyond pass it into another method. Is there a good reason to do this here?

catsby (commented):
LGTM, mostly nit picks or suggestions

internal/cli/task_cancel.go
briancain: Yep, you can only cancel tasks with this CLI, and ODR is what spawns them.

internal/server/boltdbstate/task.go
briancain: Agreed! I'll push a fix

internal/server/boltdbstate/task.go
briancain: I should leave a comment here - but I left this purposefully commented out due to my other note above `TaskCancel`. In the future if we ever fix `TaskCancel` to cancel these jobs within a single transaction, we can just uncomment this and get it most of the way there.

pkg/server/singleprocess/service_task_test.go
briancain: We do this already in the state tests! It felt weird to test that given it was already covered there, but I can!

internal/cli/task_cancel.go
briancain: Makes sense - the protobuf nesting gets a little out of hand 😄 I like this though.

internal/cli/task_cancel.go
briancain: Yeah, we can keep it to just use `c.Ctx`. There's no explicit reason here other than that's a similar pattern for other CLI commands

internal/cli/task_cancel.go
catsby: So I can't cancel a task that's done locally, if I don't have ODR going, or local things don't create tasks?

pkg/server/singleprocess/service_task_test.go
catsby: I find it valuable in that we're connecting the dots that calling this service endpoint actually cancels a task, otherwise I feel like we're just testing that the endpoint exists, not that it actually does anything. Not a huge deal if we leave it as is

internal/cli/task_cancel.go
catsby: For sure the nesting hurts the eyes after a bit trying to identify that the difference here was `Ref_Task_Id` vs. `Ref_Task_JobId` :)

internal/cli/task_cancel.go
catsby: I saw some that did, and some that didn't, which is why I asked. Feel free to leave as-is

pkg/server/singleprocess/service_task_test.go
briancain: Sounds good - I can add a few more lines of code to test this, should be pretty easy! Working on that now

internal/cli/task_cancel.go
briancain: Yeah - ODR in a local context won't work. That would create a single job run via the local runner in the CLI! So local things don't create tasks, only a remote runs.

catsby (approved):

pkg/server/proto/server.proto
```diff
@@ -1258,6 +1264,15 @@ message ListTaskResponse {
   repeated GetTaskResponse tasks = 1;
 }
 
+message CancelTaskRequest {
+  // A reference to the task to cancel
+  Ref.Task ref = 1;
+
+  // Forcefully attempt to cancel the task
+  // TODO(briancain): Unclear if this is a good idea
```
mitchellh, line 1272: I would add more commentary on this, or remove it completely to avoid confusing folks.

internal/server/boltdbstate/task.go
```diff
@@ -79,6 +79,40 @@ func (s *State) TaskDelete(ref *pb.Ref_Task) error {
...
+// state in the database. This caused a deadlock. For now we cancel each job
+// separately.
+func (s *State) TaskCancel(ref *pb.Ref_Task) error {
+	task, err := s.TaskGet(ref)
+	if err != nil {
+		return err
+	}
+
+	s.log.Debug("canceling start job for task", "task id", task.Id, "start job id", task.StartJob.Id)
+	err = s.JobCancel(task.StartJob.Id, false)
```
mitchellh, line 96: I haven't looked into it: but I thought we setup a dependency tree in the job system here. If you can cancel a parent, I think it automatically cancels all dependents (there is a setting to override that I think?). Should we just lean in on that and not manually cancel the triple, and instead just cancel the root (start)?

briancain: That's true, I'll see if we're setting up the task triple as a dependency when we schedule those jobs! Canceling a single root job and letting the internal system handle downstream dependencies sounds legit.

pkg/server/proto/server.proto
briancain: Agreed. I think given the caution we have around forcefully canceling jobs, I don't like the idea of making it easier to cancel 3 jobs via the Task API. If someone really wants to they can do a manual force cancel the hard way. I'll remove this!

internal/server/boltdbstate/task.go
briancain: Looks like currently our job system will only cancel dependent jobs off the parent if they are `QUEUED` and no other state. From that standpoint for now - I think we should explicitly cancel each of the Task triples to ensure they aren't left dangling. Then perhaps next steps is figuring out how exactly we want the job system to cancel dependent jobs and if only canceling QUEUED jobs still makes sense?

PR #3239, opened 2022-04-19
Fix nil pointer dereference

paladin-devops:
Fix nil pointer dereference for Consul/Vault tokens in Nomad platform deployer.


briancain (approved):
Ah yes, good catch on this 👍🏻 If I remember right, I had a similar issue when I was working with Nomads API and their `*int` for memory and CPU limits.

briancain:
And since I don't think your previous PR was released yet, no need to changelog! FYI.

Issue #3214, opened 2022-04-12
Waypoint entrypoint incompatible with Nomad docker driver `entrypoint`


acaloiaro:
**Describe the bug**
When waypoint's entrypoint is enabled with a Nomad jobspec task, the Nomad docker driver's `config.entrypoint` cannot be set, lest it override Waypoint entrypoint injection. 

While I understand if the project maintainers would prefer not to label this a bug, I believe it's worth noting that the result is quite surprising. That is, a task that successfully launches without any of the benefits of a waypoint entrypoint e.g. environment variables set with waypoint are not available, `waypoint exec` is unavailable, etc. 

A warning would be nice, or a mention in the documentation about waypoint configuration settings that conflict with nomad jobspec configurations. Perhaps here: https://www.waypointproject.io/plugins/nomad#entrypoint-functionality

**Steps to Reproduce**
1. Configure a waypoint task with the waypoint entrypoint enabled 
2. Launch a nomad jobspec with a nomad task that uses the docker driver
3. Set `config.entrypoint` to anything 
4. Build/deploy the app and observe that the waypoint entrypoint is not present 

**Expected behavior**
A. That this behavior is documented 
or 
B. That the waypoint-injected entrypoint launch the entrypoint specified by `config.entrypoint` 

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: 0.7.2
* Waypoint Server Platform and Version: nomad-1.2.4
* Waypoint Plugin: n/a

**Additional context**

briancain:
Hey @acaloiaro - this is a good thing to call out. I think we could at least solve this by adding some warnings in our documentation around using the nomad jobpsec plugin. A nice to have would be showing a warning in Waypoint when we detect that the nomad jobpsec file has configured an entrypoint which would conflict with Waypoints entrypoint. Anyway, thanks for the request!

acaloiaro:
I think that'd be a great middle ground solution @briancain. Thanks.

acaloiaro:
This appears complete.

PR #3242, opened 2022-04-19
Document conflict between Waypoint/Nomad entrypoint

paladin-devops:
Addresses #3214


briancain (approved):
This looks good to me! Maybe a follow up during a bug bash week could be figuring out if we can extract the entrypoint setting from a jobspec file and display a warning if both are active. This seems like a good first step for now 👍🏻

PR #3245, opened 2022-04-20
Avoid race condition reading job stream data before it's buffered.

izaaklauer:
Fixes the flaky TestServiceGetJobStream_bufferedData test 🤞 

Brief summary of the problem:

We were emitting job stream events, then connecting to a stream immediately, and assuming they would be buffered. In some cases, we read before the event finishes percolating through the tubes, so it comes through as an unbuffered event.

We could probably just sleep for a few millis after emitting the events, but this method should be more deterministic! We open a separate job stream, and wait for _it_ to see all the events. Good idea @demophoon !

Thanks Brian, Shirley, and Britt for the debug swarming


xiaolin-ninja (approved):
Much debug such awesome

izaaklauer:
It's worth noting that nobody is actually using that `Buffered` field that this is testing - it would probably be fine for us to remove it altogether.

briancain (approved):
👍🏻 🙏🏻 🐃

PR #3251, opened 2022-04-20
Update containerd

krantzinator:
Closes https://github.com/hashicorp/waypoint/security/dependabot/56, https://github.com/hashicorp/waypoint/security/dependabot/31, https://github.com/hashicorp/waypoint/security/dependabot/21, https://github.com/hashicorp/waypoint/security/dependabot/19


evanphx (approved):

briancain (approved):

krantzinator:
Actually I think this is impossible to fix, so many other dependencies are using old versions of Containerd, but at least we'll update this set.

Issue #3117, opened 2022-03-21
Bug: `docker pull` builder does not work with ODR


izaaklauer:
Currently, if you try to use the `docker pull` builder with ODR, you get an error that no docker engine is present.

### Implementation

The docker pull builder should implement `component.BuilderODR`, just like the [regular docker builder](https://github.com/hashicorp/waypoint/blob/f38e20450bd8ae31e6d8d822cee6817b39ae67c5/builtin/docker/builder.go#L59-L61) . The ODR pull builder should use kaniko.

briancain:
This was fixed in #3253!

PR #3253, opened 2022-04-20
Support remote operations for docker-pull

### LINKED ISSUE #2780: Remote operations for docker-pull
I'd like to be able to run "waypoint build" for my apps where I use the "docker-pull" plugin, so that I can keep the remote build/deploy/release workflow consistent with my apps where I use the "docker build" plugin, and also not require users to have Docker on their local machines in order to use Waypoint.

Running the Waypoint build locally, or running Waypoint "locally" on a CI/CD tool's worker such as a Jenkins agent.

**Additional context**
Forum post from where this request originated: https://discuss.hashicorp.com/t/docker-pull-plugin-w-remote-operations/31925/2

paladin-devops:
Closes #2780 & #3117.

Implementation is very similar to that of the Docker builder plugin, except the Dockerfile is created by this plugin & supplied to Kaniko with only a single line, containing `FROM` followed by the image & tag.

### DIFF
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -154,7 +154,7 @@ func (b *Builder) pullWithKaniko(
 
 	step.Done()
 
-	step = sg.Add("Image pushed to '%s:%s'", ai.Image, ai.Tag)
+	step = sg.Add("Image pull completed.")
 	step.Done()
 
 	return target, nil
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -137,7 +137,7 @@ func (b *Builder) pullWithKaniko(
 	log.Debug("executing kaniko", "args", args)
 
 	step.Done()
-	step = sg.Add("Executing kaniko...")
+	step = sg.Add("Executing Kaniko...")
 
 	// Command output should go to the step
 	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -58,6 +58,8 @@ func (b *Builder) pullWithKaniko(
 		case *wpdocker.AccessInfo_UserPass_:
 			oci.AuthConfig.Username = sv.UserPass.Username
 			oci.AuthConfig.Password = sv.UserPass.Password
+		default:
+			return nil, status.Error(codes.Unauthenticated, "Unexpected auth type.")
 		}
 	}
 
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -98,7 +98,7 @@ func (b *Builder) pullWithKaniko(
 
 		data, err := assets.Asset(assetName)
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "unable to restore custom entry point binary: %s", err)
+			return nil, status.Errorf(codes.Internal, "unable to restore custom entrypoint binary: %s", err)
 		}
 
 		step.Done()
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -1,0 +1,158 @@
+package dockerpull
+
+import (
+	"context"
+	"fmt"
+	"github.com/docker/distribution/reference"
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	wpdocker "github.com/hashicorp/waypoint/builtin/docker"
+	"github.com/hashicorp/waypoint/internal/assets"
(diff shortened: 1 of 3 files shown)


builtin/docker/pull/kaniko.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	defer li.Close()
+	go http.Serve(li, &oci)
+
+	port := li.Addr().(*net.TCPAddr).Port
+
+	localRef := fmt.Sprintf("localhost:%d/%s:%s", port, refPath, ai.Tag)
+
+	dockerfileBS := []byte(fmt.Sprintf("FROM %s:%s", target.Image, target.Tag))
```
evanphx, line 125:
```suggestion
	dockerfileBS := []byte(fmt.Sprintf("FROM %s:%s\n", target.Image, target.Tag))
```

evanphx (changes requested):
One little nit, otherwise looks great!

builtin/docker/pull/kaniko.go
```diff
@@ -0,0 +1,158 @@
...
+		// always matches the architecture of our Kaniko environment.
+		assetName, ok := assets.CEBArch[runtime.GOARCH]
+		if !ok {
+			return nil, status.Errorf(codes.FailedPrecondition,
+				"automatic entrypoint injection not supported for architecture: %s", runtime.GOARCH)
+		}
+
+		data, err := assets.Asset(assetName)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "unable to restore custom entry point binary: %s", err)
```
briancain, line 101:
```suggestion
			return nil, status.Errorf(codes.Internal, "unable to restore custom entrypoint binary: %s", err)
```

builtin/docker/pull/kaniko.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+func (b *Builder) pullWithKaniko(
+	ctx context.Context,
+	ui terminal.UI,
+	sg terminal.StepGroup,
+	log hclog.Logger,
+	ai *wpdocker.AccessInfo,
+) (*wpdocker.Image, error) {
+	step := sg.Add("Pulling Docker image with kaniko...")
```
briancain, line 30:
```suggestion
	step := sg.Add("Pulling Docker image with Kaniko...")
```

builtin/docker/pull/kaniko.go
```diff
@@ -0,0 +1,158 @@
...
+		step.Done()
+		step = sg.Add("Testing registry and uploading entrypoint layer")
+
+		err = oci.SetupEntrypointLayer(refPath, data)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "error setting up entrypoint layer: %s", err)
+		}
+	}
+
+	li, err := net.Listen("tcp", "localhost:0")
```
briancain, line 113: A note why we're setting up a net listener would be good.

I think the `localhost:0` means that it takes any available port, right? Nice 👍🏻

builtin/docker/pull/kaniko.go
```diff
@@ -0,0 +1,158 @@
...
+	args := []string{
+		"/kaniko/executor",
+		"-f", filepath.Dir("Dockerfile"),
+		"-d", localRef,
+	}
+
+	log.Debug("executing kaniko", "args", args)
+
+	step.Done()
+	step = sg.Add("Executing kaniko...")
```
briancain, line 140:
```suggestion
	step = sg.Add("Executing Kaniko...")
```

builtin/docker/pull/kaniko.go
```diff
@@ -0,0 +1,158 @@
...
+	cmd.Stderr = cmd.Stdout
+
+	// check for error from executor run
+	if err := cmd.Run(); err != nil {
+		return nil, err
+	}
+
+	step.Done()
+
+	step = sg.Add("Image pushed to '%s:%s'", ai.Image, ai.Tag)
```
briancain, line 154: I might be a little confused with the UX messages, but this is pulling with Kaniko, but the final message is pushed? I think we're pushing because we inject an entrypoint after pulling the original container? We could use some clearer messaging at the end here.

builtin/docker/pull/kaniko.go
briancain: For any user-facing messages, it's good to try to capitalize a tool or name.

briancain (commented):
Looks pretty good overall @paladin-devops - thanks for this! Most of my comments are around the StepGroup UX. I think this could tell a clearer story as to what the plugin is doing, otherwise looks great! 😄

builtin/docker/pull/kaniko.go
paladin-devops: @briancain I will definitely try to clear this up. This actually stumped me for a bit when I was reading through the Docker builder code, but I think it's because Kaniko is doing a Docker build, tag, and push all-in-one, even if that push is to a local registry.

evanphx (approved):

PR #3255, opened 2022-04-20
pkg/server: fix flaky test around buffered data

mitchellh:
To fix this, we need to retry a connection until we get buffered data.
We have to do this because the buffered data is read on connection and
immediately queued to send. It is not dependent on the client receiving
(though, it'll block due to TCP backpressure if they aren't receiving).

### DIFF
--- a/pkg/server/singleprocess/service_job_test.go
+++ b/pkg/server/singleprocess/service_job_test.go
@@ -675,49 +675,38 @@ func TestServiceGetJobStream_bufferedData(t *testing.T) {
 	}))
 
 	// Get our job stream and verify we open
-	stream, err := client.GetJobStream(ctx, &pb.GetJobStreamRequest{JobId: queueResp.JobId})
-	require.NoError(err)
-	{
-		resp, err := stream.Recv()
-		require.NoError(err)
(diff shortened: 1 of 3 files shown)


pkg/server/singleprocess/service_job_test.go
```diff
@@ -675,49 +675,38 @@ func TestServiceGetJobStream_bufferedData(t *testing.T) {
...
-	// here and get unbuffered logs.
-	{
-		resp := jobStreamRecv(t, stream, (*pb.GetJobStreamResponse_Terminal_)(nil))
-		event := resp.Event.(*pb.GetJobStreamResponse_Terminal_)
-		require.NotNil(event)
-		require.True(event.Terminal.Buffered)
-		require.Len(event.Terminal.Events, 2)
-	}
+		return true
+	}, 2*time.Second, 50*time.Millisecond)
```
izaaklauer, line 709: 👍

izaaklauer (approved):
Good catch!

briancain (approved):

xiaolin-ninja (approved):

PR #3262, opened 2022-04-20
cli: Show error message if no Waypoint contexts found

### LINKED ISSUE #3243: Feature request: improve error/warning for `ui -authenticate` when no default context

When running `waypoint ui -authenticate` with no _default/active_ context:
1. The error doesn't isn't clear on how to resolve itself.
2. The config/context is looked up with an empty (`""`) name.

The error message should include some steps on how to remediate the situation.

N/A

N/A

**Additional context**

Terminal output
```console
λ waypoint -v              
CLI: v0.8.1 (622f37bd1)

λ  waypoint context list
No contexts. Create one with `waypoint context create`.

λ  waypoint ui -authenticate
Waypoint must be configured in server mode to access the UI

» Creating invite token
This invite token will be exchanged for an authentication
token that your browser stores.
! <nil>: Configuration file not found; The configuration file
  /Users/kevin/Library/Preferences/waypoint/context/.hcl does not exist.
```

Blank name lookup:
https://github.com/hashicorp/waypoint/blob/26396c895f40276db532a368dae351f40b01d55e/internal/cli/ui.go#L55-L60

The error being thrown appears to be from:
https://github.com/hashicorp/hcl/blob/2ef09d129d968e8c6e7df29f9af103025eafd61a/hclsimple/hclsimple.go#L89-L97

briancain:
Prior to this commit, if Waypoint had no contexts locally on disk, it
would still attempt to load one with an empty name, resulting in an
error about a nil context. This commit fixes that by first seeing if any
contexts exist at all, or if there's no default, showing a better error
message to tell the user what to do about it.

Fixes #3243

I paired with @demophoon on fixing this one!

### DIFF
--- a/internal/cli/ui.go
+++ b/internal/cli/ui.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/clicontext"
 	"github.com/hashicorp/waypoint/internal/clierrors"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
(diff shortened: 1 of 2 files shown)


internal/cli/ui.go
```diff
@@ -36,6 +37,35 @@ func (c *UICommand) Run(args []string) int {
...
+				return 1
+			}
+
+			c.ui.Output("\n"+noContextFoundError, terminal.WithWarningStyle())
+			return 1
+		}
+
+		c.ui.Output("\n"+wpNoServerContext, terminal.WithWarningStyle())
+		return 1
+	}
```
catsby, line 67: Nit: I feel like this `if -> else` block isn't controlling the flow but instead doing error checking, since `name == ""` can only result in a scenario where we `return 1`, so let's move the errors into a check and then have the other things just outside of it as normal flow. The benefit of this approach is just the mental context of "we're doing error handling here" vs. "we're making a decision here and need to understand the outcome"

```suggestion
	if name == "" {
		if allContexts, err := c.contextStorage.List(); len(allContexts) == 0 || err != nil {
			if err != nil {
				c.project.UI.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
				return 1
			}

			c.ui.Output("\n"+noContextFoundError, terminal.WithWarningStyle())
			return 1
		}

		c.ui.Output("\n"+wpNoServerContext, terminal.WithWarningStyle())
		return 1
	}
  
	ctxConfig, err = c.contextStorage.Load(name)
	if err != nil {
		c.project.UI.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
		return 1
	}
```

catsby (approved):
LGTM just 1 code refactor suggestion, take it or leave it 👍

internal/cli/ui.go
briancain: That makes sense to me! 👍🏻 I'll push up a fix

briancain: I didn't end up doing `Commit suggestion` only because I wasn't sure if GitHub would do the right thing formatting wise and didn't wanna get a lint failure. Thank you for showing me the suggestion though! 🙏🏻

demophoon (approved):

PR #3265, opened 2022-04-21
Allow some state tests to be skipped

izaaklauer:
Currently, there's no good way for any other project consuming these tests to selectively skip any of them. Maybe there will be [as of go 1.19](https://github.com/golang/go/issues/41583), but until then, this introduces a new "skipTests" argument that allows users to pass in test names to skip.

### DIFF
--- a/pkg/serverstate/statetest/statetest.go
+++ b/pkg/serverstate/statetest/statetest.go
@@ -33,13 +33,30 @@ type (
 // Test runs a validation test suite for a state implementation. All
 // state implementations should pass this suite with no errors to ensure
 // the correct behavior of the state when Waypoint uses it.
-func Test(t *testing.T, f Factory, rf RestartFactory) {
+// skipTests are function names of tests in the serverstate package to skip
+// (i.e. TestJobCreate_singleton). It cannot skip sub-tests (called by
+// t.Run() inside a top-level test)
+func Test(t *testing.T, f Factory, rf RestartFactory, skipTests []string) {
 	for name, funcs := range tests {
 		t.Run(name, func(t *testing.T) {
 			for _, tf := range funcs {
 				name := runtime.FuncForPC(reflect.ValueOf(tf).Pointer()).Name()
 				if idx := strings.LastIndexByte(name, '.'); idx >= 0 {
 					name = name[idx+1:]
+				}
+
+				skip := false
+				for _, skipTest := range skipTests {
+					if name == skipTest {
+						skip = true
+						break
+					}
+				}
+				if skip {
+					t.Run(name, func(t *testing.T) {
+						t.Skip("Test is on the skip list - ignoring")
+					})
+					continue
 				}
 
 				t.Run(name, func(t *testing.T) {
--- a/pkg/serverstate/statetest/statetest.go
+++ b/pkg/serverstate/statetest/statetest.go
@@ -54,7 +54,7 @@ func Test(t *testing.T, f Factory, rf RestartFactory, skipTests []string) {
 				}
 				if skip {
 					t.Run(name, func(t *testing.T) {
-						t.Skip("Test is on the skip list - ignoring")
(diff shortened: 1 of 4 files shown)


pkg/serverstate/statetest/statetest.go
```diff
@@ -42,6 +45,20 @@ func Test(t *testing.T, f Factory, rf RestartFactory) {
...
+					if name == skipTest {
+						skip = true
+						break
+					}
+				}
+				if skip {
+					t.Run(name, func(t *testing.T) {
+						t.Skip("Test is on the skip list - ignoring")
+					})
+					continue
```
catsby, line 59: nit/suggestion: we could put a `Loop:` above the `range funcs` on line 42, then simplify this skip check to something like this:

```go
Loop:
  for _, tf := range funcs {
    // existing code
    // ...
    for _, skipTest := range skipTests {
      if name == skipTest {
        t.Run(name, func(t *testing.T) {
          t.Skip("Test is on the skip list - ignoring")
        })
        continue Loop
      }
    }
```

Just a suggestion though, take or leave as you will :)

pkg/serverstate/statetest/statetest.go
```diff
@@ -33,7 +33,10 @@ type (
 // Test runs a validation test suite for a state implementation. All
 // state implementations should pass this suite with no errors to ensure
 // the correct behavior of the state when Waypoint uses it.
-func Test(t *testing.T, f Factory, rf RestartFactory) {
+// skipTests are function names of tests in the serverstate package to skip
+// (i.e. TestJobCreate_singleton). It cannot skip sub-tests (called by
+// t.Run() inside a top-level test)
+func Test(t *testing.T, f Factory, rf RestartFactory, skipTests []string) {
```
catsby, line 39: How is `skipTests` populated? Could it be populated similar to how the current `tests` slice is?

catsby (commented):
Question on where `skipTests` comes from

pkg/serverstate/statetest/statetest.go
catsby: Ex: https://go.dev/play/p/QYk6_P15hdv

pkg/serverstate/statetest/statetest.go
izaaklauer: The workflow I envision is that you'd run the tests from a `_test` file in your project, see one that's failing, decide that it's OK from your perspective, and add that test name to the slice where you invoke this `Test` function.

catsby: 👍 for posterity, the `skipTests` slice would be added here: 

-https://github.com/hashicorp/waypoint/blob/d4245103aecefae370587efb92695398e21d1bac/internal/server/boltdbstate/state_test.go#L20

I think we need to add a `nil` or empty slice there, yeah?

pkg/serverstate/statetest/statetest.go
izaaklauer: Ha, I did think about that. I go back and forth on loop labels - they're convenient, but not my favorite for readability.

pkg/serverstate/statetest/statetest.go
```diff
@@ -42,6 +45,20 @@ func Test(t *testing.T, f Factory, rf RestartFactory) {
...
+				skip := false
+				for _, skipTest := range skipTests {
+					if name == skipTest {
+						skip = true
+						break
+					}
+				}
+				if skip {
+					t.Run(name, func(t *testing.T) {
+						t.Skip("Test is on the skip list - ignoring")
```
briancain, line 57: Can we print the name too? I think that will be helpful if anyone is ever confused why their test didn't run.

briancain (changes requested):

briancain (approved):
🧪 ✅

PR #3271, opened 2022-04-21
UI: rename `new` route to `create`

gregone:
To be more consistent with our other products, this PR renames `/new` routes to `/create`.


amyrlam (commented):
Code looks good. Have a non-blocking question, wasn't sure:

(question) Do you want to change the button to match and say "create"?

image

And maybe the breadcrumb too? (Figured the other other "new"s below made sense.)

image

For context, I just ran a `yarn start` from `/ui/` cause the Vercel is ignored. Seemed ok?

gregone:
@amyrlam Maybe, but that could be done separately, this was really more of a code concern than copywriting

amyrlam (approved):
Cool, makes sense, thanks for the explanation! :shipit:

PR #3278, opened 2022-04-22
server: Drive Task state via server state in the job system

briancain:
This commit is a follow-up to the feedback received in PR #3203. It takes the original implementation that drove the task state machine via the server API in the runners and instead manages task state internally via the job system when runners Ack and Complete jobs for on-demand runners. This is a bonus for Tasks in that we can rely on the internal system to drive the task state machine, rather than exposing `UpsertTask` and relying on the runners to make API requests to update the task state as it executes operations. Instead, we can rely on the existing functions `JobAck` and `JobComplete` to drive task state as the runner executes jobs.


internal/server/boltdbstate/job.go
```diff
@@ -749,6 +752,13 @@ func (s *State) JobAck(id string, ack bool) (*serverstate.Job, error) {
 	}
 
 	txn.Commit()
+
+	// Update the task state machine if acked job is part of an on-demand runner task.
+	if err := s.TaskAck(job.Id); err != nil {
```
mitchellh, line 757: Does this (and TaskComplete) need to be an exported API? Or can this just be a detail of the job system and unexported?

mitchellh (changes requested):
This looks way better.

My only ask is around whether `TaskAck` and `TaskComplete` should be exported at all or if they should be an implementation detail of the job system that is tested via the job APIs.

internal/server/boltdbstate/job.go
briancain: Agreed - Yeah this can live inside the job API details as unexported functions since they're an implementation detail of how the job API works! Moving now and updating my tests accordingly.

briancain: Covered in 3859b4a0f9fda70bbe8834735cde26d29979eca4 with some extra job tests for the task job triple lifecycle.

mitchellh (approved):
Great job!

PR #3283, opened 2022-04-25
Pipelines Data Model, State API, Upsert

mitchellh:
This PR is against a feature branch `f-pipelines` where I plan to stage all my work on pipelines so we don't force main to have this quite yet. This is the first PR.

This PR lays out the initial:

* Data model for pipelines and steps, covering a single type of step: arbitrary Docker execution
* State API for CRUD-ing pipelines
* Upsert gRPC API
* Validation for all of the above

Pipelines don't actually _do_ anything yet. This is just some initial data modeling work.

As a follow-up, I plan on getting step execution to work next and work my way out towards the HCL.


evanphx (approved):
Looks great! Using an explicit set-style for the Steps will definitely make it easier for future executors to reason about the steps more clearly.

briancain (approved):
🧪  looking good! We chatted about my feedback via slack, I mostly had questions. Excited to see this become functional 🚀

pkg/server/proto/server.proto
```diff
@@ -4344,6 +4375,124 @@ message EntrypointExecResponse {
...
+  // A single step within the pipeline.
+  message Step {
+    // Name of the step. This is unique within a pipeline. In the "steps"
+    // map, this must match the map key.
+    string name = 1;
+
+    // The steps that must come before this step. If this is empty,
+    // this is considered a "root" step. Only one step may be root.
+    // See "steps" on the Pipeline message for details.
+    repeated string depends_on = 2;
```
izaaklauer, line 4431: Thinking about this more, do you think we'll ever want more detail from the dependent stages? I'm imagining in the future we might want to depend on a stage even if it fails, or maybe explicitly map it's outputs to our inputs.

Something like this would be more flexible:
```
repeated DependentStep depends_on = 2;

message DependentStep {
  string name = 4;
  bool ignore_failure = 5;
  // ... other future ideas, like explicit output->input mapping
}

```

izaaklauer (changes requested):
The structure and tests all look great to me! One idea about future flexibility, but otherwise 🚀

pkg/server/proto/server.proto
mitchellh: We have some precedent for this already on jobs. The job dependencies are just specified in two separate `repeated string` fields. I do like your idea of making them structured, however. Let me think about this and I might make that change.

PR #3284, opened 2022-04-25
Upgrading to mockery v2.12.1

izaaklauer:
This fixes [this bug](https://github.com/vektra/mockery/issues/416) when running `make gen/server`.

In practice, the only thing that changes for us is the version in the mock comment headers.

This DOES mean that we all need to re-run `make tools`.


briancain (approved):
This makes sense. I think we should probably bump+re-gen the SDK as well, and maybe include that version bump in this PR before merge? Thoughts?

catsby (changes requested):
I'd like to discuss this internally briefly today before we merge this and other things. Thanks!

mitchellh (approved):

gregone:
We need to update https://github.com/hashicorp/waypoint/blob/main/ui/README.md?plain=1#L87  and make sure gen/ts still works

briancain:
Yep - we also discussed upgrading the version of Golang in both this repo and the SDK. So maybe we can use this PR for all of the version bumps?

izaaklauer:
Closing this in favor of this new PR that also upgrades us to 1.18: https://github.com/hashicorp/waypoint/pull/3286

PR #3285, opened 2022-04-26
Fix missing end quotes in configsourcer-vault.mdx

catsby:
Missing 2 end quotes in example hcls


briancain (approved):
Oof - good find! ❤️

briancain:
Ah, this might have to be fixed in the doc plugins, then generate the website mdx file with our make command?

catsby:
@briancain ah yeah good catch, I've updated the source file and ran `make gen/website-mdx`, please take another look!

PR #3286, opened 2022-04-26
Upgrade to go 1.18

izaaklauer:
This PR updates us to go 1.18. 

Closes https://github.com/hashicorp/waypoint/pull/3284


nix/overlay.nix
```diff
@@ -5,7 +5,7 @@ final: prev: {
   devShell = final.callPackage ./waypoint.nix { };
 
   # Need to manually do this since 1.16 is still the default
-  go = final.go_1_17;
+  go = final.go_1_18;
```
izaaklauer, line 8: @mitchellh Let me know if this isn't correct

mitchellh: Probably not, because we pin the index so 1.18 is prob not available. Just leave it out and I'll upgrade myself later. :)

izaaklauer:
Lets hold off on this for now - it looks like golang-ci doesn't fully support go 1.18 yet: https://github.com/golangci/golangci-lint/issues/2649

izaaklauer:
Superseded by the upgrade to go 1.19.

PR #3289, opened 2022-04-27
Docs: add allowed-redirect-url section

gregone:
Adds a note about using `-allowed-redirect-uri` when setting up UI-based OIDC auth. This was brought up in https://discuss.hashicorp.com/t/waypoint-oidc-issues/37428/5

### DIFF
--- a/website/content/docs/server/auth/oidc.mdx
+++ b/website/content/docs/server/auth/oidc.mdx
@@ -59,6 +59,17 @@ You should set the following redirect URLs:
   routable UI address you visit. And note you SHOULD set TLS (`https`) for
   this.
 
+To set the UI-based URL when setting up your auth method, add the `allowed-redirect-uri` flag:
+
+```
+$ waypoint auth-method set oidc \
+  -client-id=87a30dc693eb4db0a807c9faked4e0e643277f698aff83df5661021b6fc2d742 \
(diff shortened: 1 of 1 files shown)


website/content/docs/server/auth/oidc.mdx
```diff
@@ -59,6 +59,17 @@ You should set the following redirect URLs:
   routable UI address you visit. And note you SHOULD set TLS (`https`) for
   this.
 
+To set the UI-based URL when setting up your auth method, add the `allowed-redirect-uri` flag:
```
catsby, line 62:
```suggestion
To set the UI-based URL when setting up your auth method, add the `-allowed-redirect-uri` flag:
```

I did a quick check through some of our other docs and it seems we typically include the `-` when referencing a flag

catsby (approved):
👍 1 suggestion on adding `-`

.changelog/3289.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:docs
+Add a note about using `-allowed-redirect-uri` when setting up UI-based OIDC auth
```
briancain, line 2: I don't think we support a `docs` section in the changelog, and this is a website change rather than a change with Waypoint, so I don't think we need this changelog.

briancain (approved):
Looks good! I'd remove the changelog though

.changelog/3289.txt
gregone: oh thanks. I'll remove it then

PR #2439, opened 2021-10-07
ui: use `link:` for vendored/generated libs

jgwhite:
## Why the change?

Closes #1860

## Any other notes?

I read into the available dependency specifier types and discovered that `link:` does the thing we want: it creates a symlink to a local filepath.

[Check out the Yarn RFC](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-link-dependency-type.md)

In the video below I claim that this is exclusive to Yarn, but [apparently it works in NPM >= 5.0.0 now too](https://github.com/npm/npm/pull/15900#issuecomment-370948665), though appears to be undocumented.

## What does it look like?

https://user-images.githubusercontent.com/34030/136375067-235a740c-d0d5-4bdf-872a-f53a1656064a.mp4

## How do I test it?

1. `git checkout ui/yarn-link-vendored-libs`
2. `cd ui && yarn install`
2. Try making changes to `server.proto` and then running `make gen/ts`
3. Verify those changes are reflected in TS files that consume the protos

PR #3293, opened 2022-04-27
Set up UI toolkit

amyrlam:
## Description

_Draft, to be updated. Tagging you for reference, not for merge yet_

TODO:
- [x] Debug a new error now on `ember serve` (wasn't there before, and no changes?)
- [ ] Debug the CI error

Relates to: https://github.com/hashicorp/cloud-ui/pull/2444

## Screenshots


amyrlam:
At root, I had to `rm -rf node_modules` (not sure what happened in the yalc workflow to put them there, or maybe something else I did), as well as at `rm -rf ui/lib/waypoint-ui-toolkit/node_modules` level. So maybe this is the error @jgwhite what alluding to? Now an `ember serve` does work for me. Noting here so can better debug next time

jgwhite:
> At root, I had to `rm -rf node_modules` (not sure what happened in the yalc workflow to put them there, or maybe something else I did), as well as at `rm -rf ui/lib/waypoint-ui-toolkit/node_modules` level. So maybe this is the error @jgwhite what alluding to? Now an `ember serve` does work for me. Noting here so can better debug next time

Yeah, I should write up the exact nature of that bug. The tell tale error message is `module not found "@ember/object"` or something like that. `rm -rf ui/lib/waypoint-ui-toolkit/node_modules` will fix it. It relates to how version-detection for [ember-modules-api-polyfill](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill) works.

amyrlam:
@jgwhite I am stuck on this error and not sure how to get past it, any ideas? https://app.circleci.com/pipelines/github/hashicorp/waypoint/15459/workflows/e35797c6-8f4f-47f9-9475-7fba0e6a7d6a/jobs/161296?invite=true#step-104-25

If still stuck, I would go ahead and publish a 0.0.1-beta to npm (wild west pre-1.0 semver) to test in both here and other internal repo, and make the (belated) video on Monday. But let me know what you think, it's not 100% clear to me if you are cool with that path forward

amyrlam:
(Rebased to fix my Verified commits)

amyrlam:
From Slack, let's leave this open as a draft for now. We can explore publishing to npm with one real component instead of <Hello />, to get past the node_modules error in CI here, when pick up this task again.

jgwhite:
> Debug the CI error

This appears to be [a longstanding yarn bug](https://github.com/yarnpkg/yarn/issues/4563). Bizarrely, `yarn install` completes successfully if you run it again with no changes. 🤔

jgwhite:
After doing a bit of research, I think the best way to work around the Yarn bug is to upgrade to the latest stable (3.2.0 at the time of writing). We can break that out as a standalone task. I guess it also means adopting corepack. I guess we could _also_ consider migrating away from yarn to regular NPM.

amyrlam:
@jgwhite it's extremely hard to Google, but may need to keep Yarn for the `link:` functionality (HT to @gregone) vs npm file:

jgwhite:
> @jgwhite it's extremely hard to Google, but may need to keep Yarn for the `link:` functionality (HT to @gregone) vs npm file:

True. I remember reading somewhere that there’s a proposal for NPM to support `link:` too but my chances of finding the proposal are slim. It might be that `file:` + npm does the right thing. See #2439 for more detail on what “the right thing” is.

amyrlam:
close as stale, noted on internal ticket

PR #3297, opened 2022-04-28
docs: Document On-Demand Runners and Tasks

briancain:
This pull request builds out some basic on-demand runner documentation.

It also fixes a few bugs:

- Properly including the Nomad ODR docs
- Fixing a doc ref link for `security_context` in the kubernetes platform plugin.


website/content/docs/runner/on-demand-runner.mdx
```diff
@@ -0,0 +1,168 @@
...
+remotely. Through on-demand runner profiles and runner targeting, Waypoint can
+even utilize resources from other platforms that Waypoint has access to. For
+example, a local Kubernetes server install could be configured to use Amazon ECS
+to execute jobs.
+
+## Job Assignment
+
+When a job is submitted to the server, if on-demand runners are enabled, it
+will spawn an on-demand runner and the job is placed in a holding set. This is
+the same mechanism we use today for automatic CLI runners. When the on-demand
```
jgwhite, line 21: Given that these are stable docs rather than release notes, I’d consider dropping “today”.

```suggestion
the same mechanism we use for automatic CLI runners. When the on-demand
```

website/content/docs/runner/on-demand-runner.mdx
```diff
@@ -0,0 +1,168 @@
...
+runner starts, it will execute the normal Waypoint runner code and connect back
+to the server. On connection, the server will see the job in the holding set
+and send it immediately to the new runner.
+
+The flow of information about the job will use the existing grpc runner APIs so
+all the input and output remain the same as today.
```
jgwhite, line 27: Similar to my comment above, “as today” is good framing for release notes but less so for “timeless” docs. Perhaps contrast it with the CLI-runner again?

```suggestion
The flow of information about the job uses the normal grpc runner APIs so all
the input and output remain the same as if the work were performed by the
CLI-runner.
```

website/content/docs/runner/on-demand-runner.mdx
```diff
@@ -0,0 +1,168 @@
...
+- `StartTask`
+- `StopTask`
+
+These functions are what will launch and stop the spawned resource with the
+task launcher.
+
+Task plugins also can accept `TaskLauncherConfig` configuration which users can
+configure through an [on-demand runner profile](/docs/runner/profiles).
+
+## Task Introspection
```
jgwhite, line 97: Similar to @krantzinator’s comment in Slack, perhaps we should ship this change alongside 0.9? Readers of the documentation might not know when to expect 0.9, and therefore when to expect this feature.

jgwhite (commented):
This is awesome ❤️

website/content/docs/runner/on-demand-runner.mdx
briancain: Good call, agreed!

website/content/docs/runner/on-demand-runner.mdx
```diff
@@ -0,0 +1,96 @@
...
+Plugins expect to fulfill the SDK interface with the following functions:
+
+- `StartTask`
+- `StopTask`
+
+These functions are what will launch and stop the spawned resource with the
+task launcher.
+
+Task plugins also can accept `TaskLauncherConfig` configuration which users can
+configure through an [on-demand runner profile](/docs/runner/profiles).
```
jgwhite, line 96: I wonder if we could reframe this section as “Adding on-demand runner support for other platforms”?

```suggestion
## Adding on-demand runner support for other platforms

On-demand runner support can be implemented for additional platforms
using the [plugin SDK](/docs/extending-waypoint).

Plugins expect to fulfill the SDK interface with the following functions:

- `StartTask`
- `StopTask`

These functions are what will launch and stop the spawned resource with the
task launcher.

Task plugins also can accept `TaskLauncherConfig` configuration which users can
configure through an [on-demand runner profile](/docs/runner/profiles).
```

jgwhite (approved):
I’m gonna approve this because I think it’s already great as-is. I’ve made a few suggestions, but feel free to take-them-or-leave-them.

website/content/docs/runner/on-demand-runner.mdx
briancain: Good idea!

PR #3298, opened 2022-04-28
Add ability to embed and use OAuth creds in a waypoint token

evanphx:
This opens up the authentication modes for waypoint clients talking to a server to be able to delegate the authentication to an OAuth2 provider. This is only client support, the server does not yet expose this functionality, but it is available for future server implementations to use.


xiaolin-ninja:
Here is the [gist version](https://gist.github.com/xiaolin-ninja/1a0d47274e634c151ae0dc6ae3c2b58c) of the OAuth flow diagram above.

pkg/server/proto/server.proto
```diff
@@ -4489,6 +4489,10 @@ message Token {
     Trigger trigger = 10;
   }
 
+  // A client that sees this populated will use the details to fetch a token
+  // via oauth instead of submitting this token directly.
+  OAuthCredentials oauth_creds = 12;
```
izaaklauer, line 4494: I'm trying to think if this should be in an `ExternalAuthenticator` oneof to hedge against future non-oauth auth options, but I think this is fine.

internal/serverclient/token.go
```diff
@@ -62,5 +68,39 @@ func TokenFromContext(ctx context.Context) string {
 	}
 
 	return raw.(string)
+}
+
+const (
+	// tokenMagic is used as a byte sequence prepended to the encoded TokenTransport to identify
+	// the token as valid before attempting to decode it. This is mostly a nicity to improve
+	// understanding of the token data and error messages.
+	tokenMagic = "wp24"
```
izaaklauer, line 77: Not ideal to duplicate the token magic, but not the worst.

izaaklauer (approved):

briancain (approved):
Looks good! We went over this in a team zoom and I asked my questions there 😄

pkg/server/proto/server.proto
evanphx: Good call, did this!

internal/serverclient/token.go
evanphx: Going to leave the duplication for now, refactoring for this one thing is more ugly.

PR #3300, opened 2022-04-29
runner: static runners accept multiple jobs in parallel

mitchellh:
This modifies `internal/runner` to support accepting multiple jobs in
parallel. Not much work here since we always designed the runner struct
from the beginning to support this so there are no data races.

This modifies `internal/cli` so that runners in non-ODR run in parallel
mode by default. ODR doesn't make sense to have any parallelism since
they always run exactly one job. Non-ODR runners typically ONLY launch
ODR tasks, which are highly IO-bound, so we default to a multiple above
CPU count for concurrency.

This is a necessary pre-requisite for pipelines since they'll likely
perform blocking jobs on the static runners to "watch" tasks. Today,
tasks are launched and stopped, but not watched so this is not an issue.

### DIFF
--- a/internal/runner/accept.go
+++ b/internal/runner/accept.go
@@ -32,6 +32,7 @@ func (r *Runner) AcceptParallel(ctx context.Context, count int) {
 	defer cancel()
 
 	// Start up all the goroutines
+	r.logger.Info("accepting jobs concurrently", "count", count)
 	var wg sync.WaitGroup
 	wg.Add(count)
 	for i := 0; i < count; i++ {
--- a/internal/runner/accept.go
+++ b/internal/runner/accept.go
@@ -22,6 +22,30 @@ import (
 
 var heartbeatDuration = 5 * time.Second
 
+// AcceptParallel allows up to count jobs to be accepted and executing
+// concurrently.
+func (r *Runner) AcceptParallel(ctx context.Context, count int) {
+	// Create a new cancellable context so we can stop all the goroutines
+	// when one exits. We do this because if one exits, its likely that the
+	// unrecoverable error exists in all.
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Start up all the goroutines
+	var wg sync.WaitGroup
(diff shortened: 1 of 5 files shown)


izaaklauer (approved):
Wohoo!

briancain (approved):
Look great!

internal/runner/accept.go
```diff
@@ -22,6 +22,31 @@ import (
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Start up all the goroutines
+	r.logger.Info("accepting jobs concurrently", "count", count)
+	var wg sync.WaitGroup
+	wg.Add(count)
+	for i := 0; i < count; i++ {
+		go func() {
+			defer cancel()
```
evanphx, line 40: Just saying this out loud, should the exit of ANY of the goroutines cause them all to exit?

evanphx (approved):
Looks good, just one question about the semantics of canceling in the worker goroutine.

internal/runner/accept.go
mitchellh: So actually, yes I think this needs to do slightly better, although its edge case-y. Here is the thought:

1. If the user cancels the context, then this is effectively a no-op and all of the goroutines are canceled anyways. No issue.

2. If a goroutine has an error, its likely the error will impact all, because there is _considerable_ retry logic already in each `Accept` call -- including reconnection -- so if it actually errors it is likely unrecoverable. So we DO want to exit all the goroutines.

_However_, for #2, right now we're canceling the context which just causes a cascade effect to cancel ASAP. I think we can do better by just letting each existing job _try_ to finish gracefully, and then say "don't accept any more jobs thereafter." 

I'll work on this Monday.

mitchellh: Actually, I think this is okay for now. Its a bit non-trivial to get this fix in and looking at the possible reasons for a return from `AcceptMany`, i do think things are really broken if they exit so cancelling all is okay for now. We can improve this later. I've added a TODO to note it.

PR #3301, opened 2022-04-29
Extract most of the server handler test suite

izaaklauer:
Follows a very similar pattern to https://github.com/hashicorp/waypoint/pull/2623.

This extracts the state package test suite into a new package pkg/serverhandler/handlertest. For other waypoint server implementations, this allows them to run a broad implementation verification suite of tests to ensure they are functioning.

All I basically did was make our waypoint service tests generic, while keeping only the tests that specifically poke at the the singleprocess behavior (i.e. logs, exec, etc.) in the singleprocess package. Future service handler tests should make an effort to only use the protobuf server and client interfaces, and live inside handlertest.

This only picks the low-hanging fruit - some of the tests remaining in the singleprocess package can be refactored to be generic to any server implementation, and likely will be in the future.


pkg/server/singleprocess/service_logs_test.go
```diff
@@ -379,11 +379,12 @@ func TestServiceGetLogStream_depPlugin(t *testing.T) {
 	// Should have the job to spawn the logs plugin
 	var job *pb.Job
 	require.Eventually(t, func() bool {
-		jobs, err := testServiceImpl(impl).state(ctx).JobList(&pb.ListJobsRequest{})
+		jobsResp, err := client.ListJobs(ctx, &pb.ListJobsRequest{})
```
izaaklauer, line 382: This endpoint probably didn't exist when we wrote this test!

pkg/server/singleprocess/service_server_test.go
```diff
@@ -0,0 +1,69 @@
+package singleprocess
```
izaaklauer, line 1: These used to be in service_config_test.go, but they only call service_server apis.

pkg/serverhandler/doc.go
```diff
@@ -0,0 +1,9 @@
+// Package serverstate exports the verification harness for
+// implementing a new waypoint server protobuf (i.e. pb.WaypointServer)
+// implementations.
+//
+// Additional server implementations is not an officially supported extension
+// of Waypoint, so you do this at your own peril. It will not be supported.
+// However, this package is exported because we maintain our own private server
+// implementations for other use cases of Waypoint.
+package serverhandler
```
izaaklauer, line 9: Nothing in this package yet except the handler tests, but I think we'll put some interfaces in here before long to support some of the more complex tests that reach directly into state.

pkg/serverhandler/handlertest/test_service_deploy.go
```diff
@@ -0,0 +1,201 @@
...
+
+func init() {
+	tests["deploy"] = []testFunc{
+		TestServiceDeployment,
+		TestServiceDeployment_GetDeployment,
+		TestServiceDeployment_ListDeployments,
+	}
+}
+
+func TestServiceDeployment(t *testing.T, factory Factory) {
```
izaaklauer, line 24: These aren't new - they're from service_deploy_test.go. There was _one_ test over there that violated the interfaces, so I left it over there and brought the tests I could over here.

pkg/serverstate/serverstate.go
```diff
@@ -12,10 +12,8 @@ import (
 
 // Interface is the primary interface implemented by an implementation.
 //
-// This is an internal interface because (1) we don't expect or support
-// any external implementations and (2) we can absolutely change this interface
-// anytime we want or find it convenient, but we have to make sure so simultaneously
-// modify all our implementations.
+// Any changes to this interface will require changes to all
+// implementations in all projects.
```
izaaklauer, line 16: Updating this comment because serverstate isn't internal anymore.

mitchellh (approved):

PR #3305, opened 2022-05-02
contrib: Specify explicit Kind/K8s version in setup script 

briancain:
Due to a bug in Kaniko, there are issues with on-demand runners on the
latest version of Kubernetes, version 1.23. This commit updates the
contrib helper script to install 1.22 by default.

No changelog since this is a local dev helper script.


demophoon (approved):
Ran into this on Friday and this patch allowed my waypoint installation to build properly. Specifically when an ODR was spun up, kaniko was unable to detect that it was being ran inside of a docker container and would attempt to use the docker socket to build containers during the build phase when running `waypoint up`. Pinning to an earlier version of the Kind control plane fixed this issue locally for me.

Assuming the tests pass I'm 👍 for this update.

PR #3307, opened 2022-05-02
plugin/k8s: Add CPU/Memory resource limits to ODR

### LINKED ISSUE #3047: [plugin/k8s/task] Limit runner job resources
We are willing to build our containers using Waypoint runners instead of Github Actions but for now, we cannot limit the resource limit using Kubernetes runners.

We should be able to specify limits in the runner profile configuration.
It seems there is already a beginning of work [there](https://github.com/hashicorp/waypoint/blob/adfed1457920d547b51fee26d30c4b5d656014b1/builtin/k8s/task.go#L204-L210).

Creating a new plugin for that seems a bit over-engineered.

We are running several deployment pipelines on Github Actions and would like to migrate them on our Waypoint K8s cluster and limit the impact on the resources.

**Additional context**
N/A

demophoon:
This PR adds plumbing to enable users to add CPU and memory limits to on-demand runner profiles.

Closes #3047

### DIFF
--- a/.changelog/3307.txt
+++ b/.changelog/3307.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/k8s: Add CPU and memory resource limits to on-demand runners through runner profiles and at install time. These resource limits follow the same format that kubernetes expects within `spec.containers[].resources`.
+```
--- a/builtin/k8s/task.go
+++ b/builtin/k8s/task.go
@@ -64,6 +64,12 @@ type TaskLauncherConfig struct {
 
 	// The namespace to use for launching this task in Kubernetes
 	Namespace string `hcl:"namespace,optional"`
+
+	// Optionally define various cpu resource limits and requests for kubernetes pod containers
+	CPU *ResourceConfig `hcl:"cpu,block"`
+
+	// Optionally define various memory resource limits and requests for kubernetes pod containers
+	Memory *ResourceConfig `hcl:"memory,block"`
 }
 
 func (p *TaskLauncher) Documentation() (*docs.Documentation, error) {
@@ -118,6 +124,16 @@ task {
 			"service account is the name of the Kubernetes service account to add to the pod.",
 			"This is useful to apply Kubernetes RBAC to the application.",
 		),
+	)
+
+	doc.SetField(
+		"memory",
+		"memory resource request to be added to the task container",
+	)
+
+	doc.SetField(
+		"cpu",
+		"cpu resource request to be added to the task container",
 	)
 
 	doc.SetField(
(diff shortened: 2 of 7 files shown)


demophoon:
I've verified this change with a local kind cluster making sure that `-k8s-cpu-request` and `-k8s-mem-request` get set in the default `kubernetes` runner profile and that any on-demand runners that are spun up from the installation respect the limit and request configuration within the runner profile.

builtin/k8s/task.go
```diff
@@ -64,6 +64,12 @@ type TaskLauncherConfig struct {
 
 	// The namespace to use for launching this task in Kubernetes
 	Namespace string `hcl:"namespace,optional"`
+
+	// Optionally define various cpu resource limits and requests for kubernetes pod containers
+	CPU *ResourceConfig `hcl:"cpu,block,optional"`
+
+	// Optionally define various memory resource limits and requests for kubernetes pod containers
+	Memory *ResourceConfig `hcl:"memory,block,optional"`
```
briancain, line 72: So I think this is wrong, if we have a `block` it can be `optional` if the struct fields are also optional. Otherwise I think this is a parse error and we want this instead:

```suggestion
	Memory *ResourceConfig `hcl:"memory,block"`
```

Same with `CPU` too.

```
2022-05-04T15:20:49.722Z [DEBUG] waypoint.runner.agent.runner: plugin successfully launched and connected: job_id=01G27TG44Y9EVXBS6N6BPMMBFW job_op=*gen.Job_StartTask
2022-05-04T15:20:49.722Z [DEBUG] waypoint.runner.agent.runner: decoding task plugin config: job_id=01G27TG44Y9EVXBS6N6BPMMBFW job_op=*gen.Job_StartTask
panic: invalid hcl field tag kind "block,optional" on *struct { Requested string "hcl:\"request,optional\""; Limit string "hcl:\"limit,optional\"" } "CPU"

goroutine 54 [running]:
github.com/hashicorp/hcl/v2/gohcl.getFieldTags({0x3a07760, 0xc000a791a0})
        /go/pkg/mod/github.com/hashicorp/hcl/v2@v2.10.1-0.20210621220818-327f3ce2570e/gohcl/schema.go:176 +0x625
...
...
```

briancain (changes requested):
This looks awesome @demophoon !! The majority of my PR comments are super minor and I added some suggestions via github: it's good to use `%q` to quote variable values inside a string if they're mid-sentence for the nil case. (i.e. "failed to parse "" to quantity" versus "failed to parse  to quantity")

The only issue I found was if you don't ever specify CPU or Mem we get an HCL config parse error, which is related to the `hcl` tag on those values in config.

builtin/k8s/task.go
demophoon: Ah, gotcha. I had only added this because it seems like the docs site keys off of `optional` to determine whether or not the parameter shows up in the required parameters field or not. Is there a way for the auto gen docs to place these parameters into the optional section instead of marking them both as required?

evanphx: Also, all blocks are optional.

.changelog/3307.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+plugin/k8s: Add CPU and memory resource limits to on-demand runners through runner profiles and at install time
```
xiaolin-ninja, line 2: Let's specify that these are the same units as what the K8s API expects so there's no confusion.

builtin/k8s/task.go
```diff
@@ -254,6 +270,47 @@ func (p *TaskLauncher) StartTask(
...
+					status.Errorf(codes.InvalidArgument, "failed to parse cpu request %s to k8s quantity: %s", p.config.CPU.Requested, err)
+			}
+			resourceRequests[corev1.ResourceCPU] = q
+		}
+
+		if p.config.CPU.Limit != "" {
+			q, err := k8sresource.ParseQuantity(p.config.CPU.Limit)
+			if err != nil {
+				return nil,
+					status.Errorf(codes.InvalidArgument, "failed to parse cpu limit %s to k8s quantity: %s", p.config.CPU.Limit, err)
```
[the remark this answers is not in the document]
xiaolin-ninja: note from Brian: %q will put quotes around Nil return values.

.changelog/3307.txt
demophoon: Thanks for the feedback!

Addressed in `a474021`

builtin/k8s/task.go
```diff
@@ -64,6 +64,12 @@ type TaskLauncherConfig struct {
 
 	// The namespace to use for launching this task in Kubernetes
 	Namespace string `hcl:"namespace,optional"`
+
+	// Optionally define various cpu resource limits and requests for kubernetes pod containers
+	CPU *ResourceConfig `hcl:"cpu,block,optional"`
```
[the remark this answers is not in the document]
demophoon: This has been reverted in `10cffbf`

website/package-lock.json
```diff
@@ -1255,10 +1255,14 @@
       "version": "12.1.0",
       "resolved": "https://registry.npmjs.org/@next/swc-android-arm64/-/swc-android-arm64-12.1.0.tgz",
       "integrity": "sha512-/280MLdZe0W03stA69iL+v6I+J1ascrQ6FrXBlXGCsGzrfMaGr7fskMa0T5AhQIVQD4nA/46QQWxG//DYuFBcA==",
-      "cpu": ["arm64"],
```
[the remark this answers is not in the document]
demophoon: Reverted in `b8d55db7e`

briancain (approved):
💥 🚀

briancain:
We should backport this for the next minor release, all you need to do is apply the `backport/0.8.x` label for that to work! ✨

PR #3308, opened 2022-05-03
Update release metadata file names

sarahethompson:
Since we are releasing both waypoint and waypoint-entrypoint from this repo, we need separate release metadata files. See - https://github.com/hashicorp/waypoint-releases/pull/58.


briancain (approved):
Thanks! Looks good. I rekicked the integration test failure, looks like a circle issue.

PR #3317, opened 2022-05-05
Hide dynamic year from Percy

amyrlam:
Prevent screenshot diffs when the year changes

Link in comment: https://docs.percy.io/docs/percy-specific-css#percy-css-media-query

See the current Percy diffs to see that the year is now hidden from Percy, which seems a-ok, as not super impt to be Percy'd?


jgwhite (approved):
Super cool 👍

Issue #2697, opened 2021-11-08
Nomad ODRs batch jobs


paladin-devops:
**Is your feature request related to a problem? Please describe.**
Nomad ODR jobs live beyond the completion of their assigned tasks.

**Describe the solution you'd like**
Convert the job type of a Nomad ODR to "batch" instead of "service". 

**Additional context**
Although the service job type may not be the root cause, during testing with Nomad ODRs, the jobs still exist in Nomad after the conclusion of the task which had been assigned to them, leading to stale allocations running in Nomad. Running Nomad's garbage collection also did not clean them up. By converting from service to batch, this concern may be taken care of automatically.

briancain:
Makes sense to me 👍🏻 I also think given the nature of short lived on-demand tasks in Waypoint, a batch job makes more sense than a service job.

paladin-devops:
I was thinking more about this and maybe parameterized batch jobs could help here. I imagine that each Waypoint runner Nomad job will be pretty similar to the others, and registering a parameterized job and providing a payload or metadata to that job could be sufficient enough for a new runner job to spawn and run the right task(s).

https://www.nomadproject.io/docs/job-specification/parameterized

From what I can tell, the way the runner behaves currently is that it starts up and then waits/listens for a task from the server, so switching to batch would be not-a-small change in approach, but I think if a parameterized job is templated properly that could be a viable way to go about this!

briancain:
Thanks for the info @paladin-devops ! What are the advantages of a parameterized job when we can just make the Task plugin create a batched job with inputs already? https://github.com/hashicorp/waypoint/blob/main/builtin/nomad/task.go#L196

I'm thinking we just replace this with creating a batch job instead. Since we control the Nomad task plugin, it's essentially a parameterized batch job without needing to register the job to a users Nomad clusters. I might be missing something here though 😄

evanphx:
@paladin-devops Off hand, which nomad version are you testing this on?

paladin-devops:
@evanphx I'm on Nomad v1.1.6

& @briancain that makes sense! You're hitting the Nomad API directly with whatever needs to be passed in to the job, and anything outside of the Waypoint server most likely isn't ever going to need to dispatch a parameterized job anyways.

paladin-devops:
Here's what my Nomad cluster currently looks like btw, with most of the Nomad ODRs generated by my Waypoint server:

[...]

PR #3319, opened 2022-05-09
Fix "Nomad ODR jobs live beyond the completion of their assigned task…

pam199968:
Convert the job type of a Nomad ODR to "batch" instead of "service". (#2697)
Now the ODR jobs are automatically stopped after completion and removed by the GC.


evanphx:
I think for now it's better to have it be a service because it will be restarted if the allocation fails. We're now clearing up the tasks by calling StopTask, which will deregister the service when it's done.

PR #3321, opened 2022-05-09
internal/core: Introduce `pipeline` and `step` parsing and validation for HCL config

briancain:
This pull request implements an internal config parser and validation func for loading `pipeline` and its `step` stanzas for a `waypoint.hcl` config.

There are a few TODOs for how we intend to load Step Use labels and plugin names for plugin execution. These are left as-is for now with some comments, because this PR does not handle Step plugin execution. This will be covered in future PRs.

### DIFF
--- a/internal/config/operation.go
+++ b/internal/config/operation.go
@@ -29,6 +29,10 @@ func (op *Operation) Configure(plugin interface{}, ctx *hcl.EvalContext) hcl.Dia
 	return component.Configure(plugin, op.Use.Body, ctx)
 }
 
+func (b *Step) Operation() *Operation {
+	return mapoperation(b, true)
+}
+
 func (b *Build) Operation() *Operation {
 	return mapoperation(b, true)
 }
--- a/internal/config/operation.go
+++ b/internal/config/operation.go
@@ -29,10 +29,6 @@ func (op *Operation) Configure(plugin interface{}, ctx *hcl.EvalContext) hcl.Dia
 	return component.Configure(plugin, op.Use.Body, ctx)
 }
 
-func (b *Step) Operation() *Operation {
-	return mapoperation(b, true)
-}
-
 func (b *Build) Operation() *Operation {
 	return mapoperation(b, true)
 }
--- a/internal/config/pipeline.go
+++ b/internal/config/pipeline.go
@@ -113,13 +113,9 @@ func (c *Pipeline) Step(ctx *hcl.EvalContext) ([]*Step, error) {
 }
 
 // StepUse returns the plugin "use" value for a single step
-// TODO: Since pipelines can have N steps, where each step could be its own
-// plugin, how does this Use func work? It looks like the component creator
-// is what uses this to invoke the proper plugin, so it's expecting a single
-// plugin per stanza with build/deploy/release. But pipelines have many steps,
-// so maybe a component creator refactor is required to support multiple
-// plugins? For now I'm leaving this function commented out with its original
-// single use-stanza implementation in place.
+// TODO(briancain): We will leave this unimplemented for now  until we start
+// actually executing Step plugins. We will execute each step on their own,
+// so we'll probably need a single Step to return the plugin Use label.
 func (c *Pipeline) StepUse(ctx *hcl.EvalContext) (string, error) {
 	return "", nil
 	/*
@@ -144,6 +140,7 @@ func (c *Pipeline) StepUse(ctx *hcl.EvalContext) (string, error) {
 // every step plugin type.
 // NOTE(briancain): We could gather all of the use plugin labels this way and
 // return them as a string?
+// Nothing uses this right now, but is included for now.
 func (c *Pipeline) StepsUse(ctx *hcl.EvalContext) ([]string, error) {
 	if len(c.StepRaw) == 0 {
 		return nil, nil
--- a/internal/config/pipeline.go
+++ b/internal/config/pipeline.go
@@ -67,9 +67,9 @@ func (c *Config) Pipeline(id string, ctx *hcl.EvalContext) (*Pipeline, error) {
 	pipeline.Name = rawPipeline.Name
 	pipeline.ctx = ctx
 	pipeline.config = c
-	// TODO: config overrides?
-
-	// TODO: is this were we set step order?
+	if pipeline.config != nil {
+		pipeline.config.ctx = ctx
+	}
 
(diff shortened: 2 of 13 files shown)


internal/config/pipeline.go
```diff
@@ -0,0 +1,194 @@
...
+
+	Body hcl.Body `hcl:",body"`
+
+	ctx    *hcl.EvalContext
+	config *Config
+}
+
+type hclPipeline struct {
+	Id   string `hcl:",label"`
+	Name string `hcl:"name,optional"`
```
mitchellh, line 25: Similar feedback here to `Name` on Pipeline.

internal/config/pipeline.go
```diff
@@ -0,0 +1,194 @@
...
+		Ref: &pb.Ref_Pipeline_Id{
+			Id: &pb.Ref_PipelineId{
+				Id: c.Id,
+			},
+		},
+	}
+}
+
+// Step loads the associated section of the configuration
+func (c *Pipeline) Step(ctx *hcl.EvalContext) ([]*Step, error) {
```
mitchellh, line 89: I think we can parse all the steps when we parse the pipeline, no need to lazily do it.

internal/config/pipeline.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return steps, nil
+}
+
+// StepUse returns the plugin "use" value for a single step
+// TODO(briancain): We will leave this unimplemented for now  until we start
+// actually executing Step plugins. We will execute each step on their own,
+// so we'll probably need a single Step to return the plugin Use label.
+func (c *Pipeline) StepUse(ctx *hcl.EvalContext) (string, error) {
```
mitchellh, line 119: I don't think you need this, `StepsUse` or `StepsLabels`

internal/config/plugin.go
```diff
@@ -17,6 +17,8 @@ type Plugin struct {
 		Registry bool `hcl:"registry,optional"`
 		Platform bool `hcl:"deploy,optional"`
 		Releaser bool `hcl:"release,optional"`
+		Pipeline bool `hcl:"pipeline,optional"`
+		Step     bool `hcl:"step,optional"`
```
mitchellh, line 21: These should not be here, we're not doing any plugin types yet and these aren't what they will be called.

internal/config/stages.go
```diff
@@ -41,6 +41,34 @@ type scopedStage struct {
 	Remain hcl.Body `hcl:",remain"`
 }
 
+// Step are the step settings for pipelines
+type Step struct {
+	Labels map[string]string `hcl:"labels,optional"`
+	Hooks  []*Hook           `hcl:"hook,block"`
```
mitchellh, line 47: No labels or hooks.

internal/config/stages.go
```diff
@@ -41,6 +41,34 @@ type scopedStage struct {
...
+	Use    *Use              `hcl:"use,block"`
+
+	// Optionally give this step a name
+	// TODO: figure out how to make this an optional label rather than
+	// a variable for a Step stanza
+	Name string `hcl:"name,optional"`
```
mitchellh, line 53: Your TODO :)

internal/config/stages.go
```diff
@@ -41,6 +41,34 @@ type scopedStage struct {
...
+
+	// If set, this step will depend on the defined step. The default step
+	// will be the previously defined step in order that it was defined
+	// in a waypoint.hcl
+	DependsOn string `hcl:"depends_on,optional"`
```
mitchellh, line 58: `[]string` type, can depend on multiple.

internal/config/stages.go
```diff
@@ -41,6 +41,34 @@ type scopedStage struct {
...
+
+	// The OCI image to use for executing this step
+	ImageURL string `hcl:"image_url,optional"`
+
+	// Unused for practical reasons, but we need this here so that
+	// the decoding validates successfully (HCL doesn't error of
+	// unexpected things).
+	WorkspaceScoped []*scopedStage `hcl:"workspace,block"`
+	LabelScoped     []*scopedStage `hcl:"label,block"`
```
mitchellh, line 67: Not required for pipeline steps.

internal/core/project.go
```diff
@@ -28,8 +28,9 @@ import (
 // The Close function should be called when finished with the project
 // to properly clean up any open resources.
 type Project struct {
-	logger    hclog.Logger
-	apps      map[string]*App
+	logger hclog.Logger
+	apps   map[string]*App
+	//pipelines map[string]*Pipeline internal/core/pipeline.go needs to exist
```
mitchellh, line 33: I'd drop these temporary comments for now.

internal/config/operation.go
```diff
@@ -29,6 +29,10 @@ func (op *Operation) Configure(plugin interface{}, ctx *hcl.EvalContext) hcl.Dia
 	return component.Configure(plugin, op.Use.Body, ctx)
 }
 
+func (b *Step) Operation() *Operation {
```
mitchellh, line 32: Steps aren't operations so this shouldn't be needed.

internal/config/stages.go
```diff
@@ -29,6 +29,12 @@ type hclBuild struct {
 	LabelScoped []*scopedStage `hcl:"label,block"`
 }
 
+type hclStep struct {
```
mitchellh, line 32: I think this should go in `pipeline.go`

mitchellh (commented):
Just looking after your recent commits and def heading in the right direction!

internal/config/stages.go
briancain: Agreed! Got that moved in 1f12ea6923c56433eb71c547e1f1a57e22a40bb4

internal/config/pipeline.go
```diff
@@ -0,0 +1,114 @@
...
+
+	Body   hcl.Body `hcl:",body"`
+	Remain hcl.Body `hcl:",remain"`
+}
+
+// hclStep represents a raw HCL version of a step stanza in a pipeline config
+type hclStep struct {
+	Use    *Use     `hcl:"use,block"`
+	Body   hcl.Body `hcl:",body"`
+	Remain hcl.Body `hcl:",remain"`
```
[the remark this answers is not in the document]
briancain: Sounds good, pushed up a commit to address this, and made step labels required for now.

mitchellh (approved):
Let's go!

PR #3334, opened 2022-05-13
internal/core: Add ConfigSync for pipeline configs

briancain:
This commit introduces a way for runners to initiate a pipeline config
sync. It loads the HCL eval context, evals the config, and translates the
parsed pipeline struct into a Proto Pipeline that is upserted to the server.

I've got a couple of questions that I'll ask as a PR comment, otherwise
this PR is ready for review!

### DIFF
--- a/internal/config/pipeline.go
+++ b/internal/config/pipeline.go
@@ -4,7 +4,9 @@ import (
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/gohcl"
 	"github.com/hashicorp/waypoint-plugin-sdk/component"
+	"github.com/zclconf/go-cty/cty/function"
 
+	"github.com/hashicorp/waypoint/internal/config/dynamic"
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
 )
 
@@ -16,7 +18,7 @@ type Pipeline struct {
 	Steps   []*Step
 
 	ctx    *hcl.EvalContext
-	config *Config
+	Config *Config
 }
 
 // hclPipeline represents a raw HCL version of a pipeline config
@@ -80,9 +82,9 @@ func (c *Config) Pipeline(id string, ctx *hcl.EvalContext) (*Pipeline, error) {
 	}
 	pipeline.Name = rawPipeline.Name
 	pipeline.ctx = ctx
-	pipeline.config = c
-	if pipeline.config != nil {
-		pipeline.config.ctx = ctx
+	pipeline.Config = c
+	if pipeline.Config != nil {
+		pipeline.Config.ctx = ctx
 	}
 
 	// decode all of the defined raw steps for a pipeline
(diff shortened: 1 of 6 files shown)


internal/core/pipeline_config.go
```diff
@@ -0,0 +1,33 @@
...
+	// TODO(briancain): In the future, we can sync config vars for pipelines here
+
+	// Sync the pipeline metadata
+	p.logger.Debug("evaluating pipeline configs for syncing")
+	pipelines, err := p.config.Config.PipelineProtos()
+	if err != nil {
+		return err
+	}
+
+	// TODO(briancain): do we need a Multi upsert?
```
briancain, line 21: Seems like we might need a multi-pipeline Upsert? This could end up being a lot of API calls for a single config 🤔

internal/config/pipeline_test.go
```diff
@@ -128,6 +128,71 @@ func TestPipeline(t *testing.T) {
 	}
 }
 
+// This test is a little weird. We parse a pipeline, but only to use its config
+// to get to the HCL Eval context for loading all pipelines and generating the
+// Proto slice. This test intends to show that if you eval an HCL context,
+// the PipelineProtos func will load those and convert them to Proto Pipelines.
+// So there's probably a better way to get a pipeline config struct to access
+// PipelineProtos()?
+func TestPipelineProtos(t *testing.T) {
```
briancain, line 137: This test makes me think it might make more sense to have the pipeline `ConfigSync` take an `*App` pointer receiver instead? I wasn't able to get at `PipelineProtos()` when I did that though, so maybe this is ok for now. It might just depend on what the runner has access to when it attempts to call `ConfigSync`?

internal/config/pipeline.go
```diff
@@ -104,6 +106,58 @@ func (c *Config) Pipeline(id string, ctx *hcl.EvalContext) (*Pipeline, error) {
...
+// operations such as ConfigSync.
+func (c *Config) PipelineProtos() ([]*pb.Pipeline, error) {
+	if c == nil {
+		return nil, nil
+	}
+
+	// Build our evaluation context for the config vars
+	ctx := c.ctx
+	ctx = appendContext(ctx, &hcl.EvalContext{
+		Functions: dynamic.Register(map[string]function.Function{}),
```
mitchellh, line 120: Not sure pipelines should have access to the dynamic functions.

internal/core/pipeline.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/hashicorp/go-hclog"
+
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/config"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+// Pipeline represents a single pipeline and exposes all of the operations
+// that can be performed on a pipeline.
+type Pipeline struct {
```
mitchellh, line 15: Oh okay I actually like this model! (Making it a thing so you can do `core.pipeline.sync` etc.)

internal/config/pipeline_test.go
mitchellh: The way I'd approach this instead is probably to pass `*Config` into the `Func` field rather than `Pipeline`. Then you just call `c.PipelineProtos()` and assert on that. The rest looks fine to me...

mitchellh (changes requested):

internal/config/pipeline.go
```diff
@@ -104,6 +106,58 @@ func (c *Config) Pipeline(id string, ctx *hcl.EvalContext) (*Pipeline, error) {
 	return &pipeline, nil
 }
 
+// PipelineProtos will take the existing HCL eval context, eval the config
+// and translate the HCL result into a Pipeline Proto to be returned for
+// operations such as ConfigSync.
+func (c *Config) PipelineProtos() ([]*pb.Pipeline, error) {
+	if c == nil {
+		return nil, nil
```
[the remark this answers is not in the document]
briancain: Panic sounds good, this does seem like an internal error 👍🏻

PR #3335, opened 2022-05-13
New `runner install` command

paladin-devops:
This PR introduces a new command to the Waypoint CLI, `waypoint runner install`. This command installs a Waypoint runner to the specified platform based on flags passed to the CLI. The supported platforms include Docker, ECS, Kubernetes (which uses Helm under the hood), and Nomad.

By default, this command will "adopt" the new runner by checking the list of runners for the newly installed runner's ID, and then adopting it when detected. Following that, a runner profile is created with the target runner ID specified as the new runner's ID, making the new runner immediately capable of being targeted for ODR usage.

A new Go package has also been added, `installutil`, for code that is shared by the server install and runner install processes.

The diff is quite large, in part due to that a great deal of code formerly in the `internal/serverinstall` package has been refactored to the new `internal/installutil` package. Code that is shared by both `runner install` and `server install` exists in `internal/installutil`. Other code that is used exclusively by one or the other remains in the respective package.

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -505,7 +505,9 @@ func installRunner(
 		AuthToken:     resp.Token,
 	}
 
-	// We set the ID to be "init" since it is the initial static runner
+	// We set the ID to be "static" since it is the initial static runner
+	// Specific platform implementations should add the suffix -runner to
+	// resource names
 	id := "static"
 
 	// Install!
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -534,8 +534,7 @@ func installRunner(
 
 	err = installutil.AdoptRunner(ctx, ui, client, id, advertiseAddr.Addr)
 	if err != nil {
-		s.Update("Error adopting runner: %s", err)
-		s.Status(terminal.StatusError)
+		ui.Output("Error adopting runner: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 		return 1
 	}
 
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -530,6 +530,7 @@ func installRunner(
 		)
 		return 1
 	}
+	s.Update("Runner %q installed", id)
 	s.Done()
 
 	err = installutil.AdoptRunner(ctx, ui, client, id, advertiseAddr.Addr)
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -238,15 +238,8 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 			return 1
 		}
 
-		s = sg.Add("Creating runner profile and targeting runner %s", strings.ToUpper(id))
-		var runnerProfileOdrImage string
-		if c.runnerProfileOdrImage == "" {
-			runnerProfileOdrImage = "hashicorp/waypoint-odr:latest"
-		} else {
-			runnerProfileOdrImage = c.runnerProfileOdrImage
-		}
-
 		// Creating a new runner profile for the newly adopted runner
+		s = sg.Add("Creating runner profile and targeting runner %s", strings.ToUpper(id))
 		runnerProfile, err := client.UpsertOnDemandRunnerConfig(ctx, &pb.UpsertOnDemandRunnerConfigRequest{
 			Config: &pb.OnDemandRunnerConfig{
 				Name: c.platform[0] + "-" + strings.ToUpper(id),
@@ -257,7 +250,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 						},
 					},
 				},
-				OciUrl:               runnerProfileOdrImage,
+				OciUrl:               c.runnerProfileOdrImage,
 				EnvironmentVariables: nil,
 				PluginType:           c.platform[0],
 				PluginConfig:         nil,
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -61,7 +61,7 @@ func (c *RunnerInstallCommand) Flags() *flag.Sets {
 
 		f.StringVar(&flag.StringVar{
 			Name:    "odr-image",
-			Usage:   "Docker image for the on-demand runners. If not specified, it defaults to 'hashicorp/waypoint-odr:latest'.",
+			Usage:   "Docker image for the on-demand runners.",
 			Default: "hashicorp/waypoint-odr:latest",
 			Target:  &c.runnerProfileOdrImage,
 		})
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -250,12 +250,8 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 						},
 					},
 				},
-				OciUrl:               c.runnerProfileOdrImage,
-				EnvironmentVariables: nil,
-				PluginType:           c.platform[0],
-				PluginConfig:         nil,
-				ConfigFormat:         0,
-				Default:              false,
+				OciUrl:     c.runnerProfileOdrImage,
+				PluginType: c.platform[0],
 			},
 		})
 		if err != nil {
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -112,9 +112,12 @@ func (c *RunnerInstallCommand) Help() string {
 	return formatHelp(`
 Usage: waypoint runner install [options]
 
-  Install a Waypoint runner to the specified platform: kubernetes, nomad, ecs, or docker.
+  Install a Waypoint runner to the specified platform: kubernetes, nomad, ecs, 
+  or docker.
 
-  This command will attempt to install a runner for the server configured in the current Waypoint context. It will adopt the runner after installation, unless the `-adopt` flag is set to false.
+  This command will attempt to install a runner for the server configured in 
+  the current Waypoint context. It will adopt the runner after installation, 
+  unless the '-adopt' flag is set to false.
 
 ` + c.Flags().Help())
 }
@@ -219,7 +222,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		)
 		return 1
 	}
-	s.Update("Runner %s installed successfully to %s", id, platform[0])
+	s.Update("Runner %q installed successfully to %s", id, platform[0])
 	s.Status(terminal.StatusOK)
 	s.Done()
 
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -189,6 +189,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	}
 
 	client := c.project.Client()
+	s.Update("Finished connecting to: %s", c.serverUrl)
 	s.Status(terminal.StatusOK)
 	s.Done()
 
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -194,16 +194,14 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 
 	// We generate the ID if the user doesn't provide one
 	// This ID is used later to adopt the runner
-	var id string
+	id := c.id
 	var err error
-	if c.id == "" {
+	if id == "" {
 		id, err = server.Id()
 		if err != nil {
 			c.ui.Output("Error generating runner ID: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
 		}
-	} else {
-		id = c.id
 	}
 
 	s = sg.Add("Installing runner...")
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -115,9 +115,9 @@ Usage: waypoint runner install [options]
   Install a Waypoint runner to an existing platform. The platform should be
   specified as kubernetes, nomad, ecs, or docker.
 
-  By default, this will adopt the runner after it is installed. The install will
-  attempt to install a runner for the server configured in the current Waypoint 
-  context.
+  With the '-adopt' flag set to true (the default), this command will adopt the
+  runner after it is installed. The install will attempt to install a runner for
+  the server configured in the current Waypoint context.
 
 ` + c.Flags().Help())
 }
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -112,12 +112,9 @@ func (c *RunnerInstallCommand) Help() string {
 	return formatHelp(`
 Usage: waypoint runner install [options]
 
-  Install a Waypoint runner to an existing platform. The platform should be
-  specified as kubernetes, nomad, ecs, or docker.
+  Install a Waypoint runner to the specified platform: kubernetes, nomad, ecs, or docker.
 
-  With the '-adopt' flag set to true (the default), this command will adopt the
-  runner after it is installed. The install will attempt to install a runner for
-  the server configured in the current Waypoint context.
+  This command will attempt to install a runner for the server configured in the current Waypoint context. It will adopt the runner after installation, unless the `-adopt` flag is set to false.
 
 ` + c.Flags().Help())
 }
@@ -257,7 +254,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 			)
 			return 1
 		}
-		s.Update("Runner profile %s created successfully.", runnerProfile.Config.Name)
+		s.Update("Runner profile %q created successfully.", runnerProfile.Config.Name)
 		s.Status(terminal.StatusOK)
 		s.Done()
 		return 0
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -22,7 +22,7 @@ type RunnerInstallCommand struct {
 	adopt                 bool     `hcl:"adopt,optional"`
 	serverUrl             string   `hcl:"server_url,required"`
 	id                    string   `hcl:"id,optional"`
-	runnerProfileOdrImage string   `hcl:"id,optional"`
+	runnerProfileOdrImage string   `hcl:"odr-image,optional"`
 	serverTls             bool     `hcl:"server_tls,optional"`
 	serverTlsSkipVerify   bool     `hcl:"server_tls_skip_verify,optional"`
 	serverRequireAuth     bool     `hcl:"server_require_auth,optional"`
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -18,11 +18,11 @@ import (
 type RunnerInstallCommand struct {
 	*baseCommand
 
-	platform              []string `hcl:"platform,required"`
+	platform              []string `hcl:"platform,optional"`
 	adopt                 bool     `hcl:"adopt,optional"`
 	serverUrl             string   `hcl:"server_url,required"`
 	id                    string   `hcl:"id,optional"`
-	runnerProfileOdrImage string   `hcl:"odr-image,optional"`
+	runnerProfileOdrImage string   `hcl:"odr_image,optional"`
 	serverTls             bool     `hcl:"server_tls,optional"`
 	serverTlsSkipVerify   bool     `hcl:"server_tls_skip_verify,optional"`
 	serverRequireAuth     bool     `hcl:"server_require_auth,optional"`
@@ -47,7 +47,7 @@ func (c *RunnerInstallCommand) Flags() *flag.Sets {
 
 		f.EnumVar(&flag.EnumVar{
 			Name:   "platform",
-			Usage:  "Platform to install the Waypoint runner into.",
+			Usage:  "Platform to install the Waypoint runner into. If unset, uses the platform of the local context.",
 			Values: runnerPlatforms,
 			Target: &c.platform,
 		})
@@ -141,22 +141,29 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	log := c.Log.Named("install")
 	defer c.Close()
 
-	if len(c.platform) == 0 {
+	serverConfig, err := c.project.Client().GetServerConfig(ctx, &empty.Empty{})
+	if err != nil {
 		c.ui.Output(
-			"The -platform flag is required.",
+			"Error getting server config.",
+			clierrors.Humanize(err),
 			terminal.WithErrorStyle(),
 		)
-		c.ui.Output(c.Flags().Help(), terminal.WithInfoStyle())
 		return 1
 	}
-	p, ok := runnerinstall.Platforms[strings.ToLower(c.platform[0])]
+
+	// If the user doesn't set a platform, set platform to the platform of the user's context
+	platform := c.platform
+	if len(platform) == 0 {
+		platform = append(platform, serverConfig.Config.Platform)
+	}
+
+	p, ok := runnerinstall.Platforms[strings.ToLower(platform[0])]
 	if !ok {
 		c.ui.Output(
 			"Error installing server into %q: unsupported platform",
-			c.platform,
+			platform[0],
 			terminal.WithErrorStyle(),
 		)
-
 		return 1
 	}
 
@@ -168,15 +175,6 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 
 	var cookie string
 	if c.adopt {
-		serverConfig, err := c.project.Client().GetServerConfig(ctx, &empty.Empty{})
-		if err != nil {
-			c.ui.Output(
-				"Error getting server config.",
-				clierrors.Humanize(err),
-				terminal.WithErrorStyle(),
-			)
-			return 1
-		}
 		cookie = serverConfig.Config.Cookie
 	}
 
@@ -196,7 +194,6 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	// We generate the ID if the user doesn't provide one
 	// This ID is used later to adopt the runner
 	id := c.id
-	var err error
 	if id == "" {
 		id, err = server.Id()
 		if err != nil {
@@ -225,7 +222,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		)
 		return 1
 	}
-	s.Update("Runner %s installed successfully to %s", id, c.platform[0])
+	s.Update("Runner %s installed successfully to %s", id, platform[0])
 	s.Status(terminal.StatusOK)
 	s.Done()
 
@@ -242,7 +239,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		s = sg.Add("Creating runner profile and targeting runner %s", strings.ToUpper(id))
 		runnerProfile, err := client.UpsertOnDemandRunnerConfig(ctx, &pb.UpsertOnDemandRunnerConfigRequest{
 			Config: &pb.OnDemandRunnerConfig{
-				Name: c.platform[0] + "-" + strings.ToUpper(id),
+				Name: platform[0] + "-" + strings.ToUpper(id),
 				TargetRunner: &pb.Ref_Runner{
 					Target: &pb.Ref_Runner_Id{
 						Id: &pb.Ref_RunnerId{
@@ -251,7 +248,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 					},
 				},
 				OciUrl:     c.runnerProfileOdrImage,
-				PluginType: c.platform[0],
+				PluginType: platform[0],
 			},
 		})
 		if err != nil {
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -229,9 +229,6 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	if c.adopt {
 		err = installutil.AdoptRunner(ctx, c.ui, client, id, c.serverUrl)
 		if err != nil {
-			c.ui.Output("Error adopting runner: %s", clierrors.Humanize(err),
-				terminal.WithErrorStyle(),
-			)
 			return 1
 		}
 
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -226,7 +226,9 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	s.Status(terminal.StatusOK)
 	s.Done()
 
-	if !c.skipAdopt {
+	if c.skipAdopt {
+		c.ui.Output(runnerInstalledButNotYetAdopted, terminal.WithInfoStyle())
+	} else {
 		err = installutil.AdoptRunner(ctx, c.ui, client, id, c.serverUrl)
 		if err != nil {
 			c.ui.Output("Error adopting runner: %s", clierrors.Humanize(err))
@@ -261,3 +263,9 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	}
 	return 0
 }
+
+var (
+	runnerInstalledButNotYetAdopted = strings.TrimSpace(`The installed runner must be adopted.
+Please run "waypoint runner adopt" before the runner can start accepting jobs.
+`)
+)
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -19,7 +19,7 @@ type RunnerInstallCommand struct {
 	*baseCommand
 
 	platform              []string `hcl:"platform,optional"`
-	adopt                 bool     `hcl:"adopt,optional"`
+	skipAdopt             bool     `hcl:"skip_adopt,optional"`
 	serverUrl             string   `hcl:"server_url,required"`
 	id                    string   `hcl:"id,optional"`
 	runnerProfileOdrImage string   `hcl:"odr_image,optional"`
@@ -69,27 +69,27 @@ func (c *RunnerInstallCommand) Flags() *flag.Sets {
 		f.BoolVar(&flag.BoolVar{
 			Name:    "server-tls",
 			Target:  &c.serverTls,
-			Usage:   "If true, the Waypoint runner will connect to the server over TLS.",
+			Usage:   "Connect the runner to the server over TLS.",
 			Default: true,
 		})
 
 		f.BoolVar(&flag.BoolVar{
 			Name:   "server-tls-skip-verify",
 			Target: &c.serverTlsSkipVerify,
-			Usage:  "If true, will not validate TLS cert presented by the server.",
+			Usage:  "Skip TLS verification for runner connection to server.",
 		})
 
 		f.BoolVar(&flag.BoolVar{
 			Name:   "server-require-auth",
 			Target: &c.serverRequireAuth,
-			Usage:  "If true, will send authentication details.",
+			Usage:  "Send authentication details from runner to server.",
 		})
 
 		f.BoolVar(&flag.BoolVar{
-			Name:    "adopt",
-			Usage:   "Adopt the runner after it is installed.",
-			Default: true,
-			Target:  &c.adopt,
+			Name:    "skip-adopt",
+			Usage:   "Skip adoption of runner after it is installed.",
+			Default: false,
+			Target:  &c.skipAdopt,
 		})
 
 		f.StringVar(&flag.StringVar{
@@ -123,7 +123,7 @@ Usage: waypoint runner install [options]
 
   This command will attempt to install a runner for the server configured in 
   the current Waypoint context. It will adopt the runner after installation, 
-  unless the '-adopt' flag is set to false.
+  unless the '-skip-adopt' flag is set to true.
 
 ` + c.Flags().Help())
 }
@@ -180,7 +180,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	defer func() { s.Abort() }()
 
 	var cookie string
-	if c.adopt {
+	if !c.skipAdopt {
 		cookie = serverConfig.Config.Cookie
 	}
 
@@ -232,7 +232,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	s.Status(terminal.StatusOK)
 	s.Done()
 
-	if c.adopt {
+	if !c.skipAdopt {
 		err = installutil.AdoptRunner(ctx, c.ui, client, id, c.serverUrl)
 		if err != nil {
 			return 1
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -231,7 +231,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	} else {
 		err = installutil.AdoptRunner(ctx, c.ui, client, id, c.serverUrl)
 		if err != nil {
-			c.ui.Output("Error adopting runner: %s", clierrors.Humanize(err))
+			c.ui.Output("Error adopting runner: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
 		}
 
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -40,15 +40,17 @@ func (c *RunnerInstallCommand) Flags() *flag.Sets {
 	return c.flagSet(0, func(set *flag.Sets) {
 		f := set.NewSet("Command Options")
 
-		var runnerPlatforms []string
-		for platformName := range runnerinstall.Platforms {
-			runnerPlatforms = append(runnerPlatforms, platformName)
+		// Add platforms in alphabetical order. A consistent order is important for repeatable doc generation.
+		var sortedPlatformNames []string
+		for name := range runnerinstall.Platforms {
+			sortedPlatformNames = append(sortedPlatformNames, name)
 		}
+		sort.Strings(sortedPlatformNames)
 
 		f.EnumVar(&flag.EnumVar{
 			Name:   "platform",
 			Usage:  "Platform to install the Waypoint runner into. If unset, uses the platform of the local context.",
-			Values: runnerPlatforms,
+			Values: sortedPlatformNames,
 			Target: &c.platform,
 		})
 
@@ -97,14 +99,6 @@ func (c *RunnerInstallCommand) Flags() *flag.Sets {
 			Usage:  "If this is set, the runner will use the specified id.",
 			Target: &c.id,
 		})
-
-		// Add platforms in alphabetical order. A consistent order is important for repeatable doc generation.
-		var sortedPlatformNames []string
-		for name := range runnerinstall.Platforms {
-			sortedPlatformNames = append(sortedPlatformNames, name)
-		}
-
-		sort.Strings(sortedPlatformNames)
 
 		for _, name := range sortedPlatformNames {
 			platform := runnerinstall.Platforms[name]
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -98,9 +98,15 @@ func (c *RunnerInstallCommand) Flags() *flag.Sets {
 			Target: &c.id,
 		})
 
-		sort.Strings(runnerPlatforms)
+		// Add platforms in alphabetical order. A consistent order is important for repeatable doc generation.
+		var sortedPlatformNames []string
+		for name := range runnerinstall.Platforms {
+			sortedPlatformNames = append(sortedPlatformNames, name)
+		}
 
-		for _, name := range runnerPlatforms {
+		sort.Strings(sortedPlatformNames)
+
+		for _, name := range sortedPlatformNames {
 			platform := runnerinstall.Platforms[name]
 			platformSet := set.NewSet(name + " Options")
 			platform.InstallFlags(platformSet)
@@ -257,7 +263,6 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		s.Update("Runner profile %q created successfully.", runnerProfile.Config.Name)
 		s.Status(terminal.StatusOK)
 		s.Done()
-		return 0
 	}
 	return 0
 }
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -229,6 +229,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 	if !c.skipAdopt {
 		err = installutil.AdoptRunner(ctx, c.ui, client, id, c.serverUrl)
 		if err != nil {
+			c.ui.Output("Error adopting runner: %s", clierrors.Humanize(err))
 			return 1
 		}
 
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -131,11 +131,10 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 
 	s := sg.Add("Uninstalling runner...")
 	err := p.Uninstall(ctx, &runnerinstall.InstallOpts{
-		Log:             log,
-		UI:              c.ui,
-		ServerAddr:      c.serverUrl,
-		AdvertiseClient: nil,
-		Id:              c.id,
+		Log:        log,
+		UI:         c.ui,
+		ServerAddr: c.serverUrl,
+		Id:         c.id,
 	})
 	if err != nil {
 		c.ui.Output("Error uninstalling runner: %s", clierrors.Humanize(err),
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -142,11 +142,11 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 		)
 		return 1
 	}
-	s.Update("Runner %s uninstalled successfully", c.id)
+	s.Update("Runner %q uninstalled successfully", c.id)
 	s.Status(terminal.StatusOK)
 	s.Done()
 
-	s = sg.Add("Forgetting runner %s on server...", c.id)
+	s = sg.Add("Forgetting runner %q on server...", c.id)
 	defer func() { s.Abort() }()
 
 	_, err = c.project.Client().ForgetRunner(ctx, &pb.ForgetRunnerRequest{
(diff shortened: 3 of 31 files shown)


internal/cli/install.go
```diff
@@ -494,15 +498,20 @@ func installRunner(
 		AuthToken:     resp.Token,
 	}
 
+	// We set the ID to be "init" since it is the initial static runner
+	id := "init"
```
briancain, line 502: This is an interesting idea. I wonder if it'd be more descriptive if we called this `static-runner` instead for its id?

internal/cli/install.go
```diff
@@ -514,6 +523,12 @@ func installRunner(
 	}
 	s.Done()
 
+	err = installutil.AdoptRunner(ctx, ui, client, id)
+	if err != nil {
+		s.Update("Error adopting runner: %s", err)
+		s.Status(terminal.StatusError)
```
briancain, line 529: Should we be returning the err here?

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (c *RunnerInstallCommand) Flags() *flag.Sets {
+	return c.flagSet(0, func(set *flag.Sets) {
+		f := set.NewSet("Command Options")
+
+		f.StringVar(&flag.StringVar{
+			Name:   "platform",
+			Usage:  "Platform to install the Waypoint runner into.",
+			Target: &c.platform,
```
briancain, line 45: It'd be neat if this flag was an `EnumVar`. Then we could fail fast if they provided a unsupported platform in the flag parse step.

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+		f.StringVar(&flag.StringVar{
+			Name:   "id",
+			Usage:  "If this is set, the runner will use the specified id.",
+			Target: &c.id,
+		})
+
+		// Add platforms in alphabetical order. A consistent order is important for repeatable doc generation.
+		i := 0
+		sortedPlatformNames := make([]string, len(runnerinstall.Platforms))
+		for name := range runnerinstall.Platforms {
```
briancain, line 96:
```suggestion
		for i, name := range runnerinstall.Platforms {
```

I think you can simplify this for loop, and remove the  `i++`.

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+  attempt to install a runner for the server configured in the current Waypoint 
+  context.
+
+` + c.Flags().Help())
+}
+
+func (c *RunnerInstallCommand) Run(args []string) int {
+	ctx := c.Ctx
+	log := c.Log.Named("install")
+	defer c.Close()
```
briancain, line 127: I'd move these after we set up the CLI init after line 137, maybe a small optimization but we'd throw these away if the CLI ever fails to init.

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+	); err != nil {
+		return 1
+	}
+
+	p, ok := runnerinstall.Platforms[strings.ToLower(c.platform)]
+	if !ok {
+		if c.platform == "" {
+			c.ui.Output(
+				"The -platform flag is required.",
+				terminal.WithErrorStyle(),
```
briancain, line 144: I'd include the help output text here too! We do this in a few other places, I think it's like `c.Flags().Help()` maybe?

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+		},
+		Id: id,
+	})
+	if err != nil {
+		c.ui.Output("Error installing runner: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+	s.Update("Runner %s installed successfully", id)
```
briancain, line 210: I'd include the platform here too! Extra context will be nice.

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+			Config: &pb.OnDemandRunnerConfig{
+				Name: c.platform + "-" + strings.ToUpper(id),
+				TargetRunner: &pb.Ref_Runner{
+					Target: &pb.Ref_Runner_Id{
+						Id: &pb.Ref_RunnerId{
+							Id: strings.ToUpper(id),
+						},
+					},
+				},
+				OciUrl:               "hashicorp/waypoint-odr:latest",
```
briancain, line 235: I think this should be configurable? I believe our server install includes flags for these already.

internal/cli/runner_install.go
```diff
@@ -0,0 +1,259 @@
...
+		s.Status(terminal.StatusOK)
+		s.Done()
+		return 0
+	}
+	return 0
+}
+
+func (c *RunnerInstallCommand) Synopsis() string {
+	return "Installs a Waypoint runner to Kubernetes, Nomad, ECS, or Docker"
+}
```
briancain, line 259: I'd group this func with the Flag and Usage functions just so they're together!

internal/installutil/adoption.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/hashicorp/waypoint/internal/clierrors"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"time"
+)
+
+func AdoptRunner(ctx context.Context, ui terminal.UI, client pb.WaypointClient, id string) error {
+	sg := ui.StepGroup()
+	defer sg.Wait()
+
+	s := sg.Add("Waiting for runner to connect to server...")
```
briancain, line 16: Can we print the server address here?

briancain (changes requested):
I only looked at the CLIs on my first pass, I'll save the plugin review for later! Looking good so far 😎 Excited to see this all come together.

internal/cli/install.go
paladin-devops: @demophoon what are your thoughts here? I personally like `static-runner` better, but I remember we were discussing this when syncing up on how we uniquely identify a runner for uninstalling. So if we do change it, just wanna remain in sync 😄

demophoon: I think that static-runner makes sense here. Probably better than using init since that term would then be overloaded in some of the workflows, namely kubernetes.

paladin-devops: @demophoon great! After your PR is merged up to this branch, I'll update it to use `static-runner`, and make sure the different platforms don't add an additional suffix `-runner`, so we don't end up with `waypoint-static-runner-runner` 😃

paladin-devops: TL;DR - I'll update this field to just `static`, to avoid weird naming of resources such as `waypoint-static-runner-runner`.

@briancain @demophoon I remember why this was tricky now. We originally wanted this to be named `waypoint-init-runner`, but the problem was that the Waypoint Helm chart applies `-runner` as a suffix to the names of all the resources created in Kubernetes. For example, with `id` set to `static-runner`, `kubectl get pods` returns the following after a clean install:

```terminal-output
% kubectl get pods
NAME                              READY   STATUS    RESTARTS      AGE
waypoint-server-0                 1/1     Running   0             23m
waypoint-static-runner-runner-0   1/1     Running   3 (22m ago)   22m
```

To align with this, I added `-runner` in the naming of resources for other platforms (the job name in Nomad, etc.).

internal/runnerinstall/docker.go
```diff
@@ -0,0 +1,198 @@
...
+func (i *DockerRunnerInstaller) InstallFlags(set *flag.Set) {
+	set.StringVar(&flag.StringVar{
+		Name:    "docker-runner-image",
+		Target:  &i.Config.RunnerImage,
+		Usage:   "The Docker image for the Waypoint runner.",
+		Default: "hashicorp/waypoint:latest",
+	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "docker-runner-network",
```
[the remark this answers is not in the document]
krantzinator: On second thought, I'm getting this no matter what it seems so I'll debug a bit!

krantzinator (commented):
Dropping a few notes from my first attempt at installing a runner on Docker. I'll keep poking and adding new comment collections.

internal/cli/runner_install.go
```diff
@@ -0,0 +1,281 @@
...
+
+	sg := c.ui.StepGroup()
+	defer sg.Wait()
+
+	s := sg.Add("Connecting to: %s", c.serverUrl)
+	defer func() { s.Abort() }()
+
+	if c.adopt && c.serverCookie == "" {
+		c.ui.Output(
+			"Server cookie must be supplied for adoption.",
```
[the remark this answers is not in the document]
krantzinator: Another note here: if we require it to be supplied, we should have an earlier failure route if the server cookie is invalid.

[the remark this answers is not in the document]
paladin-devops: @krantzinator I think we can get it from the backend! I'll add the retrieval of the cookie inside the `if` block so that way we don't unnecessarily get it.

internal/runnerinstall/docker.go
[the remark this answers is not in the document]
paladin-devops: The Docker network is optional, in case you want the runner to talk to a Waypoint server in the same Docker network. It's optional because the server might not be installed in the same Docker network as the runner you're installing, so it might not be necessary to install the runner to a Docker network at all.

Regarding why you're hitting "context deadline exceeded", you might need the `-server-tls-skip-verify` flag if your Waypoint server's TLS configuration is self-signed.

internal/runnerinstall/docker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	var waypointNetwork network.NetworkingConfig
+	if i.Config.Network != "" {
+		waypointNetwork = network.NetworkingConfig{
+			EndpointsConfig: map[string]*network.EndpointSettings{
+				i.Config.Network: {},
+			},
+		}
+	}
```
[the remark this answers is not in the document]
paladin-devops: I do remember this being weird! This is because the network name (provided by the user) is the in the key portion of the map element, and the value portion of the map element is an empty struct because we have no opinion on the `EndpointSettings` for the network. Will definitely add a comment though! 😄

internal/runnerinstall/docker.go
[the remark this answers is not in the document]
krantzinator: Ah yes, I needed that flag! Can we add that flag to the example command in the docs/runner-manual section, since our default server configuration is self-signed?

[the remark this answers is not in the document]
krantzinator: This part
```
```shell-session
 waypoint runner install -platform=docker -docker-runner-network=waypoint -server-addr=waypoint-server:9701 -server-cookie=abc123
 ✓ Connecting to: waypoint-server:9701
```

internal/cli/runner_uninstall.go
```diff
@@ -0,0 +1,172 @@
...
+			"The -platform flag is required.",
+			terminal.WithErrorStyle(),
+		)
+
+		return 1
+	}
+
+	if c.id == "" {
+		c.ui.Output(
+			"Runner ID must be supplied for uninstallation",
```
krantzinator, line 111: One more quick one for today -- We should specify the flag in the error text to be more precise about what's missing. Feel free to expand on the wording if you'd like it to be more informational as well.
```suggestion
			"The -id flag is required",
```

internal/cli/runner_install.go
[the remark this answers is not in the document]
paladin-devops: Also updated this block to check the server cookie if the user provides one!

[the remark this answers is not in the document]
krantzinator: nit: we have one other command that accepts a server cookie as a flag value, and it uses `cookie` instead of `server-cookie` as the flag, so we should update either this one or that one to be consistent:
from https://github.com/hashicorp/waypoint/blob/main/internal/cli/runner_agent.go
```
Name:   "cookie",
			Target: &c.flagCookie,
			Usage: "The cookie value of the server to validate API requests. " +
				"This is required for runner adoption. If you do not already have a " +
				"runner token, this must be set.",
			EnvVar: serverclient.EnvServerCookie,
		})
```

krantzinator (commented):
Took another deeper pass through, and it's looking good. I really appreciate some of the comments in various places explaining what you're thinking.

Most of my comments are around the code itself -- specifically a few places we set or specify something that we don't need to set or specify. I also noted one open question on if we need to even have the server cookie as a flag or not.

On to the manual testing!

internal/cli/runner_install.go
```diff
@@ -0,0 +1,304 @@
...
+				terminal.WithErrorStyle(),
+			)
+			return 1
+		}
+		if c.serverCookie != serverConfig.Config.Cookie {
+			c.ui.Output(
+				"Incorrect server cookie provided.",
+				terminal.WithErrorStyle(),
+			)
+			return 1
```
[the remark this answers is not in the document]
paladin-devops: This is a great question - I think it's a good idea to remove the cookie option based on this. Retrieving the cookie from the user's context would make for a less-clunky experience, one less incorrect value that could possibly be entered for installing a runner, when we have the correct one ready!

internal/cli/runner_install.go
```diff
@@ -0,0 +1,304 @@
...
+	var err error
+	if c.id == "" {
+		id, err = server.Id()
+		if err != nil {
+			c.ui.Output("Error generating runner ID: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		}
+	} else {
+		id = c.id
+	}
```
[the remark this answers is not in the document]
paladin-devops: It's been a while since I fought with this! But I did try out this suggestion locally and the compiler actually thinks that `id` on line 222 in your suggestion is a new, unused variable `id` 😅 even though it is indeed used later on `p.Install`'s options.

```terminal-output
# github.com/hashicorp/waypoint/internal/cli
internal/cli/runner_install.go:199:3: id declared but not used
make: *** [bin] Error 2
```

But if I partially implement your suggestion, and still declare `err` as a var of type `error` before the if block, I don't need to use `:=` in the if block, and the compiler is happy.
```suggestion
	id := c.id
	var err error
	if id == "" {
		id, err = server.Id()
		if err != nil {
			c.ui.Output("Error generating runner ID: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
			return 1
		}
	} 
```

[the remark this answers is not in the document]
krantzinator: Sounds good! Def trust the compiler over me 😄

internal/runnerinstall/k8s.go
```diff
@@ -0,0 +1,347 @@
...
+	client.GenerateName = false
+	client.NameTemplate = ""
+	client.OutputDir = ""
+	client.Atomic = false
+	client.SkipCRDs = false
+	client.SubNotes = true
+	client.DisableOpenAPIValidation = false
+	client.Replace = false
+	client.Description = "Static runner for executing remote operations for Hashicorp Waypoint."
+	client.CreateNamespace = true
```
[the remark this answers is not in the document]
krantzinator: Agree!

internal/cli/install.go
briancain: Cool, that makes a lot of sense @paladin-devops !

internal/cli/install.go
```diff
@@ -494,15 +498,20 @@ func installRunner(
 		AuthToken:     resp.Token,
 	}
 
+	// We set the ID to be "init" since it is the initial static runner
+	id := "static"
```
briancain, line 502: Ahh yeah, I saw your notice about not adding runner here. I'd make a comment that it's expected from the supported platforms that it adds a `-runner` to the name later.

internal/cli/install.go
```diff
@@ -514,6 +523,13 @@ func installRunner(
 	}
 	s.Done()
 
+	err = installutil.AdoptRunner(ctx, ui, client, id, advertiseAddr.Addr)
+	if err != nil {
+		s.Update("Error adopting runner: %s", err)
+		s.Status(terminal.StatusError)
+		return 1
```
briancain, line 530: We typically do `c.ui.Output` instead of step groups on error returns, because step groups at the moment can be messy and it's not going to output the error as the "last" thing, but where the original step group was started. I recommend using `c.ui.Output` for any err handling. We also have a helper for humanizing errors:

https://github.com/hashicorp/waypoint/blob/f9b5e91a9ab0e9798f7c88428e6d1c36ef589155/internal/cli/status.go#L49-L56

This feedback goes for any where else too where we are returning errors in the CLI. Sorry I missed this feedback the first review around! ❤️

internal/cli/runner_install.go
```diff
@@ -0,0 +1,269 @@
...
+		f.StringVar(&flag.StringVar{
+			Name:   "server-addr",
+			Usage:  "Address of the Waypoint server.",
+			EnvVar: "WAYPOINT_ADDR",
+			Target: &c.serverUrl,
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "odr-image",
+			Usage:   "Docker image for the on-demand runners. If not specified, it defaults to 'hashicorp/waypoint-odr:latest'.",
```
briancain, line 64: No need for the default verbage. If you notice, the flag package actually adds that for you if you set a default on a flag!

internal/cli/runner_install.go
```diff
@@ -0,0 +1,269 @@
...
+)
+
+type RunnerInstallCommand struct {
+	*baseCommand
+
+	platform              []string `hcl:"platform,required"`
+	adopt                 bool     `hcl:"adopt,optional"`
+	serverUrl             string   `hcl:"server_url,required"`
+	id                    string   `hcl:"id,optional"`
+	runnerProfileOdrImage string   `hcl:"id,optional"`
```
xiaolin-ninja, line 25: copy pasta here 
```suggestion
	runnerProfileOdrImage string   `hcl:"odr-image,optional"`
```

internal/cli/runner_install.go
```diff
@@ -0,0 +1,269 @@
...
+		},
+		Id: id,
+	})
+	if err != nil {
+		c.ui.Output("Error installing runner: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+	s.Update("Runner %s installed successfully to %s", id, c.platform[0])
```
briancain, line 228:
```suggestion
	s.Update("Runner %q installed successfully to %s", id, c.platform[0])
```

internal/cli/runner_install.go
```diff
@@ -0,0 +1,269 @@
...
+				PluginType: c.platform[0],
+			},
+		})
+		if err != nil {
+			c.ui.Output("Error creating runner profile: %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return 1
+		}
+		s.Update("Runner profile %s created successfully.", runnerProfile.Config.Name)
```
briancain, line 263:
```suggestion
		s.Update("Runner profile %q created successfully.", runnerProfile.Config.Name)
```

internal/cli/runner_uninstall.go
```diff
@@ -0,0 +1,171 @@
...
+		WithFlags(c.Flags()),
+		WithNoLocalServer(), // no auth in local mode
+		WithNoConfig(),
+	); err != nil {
+		return 1
+	}
+
+	if c.platform == "" {
+		c.ui.Output(
+			"The -platform flag is required.",
```
briancain, line 102: Instead of requiring a platform flag. we should load the CLI context and read the platform out of there. I believe that's how the server uninstall works today! If there is no platform defined in the CLI context, we can fail if the platform flag wasn't specified.

internal/cli/runner_uninstall.go
```diff
@@ -0,0 +1,171 @@
...
+		ServerAddr: c.serverUrl,
+		Id:         c.id,
+	})
+	if err != nil {
+		c.ui.Output("Error uninstalling runner: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
+	s.Update("Runner %s uninstalled successfully", c.id)
```
briancain, line 145:
```suggestion
	s.Update("Runner %q uninstalled successfully", c.id)
```

internal/cli/runner_uninstall.go
```diff
@@ -0,0 +1,171 @@
...
+	s.Status(terminal.StatusOK)
+	s.Done()
+
+	s = sg.Add("Forgetting runner %s on server...", c.id)
```
briancain, line 149:
```suggestion
	s = sg.Add("Forgetting runner %q on server...", c.id)
```

internal/cli/runner_uninstall.go
```diff
@@ -0,0 +1,171 @@
...
+	defer func() { s.Abort() }()
+
+	_, err = c.project.Client().ForgetRunner(ctx, &pb.ForgetRunnerRequest{
+		RunnerId: c.id,
+	})
+	if err != nil {
+		s.Update("Couldn't forget runner: %s", clierrors.Humanize(err))
+		s.Status(terminal.StatusWarn)
+	} else {
+		s.Update("Runner %s forgotten on server", c.id)
```
briancain, line 159:
```suggestion
		s.Update("Runner %q forgotten on server", c.id)
```

internal/installutil/adoption.go
```diff
@@ -0,0 +1,69 @@
...
+		RunnerId: id,
+		Adopt:    true,
+	})
+	if err != nil {
+		ui.Output("Error adopting runner: %s", clierrors.Humanize(err),
+			terminal.WithErrorStyle(),
+		)
+		return err
+	}
+	s.Update("Runner %s adopted successfully.", id)
```
briancain, line 65:
```suggestion
	s.Update("Runner %q adopted successfully.", id)
```

briancain (changes requested):
Another round of CLI feedback! I plan on testing the k8s flow next today :) Looking good!

internal/cli/runner_install.go
```diff
@@ -0,0 +1,269 @@
...
+func (c *RunnerInstallCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint runner install [options]
+
+  Install a Waypoint runner to an existing platform. The platform should be
+  specified as kubernetes, nomad, ecs, or docker.
+
+  With the '-adopt' flag set to true (the default), this command will adopt the
+  runner after it is installed. The install will attempt to install a runner for
+  the server configured in the current Waypoint context.
```
xiaolin-ninja, line 120: Sorry, wording nit.

```suggestion
  Install a Waypoint runner to the specified platform: kubernetes, nomad, ecs, or docker.

  This command will attempt to install a runner for the server configured in the current Waypoint context. It will adopt the runner after installation, unless the `-adopt` flag is set to false.
```

internal/cli/runner_uninstall.go
paladin-devops: Great minds! 😄

internal/cli/runner_install.go
```diff
@@ -0,0 +1,269 @@
...
+				"Error getting server config.",
+				clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return 1
+		}
+		cookie = serverConfig.Config.Cookie
+	}
+
+	if c.serverUrl == "" {
```
xiaolin-ninja, line 183: I think like the platform, we can get the server address from context. The current `runner agent` command does this.

briancain:
Just in case you haven't dug through this yet, you'll need to include this new website change for Vercel and our docs website to be happy:

```
> Build error occurred
07:42:57.659 | Error:
07:42:57.659 | Error: Unlinked pages found in the ../content/commands directory.
07:42:57.659 |  
07:42:57.659 | Please add these paths to the "../content/commands" nav data file, or remove the .mdx files. If you want the pages to exist but not be linked in the navigation, you can add a "hidden" property to the page object in the navigation structure.
07:42:57.659 |  
07:42:57.660 | [
07:42:57.660 | "runner-uninstall"
07:42:57.660 | ]
07:42:57.660 |  
07:42:57.660

<br class="Apple-interchange-newline">
```

briancain:
The error doesn't tell you which nav data json file, I think it's this one: https://github.com/hashicorp/waypoint/blob/main/website/data/commands-nav-data.json

internal/cli/server_upgrade.go
```diff
@@ -375,7 +381,7 @@ func (c *ServerUpgradeCommand) upgradeRunner(
 	}
 
 	s.Update("Runner found. Uninstalling previous runner...")
```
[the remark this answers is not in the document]
xiaolin-ninja: hmmm yeah that makes sense. I was thinking we should be able to pull information necessary for upgrade from the context, but didn't think about the persistent storage (which I think we could also pull from the existing runner, but it would be a whole extra level of complexity.) Since we didn't handle the install for the runner, it's a good idea to let the user handle the update too.

internal/installutil/adoption.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+func AdoptRunner(ctx context.Context, ui terminal.UI, client pb.WaypointClient, id string, addr string) error {
+	sg := ui.StepGroup()
+	defer sg.Wait()
+
+	s := sg.Add("Waiting for runner to connect to server at %s...", addr)
+	defer func() { s.Abort() }()
+
+	// Waits 5 minutes for the server to detect the new runner before timing out
```
xiaolin-ninja, line 19: How did we decide on 5 min?

internal/installutil/adoption.go
```diff
@@ -0,0 +1,69 @@
...
+	ticker := time.NewTicker(5 * time.Second)
+	found := false
+	for !found {
+		select {
+		case <-ticker.C:
+		case <-ctx.Done():
+			ui.Output("Cancelled.",
+				terminal.WithErrorStyle(),
+			)
+			return errors.New("context canceled")
```
xiaolin-ninja, line 32:
```suggestion
			return errors.New("context cancelled")
```

internal/installutil/nomad/nomad.go
```diff
@@ -0,0 +1,247 @@
...
+		qopts.WaitIndex = meta.LastIndex
+
+		switch eval.Status {
+		case "pending":
+			s.Update("Nomad allocation pending...")
+		case "complete":
+			s.Update("Nomad allocation created")
+
+			return eval, meta, nil
+		case "failed", "canceled", "blocked":
```
xiaolin-ninja, line 156:
```suggestion
		case "failed", "cancelled", "blocked":
```

internal/installutil/adoption.go
paladin-devops: @xiaolin-ninja this was arbitrarily chosen, but I could see it taking more time depending on the user's environment, which would lead this to unnecessarily time out.

Think we could add a flag like `-adoption-timeout` with a default of 5 mins?

internal/installutil/adoption.go
paladin-devops: Canceled! 😎

krantzinator:
One other suggestion -- on the new `server install` for Kubernetes, there's a lot of redundant success messages here:
```
✓ Helm settings retrieved
✓ Helm action initialized
✓ Helm install created
✓ Helm chart located
✓ Helm chart loaded
```

I suggest consolidating these. It's good to have separate, specific `if err` failure messages for each of these steps along the way, but imo there should just be one `Helm install created and chart loaded` success message at the end of this section.

internal/runnerinstall/k8s.go
```diff
@@ -0,0 +1,324 @@
...
+	s.Status(terminal.StatusOK)
+	s.Done()
+
+	return nil
+}
+
+func (i *K8sRunnerInstaller) InstallFlags(set *flag.Set) {
+	set.StringVar(&flag.StringVar{
+		Name:   "k8s-Config-path",
+		Usage:  "Path to the kubeconfig file to use,",
```
krantzinator, line 174:
```suggestion
		Name:   "k8s-config-path",
		Usage:  "Path to the kubeconfig file to use.",
```

internal/installutil/helm/helm.go
```diff
@@ -0,0 +1,195 @@
...
+	"k8s.io/client-go/discovery"
+	memcached "k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/client-go/tools/clientcmd"
+	"net/url"
+	"strings"
+)
+
+const DefaultHelmChartVersion = "0.1.8"
```
krantzinator, line 22: This will need to be updated each time we release a new version of Waypoint and update the version of our Helm chart. Does Helm not have an understanding of `latest`? I'm going to put some code suggestions over in `internal/runnerinstall/k8s.go`.

internal/runnerinstall/k8s.go
```diff
@@ -0,0 +1,324 @@
...
+		Target:  &i.Config.MemRequest,
+		Default: defaultRunnerMemory,
+		Usage:   "Requested amount of memory for Waypoint runner.",
+	})
+
+	set.BoolVar(&flag.BoolVar{
+		Name:    "k8s-runner-service-account-init",
+		Target:  &i.Config.CreateServiceAccount,
+		Default: true,
+		Usage:   "Create the service account if it does not exist. The default is true.",
```
krantzinator, line 231: You actually don't need this! Our flag package automatically informs the user what the `Default` is if that field is set.
```suggestion
		Usage:   "Create the service account if it does not exist.",
```
For example, right now this is returning:
```
 -k8s-runner-service-account-init
      Create the service account if it does not exist. The default is true.
      The default is true.
```

krantzinator: Oooh fascinating though, because our docs generation does not include what is generated for the `Default` field, just our cli `-h` does. I still say remove this here, and including the `Default` infos will be part of the revamp of how we generate documentation.

internal/installutil/adoption.go
krantzinator: I think past discussions have noted that for things that _should_ be snappy, like the server detecting a new runner, 5 minutes is the safe "it really shouldn't take this long at all, if it does, there are problems" length of time.

internal/cli/server_upgrade.go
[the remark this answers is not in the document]
demophoon: Yeah, I think this is definitely a case we should be able to handle. I think within the scope of runner installation/uninstallation right now, we should plan on getting to that in a future PR.

internal/cli/runner_uninstall.go
```diff
@@ -0,0 +1,172 @@
...
+		s.Status(terminal.StatusOK)
+	}
+	s.Done()
+
+	// TODO: Remove runner profiles associated solely with this runner
```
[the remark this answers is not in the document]
demophoon: Opened #3495

internal/runnerinstall/k8s.go
demophoon: Good call out, I went ahead and removed the snippet in `4dc7f1ef6`.

internal/runnerinstall/k8s.go
demophoon: Resolved in `529a857b7`

internal/installutil/helm/helm.go
briancain: That is a super good idea. Especially because if we mess up releasing a version of Helm, we'd now have to do yet-another release of Waypoint too.

krantzinator: Wow, the helm Golang package does not make this easy, do they.

internal/runnerinstall/k8s.go
```diff
@@ -0,0 +1,324 @@
...
+	s.Status(terminal.StatusOK)
+	s.Done()
+
+	version := i.Config.Version
+	if version == "" {
+		version = helminstallutil.DefaultHelmChartVersion
+	}
+
+	s = sg.Add("Locating chart...")
+	path, err := client.LocateChart("https://github.com/hashicorp/waypoint-helm/archive/refs/tags/v"+version+".tar.gz", settings)
```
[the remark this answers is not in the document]
krantzinator: OK I am realizing this is not easy with the helm Golang package. Hm. 🤔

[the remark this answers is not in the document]
krantzinator: Maybe do a manual pull of all the tags and use the latest tag in the version, unless the user specifies a version.

internal/runnerinstall/k8s.go
```diff
@@ -101,22 +100,25 @@ func (i *K8sRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
 
 	version := i.Config.Version
 	if version == "" {
-		version = helminstallutil.DefaultHelmChartVersion
+		tags, err := helminstallutil.GetLatestHelmChartVersion(ctx)
+		if err != nil {
+			opts.UI.Output("Error getting latest tag of Waypoint helm chart.", terminal.WithErrorStyle())
+			return err
+		}
+		version = *tags[0].Name
```
krantzinator, line 108: I was just thinking "Oh we should only do this if the user doesn't provide an error so we don't fail out unnecessarily" but then I saw that's exactly what you did 🎉

internal/runnerinstall/k8s.go
```diff
@@ -101,22 +100,25 @@ func (i *K8sRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
...
 	}
 
 	s.Update("Locating chart...")
-	path, err := client.LocateChart("https://github.com/hashicorp/waypoint-helm/archive/refs/tags/v"+version+".tar.gz", settings)
+	path, err := client.LocateChart("https://github.com/hashicorp/waypoint-helm/archive/refs/tags/"+version+".tar.gz", settings)
```
krantzinator, line 112: Does it matter for this API if the `v` is in front of the version or not? I assume so because it's looking for a specific tag. We should add a note of the expected format to the `k8s-helm-version` flag description, both here and on `server install` flag.

internal/installutil/adoption.go
```diff
@@ -0,0 +1,69 @@
...
+	sg := ui.StepGroup()
+	defer sg.Wait()
+
+	s := sg.Add("Waiting for runner to connect to server at %s...", addr)
+	defer func() { s.Abort() }()
+
+	// Waits 5 minutes for the server to detect the new runner before timing out
+	d := time.Now().Add(time.Minute * time.Duration(5))
+	ctx, cancel := context.WithDeadline(ctx, d)
+	defer cancel()
```
catsby, line 22: nit: could probably just use `time.NewTimer()` here: https://pkg.go.dev/time#NewTimer

internal/cli/runner_install.go
```diff
@@ -0,0 +1,263 @@
...
+			Usage:   "Docker image for the on-demand runners.",
+			Default: "hashicorp/waypoint-odr:latest",
+			Target:  &c.runnerProfileOdrImage,
+		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:    "server-tls",
+			Target:  &c.serverTls,
+			Usage:   "If true, the Waypoint runner will connect to the server over TLS.",
+			Default: true,
```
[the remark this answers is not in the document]
briancain: We've run into this before. The issue is our docs cli tool doesn't add the default string text that our CLI flag package does. If we add the defaults manually here and else where, our CLI strings will have duplicate default texts when you get the CLI help text. We should fix the website docs tool to include defaults and then we'll have it for all these commands instead.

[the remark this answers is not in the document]
briancain: I went ahead and fixed our docs generator: https://github.com/hashicorp/waypoint/pull/3501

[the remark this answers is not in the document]
catsby: Sounds good, we don't need to solve this problem here as it affects the codebase as a whole

internal/installutil/adoption.go
```diff
@@ -0,0 +1,69 @@
...
+		runners, err := client.ListRunners(ctx, &pb.ListRunnersRequest{})
+		if err != nil {
+			ui.Output("Error getting runners: %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+			return err
+		}
+		for _, myRunner := range runners.Runners {
+			if myRunner.Id == id {
+				found = true
```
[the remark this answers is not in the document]
paladin-devops: @catsby I went with using `found` here since I'm using `break` on [line 46](https://github.com/hashicorp/waypoint/blob/4ea71702338d96cd41997521ee2a77ba4564b13f/internal/installutil/adoption.go#L46), that breaks us out of this loop, but not the "parent" loop started on [line 25](https://github.com/hashicorp/waypoint/blob/4ea71702338d96cd41997521ee2a77ba4564b13f/internal/installutil/adoption.go#L25).

internal/installutil/adoption.go
```diff
@@ -68,3 +70,10 @@ func AdoptRunner(ctx context.Context, ui terminal.UI, client pb.WaypointClient,
 	s.Done()
 	return nil
 }
+
+var (
+	runnerFailedToConnectToServer = strings.TrimSpace(`
+The Waypoint runner was unable to connect to Waypoint server. Maybe the
+-server-addr specified is not accessible from the Waypoint runner?
+`)
```
krantzinator, line 78: Love this

krantzinator (approved):
OK! We've talked a lot in various places about improvements and potential bugs, and I'm confident we've covered everything I hit so far. I think there's a couple suggestions left from Clint, and then this LGTM! 🎊

internal/installutil/adoption.go
paladin-devops: @catsby upon updating this to the below using a timer, I get stuck in the loop 🤔 

```suggestion
	// Waits 5 minutes for the server to detect the new runner before timing out
	timer := time.NewTimer(time.Minute * time.Duration(5))
	defer timer.Stop()
	found := false
	for !found {
		select {
		case <-timer.C:
		case <-ctx.Done():
			ui.Output("Cancelled.",
				terminal.WithErrorStyle(),
			)
			return errors.New("context canceled")
		}
```

catsby:
I noticed a few files like `website/content/partials/commands/runner-install_desc.mdx` that are empty, is that expected?

catsby (commented):
A few more things in the documentation part and some clarifications on the timer/loop stuff. The docs questions/changes should be addressed but the rest you can take or leave

website/content/commands/install.mdx
```diff
@@ -79,15 +79,16 @@ and disable the UI, the command would be:
...
 - `-k8s-annotate-service=<key=value>` - Annotations for the Service generated.
 - `-k8s-context=<string>` - The Kubernetes context to install the Waypoint server to. If left unset, Waypoint will use the current Kubernetes context.
+- `-k8s-helm-version=<string>` - The version of the Helm chart to use for the Waypoint runner install.The required version number format is: 'vX.Y.Z'.
 - `-k8s-cpu-request=<string>` - Configures the requested CPU amount for the Waypoint server in Kubernetes.
 - `-k8s-mem-request=<string>` - Configures the requested memory amount for the Waypoint server in Kubernetes.
 - `-k8s-cpu-limit=<string>` - Configures the CPU limit for the Waypoint server in Kubernetes.
 - `-k8s-mem-limit=<string>` - Configures the memory limit for the Waypoint server in Kubernetes.
 - `-k8s-namespace=<string>` - Namespace to install the Waypoint server into for Kubernetes.
-- `-k8s-openshift` - Enables installing the Waypoint server on Kubernetes on Red Hat OpenShift. If set, auto-configures the installation.
```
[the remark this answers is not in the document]
paladin-devops: The `waypoint-helm` chart (currently) has no configs related to Openshift, and since we're installing the server and runner with the Helm chart as of this PR, this config is no longer needed.

internal/serverinstall/k8s.go
```diff
@@ -83,176 +79,145 @@ func (i *K8sInstaller) Install(
...
-
-	clientset, err := i.newClient()
+	settings, err := helminstallutil.SettingsInit()
 	if err != nil {
-		ui.Output(err.Error(), terminal.WithErrorStyle())
 		return nil, err
 	}
+	s.Update("Helm settings retrieved")
+	s.Status(terminal.StatusOK)
+	s.Done()
```
briancain, line 96: This is extra noise I think, we don't have to `Done` here. I'd group all of the `helm` related output into a single step group. Otherwise this feels noiser than it should:

```
brian@localghost:waypoint-webapp-go λ waypoint install -platform=kubernetes -accept-tos -k8s-server-image="localhost:5000/waypoint:dev"                               ±[●][main]
✓ Helm settings retrieved
✓ Helm action initialized
✓ Helm install created
✓ Helm chart located
✓ Helm chart loaded
```

internal/serverinstall/k8s.go
```diff
@@ -83,176 +79,145 @@ func (i *K8sInstaller) Install(
...
-	// If this is kind, then we want to warn the user that they need
-	// to have some loadbalancer system setup or this will not work.
-	_, err = clientset.AppsV1().DaemonSets("kube-system").Get(
-		ctx, "kindnet", metav1.GetOptions{})
-	isKind := err == nil
-	if isKind {
-		s.Update(warnK8SKind)
-		s.Status(terminal.StatusWarn)
-		s.Done()
-		s = sg.Add("")
```
briancain, line 110: We need to keep this warning here.

internal/serverinstall/k8s.go
```diff
@@ -83,176 +79,145 @@ func (i *K8sInstaller) Install(
...
-		namespaceClient := clientset.CoreV1().Namespaces()
-		_, err := namespaceClient.Get(context.TODO(), "openshift", metav1.GetOptions{})
-		isOpenShift := err == nil
-
-		// Default namespace in OpenShift acts like a regular K8s namespace, so we don't want
-		// to remove fsGroup in this case.
-		if isOpenShift && i.config.namespace != "default" {
-			s.Update("OpenShift detected. Switching configuration...")
-			i.config.openshift = true
-		}
```
briancain, line 167: We should probably keep this here for OpenShift too...

internal/serverinstall/k8s.go
```diff
@@ -83,176 +79,145 @@ func (i *K8sInstaller) Install(
...
-			return nil, err
-		}
-		s = sg.Add("")
-	}
-
-	s.Update("Creating Kubernetes resources...")
-
-	serviceClient := clientset.CoreV1().Services(i.config.namespace)
-	_, err = serviceClient.Create(context.TODO(), service, metav1.CreateOptions{})
+	imageRef, err := dockerparser.Parse(i.config.serverImage)
```
briancain, line 172: Just a guess, but I think this is trimming off the registry address for an image. We need to preserve this otherwise you won't be able to pull from custom container registries

internal/serverinstall/k8s.go
```diff
@@ -293,7 +247,7 @@ func (i *K8sInstaller) Install(
 				grpcPort = spec.Port
 			}
 
-			if spec.Name == "http" {
+			if spec.Name == "https-2" {
```
briancain, line 250: Why the `https-2` change here?

internal/serverinstall/k8s.go
```diff
@@ -309,12 +263,13 @@ func (i *K8sInstaller) Install(
 
 		// Set the grpc address
 		grpcAddr = fmt.Sprintf("%s:%d", addr, grpcPort)
-		log.Info("server service ready", "addr", addr)
+		log.Info("server service ready: %s", addr)
```
briancain, line 266: This unfortunately doesn't work with `hclog`. The log message here will say like `EXTRA_END VALUE` or something. Basically the hclog functions aren't formatters in the way `ui.Output` is so you'll still want `log.Info("msg", "key", value)`

internal/serverinstall/k8s.go
```diff
@@ -335,7 +290,7 @@ func (i *K8sInstaller) Install(
 		// since pods can't reach this.
 		if i.config.advertiseInternal || strings.HasPrefix(grpcAddr, "localhost:") {
 			advertiseAddr.Addr = fmt.Sprintf("%s:%d",
-				serviceName,
+				"waypoint-server",
```
briancain, line 293: Can we keep this `serviceName`? Seems like it would be good to have some `const` in this file for repeat config values in the case where we wanna change something, we can change it once.

internal/serverinstall/k8s.go
```diff
@@ -568,16 +503,23 @@ func (i *K8sInstaller) Upgrade(
 
 		// Set the grpc address
 		grpcAddr = fmt.Sprintf("%s:%d", addr, grpcPort)
-		log.Info("server service ready", "addr", addr)
+		log.Info("server service ready: %s", addr)
```
briancain, line 506: Same comment here, gotta put it back to `"key", value`

internal/serverinstall/k8s.go
```diff
@@ -601,28 +543,17 @@ func (i *K8sInstaller) Upgrade(
...
-			"'RollingUpdate' strategy. This means the pod might not be updated immediately.",
-			terminal.WithWarningStyle(),
-		)
-		s.Update("Update Strategy is 'RollingUpdate'; once the upgrade completes, you may need to restart the pod to update the server image")
-		s.Status(terminal.StatusWarn)
-		s.Done()
-		s = sg.Add("")
-	}
-	s.Update("Upgrade complete!")
-	s.Done()
```
briancain, line 624: Is this not true anymore?

internal/serverinstall/k8s.go
```diff
@@ -635,470 +566,106 @@ func (i *K8sInstaller) Upgrade(
...
-		ui.Output(
-			"Error creating deployment %s", clierrors.Humanize(err),
-			terminal.WithErrorStyle(),
-		)
-		return err
+	runnerInstaller := runnerinstall.K8sRunnerInstaller{
+		Config: runnerinstall.K8sConfig{
+			K8sContext:           i.config.k8sContext,
+			Namespace:            i.config.namespace,
+			RunnerImage:          ref.ShortName(),
```
briancain, line 638: This has gotta be the full name, not short name, I suspect!

internal/serverinstall/k8s.go
```diff
@@ -379,139 +337,127 @@ func (i *K8sInstaller) Upgrade(
...
-	s.Update("Image set to update!")
-	s.Done()
-
-	s = sg.Add("Waiting for server to be ready...")
-	log.Info("waiting for waypoint server to become ready after image refresh")
+	values := map[string]interface{}{
+		"server": map[string]interface{}{
+			"enabled": true,
+			"image": map[string]interface{}{
+				"repository": imageRef.ShortName(),
```
briancain, line 433: This should probably be `imageRef.Name()` and same for else where in this file.

briancain: Although maybe not, I'm not sure how docker parses images but you might also need `imageRef.Repository()` too

internal/runnerinstall/nomad.go
```diff
@@ -0,0 +1,388 @@
...
+		{
+			Mode: "host",
+		},
+	}
+
+	// Preserve disk, otherwise upgrades will destroy previous allocation and the disk along with it
+	volumeRequest := api.VolumeRequest{ReadOnly: false}
+	if c.CsiVolumeProvider != "" {
+		volumeRequest.Type = "csi"
+		volumeRequest.Source = "waypoint-runner-" + opts.Id
```
krantzinator, line 137: Does this format need to match line 121 above?

internal/runnerinstall/ecs.go
```diff
@@ -0,0 +1,686 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	// Find clusterArn which waypoint runner is installed into
+	ecsSvc := ecs.New(sess)
+	services, err := ecsSvc.DescribeServices(&ecs.DescribeServicesInput{
+		Cluster:  aws.String(i.Config.Cluster),
+		Services: []*string{aws.String("waypoint-runner-" + opts.Id)},
```
krantzinator, line 284: Does this need to match the format of the other things? And why did we update the format on Docker and Nomad from this format to `waypoint-[opts.Id]-runner`?

internal/serverinstall/k8s.go
paladin-devops: The `waypoint-ui` service created by the Helm chart maps port 9702 to the named port `https-2`:

```terminal-output
$ kubectl get svc waypoint-ui -o yaml
// other service yaml
  ports:
  - name: http
    nodePort: 32284
    port: 80
    protocol: TCP
    targetPort: http
  - name: https
    nodePort: 30217
    port: 443
    protocol: TCP
    targetPort: https
  - name: grpc
    nodePort: 30730
    port: 9701
    protocol: TCP
    targetPort: grpc
  - name: https-2
    nodePort: 31750
    port: 9702
    protocol: TCP
    targetPort: https
```

internal/runnerinstall/ecs.go
paladin-devops: @krantzinator the naming convention for the runner was changed [after some feedback](https://github.com/hashicorp/waypoint/pull/3335#discussion_r900435044), but changed to append `-runner` to the end (instead of the ID) only after observing how the Helm chart was adding the suffix `-runner` too. 

That has been resolved for the k8s install, and Nomad/ECS can and should be made more consistent with the other platforms! Updating.

internal/serverinstall/k8s.go
```diff
@@ -379,139 +337,127 @@ func (i *K8sInstaller) Upgrade(
...
-		namespaceClient := clientset.CoreV1().Namespaces()
-		_, err := namespaceClient.Get(context.TODO(), "openshift", metav1.GetOptions{})
-		isOpenShift := err == nil
-
-		// Default namespace in OpenShift acts like a regular K8s namespace, so we don't want
-		// to remove fsGroup in this case.
-		if isOpenShift && i.config.namespace != "default" {
-			s.Update("OpenShift detected. Switching configuration...")
-			i.config.openshift = true
-		}
```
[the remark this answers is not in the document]
paladin-devops: @briancain the Openshift flag/auto-detection is no longer useful since we're installing to k8s with the Helm chart. Before this branch, it appears that not detecting Openshift influenced the pod security context for the k8s deployment that was created:

```
	securityContext := &apiv1.PodSecurityContext{}
	if !c.openshift {
		securityContext.FSGroup = int64Ptr(1000)
	}
```

I also double checked in the Helm chart and it looks like we don't have any values to set related to Openshift or pod security context, so I think this is safely removed:

https://github.com/hashicorp/waypoint-helm/blob/main/values.yaml

briancain (approved):
Let's gooooo! 💨🚀

PR #3337, opened 2022-05-13
Update server install for k8s to use Helm

paladin-devops:
This updates the server install for Kubernetes to use the Waypoint Helm chart to install the server, instead of using the Kubernetes API directly to create the resources.

In this implementation, the Helm installation of the server is separate from the Helm installation of the initial static runner that is deployed alongside the server. It has been so written because the `InstallRunner` function of the Kubernetes server  `Installer` uses the `Install` function of the `K8sRunnerInstaller` type, which deploys a runner with Helm, without a server.

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"github.com/hashicorp/waypoint/internal/installutil"
 	"github.com/hashicorp/waypoint/internal/runnerinstall"
-	"github.com/hashicorp/waypoint/pkg/server"
 	"sort"
 	"strings"
 	"time"
@@ -499,12 +498,8 @@ func installRunner(
 		AuthToken:     resp.Token,
 	}
 
-	// Get ID for runner
-	id, err := server.Id()
-	if err != nil {
-		ui.Output("Error generating runner ID: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
-		return 1
-	}
+	// We set the ID to be "init" since it is the initial static runner
+	id := "init"
 
 	// Install!
 	s.Update("Installing runner...")
--- a/internal/runnerinstall/k8s.go
+++ b/internal/runnerinstall/k8s.go
@@ -59,7 +59,7 @@ func (i *K8sRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
 	client.DependencyUpdate = false
 	client.Timeout = 300 * time.Second
 	client.Namespace = chartNS
-	client.ReleaseName = "waypoint-" + strings.ToLower(opts.Id)
+	client.ReleaseName = "waypoint-runner" + strings.ToLower(opts.Id)
 	client.GenerateName = false
 	client.NameTemplate = ""
 	client.OutputDir = ""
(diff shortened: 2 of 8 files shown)


internal/cli/install.go
```diff
@@ -499,12 +498,8 @@ func installRunner(
...
 	}
 
-	// Get ID for runner
-	id, err := server.Id()
-	if err != nil {
-		ui.Output("Error generating runner ID: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
-		return 1
-	}
+	// We set the ID to be "init" since it is the initial static runner
+	id := "init"
```
demophoon, line 502: Maybe a nitpick but I'm wondering if we should somehow indicate this is a runner since just listing helm deployments shows the default runner as `waypoint-init`, maybe the k8s runner installer can use `waypoint-runner-%s` instead of just `waypoint-%s` or just a simple `waypoint-runner` for the initial runner.

May make the deployment a bit clearer to an operator after they have forgotten about this in the future after an initial installation.

internal/runnerinstall/k8s.go
```diff
@@ -3,7 +3,7 @@ package runnerinstall
 import (
 	"context"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
-	"github.com/hashicorp/waypoint/internal/installutil/helm"
+	installutil "github.com/hashicorp/waypoint/internal/installutil/helm"
```
demophoon, line 6: I can take care of this within the `uninstall` implementation so this is more a question for me I suppose, but should we include helm within the `installutil` import here?
I only mention this because we will also likely need to import a `k8s` specific installutil as well which will help DRY up the similarities between server and runner uninstallation implementations regarding persistent volumes whose cleanup is not managed by helm.

website/content/commands/runner-install.mdx
```diff
@@ -35,6 +35,7 @@ Usage: `waypoint runner install [options]`
 #### docker Options
 
 - `-docker-runner-image=<string>` - The Docker image for the Waypoint runner.
+- `-docker-runner-network=<string>` - The Docker network in which to deploy the Waypoint runner.
```
demophoon, line 38: I assume this is unrelated but needed so that docs tests are happy?

demophoon (dismissed):
This all looks great and server is installed for me within a local kubernetes cluster! I had a few comments that are just for clarifications but overall i'm 👍 for merging into `f-runner-install`

internal/cli/install.go
paladin-devops: Good catch! I missed that here:

https://github.com/hashicorp/waypoint/blob/bf26572071b113d5a24ae6fb7605363a4a24ae7b/internal/runnerinstall/k8s.go#L62

Updating to prefix the runner's ID with `waypoint-runner-` instead!

internal/runnerinstall/k8s.go
paladin-devops: Good point! I hadn't yet added such an `installutil` for non-Helm k8s stuff specifically, so I will leave this as-is for this PR, but for the alias of that import in other places I will phase in specifying which "installutil" it is, since there could be multiple.

website/content/commands/runner-install.mdx
paladin-devops: Yes! 😅 I branched before updating the website docs for my changes upstream.

PR #3346, opened 2022-05-17
RunPipeline API

mitchellh:
This introduces the RunPipeline API. This works by pre-queueing all the jobs necessary for the pipeline and setting up the proper dependencies. 

I fixed a couple core issues in this PR as well, the most important being that if a job was queued with dependencies _and_ ODR, the dependencies would get wiped out. Additionally, I made the "start task" job for ODR depend on the same dependencies as the main job. This way, we don't start the ODR task until the primary task is ready to run.

This is only tested with unit tests. I want to merge this to `f-pipelines` so @briancain can maybe do some `waypoint pipeline run` CLI stuff :)


pkg/server/proto/server.proto
```diff
@@ -4625,6 +4636,36 @@ message GetPipelineResponse {
...
+  string job_id = 1;
+
+  // All job IDs are the list of all jobs that were queued for this pipeline.
+  // They are in topologically sorted order. This means that a job's dependencies
+  // are guaranteed to be satisfied by previous jobs in the list, but all
+  // previous jobs may not be necessary. This is because two jobs may be
+  // totally independent of each other and can execute concurrently. To get
+  // a better view, a graph should be built using step dependencies, and
+  // the job IDs can be correlated using the job_map field.
+  repeated string all_job_ids = 2;
```
briancain, line 4661: Will `all_job_ids` include the root `job_id`? Would `AllJobIds[0] == JobId`? Mostly curious if that's something we could rely on, or if there's a good reason to keep the root job id in its own field.

briancain (approved):
Looks good to me from a code standpoint! Had a question that doesn't block on merge into the feature branch. I'll be sure to integrate this into my CLI branch and give it a full test that way.

pkg/server/proto/server.proto
mitchellh: Yes, it will, but I thought for people who just want the root, separating it out would be cleaner in case we do every introduce the like... "get child jobs" API.

PR #3349, opened 2022-05-17
Make the log tests consumable

izaaklauer:


pkg/serverhandler/handlertest/test_service_logs.go
```diff
@@ -453,10 +436,19 @@ func TestServiceGetLogStream_depPlugin(t *testing.T) {
 			})
 		}
 
-		logSendClient.Send(&pb.EntrypointLogBatch{
-			InstanceId: instanceId,
-			Lines:      entries,
-		})
+		// Our instance log buffer might not be registered, so it's OK to retry
+		require.Eventually(t, func() bool {
+			err := logSendClient.Send(&pb.EntrypointLogBatch{
```
izaaklauer, line 441: This was the only real change required to make these tests consumable - we can't couple directly to the inmemstate serverstate, so we need a different way to make sure the buffer is initialized.

This is a little less precise of a check, but makes sense to me and seems to work in practice.

pkg/server/testing.go
```diff
@@ -240,3 +241,39 @@ func TestCookieContext(ctx context.Context, t testing.T, c pb.WaypointClient) co
 	md := metadata.New(map[string]string{"wpcookie": resp.Config.Cookie})
 	return metadata.NewOutgoingContext(ctx, md)
 }
+
+// TestRunner registers a runner and returns the ID and a function to
+// deregister the runner. This uses t.Cleanup so that the runner will always
+// be deregistered on test completion.
+func TestRunner(t testing.T, client pb.WaypointClient, r *pb.Runner) (string, func()) {
```
izaaklauer, line 248: This was all moved from internal.

briancain (approved):
🪵 ✅

PR #3351, opened 2022-05-17
internal/datasource: git supports cloning submodules recursively

### LINKED ISSUE #3350: cloning recursive git repos w remove runner

If building a project relies on submodules there should be an option for the remote runner config to clone the repo recursively, or maybe just a way to pass arbitrary git flags to the command that is run.

mitchellh:
The `recurse_submodules` configuration can be set to a non-zero value to specify the maximum depth of the recursion. 

The test verifies this does work, I'm not sure how this works if submodules change across refs significantly from HEAD. That's another test case we should probably do, but looks significantly more involved with `go-git` so I propose we just get this merged and then iterate on any issues from there.

Fixes #3350


briancain (approved):

mitchellh:
I've added tests and resolved all the issues with checking out a ref with different or no submodules. 😄

briancain (approved):

PR #3353, opened 2022-05-17
pipelines: Add `waypoint pipeline list`, `waypoint pipeline inspect`, `GetPipeline` and `ListPipelines`

briancain:
This pull request introduces a few CLI commands and API endpoints to interface with pipeline data in Waypoint. Waypoint pipeline list will take a project ref and list all known pipelines for the project, where as pipeline inspect will give a detailed view of a specific pipeline by id.

TODOs that I think should probably be in their own follow-up PRs that I will tackle after this one:

- `waypoint pipeline run`: Lots of good UX to think about here so it's worth its own PR
- `waypoint pipeline config-sync`:
  + My idea here is it could work similar to `waypoint config sync`. Which is use `DoApp` and use the client to queue a new `PipelineConfigSync` operation for syncing Pipeline configs. There might be a better way to do this, but that's just a thought before starting the implementation.
  + If possible, we could also enhance `waypoint config sync` to sync app _and_ pipeline configs.


pkg/server/proto/server.proto
```diff
@@ -4666,6 +4680,23 @@ message RunPipelineResponse {
...
+  // All of the listed pipelines
+  repeated Pipeline pipelines = 1;
+}
+
+message ConfigSyncPipelineRequest {
+  // The project to sync all pipeline configs on
+  Ref.Project project = 1;
+}
+
+message ConfigSyncPipelineResponse {}
```
mitchellh, line 4698: Would be cool if we can get the IDs of the synced pipelines. Similar to `RunPipeline` if we can respond with like a `map<string, Ref.PipelineId> pipelines` that maps the pipeline name in the given project to the ID, that'd be cool.

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -26,6 +26,45 @@ func (s *Service) UpsertPipeline(
...
+	}
+
+	p, err := s.state(ctx).PipelineGet(req.Pipeline)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pb.GetPipelineResponse{
+		Pipeline: p,
+		//RootStep: "", // TODO figure out root step
```
mitchellh, line 45: Might as well fill this in here, rebase on f-pipelines and use `ptypes.PipelineGraph`, then call `KahnSort`, and its element 0.

mitchellh (approved):

PR #3354, opened 2022-05-17
Add k8s runner uninstallation

demophoon:
This PR adds a runner uninstall implementation for kubernetes

### DIFF
--- a/internal/installutil/k8s/k8s.go
+++ b/internal/installutil/k8s/k8s.go
@@ -1,0 +1,82 @@
+package k8s
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/waypoint/internal/clierrors"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type K8sConfig struct {
+	serverImage        string            `hcl:"server_image,optional"`
+	namespace          string            `hcl:"namespace,optional"`
+	serviceAnnotations map[string]string `hcl:"service_annotations,optional"`
+
+	odrImage              string `hcl:"odr_image,optional"`
+	odrServiceAccount     string `hcl:"odr_service_account,optional"`
+	odrServiceAccountInit bool   `hcl:"odr_service_account_init,optional"`
+
+	advertiseInternal bool   `hcl:"advertise_internal,optional"`
+	imagePullPolicy   string `hcl:"image_pull_policy,optional"`
+	k8sContext        string `hcl:"k8s_context,optional"`
+	openshift         bool   `hcl:"openshft,optional"`
+	cpuRequest        string `hcl:"cpu_request,optional"`
+	memRequest        string `hcl:"mem_request,optional"`
+	storageClassName  string `hcl:"storageclassname,optional"`
+	storageRequest    string `hcl:"storage_request,optional"`
+	secretFile        string `hcl:"secret_file,optional"`
+	imagePullSecret   string `hcl:"image_pull_secret,optional"`
+}
+
+type K8sInstaller struct {
+	config K8sConfig
+}
+
+// newClient creates a new K8S client based on the configured settings.
+func (i *K8sInstaller) NewClient() (*kubernetes.Clientset, error) {
+	// Build our K8S client.
+	configOverrides := &clientcmd.ConfigOverrides{}
+	if i.config.k8sContext != "" {
+		configOverrides = &clientcmd.ConfigOverrides{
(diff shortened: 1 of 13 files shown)


internal/installutil/k8s/k8s.go
```diff
@@ -0,0 +1,82 @@
...
+	serviceAnnotations map[string]string `hcl:"service_annotations,optional"`
+
+	odrImage              string `hcl:"odr_image,optional"`
+	odrServiceAccount     string `hcl:"odr_service_account,optional"`
+	odrServiceAccountInit bool   `hcl:"odr_service_account_init,optional"`
+
+	advertiseInternal bool   `hcl:"advertise_internal,optional"`
+	imagePullPolicy   string `hcl:"image_pull_policy,optional"`
+	k8sContext        string `hcl:"k8s_context,optional"`
+	openshift         bool   `hcl:"openshft,optional"`
```
paladin-devops, line 23: Unless this needs to be used, can we remove it?

internal/runnerinstall/k8s.go
```diff
@@ -243,6 +390,7 @@ type K8sConfig struct {
 	CpuRequest           string
 	MemRequest           string
 	CreateServiceAccount bool
+	KeepPV              bool
```
paladin-devops, line 393: Is there anything in the persistent volume that the user might want to keep after uninstalling?

internal/runnerinstall/k8s.go
```diff
@@ -224,15 +237,149 @@ func (i *K8sRunnerInstaller) InstallFlags(set *flag.Set) {
...
+
+	_, err = client.Run(helmRunnerId)
+	if err != nil {
+		return err
+	}
+	s.Update("Runner Uninstalled")
+	s.Status(terminal.StatusOK)
+	s.Done()
+
+	if ! i.Config.KeepPV {
```
paladin-devops, line 296: This is a great short-term solution! I think in the long term, if/when we update the Helm chart to do it, we should leverage Helm to do that for us on uninstall.

internal/installutil/k8s/k8s.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/waypoint/internal/clierrors"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type K8sConfig struct {
```
paladin-devops, line 11: I like the approach of having the K8sConfig here, since runner install and server install currently both have the same elements in their config structs! However, this has me thinking, do they (the server installer's and the runner installer's config) need to be the same, or could some of these elements be removed from either one?

paladin-devops (commented):
This is looking great! I left some feedback with some ideas on how we could refactor out a few more little things.

Also, I think that the `runner_uninstall.go` file may have been missed under `internal/cli`. I was having trouble building, but I think adding that into this PR would do the trick!

internal/installutil/k8s/k8s.go
demophoon: Oh good catch, I think I just missed this in my rebase!

internal/runnerinstall/k8s.go
demophoon: Probably not, I only thought about adding this in because helm won't delete PVs regardless of uninstall flags and leaves it up to the cluster to reap them after a period of time which an operator can configure. I'll go ahead and remove it since its really just for storing that token which really has no use after the runner is uninstalled.

internal/installutil/k8s/k8s.go
demophoon: That's a good point, there is a lot of stuff in this struct that no longer needs to be here now that we've ripped out most of the k8s bits. And really its only for uninstalls only now that the rest of the installation is handled by helm. I'll remove the install specific configurations.

demophoon: Resolved in https://github.com/hashicorp/waypoint/pull/3354/commits/36e943e2457c4eac4a0cedcc2e2c79f82a5d0eb1

internal/installutil/k8s/k8s.go
demophoon: Resolved in https://github.com/hashicorp/waypoint/commit/36e943e2457c4eac4a0cedcc2e2c79f82a5d0eb1

paladin-devops (approved):
After the latest updates, I re-tested `runner uninstall` with Nomad w/CSI persistence, & with k8s, and both are working. Approved! 🚀

PR #3355, opened 2022-05-18
WAYP-190 Add jobs and ui endpoints to gateway and re-gen json.

teresamychu:


.changelog/3355.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
+core: update swagger and gateway yml to include jobs and ui endpoints.
+
```
briancain, line 3: Hey! Noting that we typically reserve changelog entries for user facing bugs/features/improvements. This is more of a dev thing, so no need for a changelog 👍🏻

teresamychu: OH ok! Hmmm i added this because the build was failing on the changelog check? Is there something I need to do to make it skip this step in the build?

briancain: Yep! We run that check on all PRs, so you can opt out of it for ones that don't need it with `pr/no-changelog`.

paladin-devops: @teresamychu if it's a "no changelog" kind of PR, you can just add the `pr/no-changelog` label! 😄

teresamychu: ooooo ok got it! I'll do that now and remove this file. Thanks guys!

izaaklauer (changes requested):
Looks good overall! Good selection of path-based routes on the list endpoint. For posterity, all permutations of the ListJobsRequest proto are possible through the gateway, even if they're not represented as path tokens - the other fields in the request proto can be passed through as query parameters.

Just one note on rerunning `make gen/server`

pkg/server/gen/server.swagger.json
```diff
@@ -4175,6 +4175,887 @@
...
+            ],
+            "default": "UNKNOWN"
+          }
+        ],
+        "tags": [
+          "Waypoint"
+        ]
+      }
+    },
+    "/waypoint/2022-02-03/namespace/*/ui/deployments/project/{application.project}": {
```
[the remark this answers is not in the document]
teresamychu: Ah right! Ok done!

PR #3359, opened 2022-05-18
Refactor Nomad canary releaser into separate package

### LINKED ISSUE #3276: The release did not provide a URL exit code
Hello.

I am using local waypoint to deploy to a nomad in CI

The main issue is with exit code. CI mark job as fail

Also I don't understand: why for some jobs where is a release URL, for some not. Probably depends on nomad IPs. 

```» Deploying app...
  Performing operation locally
-> Initializing the Nomad client...
-> Parsing the job specification...
-> Registering job "xxx"...
Monitoring evaluation "3205a1b3-e16d-e794-250e-ad24cadfad1b"
Monitoring evaluation "3205a1b3-e16d-e794-250e-ad24cadfad1b"
 +: Allocation "f5882aeb-a9a4-d466-e316-ca333b2b1e36" created: node "587181d7-7a62-4be2-0d63-97c15776c839", group "mygento"
Monitoring evaluation "3205a1b3-e16d-e794-250e-ad24cadfad1b"
 +: Evaluation status changed: "pending" -> "complete"
 +: Evaluation "3205a1b3-e16d-e794-250e-ad24cadfad1b" finished with status "complete"
 +: Deployment successfully rolled out!
  Performing operation locally
The deploy was successful!
The release did not provide a URL and the URL service is disabled on the
server, so no further URL information can be automatically provided. If
this is unexpected, please ensure the Waypoint server has both the URL service
enabled and advertise addresses set.

### LINKED ISSUE #3316: CLI deploy/deployment destroy deletes job in Nomad
When I invoke _waypoint deploy_ (it destroys old deployments during the release phase) or _waypoint deployment destroy_ then it also deletes the job in Nomad:

image

**Steps to Reproduce**
1. waypoint deployment list
image

2. job tim is properly running in Nomad
3. waypoint deployment destroy 267
4. job tim is deleted
image

* Waypoint CLI Version: 0.8.1
* Waypoint Server Platform and Version: Nomad 0.8.1
* Waypoint Plugin: nomad-jobspec

### PR DESCRIPTION
**Pull request by paladin-devops**
The Nomad `jobspec-canary` releaser plugin is being refactored into a separate plugin from the `jobspec` plugin. Presently, the canary releaser does no harm to any Nomad job deployments which do not have canaries, as the releaser will just exit early. 

However, the default behavior of Waypoint is to "prune" old deployments after a new release has occurred. For "mutable deployment" plugins, such as the `jobspec` plugin, the ID of the deployment is equal to the ID of the Nomad job. This means that even though there is a new version (or "generation") of the Nomad job, deployed by Waypoint, Waypoint by default would delete it right after. [This is called out](https://www.waypointproject.io/plugins/nomad#nomad-jobspec-canary-releasemanager) in the `jobspec-canary` documentation already, and users are recommended to run their `release` operations with the plugin with the `prune=false` flag.

This change is being introduced so that users of the `jobspec` plugin, who have not decided to opt-in to the `jobspec-canary` releaser, are not subject to that behavior. Without it, the `jobspec-canary` plugin is invoked by default, and thus automatic pruning (if not running with the aforementioned prune flag).

Closes #3276 and closes #3316.

### CODE REVIEW

File: website/content/partials/components/platform-nomad-jobspec-canary.mdx [hunk 1]
```diff
@@ -1,142 +1,11 @@
 ## nomad-jobspec-canary (platform)
 
-Deploy to a Nomad cluster from a pre-existing Nomad job specification file.
```
briancain, line 3: Do we need to add a new partial mdx file to account for the plugin moving? I think the docs are still around, just not being included because it's a new partial?

briancain: Sweet!

briancain (approved):
🚀

xiaolin-ninja (approved):
Looks good! The integration test failure is the same problem from yesterday

builtin/nomad/jobspec/platform.go
```diff
@@ -74,6 +74,10 @@ func (p *Platform) resourceManager(log hclog.Logger, dcr *component.DeclaredReso
 	)
 }
 
+type nomadClient struct {
+	NomadClient *api.Client
+}
```
catsby, line 79: After doing some digging, I see this comes from the former `releaser.go` file that was in here and both the platform and releaser shared this type. 

That said, the `builtin/nomad` tree now defines `getNomadClient` 4 times: 

```
[waypoint][main](4)$ ack getNomadClient builtin/nomad | grep func

builtin/nomad/task.go:286:func (p *TaskLauncher) getNomadClient() (*api.Client, error) {
builtin/nomad/platform.go:96:func (p *Platform) getNomadClient() (*api.Client, error) {
builtin/nomad/jobspec/releaser.go:73:func (r *Releaser) getNomadClient() (*nomadClient, error) {
builtin/nomad/jobspec/platform.go:79:func (p *Platform) getNomadClient() (*nomadClient, error) {
```

That itself might not be a problem but 2 of them return a `nomadClient` type that wraps the client, and the other 2 do not, and I don't understand why. Any thoughts on consolidating?

catsby (changes requested):
I'd like to understand and document any backwards compatibility concerns here before we proceed.

builtin/nomad/jobspec/platform.go
catsby: Updated my original comment above to strike the "now" part; it seems before this change the `builtin/nomad` tree already defined the method 4 times :P

catsby (changes requested):

builtin/nomad/jobspec/plugin.proto
```diff
@@ -12,11 +12,6 @@ message Deployment {
   opaqueany.Any resource_state = 3;
 }
 
-message Release {
-  string Url = 1;
-  opaqueany.Any resource_state = 2;
-}
```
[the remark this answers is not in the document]
paladin-devops: @catsby regarding backwards-compatibility, a waypoint.hcl file using the `nomad-jobspec-canary` releaser will work as-is with this change. However, I have not tested the path where a deployment was created with 0.8.x, and then attempted to be destroyed with this branch. I can try that out, but I'm okay with un-backporting this.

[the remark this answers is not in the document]
evanphx: Deleting the definition won't break anything. I presume (but don't see?) that nothing is using this message as a type in another message anymore.

With protobuf, changing the type of existing fields is the number 1 backward compat issue. Deleting types just means you can't decode them in the future (but since you deleted the type, it's assumed you don't need to decode them either).

builtin/nomad/jobspec/platform.go
paladin-devops: @catsby after taking a look at this, I'm in favor of consolidating the two implementations of this in the `builtin/nomad` package, in `platform.go` and `task.go`. 

I'd like to leave the implementations in `builtin/nomad/jobspec` and `builtin/nomad/canary` separate though, just for the purpose of keeping the plugin packages more independent of each other (given that at some point we may move the built in plugins to their own repos altogether). Though it's not exactly the same situation, I intentionally did not update the `runnerinstall` package to use any utility methods under `builtin`, and likewise did not update anything in `builtin` to depend on the `installutil` package's helper methods, just to keep the plugin code more independent.

Regarding why the client is wrapped, when I initially implemented the resource manager framework for the jobspec & canary plugins earlier this year, I noticed that the k8s resource manager implementation wrapped the k8s client set in a struct. Albeit in that situation, the clientset, namespace and client config are provided in that struct (copied below for reference):

```go
// getClientset is a value provider for our resource manager and provides
// the connection information used by resources to interact with Kubernetes.
func (p *Platform) getClientset() (*clientsetInfo, error) {
	// Get our client
	clientSet, ns, config, err := Clientset(p.config.KubeconfigPath, p.config.Context)
	if err != nil {
		return nil, err
	}

	return &clientsetInfo{
		Clientset: clientSet,
		Namespace: ns,
		Config:    config,
	}, nil
}
```

Given that, since for Nomad we only need `*api.Client`, I think that `getNomadClient` in the jobspec and canary packages can be updated to return just that, instead of a struct wrapping it.

catsby: Sounds rational, thanks 👍

briancain:
I think this is good to merge now @paladin-devops ?

catsby (approved):

PR #3360, opened 2022-05-18
internal/cli: Upgrade go-getter library to latest

briancain:
This commit upgrades the go-getter library to the latest release, and
explicitly configures the Getters for its client in Waypoint CLI.


eastebry (approved):
Awesome! Thanks for the quick turnaround!

PR #3365, opened 2022-05-19
Option 1 of 2: use github pkg for protoc-gen-go for plugin support

krantzinator:
Option 1 of 2 to fix our plugin generation (option 2 is PR #3366):

Installing `protoc-gen-go` via `go install google.golang.org/protobuf/cmd/protoc-gen-go` results in the following errors when running `make gen/plugins`:
```
See https://grpc.io/docs/languages/go/quickstart/#regenerate-grpc-code for more information.
builtin/k8s/k8s.go:8: running "protoc": exit status 1
--go_out: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC
```

Installing `protoc-gen-go` via `go get github.com/golang/protobuf/protoc-gen-go` fixes this issue. We do receive the warning:
```
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
go get: installing executables with 'go get' in module mode is deprecated.
        To adjust and download dependencies of the current module, use 'go get -d'.
        To install using requirements of the current module, use 'go install'.
        To install ignoring the current module, use 'go install' with a version,
        like 'go install example.com/cmd@latest'.
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.
```

PR #3366, opened 2022-05-19
Fix `make gen/plugins`

krantzinator:
Option 2 of 2 to fix our plugin generation (option 1 is PR #3365):

Installing `protoc-gen-go` via `go install google.golang.org/protobuf/cmd/protoc-gen-go` results in the following errors when running `make gen/plugins`:
```
See https://grpc.io/docs/languages/go/quickstart/#regenerate-grpc-code for more information.
builtin/k8s/k8s.go:8: running "protoc": exit status 1
--go_out: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC
```

This PR follows the suggested updates from the error message to enable us to continue generating our plugins.


briancain (commented):
We could talk about this one in our next team sync. The option 1 feels like an easy temp fix for "today", where as this one feels like probably the right fix going forward should we choose to upgrade protoc soon.

krantzinator:
Per discussion, I'll move the `go generate` stuff into its own `scripts/` location so that we don't have to update each plugin separately. Also see if the updates to `path=relative` lessen the requirement for endless `../`.

denysvitali:
Any update here?
It's always painful to compile the `pb` files for this project :(

briancain:
I think maybe we can close this now that #4444 is merged!

PR #3367, opened 2022-05-19
Docs: Add Task introspection to docs

briancain:
This was content that we cut for the last point release because the feature hadn't been released yet, but here it is now!


briancain:
Oops, I misread the changelog. not going in 0.8.2. Closing!

PR #3370, opened 2022-05-19
Update changelog

briancain:


krantzinator (approved):
Looks good! Looks like we'll need to manually update the 0.8.1 changelog as well on that branch, but we can do that after the release.

PR #3373, opened 2022-05-19
Extract `internal/jobstream` from `internal/client`

mitchellh:
This extracts job stream handling from `internal/client` into a dedicated, reusable package. See the tests and `internal/client` for usage. The goal here for pipelines is to be able to build on this to implement `pipeline run`. 

@briancain For an _initial_ step at `pipeline run`, I was thinking we just for loop over the returned IDs and just stream them _sequentially_. It isn't perfect but it gets us started. We can do smarter stuff later.

### DIFF
--- a/internal/client/job.go
+++ b/internal/client/job.go
@@ -3,7 +3,6 @@ package client
 import (
 	"context"
 	"fmt"
-	"io"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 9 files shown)


briancain (approved):
Nice! Good extraction of what we had into a consumable package, looks like it'll be easier to consume for writing the run CLI 👍🏻

PR #3380, opened 2022-05-20
Docs: Add ConfigSourcer plugin interface docs for extending Waypoint

### LINKED ISSUE #3083: ConfigSourcer interface documentation is missing from the public docs
**Describe the bug**
The [plugin interfaces docs](https://www.waypointproject.io/docs/extending-waypoint/plugin-interfaces) on waypointproject.io has documentation regarding many of the interfaces that may be needed for plugin development. It appears though that docs for the `ConfigSourcer` interface are missing from those docs.

briancain:
Fixes #3083

### DIFF
--- a/website/content/docs/extending-waypoint/plugin-interfaces/configsourcer.mdx
+++ b/website/content/docs/extending-waypoint/plugin-interfaces/configsourcer.mdx
@@ -74,3 +74,17 @@ func (cs *ConfigSourcer) read(
         // Return the results
 }
 ```
+
+## Getting your plugin into the CEB
+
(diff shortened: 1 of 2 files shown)


evanphx:
Can we include some text about where the code would go? Because these are run inside the CEB, they're not like "normal" plugins, they have to be embedded in the CEB today.

briancain:
Got it @evanphx ! And opened a issue/feature request to give the CEB the ability to actually load plugins: https://github.com/hashicorp/waypoint/issues/3411

paladin-devops (approved):
Looks great @briancain! I will be using this guide to write a ConfigSourcer 😄

PR #3389, opened 2022-05-26
Remove intro sections as devdot content includes it

tunzor:


BRKalow:
@tunzor Since this is targeting the `dev-portal` branch it'll only impact dev-portal and not the live .io site, so this should be safe!

topfunky (approved):
👍🏻

PR #3397, opened 2022-05-31
feature/{server,cli}: Pipeline configurations can be synchronized via the CLI or a config sync operation

briancain:
This pull request introduces the config sync API for synchronizing pipeline config for Waypoint. It has been wired into the config sync job operation and can be manually synced using `waypoint config sync` just as you would app variable syncing. It updates the UX of that CLI command to support telling the user that their pipelines were synced, if any.

This PR also introduces a new way to reference a given Pipeline beyond its ID entry in bolt. A pipeline can be referenced via its owner and the pipeline configs name. Pipeline names are expected to be unique across a project.

Peek 2022-05-31 12-00


internal/core/pipeline_config.go
```diff
@@ -18,10 +20,43 @@ func (p *Pipeline) ConfigSync(ctx context.Context) error {
...
+		if !ok {
+			return status.Error(codes.FailedPrecondition,
+				"failed to determine pipeline owner. Only expecting pipeline owner to be a Pipeline_Project type.")
+		}
+
+		p.logger.Trace("looking up pipeline by owner and name", "owner", pipeOwner.Project, "name", pipeline.Name)
+
+		// Look up if this pipeline already exists by Pipeline Owner (project and pipeline name)
+		// If so, use the existing id to upsert into an existing config. Otherwise
+		// create a new pipeline entry.
```
mitchellh, line 35: I think we should do this automatically on the backend. We can chat about it on Monday but that way I think we can do this all in a single transaction.

mitchellh (changes requested):
Looks great!

I'm not 100% sure config sync is the right place for this to happen but I think its a fine place for it to be for now and then if we change our thoughts we can always modify that later. 

I only have one change I think we should make which is making the "upsert" behavior on owner happen in the state package rather than on the frontend so that we can wrap that in a transaction. I'm not 100% sure this is the right approach so maybe we can talk about it but that's what I always had in mind.

internal/core/pipeline_config.go
briancain: Roger that. That shouldn't be too difficult to add here!

briancain: I think that's covered via [4789fa5](https://github.com/hashicorp/waypoint/pull/3397/commits/4789fa5b7338b2eef6c3682a1e34091ab081b174) now

mitchellh (approved):
Awesome work!

Issue #3394, opened 2022-05-30
Error pulling image: Cannot connect to the Docker daemon at unix:///var/run/docker.sock


rhabbachi:
**Describe the bug**
Kaniko is failing to build or pull docker image in a kubernetes context. `waypoint build` falls back to the docker client which fails as expected.

**Steps to Reproduce**
1. Waypoint server deployed in a kubernetes cluster (EKS for my case).
2. snippets from waypoint.hcl
```
variable "keycloak_admin_password" {
  type = string
  default = dynamic("kubernetes", {
    name      = "ovh.med.waypoint.keycloak.keycloak-admin-password"
    key       = "develop"
    secret    = true
    namespace = "default"
  })
}

app "keycloak" {
  url {
    auto_hostname = false
  }

  config {
    env = {
      KC_LOG_LEVEL            = "DEBUG"
      KEYCLOAK_ADMIN          = "admin"
      KEYCLOAK_ADMIN_PASSWORD = base64encode(workspace.name)
    }

    workspace "develop" {
      env = {
        KC_LOG_LEVEL   = "DEBUG"
        KEYCLOAK_ADMIN = "admin"
        # TODO: support custom passwords from cluster secrets.
        KEYCLOAK_ADMIN_PASSWORD = base64encode(workspace.name)
      }
    }
  }

  # Build specifies how an application should be deployed. In this case,
  # we'll build using a Dockerfile and keeping it in a local registry.
  build {
    use "docker-pull" {
        image = var.registry_image
        tag   = var.registry_tag

        encoded_auth = base64encode(
          jsonencode({
            username = var.registry_username,
            password = var.registry_password
          })
        )
    }

    registry {
      use "docker" {
        image = var.registry_image
        tag   = var.registry_tag

        encoded_auth = base64encode(
          jsonencode({
            username = var.registry_username,
            password = var.registry_password
          })
        )
      }
    }
  }

  deploy {
    use "helm" {
      context   = var.k8s_context
      namespace = var.k8s_namespace

      name  = "keycloak-${workspace.name}"
      chart = "${path.app}/waypoint/keycloak/helm/chart"

      values = [
        file(templatefile("${path.app}/waypoint/keycloak/helm/values.yaml.tmpl",
          {
            global : "null"
            keycloakx : {
              hostname = "keycloak-${workspace.name}.${var.ingress.zone}"
              dbchecker : {
                enabled : false
              }
              database : ""
            }
            keycloak-config-cli : {
[...]

PR #3398, opened 2022-06-01
Fix `docker-pull` remote operations build & push

paladin-devops:
In version 0.8.2 of Waypoint, using remote operations with the `docker-pull` plugin exits without building the app. The error produced is indicated in this log message:
```
[DEBUG] waypoint.runner.agent.runner.app.example-nodejs.mapper.stdio: received EOF, stopping recv loop: job_id=01G4FSXDE0HJMNADYC94CWZ8AN job_op=*gen.Job_Build err="rpc error: code = Unavailable desc = error reading from server: read unix @->/kaniko/tmp/plugin25197406: read: connection reset by peer
```

This PR corrects that problem by updating the OCI registry that's created to align with the `registry` stanza details. The Docker context is now also explicitly provided to Kaniko.

#3394 has more details on the issue with remote operations. [This](https://gist.github.com/paladin-devops/e3cb17ec243a592af38c1a4505ca731d) is a gist of the build log and `waypoint.hcl` where I tested this out.

### DIFF
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -3,12 +3,14 @@ package dockerpull
 import (
 	"context"
 	"fmt"
+	"github.com/docker/cli/cli/command/image/build"
 	"github.com/docker/distribution/reference"
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	wpdocker "github.com/hashicorp/waypoint/builtin/docker"
 	"github.com/hashicorp/waypoint/internal/assets"
 	"github.com/hashicorp/waypoint/internal/pkg/epinject/ociregistry"
+	"github.com/pkg/errors"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	empty "google.golang.org/protobuf/types/known/emptypb"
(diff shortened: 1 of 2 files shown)


builtin/docker/pull/kaniko.go
```diff
@@ -79,16 +80,23 @@ func (b *Builder) pullWithKaniko(
 		// is "index.docker.io".
 		host = "index.docker.io"
 	}
-	log.Trace("auth host", "host", host)
 
+	var insecure bool
```
briancain, line 84: Instead of the bool here for using it later, could we rely on `ai.Insecure` again?

briancain (changes requested):
Minor comment from me! Otherwise looks pretty straight forward 😄

briancain (approved):
Code looks good to me! 👍🏻

demophoon (changes requested):
This looks good and remote building and pushing appear to work. However when testing this out it looks like we are still relying on docker for tagging images built with kaniko on remote runners.

```
✓ Running push build v2
❌ Tagging Docker image: registry-service/demophoon/nodejs-example:latest => registry-service/demophoon/nodejs-example:latest
! unable to tag image:Cannot connect to the Docker daemon at
  unix:///var/run/docker.sock. Is the docker daemon running?
```

I think this is automatically handled via kaniko when kaniko pushes the image to the registry so we might just need to skip that step for kaniko images.

demophoon (approved):

PR #3399, opened 2022-06-02
Update CODEOWNERS for Ecosystem

paladin-devops:
Update CODEOWNERS to set the ecosystem team as an automatic reviewer for plugin changes.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,3 +2,4 @@
 # See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
 /ui/ @hashicorp/waypoint-frontend
+/builtin/ @hashicorp/waypoint-ecosystem


.github/CODEOWNERS
```diff
@@ -2,3 +2,4 @@
 # See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
 /ui/ @hashicorp/waypoint-frontend
+/builtin/ @hashicorp/waypoint-ecosystem
```
demophoon, line 5: Should we also add `/internal/plugin/`? I'm also totally happy keeping it simple :)

demophoon: Thinking about this more, I'm going to go ahead and dismiss this for now and keep it as is. Chances are if its in `/internal/plugin/` we will probably be the ones doing the work and less extending plugin functionality which is I think at the root of us adding this line.

demophoon (approved):

PR #3410, opened 2022-06-03
internal/cli: Implement `get-stream` for job CLI

briancain:
This commit implements the get stream subcommand for the job CLI so that
users can attach to completed or currently running job streams.

### DIFF
--- a/internal/cli/job_get_stream.go
+++ b/internal/cli/job_get_stream.go
@@ -1,13 +1,10 @@
 package cli
 
 import (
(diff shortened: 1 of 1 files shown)


briancain:
Notice that this is going into the `f-pipelines` branch because that's the branch with the extracted job stream client that makes implementing this CLI super easy. It's also going to be a fairly critical feature of pipelines so users can attach to jobs as they execute in a step out of band from `waypoint pipeline run`.

mitchellh (approved):

PR #3412, opened 2022-06-03
Make vercel less noisy

amyrlam:
Ref: https://vercel.com/support/articles/how-to-prevent-vercel-github-comments

I think it should be here vs root, similar to https://github.com/hashicorp/waypoint/blob/main/website/vercel.json


krantzinator (approved):
FWIW, LGTM 😄

PR #3416, opened 2022-06-03
Decouple the server from the boltdb instanceExec functionality

evanphx:
The bits within boltdb related to the instanceExec interface have been moved to serverstate, and then boltdb implements that interface.

### DIFF
--- a/pkg/serverstate/instance.go
+++ b/pkg/serverstate/instance.go
@@ -53,13 +53,37 @@ type InstanceExec struct {
 	Context context.Context
 }
 
+// InstanceExecHandler is an optional interface that the state interface can implement.
+// When it does, the functionality associated with `waypoint exec` will be available.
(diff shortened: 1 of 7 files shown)


pkg/serverstate/instance.go
```diff
@@ -51,3 +52,14 @@ type InstanceExec struct {
 	// around or not.
 	Context context.Context
 }
+
+type InstanceExecHandler interface {
```
izaaklauer, line 56: Could probably use a header comment here.

izaaklauer (approved):
Very nice! Great way to abstract this out.

pkg/serverstate/instance.go
evanphx: Oh good call. I ran into a few nits when using that I'll push up shortly too.

PR #3423, opened 2022-06-06
internal/server: Ignore WatchTask on job ack and complete

briancain:
Prior to this commit, when a job when to ack or complete its state, it
would return an error because it didn't account for the WatchJob id.
This commit fixes that by ignoring any WatchJob acks given it does not
affect the task state machine.

This was supposed to be included in https://github.com/hashicorp/waypoint/pull/3420 but was missed. This will fix any job ack job complete failures on `main`.


briancain:
Small PR @mitchellh - forgot to include this in the previous PR. WatchTask jobs were failing to be marked complete and acked without this patch.

mitchellh (approved):

PR #3425, opened 2022-06-07
Fix panic in `waypoint status`

catsby:
If an application has not been deployed, there will be no initial status report for it, and running `waypoint status -app=<name>` will panic. We fix that here by guarding on `nil`


briancain (approved):
👍🏻

PR #3432, opened 2022-06-08
fix warn w/ git branch

cicoyle:
Fixing the warning seen when running waypoint build -vv. The git dirty check seemed to be failing due to the -> in the output of the git cli command. Now with the change, this warning no longer shows up bc the parsing is corrected by taking the -> into account.

```
$git branch -r
 origin/HEAD -> origin/main
```
```
$waypoint build -vv
2022-06-07T15:49:35.531-0500 [WARN]  waypoint.setupLocalJobSystem: failed to determine if local vcs is dirty: err="failed to diff repo at "/Users/cassiecoyle/go/src/github.com/waypoint-examples" subpath "/Users/cassiecoyle/go/src/github.com/waypoint-examples/kubernetes/nodejs" against remote with url "https://github.com/hashicorp/waypoint-examples.git" ref "HEAD": remote origin does not have specified branch "HEAD". To fix this, try running `git fetch origin`"
```

It is weird this didn't get caught in develop. Maybe git changed output formats & added the -> recently? If this kind of problem comes up again it might be working switching to the go -git.

### DIFF
--- a/.changelog/3432.txt
+++ b/.changelog/3432.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: fix git dirty check that was broken for some versions of the git cli
+```


izaaklauer (approved):
Looks good! This is also probably a signal that we need better testing around the functionality that depends on the output of the `git` command. I had assumed `git`'s output would be very stable, but maybe that's not true!

briancain (approved):

briancain:
This should be backported too?

evanphx:
Yes, it should be backported.

PR #2402, opened 2021-10-01
Server Telemetry

izaaklauer:
This PR enables server telemetry, which creates OpenCensus traces and timing/count stats for each grpc request.

Closes https://github.com/hashicorp/waypoint/issues/2364

Traces in datadog:

Screen Shot 2021-10-01 at 9 45 15 AM

Traces in jeager:
jeager-traces

Stats in datadog:

Screen Shot 2021-10-01 at 9 47 44 AM

### What changed

- Instruments the waypoint server with the ocgrpc stats handler.
- Adds a new `telemetry` package, responsible for configuring opencensus and its exporters
- Adds new flags to the `waypoint server run` for selecting and configuring either an opencensus agent exporter, or a direct datadog exporter
- Adds a new `telemetry` top-level directory that contains a docker-compose with example infra (an oc-agent, an oc-collector, a datadog agent, and jeager), and an example of how to use it.

### How to verify

Check out the `telemetry` top-level directory, and try running the example!

PR #3440, opened 2022-06-08
Initial metrics spike for gathering metrics about Waypoint operations

catsby:
This PR is a spiritual continuation of https://github.com/hashicorp/waypoint/pull/2402 where we initially introduce telemetry. Here we build off of the inclusion of the DataDog exporter and enable collecting basic metrics for our operations. 

Example of collected operation metrics:

metrics

There are some caveats here, in that the operation times are reported as gauges and not distributions (see https://github.com/DataDog/opencensus-go-exporter-datadog/issues/17). This pattern is however consistent with some of our internal (private) packages from other projects. 

Additional metrics can be collected by adding new OpenCensus Views and Stats in `metrics/stats.go` and exposed to Waypoint as public methods in the `metrics` package.

### DIFF
--- a/internal/cli/server_run.go
+++ b/internal/cli/server_run.go
@@ -115,7 +115,7 @@ func (c *ServerRunCommand) Run(args []string) int {
 	}
 	path := c.config.DBPath
 	log.Info("opening DB", "path", path)
-	db, err := bolt.Open(path, 0600, nil)
+	db, err := bolt.Open(path, 0o600, nil)
 	if err != nil {
 		c.ui.Output(
 			"Error opening database: %s", err.Error(),
@@ -271,7 +271,8 @@ func (c *ServerRunCommand) Run(args []string) int {
 	}
 	if httpInsecureLn != nil {
 		values = append(values, terminal.NamedValue{
-			Name: "HTTP Address (Insecure)", Value: httpInsecureLn.Addr().String()})
+			Name: "HTTP Address (Insecure)", Value: httpInsecureLn.Addr().String(),
+		})
 	}
 	if auth {
 		values = append(values, terminal.NamedValue{Name: "Auth Required", Value: "yes"})
@@ -374,6 +375,7 @@ This command will bootstrap the server and setup a CLI context.
 			telemetryOptions = append(telemetryOptions, telemetry.WithDatadogExporter(
 				datadog.Options{
 					TraceAddr: c.flagTelemetryDatadogTraceAddr,
+					StatsAddr: c.flagTelemetryDatadogTraceAddr,
 					Service:   "waypoint",
 				},
 			))
--- a/internal/cli/server_run.go
+++ b/internal/cli/server_run.go
@@ -115,7 +115,7 @@ func (c *ServerRunCommand) Run(args []string) int {
 	}
 	path := c.config.DBPath
 	log.Info("opening DB", "path", path)
-	db, err := bolt.Open(path, 0o600, nil)
+	db, err := bolt.Open(path, 0600, nil)
 	if err != nil {
 		c.ui.Output(
 			"Error opening database: %s", err.Error(),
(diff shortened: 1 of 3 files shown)


internal/cli/server_run.go
```diff
@@ -374,7 +375,9 @@ This command will bootstrap the server and setup a CLI context.
 			telemetryOptions = append(telemetryOptions, telemetry.WithDatadogExporter(
 				datadog.Options{
 					TraceAddr: c.flagTelemetryDatadogTraceAddr,
+					StatsAddr: c.flagTelemetryDatadogTraceAddr,
```
catsby, line 378: Here I simply re-use the current trace address, but we could easily add a separate flag for stats.

pkg/server/singleprocess/service_runner.go
```diff
@@ -874,3 +880,50 @@ func (s *Service) runnerVerifyToken(
 
 	return nil
 }
+
+func operationString(job *pb.Job) string {
```
catsby, line 884: we seem to do this in more than one place in Waypoint, so I'm open to suggestions to where we could place this and use it. Maybe [/pkg/server/ptypes/job.go](https://github.com/hashicorp/waypoint/blob/bc8959caca4047ab142dff101ad2d32a43042431/pkg/server/ptypes/job.go)?

pkg/server/singleprocess/service_runner.go
```diff
@@ -874,3 +880,50 @@ func (s *Service) runnerVerifyToken(
...
+	// Types that are assignable to Operation:
+	switch job.Operation.(type) {
+	case *pb.Job_Noop_:
+		return "Noop"
```
evanphx, line 888:
```suggestion
		return "noop"
```

evanphx (approved):
Just the one upper case change that makes my eye twitch.

pkg/server/singleprocess/service_runner.go
catsby: Whoops, thanks!

pkg/server/singleprocess/service_runner.go
briancain: The CLI client needs something like this too, so maybe `pkg/server/...` isn't the right place. Maybe it could go some where in `internal/...`? 🤔

pkg/server/singleprocess/service_runner.go
```diff
@@ -874,3 +880,50 @@ func (s *Service) runnerVerifyToken(
...
+	case *pb.Job_Poll:
+		return "poll"
+	case *pb.Job_StatusReport:
+		return "status_report"
+	case *pb.Job_StartTask:
+		return "start_task"
+	case *pb.Job_StopTask:
+		return "stop_task"
+	case *pb.Job_WatchTask:
+		return "watch_task"
```
briancain, line 924: What do you think about making these PascalCase so they match the proto go definition? Like `WatchTask`, `ConfigSync`, `Poll`, etc.

pkg/server/singleprocess/service_runner.go
```diff
@@ -535,7 +537,11 @@ func (s *Service) RunnerJobStream(
 	}
 	log.Trace("loaded config sources for job", "total_sourcers", len(cfgSrcs))
 
-	log.Debug("sending job assignment to runner")
```
briancain, line 538: Was this intentionally deleted?

briancain (approved):
Noice! ✨  Looks pretty good to me 😄 I had a few minor comments.

pkg/server/singleprocess/service_runner.go
catsby: From what I've seen internally, lowercase with `_` is the common style, so I chose this to be consistent

pkg/server/singleprocess/service_runner.go
catsby: Nope, thanks! Patched up

pkg/server/singleprocess/service_runner.go
briancain: Ooh ok! Makes sense from that standpoint 👍🏻 Sounds good.

briancain (approved):

PR #3444, opened 2022-06-10
Adding the rest of the relevant rpcs to gateway.yml

izaaklauer:
Also upgrading mockery version. It's time, and this matches other projects.

Apologies for the diff - I also rearranged the endpoints both in server.proto and gateway.yml so they matched, and so that endpoints were grouped by resource (which they mostly were already)


krantzinator (approved):
🟢

PR #3456, opened 2022-06-13
update mockery version for make tools

krantzinator:
This is just upgrading the tools version for the upgrade that already happened https://github.com/hashicorp/waypoint/pull/3444. After making this change, running `make tools`, verifying the `mockery` version had upgraded, and running `make gen/server`, there were no generated changes to our pbs.

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -17,7 +17,7 @@ package tools
 //go:generate go install github.com/evanphx/grpc-gateway/protoc-gen-swagger
 
 //go:generate go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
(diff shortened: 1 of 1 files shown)


briancain (commented):
If I remember right, there's a specific reason we haven't upgraded to the latest mockery. I think this new version means we have to fix all of the places where mockery code gen happens since 1.x and 2.x aren't backwards compatible, and it'll break anyone who isn't on mockery 2.x+ going forward when we do.

krantzinator:
@briancain ah yeah I should have mentioned this is just upgrading the tools version for the upgrade that already happened https://github.com/hashicorp/waypoint/pull/3444

briancain (approved):

PR #3458, opened 2022-06-14
Helper for externalizing top-level server errors

izaaklauer:
## What does this change?

This introduces a new server package `hcerr`, with one function: `Externalize`. The idea is that, rather than return any `err` directly to a client in a rpc handler, we'll instead call `hcerr.Externalize(log, err, "failed doing thing", ...)`.

This also allows you to add "args" to an error, ala hclog, which are added to the "details" field of the status protobuf, and then rendered by the client.

If you return this in the server:
```
hcerr.Externalize(log, err, "failed to get on-demand runner config", "id", id, "name", name)
```

It'll look something like this in the client:

Screen Shot 2022-06-13 at 10 05 37 PM

### Deep status

For better or worse, we're not just using grpc status errors at the highest level rpc handlers - we're using them all over the code. It's not that uncommon for function way down the callstack to want to set the grpc status error code.

Example: Someone calls GetOnDemandRunnerConfig, which calls something, which calls something else, which makes a DB call, which can't find the record. The lowest-level function we control returns a status error with `codes.NotFound`. This works, until somebody in the middle of the call chain wraps that error (i.e. `errors.Wrapf(err, "failed while trying to BLAH")`). If this wrapped status error is returned out of the framework, it won't be treated as a status error, and the `codes.NotFound` status won't be shown to the client. The clients often key off of this status to change behavior.

I hit this in practice - I wrapped an error somewhere in the middle of the call stack, and then spent a good long while debugging why the client wasn't behaving correctly, only to realize my error wrapping resulted in a lower-level status falling off.

`hcerr.Externalize` solves this by explicitly checking for a status error in the error stack, and using that code in the final error returned.

### Leaking too much context

Generally, when we return an error, it's nice and helpful to add as much context as possible to the error. Rather than `if err != nil { return err}`, it's much nicer and more helpful to do `if err != nil { return errors.Wrapf("failed to update user with database id %q", id)}` (or something). Without this, it's often impossible to figure out what the code was trying to do without modifying it to add logs or debugging through it.

This context, while helpful, isn't always appropriate for clients to see. It's leaking too much detail to tell clients about things like database IDs - it's also not great security as we don't necessarily trust the clients.

I've been working around this recently by logging all the detail for an error, then returning a briefer, sanitized error. That's fine, but it's a fair bit of boilerplate.

`hcerr.Externalize()` solves this by accepting the error and a short message, logging the full context, and returning the short message to the client.

Old bad pattern:
```
	if err := s.state(ctx).TaskPut(result); err != nil {
		return nil, err // Err might not be safe to return!
	}
```

Old annoying sanitization:
```
	if err := s.state(ctx).TaskPut(result); err != nil {
		msg := "failed to upsert task"
		log.Err(msg, "id", result.Id)
		return nil, status.Errorf(codes.Internal, msg + fmt.Sprintf(" id: %d", id))
	}
```

With `hcerr.Externalize`:
```
        if err := s.state(ctx).TaskPut(result); err != nil {
		return nil, hcerr.Externalize(log, err, "failed to upsert task", "id", req.Task.Id)
	}
```

## Future work
I updated a few of the service handlers to use `hcerr.Externalize`, but not all of them. That'll come in future PRs.


pkg/server/hcerr/hcerr.go
```diff
@@ -0,0 +1,64 @@
...
+	} else {
+		// And if all else fails, default to internal error
+		code = codes.Internal
+	}
+
+	var details []*anypb.Any
+	if len(args) > 0 {
+
+		// Even out the number of pairs
+		if len(args)%2 != 0 {
```
briancain, line 40: Took me a sec to realize what was going on here...so maybe a doc comment with an example for why we handle pairs in args would be awesome!

pkg/server/singleprocess/service_ondemand_runner.go
```diff
@@ -24,7 +27,7 @@ func (s *Service) UpsertOnDemandRunnerConfig(
 	}
 	result := req.Config
 	if err := s.state(ctx).OnDemandRunnerConfigPut(result); err != nil {
-		return nil, err
+		return nil, hcerr.Externalize(log, err, "failed setting on-demand runner config", err, "id", result.Id, "name", result.Name)
```
briancain, line 30: Looks like `err` is here twice!

pkg/server/hcerr/hcerr_test.go
```diff
@@ -0,0 +1,48 @@
...
+	})
+
+	t.Run("Can handle uneven number of args", func(t *testing.T) {
+		externalError := Externalize(hclog.Default(), errors.New("original error"), "message", "key", "value", "whoops")
+		statusErr, isStatusErr := status.FromError(externalError)
+		require.True(isStatusErr)
+		details := statusErr.Details()
+		require.Len(details, 2)
+	})
+}
```
briancain, line 48: Should we include a test that raises a grpc error and the status code is still preserved/reconstructed by the `int` we have in `Code`? Is that too heavy handed here? 🤔 Basically something that is like

```golang
require.Equal(codes.FailedPrecondition, statusErr.Code)
```

briancain (changes requested):
Noice! This makes sense to me I think. Just a couple of super minor things to address ✨

pkg/server/hcerr/hcerr.go
izaaklauer: Good point. Added a much better doc comment.

pkg/server/hcerr/hcerr_test.go
izaaklauer: Good idea. Added a new test, and added that condition to an existing test.

briancain (approved):
🚀 !!

PR #3459, opened 2022-06-15
Pipelines: Add first reserved Step operation "Build"

briancain:
This pull request introduces the first implementation of reserved Step operations "Build". It accepts a single option for whether or not the build operation should push its artifact. This matches the BuildOp job proto as well as the configurable option when running `waypoint build`.


internal/config/pipeline.go
```diff
@@ -127,11 +127,14 @@ func (c *Config) PipelineProtos() ([]*pb.Pipeline, error) {
 			},
 		}
 
+		// TODO(briancain): Use the default ODR image for image_url so that it isn't required
+		// We do this already maybe in the ODR task launcher.
```
mitchellh, line 131: Yeah, you can just make it empty and I think we should pick up the default that is configured with the server... that maybe won't work today but that's probably how it should work.

internal/runner/operation_pipeline_step.go
```diff
@@ -29,67 +29,26 @@ func (r *Runner) executePipelineStepOp(
 	case *pb.Pipeline_Step_Exec_:
 		return r.executePipelineStepExec(ctx, log, job, project, kind.Exec)
 
+	case *pb.Pipeline_Step_Build_:
```
mitchellh, line 32: Ah! I would actually never hit this. I would change `RunPipeline` on the server to transform this into a build job _there_ so we never have to spin up this superflous ODR/task.

mitchellh (changes requested):

internal/runner/operation_pipeline_step.go
briancain: Ahhh ok, I'll look into making that work 👍🏻

briancain: Got that working in [b4a036a](https://github.com/hashicorp/waypoint/pull/3459/commits/b4a036a2350837512f4d951f6ce907735f1ef319)

mitchellh (approved):
This looks great!

PR #3460, opened 2022-06-15
internal/server: Set expiry time on runner accept

briancain:
Prior to this commit, Waypoint server only expected queued jobs to wait
for 60 seconds before a runner accepted the job and began executing it
from its creation time. With the introduction of pipelines, this no
longer makes sense because a pipeline will queue all jobs in a pipeline
right away. This means downstream jobs could be waiting around to be
executing for a while, much longer than 1 minute. This is expected.

Given that, this commit updates the server behavior to set a jobs expiry
time to something much longer, 60 minutes, and then when the runner goes
to accept the job, we update the expiry time to the original expected
behavior of 60 seconds, because if the runner has accepted a job to work
on, it should run that job fairly quickly like the original intended
behavior.

### DIFF
--- a/pkg/server/singleprocess/service_job.go
+++ b/pkg/server/singleprocess/service_job.go
@@ -508,19 +508,6 @@ func (s *Service) onDemandRunnerStartJob(
 	}
 	job.Id = id
 
-	// NOTE(briancain): We set this value to something long for now, because it's
-	// possible during a pipeline run for jobs to be waiting in the QUEUED status
-	// while upstream jobs continue to run. Once a runner accepts a job however,
-	// we will update the jobs expiration time to be 60 seconds because by that
-	// point it should be executing quickly.
-	dur, err := time.ParseDuration("60m")
-	if err != nil {
-		return nil, "", status.Errorf(codes.FailedPrecondition,
-			"Invalid expiry duration: %s", err.Error())
-	}
-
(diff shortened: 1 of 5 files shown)


pkg/server/singleprocess/service_job.go
```diff
@@ -508,14 +508,11 @@ func (s *Service) onDemandRunnerStartJob(
...
-	// and it's because we could queue a LOT of jobs. Downstream jobs could be waiting
-	// for a while, like say if a pipeline has a few build steps and a few deploy
-	// steps, a release step might not run within ~60 seconds.
-	// NOTE(briancain): Setting this to 60 minutes for pipelines for now
-	// We should actually set this when the runner peaks at the job
+	// NOTE(briancain): We set this value to something long for now, because it's
+	// possible during a pipeline run for jobs to be waiting in the QUEUED status
+	// while upstream jobs continue to run. Once a runner accepts a job however,
+	// we will update the jobs expiration time to be 60 seconds because by that
+	// point it should be executing quickly.
```
mitchellh, line 515: I think we can just remove expiry complete now.

mitchellh (approved):
Looks good but one suggestion is to remove the 60m (minute) expiry completely. I think its fine for jobs to be queued indefinitely. If the start task fails, it will cascade the failure down.

PR #3461, opened 2022-06-16
Nomad plugin: Nomad service discovery

paladin-devops:
Update Nomad plugin to support specifying the service provider, which enables the use of Nomad service discovery. Consul remains the default service provider, which is consistent with Nomad.


krantzinator (approved):
Looks cut-and-dry to me! 🎉

briancain (approved):
🔍 🎉 Looks good to me too!

PR #3462, opened 2022-06-16
Backport of sync lambda plugin versions and protoc cmd format into release/0.8.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #3358 to be assessed for backporting due to the inclusion of the label backport/0.8.x.

---

In chasing down an error that ended up being a toolchain hiccup, I figured I'd update the lambda plugins to be the same version of things as everything else, and also make the `protoc` command use the same format.


krantzinator:
Not back porting because the `function-url` plugin isn't in 0.8.x.

PR #3463, opened 2022-06-16
internal/cli: Add a `-json` flag for project list

briancain:
This commit adds a simple -json flag to output the project list as json

### DIFF
--- a/internal/cli/project_list.go
+++ b/internal/cli/project_list.go
@@ -1,8 +1,10 @@
 package cli
 
 import (
+	"fmt"
 	"sort"
 
+	"github.com/golang/protobuf/jsonpb"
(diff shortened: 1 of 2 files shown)


xiaolin-ninja (approved):
LGTM, but what is the need for `-json` mode if the regular output is just one string anyway?

briancain:
Yeah, without json output the CLI for `project list` basically a `\n` separated list of project names. This was a request from @eveld who has been using some of the `json` output for a demo project.

eveld:
When dealing with the out programmatically it is nice to have a consistent way to parse the output. Many of the commands already output json. This makes it much easier to use in scripted flows in combination with tools like `jq`.

eveld (approved):

PR #3464, opened 2022-06-16
Allow debugging task launcher plugins

izaaklauer:
Task launcher plugins are the only ones that use plugin.Open at present.


krantzinator (approved):
🍟

Issue #2884, opened 2022-01-06
tech debt: Update `pb.Job_DeployOp` to take a reference to a pushed artifact stead of the full PushedArtifact message


briancain:
**Is your feature request related to a problem? Please describe.**

The `Job_DeployOp` operation requires the full `pb.PushedArtifact` message to be set to properly queue and execute a Deployment job. This is not ideal when you are trying to queue a job and already have all of the references you'd need (i.e. the workspace/project/app refs and artifact sequence ID or just wanting to deploy latest). Additionally, if you specify your own `pb.PushedArtifact` that only includes _some_ of the attributes on that message, the Deployment operation fails in an odd way where the Deployment some how gets Upserted without an ID (this is probably a separate bug).

**Describe the solution you'd like**

Ideally, this operation should take an Artifact reference like how our other operations take references to workspaces or projects, etc. So one could queue a Deployment operation with a `pb.Ref_Artifact`.

We could also make it a `oneof` Target so you could easily specify you want the _latest_ build artifact without needing to query for it from the database.

```golang
Artifact: &pb.Ref_Artifact{
  Target: &pb.Artifact_Latest{}
}
// Or by ID
Artifact: &pb.Ref_Artifact{
  Target: &pb.Artifact_BySeq{
    Application: &pb.Ref_Application{ ... },
    Workspace: &pb.Ref_Workspace{ ... },
    Sequence: 123,
  },
}
```

Then in our execute deployment operation, we can take the artifact reference and look up which one we should use based on the ref supplied.

This now means the caller/requester queueing a Deployment operation no longer needs to supply the full Artifact message, and can just supply a Ref similar to other operations.

**Describe alternatives you've considered**

Nothing really 😅 

**Explain any additional use-cases**

I could see this being important as we add more ways to queue jobs with Waypoint beyond the CLI and Triggers.

evanphx:
Totally support this, let's add references along side where there are the full values, with the reader code supporting both refs and full values. That gives us a smooth upgrade path, then we can drop the full values in a release or two.

briancain:
Just a future note to self: There are a handful of other Job operations that have the same issue that can benefit from the same solution! Like the status report operations, pushed artifact ops, etc. Keep an eye on the other operations when implementing this fix so they all get the same resolution.

PR #3470, opened 2022-06-17
internal/server: Support reserved pipeline step "deploy"

briancain:
This commit introduces the deploy reserved built-in step for Pipelines.
It supports a Release bool, but intentionally does not implement release
yet as that will be coming in a follow-up commit/pull request to this
one.

It also introduces a change in the deploy operation in that if there was
no artifact set on the jobs Deploy operation, it looks up the latest
pushed artifact to use that. This change is important because pipelines
queues its jobs ahead of time all at once, which means if at any point a
Build step runs prior, we couldn't look up the artifact without it
existing. So the Deploy step will not set the artifact, making it nil,
and when the Deploy goes to run it will look up the latest.


internal/runner/operation_deploy.go
```diff
@@ -24,6 +24,19 @@ func (r *Runner) executeDeployOp(
 		panic("operation not expected type")
 	}
 
+	if op.Deploy.Artifact == nil {
```
briancain, line 27: Not sure of the consequences of this change, but the reasoning behind this is Pipelines queue all of its Step operations at once ahead of time. That means we can't attach the "latest" build artifact then because it won't exist if there is a Build step prior that hasn't run yet. Pipelines will set the artifact to nil on the job to be queued and then the deploy can look it up here.

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -121,6 +123,19 @@ func (s *Service) RunPipeline(
...
+				Deploy: &pb.Job_DeployOp{},
+			}
+
+			if o.Deploy.Release {
+				// TODO(briancain): do it
+				// copy `job` and update Operation to be release I think. then append
+				// job to stepJobs
+
+				// Queue a release job too
+				log.Warn("Currently not queueing a release job yet....sry!!!")
```
briancain, line 137: I will implement this in a follow-up PR for the reserved built-in Release step!

internal/runner/operation_deploy.go
briancain: I think doing this also gets us one step closer to being able to update the `DeployOp` to use a `Ref_Artifact` instead of requiring the full artifact proto for a deploy:

https://github.com/hashicorp/waypoint/blob/d557c0a5f5b9fe8fad599d6fad8ea11b97da4173/pkg/server/proto/server.proto#L1965-L1968

https://github.com/hashicorp/waypoint/issues/2884

briancain:
Rebased this off `f-pipelines` which was recently rebased off `main`!

internal/runner/operation_deploy.go
paladin-devops: [Question] If an earlier build step in a pipeline fails, will the pipeline stop in its tracks (i.e. no remaining queued steps are run), or will it continue on failure? 

I'm thinking that if the pipeline continues to run, then what could possibly happen is with an earlier build phase failure, the pipeline could then unexpectedly re-deploy an old artifact by looking up the "latest" artifact here, since the one that was just attempted to be built failed (in this scenario).

Setting that question aside, I think this is a good way to solve the problem!

briancain: Good question! This is definitely and edge case I think, but possible if you tried hard enough with how pipelines are right now.

All steps are translated by Waypoint into jobs that run through the job system. By default, steps depend on the previous step in the pipeline unless it's the "root" or first step. So if that build step fails, the downstream jobs will be cancelled due to the job failure. You'd need to set the `depends_on` field in a weird way where the deploy relies on a successful step prior to the build failure for that to possibly happen but it wouldn't be the default case.

If we end up implementing #2884 too we can make a Deploy step include a `Ref_Artifact` that might reference a prior pipeline step and combined with explicitly having your deploy step `depends_on` the build, this might help in this case. 🤔

paladin-devops (approved):

xiaolin-ninja (approved):

PR #3473, opened 2022-06-20
Allow calling state from handler tests

izaaklauer:
Currently, the only way to call state from a handler test is to cast the server impl, and then reach in and call it's specific methods. This doesn't work in tests that we want to share with other server impls, because the casting is type-specific.

The workaround here is: server impls that want to use our tests will need to implement a TestServerImpl interface, that creates public equivalents to their private methods.

Example usage:

```
client, impl := factory(t)
_, err := impl.State(context.Background()).JobList(&pb.ListJobsRequest{})
require.NoError(t, err)
```


briancain (approved):
🙏🏻 thank you!! Looks great, makes sense to me 👍🏻

PR #3475, opened 2022-06-20
internal/server: Ensure task plugin is properly parsed when invoked

briancain:
Prior to this commit, when the server install function went to json
marshall its config into a k8s task config, it would translate the k8s
config to be `Requested`, but the k8s platform config expected `request`
and would fail to parse the profile. This commit ensures that the task
runner profile properly sets request for both mem and CPU so that it can
be parsed.

Additionally, instead of small resource limits and requests, the default
ODR profile for k8s will instead set each value to `"0"`, which according
to the k8s API is the same as being unset.

### DIFF
--- a/.changelog/3475.txt
+++ b/.changelog/3475.txt
@@ -1,0 +1,8 @@
+```release-note:bug
+plugin/k8s: Properly parse kubernetes task launcher config on plugin invoke.
+```
+
+```release-note:improvement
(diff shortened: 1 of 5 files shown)


krantzinator (approved):
🎉

briancain:
Weird, none of the go test or integration tests ran. Going to hold off on merging for a bit.

PR #3477, opened 2022-06-21
pulled out the restart functionality to its own function: TestJobHear…

cicoyle:
Pulled out the test for time out if running state loaded on restart into its own function `TestJobHeartbeatOnRestart` for cases where testing the timeout functionality for a restart may be less relevant.


teresamychu:
LGTM! I think you need to add a nochangelog label to the PR for the changelog-check to pass. But otherwise all good!

teresamychu (approved):

briancain (approved):
👍🏻 🧪

PR #3478, opened 2022-06-21
Test serverstate deploy filtering

izaaklauer:
We weren't testing this elsewhere, and it caught a bug for me.

We could also test this in all the other ops, but this is enough of an improvement for me for now.


briancain (approved):
Nice! Looks like the previous test does test some operation status filtering, but not these specific ones! Good catch

PR #3479, opened 2022-06-21
More error wrapping around ops

izaaklauer:
Due to an unrelated bug in the waypoint server, we were exiting early when creating status reports. The only error the client cound see was:

```
unknown error
```

Which is not helpful. This wrapps the errors in that flow to give more context as to what the client was attempting to do when it encountered the error.

### DIFF
--- a/internal/core/app_build.go
+++ b/internal/core/app_build.go
@@ -2,11 +2,11 @@ package core
 
 import (
 	"context"
-	"errors"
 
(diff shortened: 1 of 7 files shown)


briancain (approved):
Ahhh, this makes sense 👍🏻 Is the status report failing to execute bug fixed already or do we need a GH issue to track it? I think I noticed that the other day in my logs....

PR #3487, opened 2022-06-22
bugfix: Pass region to lambda-function releaser

demophoon:
Before this commit the aws configuration which was initialized as part
of the lambda-function releaser didn't inherit the region from the
lambda deploy step. This caused the lambda-function releaser to fail
with a "Region not specified" error from AWS during the deployment.

This commit grabs the region from the deployment phase so that we don't
have to rely on a users environment when initializing the AWS config for
determining the correct region the lambda is in for configuring the
lambda function url.


krantzinator (approved):
🎸

briancain (approved):

PR #3489, opened 2022-06-22
Add ECR Pull Plugin docs to Lambda page

paladin-devops:


briancain (approved):
Ahhh yes, this got us again! Looks good 👍🏻

PR #3490, opened 2022-06-22
Update runner profile during `server upgrade`

paladin-devops:


internal/cli/server_upgrade.go
```diff
@@ -389,10 +389,44 @@ func (c *ServerUpgradeCommand) upgradeRunner(
...
+		}
+
+		// Now that the new runner is installed, we can delete the old default runner profile
+		_, err = client.DeleteOnDemandRunnerConfig(ctx, &pb.DeleteOnDemandRunnerConfigRequest{
+			Config: &pb.Ref_OnDemandRunnerConfig{
+				Id:   oldRunnerConfig.Config.Id,
+				Name: oldRunnerConfig.Config.Name,
+			}})
+		if err != nil {
+			c.ui.Output("Error deleting previous default runner profile.", clierrors.Humanize(err), terminal.WithErrorStyle())
```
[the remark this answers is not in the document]
briancain: +1 to that, we could maybe also include details on how to manually delete and then suggest upgrading again too!

internal/cli/server_upgrade.go
```diff
@@ -389,10 +390,62 @@ func (c *ServerUpgradeCommand) upgradeRunner(
...
+		// Deleting on-demand runner configs isn't possible prior to version 0.9.0
+		if cliVersion.GreaterThanOrEqual(okVersion) {
+			// Now that the new runner is installed, we can delete the old default runner profile
+			_, err = client.DeleteOnDemandRunnerConfig(ctx, &pb.DeleteOnDemandRunnerConfigRequest{
+				Config: &pb.Ref_OnDemandRunnerConfig{
+					Id:   oldRunnerConfig.Config.Id,
+					Name: oldRunnerConfig.Config.Name,
+				}})
+			if err != nil {
+				c.ui.Output("Error deleting previous default runner profile. ID: %s, Name: %s", oldRunnerConfig.Config.Id, oldRunnerConfig.Config.Name, clierrors.Humanize(err), terminal.WithErrorStyle())
```
briancain, line 443: Ahh, this might of got lost in the review but it would be good if we could add some CLI suggestions (if possible) on how to delete or remove the default profile to accomplish a successful upgrade.

internal/cli/install.go
```diff
@@ -520,21 +522,26 @@ func installRunner(
...
-				s.Status(terminal.StatusError)
-			} else {
-				s.Update("Registered ondemand runner!")
-				s.Status(terminal.StatusOK)
-			}
+		version, err := client.GetVersionInfo(ctx, &empty.Empty{})
+		// The expected naming convention is "<platform>-default-<waypoint_version>
+		odrConfig.Name = odrConfig.Name + "default-" + version.Info.Version
+		if err != nil {
+			ui.Output("Error getting version: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
```
briancain, line 533: Return `err` or `1` here?

internal/cli/install.go
```diff
@@ -520,21 +522,26 @@ func installRunner(
...
+		}
+
+		_, err = client.UpsertOnDemandRunnerConfig(ctx, &pb.UpsertOnDemandRunnerConfigRequest{
+			Config: odrConfig,
+		})
+		if err != nil {
+			s.Update("Error creating ondemand runner: %s", err)
+			s.Status(terminal.StatusError)
```
briancain, line 541: Same feedback here, return err? I recommend making this a `ui.Output` instead of a step group too

internal/cli/server_upgrade.go
```diff
@@ -389,10 +389,42 @@ func (c *ServerUpgradeCommand) upgradeRunner(
...
+			c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		} else if err != nil && status.Code(err) == codes.NotFound {
+			c.ui.Output("Waypoint default runner profile not found, creating new profile", terminal.WithWarningStyle())
+		} else {
+			odr = &pb.OnDemandRunnerConfig{
+				Id:                   oldRunnerConfig.Config.Id,
+				Name:                 oldRunnerConfig.Config.Name,
+				TargetRunner:         oldRunnerConfig.Config.TargetRunner,
+				OciUrl:               "hashicorp/waypoint:latest",
```
briancain, line 413: This image URL can come from a CLI flag too, so we should use that instead of the static string here in case they provide their own image URL

briancain (changes requested):

briancain:
This PR warrants a changelog too I think!

xiaolin-ninja (approved):

.changelog/3490.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+upgrade: Update existing runner profile
```
briancain, line 2: This could use more detail here!

briancain (approved):
Looks good! I'd add more details to the changelog as to what this change means, but otherwise nice work! 🙏🏻 🎉

PR #3491, opened 2022-06-22
pipelines: Add Release as a built-in reserved step

briancain:
This commit introduces parsing the reserved release step. It also
introduces two new kinds of references: `Ref_Artifact` and `Ref_Deployment`.
This will be important because we need to let users reference a
Deployment either via latest (the default) or with a specific deployment
sequence number.

This PR fixes a unintended bug/coupling issue with a Release job in that it expected a full Deployment proto to exist prior to the job being queued, which is a similar issue to #2884.


briancain:
I'm going to go ahead and merge this into the parent PR.

PR #3493, opened 2022-06-23
pipelines: Add final reserved built-in pipeline step `Up`

briancain:
This commit introduces a built in Step for pipelines that executes a
full Build Deploy Release via the Up operation.


briancain:
This builds on top of the work from https://github.com/hashicorp/waypoint/pull/3491. Figured I'd make it a separate PR because the dependent PR has some major code changes that are unrelated to implementing the Up operation.

PR #3500, opened 2022-06-27
Server install: Nomad service discovery

### LINKED ISSUE #3376: Nomad 1.3.0: Node excluded because "${attr.consul.version} semver >= 1.7.0"
Following the official [tutorial](https://learn.hashicorp.com/tutorials/waypoint/get-started-nomad?in=waypoint/get-started-nomad#install-the-waypoint-server), when deploying the example nodejs application,  Waypoint errors out with:

```
* Constraint "${attr.consul.version} semver >= 1.7.0": 1 nodes excluded by filter
```

  Full logs below.  And my consul version is above 1.7.0 (at 1.12.0). 

**Steps to Reproduce**
Follow the [tutorial](https://learn.hashicorp.com/tutorials/waypoint/get-started-nomad?in=waypoint/get-started-nomad#install-the-waypoint-server)

```
nomad agent -dev -network-interface=enp8s0 -config=nomad.config
```

with this nomad.config to account for the host volume:
```
client {
  host_volume "waypoint-server" {
    path = "/data/waypoint-server"
    read_only = false
  }
}
```

with a waypoint install command (adjusted for host volume):
```
waypoint install -platform=nomad -accept-tos -nomad-host-volume=waypoint-server -nomad-consul-service=false  
```

waypoint, indeed, creates it's db:
```
root@vagrant: ~# ls -Rlh /data                         
/data:
total 4.0K
drwxr-xr-x 2 systemd-network vagrant 4.0K May 22 22:35 waypoint-server

/data/waypoint-server:
total 44K
-rw------- 1 systemd-network vagrant 64K May 22 22:42 data.db
-rw------- 1 systemd-network vagrant   0 May 22 22:35 waypoint-restore.db.lock
```

the official example nodejs waypoint was used --> https://github.com/hashicorp/waypoint-examples/blob/main/nomad/nodejs/waypoint.hcl 

**Expected behavior**
The nodejs application to start.

* Waypoint CLI Version:

```
root@vagrant:~# waypoint version
CLI: v0.8.2 (5b8885480)
Server: v0.8.2
root@vagrant:~# consul version
Consul v1.12.0
Revision 09a8cdb4
Protocol 2 spoken by default, understands 2 to 3 (agent will automatically use protocol >2 when speaking to compatible agents)

root@vagrant:~# nomad version
Nomad v1.3.0 (52e95d64113e01be05d585d8b4c07f6f19efebbc)
root@vagrant:~# vault version
Vault v1.10.3 (af866591ee60485f05d6e32dd63dde93df686dfb)
root@vagrant:~# docker info
Client:
 Debug Mode: false

Server:
 Containers: 2
  Running: 2
  Paused: 0
  Stopped: 0
 Images: 13
 Server Version: 19.03.11
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 7eba5930496d9bbe375fdf71603e610ad737d2b2
 runc version: v1.0.0-0-g84113ee
 init version: fec3683
 Security Options:
  apparmor
  seccomp
   Profile: default
 Kernel Version: 5.4.0-80-generic
 Operating System: Ubuntu 20.04.2 LTS
 OSType: linux
 Architecture: x86_64
 CPUs: 2
 Total Memory: 7.775GiB
 Name: vagrant
 ID: 7CYX:MNFM:YUVU:RI3M:XT3C:YZSH:R3ZP:VSOQ:M55N:QAW2:JSFU:74TB
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Registry: https://index.docker.io/v1/
 Labels:
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
```

**Additional context**

logs from a generic ubuntu focal:
```
root@vagrant:~/waypoint-examples/nomad/nodejs# ls
index.js  package.json  public  README.md  views  waypoint.hcl
root@vagrant:~/waypoint-examples/nomad/nodejs# waypoint init
✓ Project "example-nodejs" and all apps are registered with the server.

Project initialized!

You may now call 'waypoint up' to deploy your project or
commands such as 'waypoint build' to perform steps individually.
root@vagrant:~/waypoint-examples/nomad/nodejs# waypoint up
  Performing operation locally

» Building example-nodejs...
✓ Running build v1
Warning! Buildpacks are known to have issues on architectures other than amd64. The architecure being reported by the Docker server is "x86_64". We will still attempt to build the image,
but may run into issues.
Creating new buildpack-based image using builder: heroku/buildpacks:20
✓ Creating pack client
✓ Building image
 │ [exporter] Adding layer 'process-types'
 │ [exporter] Adding label 'io.buildpacks.lifecycle.metadata'
 │ [exporter] Adding label 'io.buildpacks.build.metadata'
 │ [exporter] Adding label 'io.buildpacks.project.metadata'
 │ [exporter] Setting default process type 'web'
 │ [exporter] Saving example-nodejs...
 │ [exporter] *** Images (42f4521666c2):
 │ [exporter]       example-nodejs
 │ [exporter] Adding cache layer 'heroku/nodejs-engine:dist'
 │ [exporter] Adding cache layer 'heroku/nodejs-npm:toolbox'
✓ Injecting entrypoint binary to image

Generated new Docker image: example-nodejs:latest
✓ Running push build v1
✓ Tagging Docker image: example-nodejs:latest => nodejs-example:1
✓ Docker image pushed: nodejs-example:1

» Deploying example-nodejs...
✓ Running deploy v1
✓ Job registration successful
✓ Evaluation status changed: "pending" -> "complete"
⚠️ Evaluation "59459995-c2e7-30cc-dbcf-3a3c04c51400" finished with status "complete" but failed to place all allocations
⚠️ Task Group "example-nodejs-01g3gwqgmrh02msgt0jhvg2v5t" (failed to place 1 allocation):
  * Constraint "${attr.consul.version} semver >= 1.7.0": 1 nodes excluded by filter
⚠️ Evaluation "860af1f2-6dd1-d4a1-ea55-ba161bda2c6b" waiting for additional capacity to place remainder
example-nodejs-01g3gwqgmrh02msgt0jhvg2v5t
! Failed to schedule all allocations
```

[full focal -vvv waypoint deploy log](https://gist.github.com/joshuacox/33d0ee83025cdf8a7a6faf2e7ebb132d)
[archlinux full debug](https://gist.github.com/joshuacox/0fbc21f55c149ca2eb128260b6f674f8)

[discuss.hashicorp topic](https://discuss.hashicorp.com/t/node-excluded-because-attr-consul-version-semver-1-7-0/39818)

EDIT:
I upgraded nomad to 1.3.1 and reran on a fresh focal install, pretty much the same results, I have a gist of the `waypoint deploy -vvv` step and the accompanying nomad stdout [here](https://gist.github.com/joshuacox/65431d3da42f24ff5cb2dfcb89b32087).

[Vagrantfile](https://gist.github.com/joshuacox/267062259e38de8a45545faca8b50467)

EDIT: workaround with consul [Vagrantfile](https://gist.github.com/joshuacox/7f9e45f6741a1ca18e059116650734f4)

```
vagrant up && vagrant ssh
```

then

```
sudo -i
./start.sh
```

paladin-devops:
This PR updates the server installation for Nomad to support service discovery for Nomad or Consul by specifying the service discovery provider as a flag for the `waypoint install` command. 

Closes #3376.

NOTE: The old config, `-nomad-consul-service`, is still supported with this PR update, but it should be phased out eventually. The new `-nomad-service-provider` config supports pre- and post-1.3.0 Waypoint installs to Nomad.

### DIFF
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -3,6 +3,7 @@ package serverinstall
 import (
 	"context"
 	json "encoding/json"
+	"errors"
 	"fmt"
 	"github.com/hashicorp/waypoint/internal/installutil"
 	"os"
@@ -350,8 +351,9 @@ func (i *NomadInstaller) Upgrade(
 				clierrors.Humanize(err),
 				terminal.WithErrorStyle(),
 			)
-		} else if strings.ToLower(proceed) != "yes" {
 			return nil, err
+		} else if strings.ToLower(proceed) != "yes" {
+			return nil, errors.New("upgrade aborted")
 		}
 	}
 
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -244,6 +244,9 @@ func (i *NomadInstaller) Install(
 	// if Nomad restarts the server allocation, a new IP will be assigned and any
 	// configured clients will be invalid
 	httpAddr, addr.Addr, err = i.getWaypointAddress(client, allocID)
+	if err != nil {
+		return nil, err
+	}
 
 	clicfg = clicontext.Config{
 		Server: serverconfig.Client{
@@ -507,6 +510,9 @@ func (i *NomadInstaller) Upgrade(
 	// if Nomad restarts the server allocation, a new IP will be assigned and any
 	// configured clients will be invalid
 	httpAddr, addr.Addr, err = i.getWaypointAddress(client, allocID)
+	if err != nil {
+		return nil, err
+	}
 
 	clicfg = clicontext.Config{
 		Server: serverconfig.Client{
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -844,8 +844,8 @@ func waypointNomadJob(c nomadConfig, rawRunFlags []string, upgrade bool) *api.Jo
 
 	// Include services to be registered. Currently configured to happen with Consul by default
 	// One service added for Waypoint UI, and one for Waypoint backend port
-	services := []*api.Service{}
-	if c.serviceProvider == "consul" || (c.consulService && c.serviceProvider == "consul") || (c.consulService && upgrade) {
+	var services []*api.Service
+	if (c.serviceProvider == "consul" && (c.consulService || upgrade)) || (c.consulService && (c.serviceProvider == "consul" || upgrade)) {
 		token := ""
 		if c.consulToken == "" {
 			token = os.Getenv("CONSUL_HTTP_TOKEN")
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -304,21 +304,17 @@ func (i *NomadInstaller) Install(
 
 	switch i.config.serviceProvider {
 	case "consul":
-		s = sg.Add("The CLI has been configured to automatically install a Consul service for\n" +
+		opts.UI.Output("The CLI has been configured to automatically install a Consul service for\n" +
 			"the Waypoint service backend and ui service in Nomad.")
-		s.Done()
 	case "nomad":
-		s = sg.Add("The CLI has been configured to automatically install a Nomad service for\n" +
+		opts.UI.Output("The CLI has been configured to automatically install a Nomad service for\n" +
 			"the Waypoint service backend and ui service in Nomad.")
-		s.Done()
 	default:
-		s = sg.Add("Waypoint server running on Nomad is being accessed via its allocation IP and port.\n" +
+		opts.UI.Output("Waypoint server running on Nomad is being accessed via its allocation IP and port.\n" +
 			"This could change in the future if Nomad creates a new allocation for the Waypoint server,\n" +
 			"which would break all existing Waypoint contexts.\n\n" +
 			"It is recommended to use Consul for determining Waypoint servers IP running on Nomad rather than\n" +
 			"relying on the static IP that is initially set up for this allocation.")
-		s.Status(terminal.StatusWarn)
-		s.Done()
 	}
 
 	return &InstallResults{
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -37,7 +37,8 @@ type nomadConfig struct {
 	namespace          string            `hcl:"namespace,optional"`
 	serviceAnnotations map[string]string `hcl:"service_annotations,optional"`
 
-	serviceProvider          string   `hcl:"service_provider,optional"`
+	serviceProvider          []string `hcl:"service_provider,optional"`
+	consulService            bool     `hcl:"consul_service,optional"`
 	consulServiceUITags      []string `hcl:"consul_service_ui_tags:optional"`
 	consulServiceBackendTags []string `hcl:"consul_service_backend_tags:optional"`
 	consulDatacenter         string   `hcl:"consul_datacenter,optional"`
@@ -239,8 +240,7 @@ func (i *NomadInstaller) Install(
 	// than the direct static IP for the CLI context and server config. Otherwise
 	// if Nomad restarts the server allocation, a new IP will be assigned and any
 	// configured clients will be invalid
-	switch i.config.serviceProvider {
-	case "consul":
+	if i.config.serviceProvider[0] == "consul" && i.config.consulService {
 		s.Update("Configuring the server context to use Consul DNS hostname")
 		if i.config.consulDatacenter == "" {
 			i.config.consulDatacenter = defaultConsulDatacenter
@@ -261,7 +261,7 @@ func (i *NomadInstaller) Install(
 			addr.Addr = fmt.Sprintf("%s:%d", i.config.consulServiceHostname, grpcPort)
 			httpAddr = fmt.Sprintf("%s:%d", i.config.consulServiceHostname, httpPort)
 		}
-	case "nomad":
+	} else if i.config.serviceProvider[0] == "nomad" {
 		backendService, _, err := client.Services().Get(waypointBackendServiceName, nil)
 		if err != nil {
 			return nil, err
@@ -274,7 +274,7 @@ func (i *NomadInstaller) Install(
 
 		httpAddr = fmt.Sprintf("%s:%d", uiService[0].Address, uiService[0].Port)
 		addr.Addr = fmt.Sprintf("%s:%d", backendService[0].Address, backendService[0].Port)
-	default:
+	} else {
 		s.Update("Configuring the server context to use the static IP address from the Nomad allocation")
 
 		serverAddr, err := getAddrFromAllocID(allocID, client)
@@ -302,14 +302,13 @@ func (i *NomadInstaller) Install(
 	s.Update("Waypoint server ready")
 	s.Done()
 
-	switch i.config.serviceProvider {
-	case "consul":
+	if i.config.serviceProvider[0] == "consul" && i.config.consulService {
 		opts.UI.Output("The CLI has been configured to automatically install a Consul service for\n" +
 			"the Waypoint service backend and ui service in Nomad.")
-	case "nomad":
+	} else if i.config.serviceProvider[0] == "nomad" {
 		opts.UI.Output("The CLI has been configured to automatically install a Nomad service for\n" +
 			"the Waypoint service backend and ui service in Nomad.")
-	default:
+	} else {
 		opts.UI.Output("Waypoint server running on Nomad is being accessed via its allocation IP and port.\n" +
 			"This could change in the future if Nomad creates a new allocation for the Waypoint server,\n" +
 			"which would break all existing Waypoint contexts.\n\n" +
@@ -335,7 +334,7 @@ func (i *NomadInstaller) Upgrade(
 	sg := ui.StepGroup()
 	defer sg.Wait()
 
-	if i.config.serviceProvider == "" {
+	if i.config.serviceProvider[0] == "none" {
 		// By default, we don't auto-enable the consul service because prior to Waypoint
 		// version 0.6.2, we did not enable it by default.
 		sw := sg.Add("Service discovery is disabled for the Waypoint Nomad job. If you had previously enabled " +
@@ -493,8 +492,7 @@ func (i *NomadInstaller) Upgrade(
 	// than the direct static IP for the CLI context and server config. Otherwise
 	// if Nomad restarts the server allocation, a new IP will be assigned and any
 	// configured clients will be invalid
-	switch i.config.serviceProvider {
-	case "consul":
+	if i.config.serviceProvider[0] == "consul" && i.config.consulService {
 		s.Update("Configuring the server context to use Consul DNS hostname")
 		if i.config.consulDatacenter == "" {
 			i.config.consulDatacenter = defaultConsulDatacenter
@@ -515,7 +513,7 @@ func (i *NomadInstaller) Upgrade(
 			addr.Addr = fmt.Sprintf("%s:%d", i.config.consulServiceHostname, grpcPort)
 			httpAddr = fmt.Sprintf("%s:%d", i.config.consulServiceHostname, httpPort)
 		}
-	case "nomad":
+	} else if i.config.serviceProvider[0] == "nomad" {
 		backendService, _, err := client.Services().Get(waypointBackendServiceName, nil)
 		if err != nil {
 			return nil, err
@@ -528,7 +526,7 @@ func (i *NomadInstaller) Upgrade(
 
 		httpAddr = fmt.Sprintf("%s:%d", uiService[0].Address, uiService[0].Port)
 		addr.Addr = fmt.Sprintf("%s:%d", backendService[0].Address, backendService[0].Port)
-	default:
+	} else {
 		s.Update("Configuring the server context to use the static IP address from the Nomad allocation")
 
 		serverAddr, err := getAddrFromAllocID(allocID, client)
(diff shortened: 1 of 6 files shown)


internal/serverinstall/nomad.go
```diff
@@ -1394,11 +1428,11 @@ func (i *NomadInstaller) UpgradeFlags(set *flag.Set) {
...
 
-	set.BoolVar(&flag.BoolVar{
-		Name:    "nomad-consul-service",
-		Target:  &i.config.consulService,
+	set.StringVar(&flag.StringVar{
+		Name:    "nomad-service-provider",
+		Target:  &i.config.serviceProvider,
 		Usage:   "Create service for Waypoint UI and Server in Consul.",
-		Default: false, // default to false, make sure people opt into this for upgrades
+		Default: "consul",
```
briancain, line 1435: This could be a good case for an `EnumVar` so we only ever allow `nomad` or `consul` and can fail early if they give us an unsupported option!

internal/serverinstall/nomad.go
```diff
@@ -285,17 +299,21 @@ func (i *NomadInstaller) Install(
...
+	case "nomad":
+		s = sg.Add("The CLI has been configured to automatically install a Nomad service for\n" +
+			"the Waypoint service backend and ui service in Nomad.")
+		s.Done()
+	default:
+		s = sg.Add("Waypoint server running on Nomad is being accessed via its allocation IP and port.\n" +
+			"This could change in the future if Nomad creates a new allocation for the Waypoint server,\n" +
+			"which would break all existing Waypoint contexts.\n\n" +
+			"It is recommended to use Consul for determining Waypoint servers IP running on Nomad rather than\n" +
+			"relying on the static IP that is initially set up for this allocation.")
```
briancain, line 316: Hmm, these three outputs might be better suited for `ui.Output` instead of a step group since they are single output! I think @evanphx has mentioned to be before that doing `\n` new lines in step groups might be a bad idea too.

briancain (changes requested):
Haven't tried the full install path for these yet, but I have a few minor code comments! Changes make sense to me otherwise 👍🏻

internal/serverinstall/nomad.go
evanphx: We were doing it previously, so I think it's fine. If it's weird we should probably just fix it so it works right too.

internal/serverinstall/nomad.go
```diff
@@ -1394,11 +1428,11 @@ func (i *NomadInstaller) UpgradeFlags(set *flag.Set) {
 		Usage:  "Nomad host volume name.",
 	})
 
-	set.BoolVar(&flag.BoolVar{
-		Name:    "nomad-consul-service",
-		Target:  &i.config.consulService,
+	set.StringVar(&flag.StringVar{
+		Name:    "nomad-service-provider",
```
evanphx, line 1432: Might be a good idea to support the old flag for a release or two rather than remove it immediately.

internal/serverinstall/nomad.go
paladin-devops: @briancain in order to still support an install method without a service discovery provider (pre-Nomad 1.3.0), I'll probably name this 3rd option (no service discovery provider) `none`.

internal/serverinstall/nomad.go
paladin-devops: @evanphx updated!

internal/serverinstall/nomad.go
```diff
@@ -813,29 +839,48 @@ func waypointNomadJob(c nomadConfig, rawRunFlags []string) *api.Job {
 	grpcPort, _ := strconv.Atoi(defaultGrpcPort)
 	httpPort, _ := strconv.Atoi(defaultHttpPort)
 
-	// Include services to be registered in Consul. Currently configured to happen by default
+	// Include services to be registered. Currently configured to happen with Consul by default
 	// One service added for Waypoint UI, and one for Waypoint backend port
-	if c.consulService {
+	var services []*api.Service
+	if (c.serviceProvider == "consul" && (c.consulService || upgrade)) || (c.consulService && (c.serviceProvider == "consul" || upgrade)) {
```
demophoon, line 845: Would it be worth while to emit a warning if we find that a user is using the `-nomad-consul-service` flag instead of the `-nomad-service-provider` flag?

demophoon: Oh, i see, its above :D ignore me!

internal/serverinstall/nomad.go
```diff
@@ -319,11 +332,11 @@ func (i *NomadInstaller) Upgrade(
...
 	defer sg.Wait()
 
-	if !i.config.consulService {
+	if i.config.serviceProvider == "none" && !i.config.consulService {
 		// By default, we don't auto-enable the consul service because prior to Waypoint
 		// version 0.6.2, we did not enable it by default.
-		sw := sg.Add("Nomad Consul Service is disabled. If you had previously enabled " +
-			"it in the last installation, please stop this upgrade and re-run with -nomad-consul-service=true.")
+		sw := sg.Add("Service discovery is disabled for the Waypoint Nomad job. If you had previously enabled " +
+			"it in the last installation, please stop this upgrade and re-run with -nomad-service-provider flag.")
```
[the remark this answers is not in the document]
paladin-devops: @demophoon after giving this some thought, I think an interactive terminal event could be useful here. If the user is OK with proceeding, we let them, otherwise they can abort!

```suggestion
		proceed, err := opts.UI.Input(&terminal.Input{
			Prompt: "Service discovery is disabled for the Waypoint Nomad job. If you had previously enabled " +
				"it in the last installation, please stop this upgrade and re-run with the -nomad-service-provider flag. " +
				"Otherwise, enter 'yes' to continue the upgrade: ",
			Style:  "",
			Secret: false,
		})
		if err != nil {
			opts.UI.Output(
				"Error uninstalling server: %s",
				clierrors.Humanize(err),
				terminal.WithErrorStyle(),
			)
		} else if strings.ToLower(proceed) != "yes" {
			return nil, err
		}
```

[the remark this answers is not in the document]
demophoon: This is a good solution to this particular concern!

internal/serverinstall/nomad.go
```diff
@@ -813,29 +853,61 @@ func waypointNomadJob(c nomadConfig, rawRunFlags []string) *api.Job {
 	grpcPort, _ := strconv.Atoi(defaultGrpcPort)
 	httpPort, _ := strconv.Atoi(defaultHttpPort)
 
-	// Include services to be registered in Consul. Currently configured to happen by default
+	// Include services to be registered. Currently configured to happen with Consul by default
 	// One service added for Waypoint UI, and one for Waypoint backend port
-	if c.consulService {
+	var services []*api.Service
+	if (c.serviceProvider == "consul" && (c.consulService || upgrade)) || (c.consulService && (c.serviceProvider == "consul" || upgrade)) {
```
catsby, line 859: This line is confusing for me so I tried to simplify it. I haven't been successful yet but I did find an edge case I wanted to ask about; could `serviceProvider` be `"nomad"`, `consulService` be `true`, while also `upgrade` is `true`? If so, then this top block will execute and not the `if c.serviceProvider == "nomad"` one.

See https://go.dev/play/p/3lLyP9eqHa1

internal/serverinstall/nomad.go
```diff
@@ -813,29 +853,61 @@ func waypointNomadJob(c nomadConfig, rawRunFlags []string) *api.Job {
...
 		token := ""
 		if c.consulToken == "" {
 			token = os.Getenv("CONSUL_HTTP_TOKEN")
 		} else {
 			token = c.consulToken
 		}
 		job.ConsulToken = &token
```
catsby, line 866:
```suggestion
		token := c.consulToken
		if token == "" {
			token = os.Getenv("CONSUL_HTTP_TOKEN")
		}
		job.ConsulToken = &token

```

I'm not sure if we should check if token is empty before storing it in `job.ConsulToken` or if it's fine to put something empty in there ¯\\_(ツ)_/¯

catsby (commented):
Some nits and a question about some settings regarding service providers

internal/serverinstall/nomad.go
paladin-devops: @catsby yes, the top block would indeed execute. However since `consulService` is false by default _for upgrades only_, a user would have to explicitly set `consulService` to true, & set `serviceProvider` to `nomad`. In light of @evanphx's earlier feedback for supporting the old configuration options, I felt that it was "okay" to favor the old flag being set over the new one.

Maybe it would make sense to print out some kind of warning or prompt for input if the input options are colliding...
```go
if c.consulService && c.serviceProvider == "nomad" {
  // warn the user this combo isn't valid
}
```

catsby: I don't have operational experience here so I simply posed the question of whether it's valid or not after I noticed the possibility. I'll defer to you for the decision on if it's correct behavior or not 👍

internal/serverinstall/nomad.go
```diff
@@ -491,13 +518,26 @@ func (i *NomadInstaller) Upgrade(
...
-				waypointConsulBackendName, i.config.consulDatacenter, i.config.consulDomain, grpcPort)
+				waypointBackendServiceName, i.config.consulDatacenter, i.config.consulDomain, grpcPort)
 			httpAddr = fmt.Sprintf("%s.service.%s.%s:%d",
-				waypointConsulUIName, i.config.consulDatacenter, i.config.consulDomain, httpPort)
+				waypointUIServiceName, i.config.consulDatacenter, i.config.consulDomain, httpPort)
 		} else {
 			addr.Addr = fmt.Sprintf("%s:%d", i.config.consulServiceHostname, grpcPort)
 			httpAddr = fmt.Sprintf("%s:%d", i.config.consulServiceHostname, httpPort)
 		}
+	} else if i.config.serviceProvider == "nomad" {
```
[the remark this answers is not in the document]
paladin-devops: So refactored!

internal/serverinstall/nomad.go
```diff
@@ -319,13 +334,25 @@ func (i *NomadInstaller) Upgrade(
...
+		})
+		if err != nil {
+			opts.UI.Output(
+				"Error upgrading server: %s",
+				clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+		} else if strings.ToLower(proceed) != "yes" {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
paladin-devops: You summarized it correctly, but I think this needs a change! We should return the error if there was an error getting the input.

```suggestion
		if err != nil {
			opts.UI.Output(
				"Error upgrading server: %s",
				clierrors.Humanize(err),
				terminal.WithErrorStyle(),
			)
			return nil, err
		} else if strings.ToLower(proceed) != "yes" {
			return nil, err
		}
```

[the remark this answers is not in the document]
paladin-devops: You summarized this correctly, but it should be changed!

```suggestion
		if err != nil {
			opts.UI.Output(
				"Error upgrading server: %s",
				clierrors.Humanize(err),
				terminal.WithErrorStyle(),
			)
			return nil, err
		} else if strings.ToLower(proceed) != "yes" {
			return nil, errors.New("upgrade aborted")
		}
```

briancain (approved):
🔍 Looks good to me code wise 👍🏻

izaaklauer (approved):

internal/serverinstall/nomad.go
```diff
@@ -286,19 +255,18 @@ func (i *NomadInstaller) Install(
 	s.Update("Waypoint server ready")
 	s.Done()
 
-	if i.config.consulService {
-		s = sg.Add("The CLI has been configured to automatically install a Consul service for\n" +
+	if i.config.serviceProvider == "consul" && i.config.consulService {
```
xiaolin-ninja, line 258: I recall from a code review meeting that this was supposed to be an "or" instead of an "and"?

catsby (changes requested):

internal/serverinstall/nomad.go
```diff
@@ -237,42 +241,7 @@ func (i *NomadInstaller) Install(
...
-		}
-		hAddr, err := getHTTPFromAllocID(allocID, client)
-		if err != nil {
-			return nil, err
-		}
-
-		httpAddr = hAddr
-		addr.Addr = serverAddr
-	}
+	httpAddr, addr.Addr, err = i.getWaypointAddress(client, allocID)
```
catsby, line 244: We are not checking this `err`, we should, or use `_` if it's something we can ignore

internal/serverinstall/nomad.go
```diff
@@ -477,42 +504,7 @@ func (i *NomadInstaller) Upgrade(
...
```
catsby, line 507: We are not checking this `err`, we should, or use `_` if it's something we can ignore

internal/serverinstall/nomad.go
paladin-devops: @xiaolin-ninja since the default values are "consul" and `true` for `serviceProvider` and `consulService`, respectively, we want this to be "and" because if the user sets something other than "consul" for `serviceProvider`, or sets `consulService` to false, we then do not want to create the job with a Consul service.

briancain:
You might need to rebase off `main` to get the integration tests to pass, btw!

catsby (approved):

PR #3501, opened 2022-06-27
internal/cli: Include default value for a flag for docs

briancain:
Prior to this commit, our docs CLI generator would not include the
default value for a flag when auto-generating our website docs. The CLI
docs include it, so now as of this commit, our website docs for the CLI
will also include it.

### DIFF
--- a/internal/cli/docs.go
+++ b/internal/cli/docs.go
@@ -199,9 +199,21 @@ description: "%s"
 				if len(f.Aliases) > 0 {
 					aliases := strings.Join(f.Aliases, "`, `-")
 
-					fmt.Fprintf(w, "- `-%s` (`-%s`) - %s\n", name, aliases, f.Usage)
+					str := fmt.Sprintf("- `-%s` (`-%s`) - %s", name, aliases, f.Usage)
+					if f.Default != "" {
+						str = fmt.Sprintf("%s The default is %s.", str, f.Default)
+					}
+					str += "\n"
+
+					fmt.Fprint(w, str)
 				} else {
-					fmt.Fprintf(w, "- `-%s` - %s\n", name, f.Usage)
+					str := fmt.Sprintf("- `-%s` - %s", name, f.Usage)
+					if f.Default != "" {
+						str = fmt.Sprintf("%s The default is %s.", str, f.Default)
+					}
+					str += "\n"
+
+					fmt.Fprint(w, str)
 				}
 			})
 		})
--- a/internal/cli/docs.go
+++ b/internal/cli/docs.go
@@ -205,7 +205,7 @@ description: "%s"
 
 				// Add a period at the end of the doc sentence if the field didn't add
 				// one already.
-				if str[len(str)-1:] != "." {
+				if !strings.HasSuffix(str, ".") {
 					str += "."
 				}
 
--- a/internal/cli/docs.go
+++ b/internal/cli/docs.go
@@ -196,25 +196,19 @@ description: "%s"
 					}
 				}
 
+				str := fmt.Sprintf("- `-%s` - %s", name, f.Usage)
 				if len(f.Aliases) > 0 {
 					aliases := strings.Join(f.Aliases, "`, `-")
 
-					str := fmt.Sprintf("- `-%s` (`-%s`) - %s", name, aliases, f.Usage)
-					if f.Default != "" {
-						str = fmt.Sprintf("%s The default is %s.", str, f.Default)
-					}
-					str += "\n"
-
-					fmt.Fprint(w, str)
-				} else {
-					str := fmt.Sprintf("- `-%s` - %s", name, f.Usage)
-					if f.Default != "" {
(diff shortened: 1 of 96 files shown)


briancain:
Looks like our `website-mdx` circle check will be indefinetly broken because these flags have dynamic default values: https://github.com/hashicorp/waypoint/pull/3501/commits/7d25acde6c81084683276bdffe6980da61d31bdf

Not sure what we should do about this.

website/content/commands/docs.mdx
```diff
@@ -41,8 +41,8 @@ The flags can change which plugins are listed and in which format.
 
 #### Command Options
 
-- `-builtin` - Show documentation on all builtin plugins
-- `-markdown` - Show documentation in markdown format
+- `-builtin` - Show documentation on all builtin plugins The default is false.
+- `-markdown` - Show documentation in markdown format The default is false.
```
catsby, line 45: I see a handful of these cases where the docs don't end in a period, and this new auto-generated part doesn't start with one. I'm fine with just fixing the isolated cases and adding a period at the end instead of more code that checks for it, but we could do that as well

internal/cli/docs.go
```diff
@@ -199,9 +199,21 @@ description: "%s"
...
+					fmt.Fprint(w, str)
 				} else {
-					fmt.Fprintf(w, "- `-%s` - %s\n", name, f.Usage)
+					str := fmt.Sprintf("- `-%s` - %s", name, f.Usage)
+					if f.Default != "" {
+						str = fmt.Sprintf("%s The default is %s.", str, f.Default)
+					}
+					str += "\n"
+
+					fmt.Fprint(w, str)
```
catsby, line 216: Nit: It looks like this block and the one above on ~202 are identical, so we could move them outside of the `if-else` block. ex:

```go
var str string
if len(f.Aliases) > 0 {
  // alias stuff
  str = fmt.Sprintf(//things)
} else {
  str = fmt.Sprintf(//things)
}
if f.Default != "" {
	str = fmt.Sprintf("%s The default is %s.", str, f.Default)
}
str += "\n"
fmt.Fprint(w, str)
```

catsby (approved):
LGTM minor nit and large ugh to broken test scenario :/

internal/cli/docs.go
briancain: Agreed! I think I beatcha by a few seconds. Got this simplified in [315ab8d](https://github.com/hashicorp/waypoint/pull/3501/commits/315ab8d5ccc858884ab8dd7375979cc1134182df)

website/content/commands/docs.mdx
briancain: True....I'll do a quick scan and fix these. It might be nice to have the auto-check for future docs in the case where we forget to add periods.

briancain: Done in [4bcbc1c](https://github.com/hashicorp/waypoint/pull/3501/commits/4bcbc1ce28c7869cfcd015def8659ebdb299e175)!

briancain:
Vercel failure looks unrelated to this PR 🤔 /cc @BRKalow 

```
./node_modules/@reach/tooltip/node_modules/@reach/auto-id/dist/reach-auto-id.esm.js
--
09:34:46.431 | Attempted import error: 'useId' is not exported from 'react' (imported as 'useId$1').
09:34:46.432 |  
09:34:46.432 | Import trace for requested module:
09:34:46.432 | ./node_modules/@reach/tooltip/dist/reach-tooltip.esm.js
09:34:46.432 | ./src/components/tooltip/index.tsx
09:34:46.432 | ./src/components/tutorial-collection-cards/components/card-badges/index.tsx
09:34:46.432 | ./src/components/tutorial-collection-cards/components/index.tsx
09:34:46.432 | ./src/components/tutorial-collection-cards/index.tsx
09:34:46.432 | ./src/components/collection-card/index.tsx
09:34:46.432 | ./src/views/homepage/components/learn-section/index.tsx
09:34:46.432 | ./src/views/homepage/index.tsx
09:34:46.432 | ./src/pages/index.tsx
09:34:46.432 |  
09:34:46.433 | ./node_modules/@reach/tooltip/node_modules/@reach/auto-id/dist/reach-auto-id.esm.js
09:34:46.433 | Attempted import error: 'useId' is not exported from 'react' (imported as 'useId$1').
```

internal/cli/install.go
```diff
@@ -359,7 +364,7 @@ func (c *InstallCommand) Flags() *flag.Sets {
 		f.StringVar(&flag.StringVar{
 			Name:    "context-create",
 			Target:  &c.contextName,
-			Default: fmt.Sprintf("install-%d", time.Now().Unix()),
+			Default: "install-timestamp",
```
catsby, line 367: Why the sentinel value here, instead of removing the `Default` and adding to the `Usage` something like `"The default value is 'install-' with a timestamp suffix, ex: 'install-123456'"`

briancain: Sure, I can do that!

PR #3502, opened 2022-06-28
Validating fields that can be used as path tokens

izaaklauer:
Fields that can be used as GRPC gateway path tokens need some additional validation. For now, we ensure they don't contain a path traversal.

### DIFF
--- a/pkg/serverhandler/handlertest/test_service_project.go
+++ b/pkg/serverhandler/handlertest/test_service_project.go
@@ -16,6 +16,7 @@ func init() {
 		TestServiceProject,
 		TestServiceProject_GetApplication,
 		TestServiceProject_UpsertApplication,
+		TestServiceProject_InvalidName,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pkg/serverhandler/handlertest/test_service_project.go
```diff
@@ -210,3 +211,21 @@ func TestServiceProject_UpsertApplication(t *testing.T, factory Factory) {
...
+
+func TestServiceProject_InvalidName(t *testing.T, factory Factory) {
+	ctx := context.Background()
+	require := require.New(t)
+	client, _ := factory(t)
+
+	// GRPC Gateway interprets ../ as a path traversal, and therefore we cannot allow
+	// '../' in any fields we use as path tokens.
+	project := ptypes.TestProject(t, &pb.Project{
+		Name: "../../",
```
briancain, line 223: We should probably also enforce this at the state level when we go to upsert a project name, if we don't already have some kind of name validation there? 🤔

briancain (approved):

NodyHub (approved):
Should work to fix that one issue from the security review. 

General recommendation from security is to switch to an allow-listing approach for user-input, e.g., allowed charset `[a-zA-Z0-9]`. A project name like `Test Project'); DROP TABLE PROJECTS; --` might be possible, but for what reason? :) 

We might run in future situation where that missing validation lead to a bigger issue. Since we are still in beta-phase it should be IMHO not a big thing to introduce that change before GA.

pkg/serverhandler/handlertest/test_service_project.go
briancain: The main thing is going forward, any new API endpoints now have to know about this path validation. If it's in the state package, every current and future API endpoint will behave this way no matter how many new ones we add.

PR #3503, opened 2022-06-28
Check plugin config during upgrade

paladin-devops:
Updates the upgrade process to verify that the plugin configuration for the pre-0.9 default runner profile is correct, and warn the user if it is not.

NOTE: this does not automatically fix the runner profile's configuration, nor does it create an example plugin configuration file for the user to supply this command. In its current state, it simply warns the user that their plugin config is not correct for the given platform.


internal/cli/server_upgrade.go
```diff
@@ -446,6 +448,24 @@ func (c *ServerUpgradeCommand) upgradeRunner(
 					runnerDefaultNames = append(runnerDefaultNames, fmt.Sprintf(runnerDefaultName, cfg.Name))
 					runnerUnsetStr = append(runnerUnsetStr, fmt.Sprintf(runnerUnsetDefault, cfg.PluginType, cfg.Name))
 				}
+
+				// Checking if the default runner profile for a platform, pre-0.9, has the correct task launcher configs
+				// If incorrect, update them
```
paladin-devops, line 453: This comment is (currently) inaccurate, but will leave it until we decide what the upgrade should do here.

internal/cli/server_upgrade.go
```diff
@@ -586,5 +606,11 @@ Please run the following commands if you wish to unset these runner profiles
 from being the default:
 
 %[2]s
+`)
+
+	runnerProfileUpgradeConfigError = strings.TrimSpace(`
+The plugin config for runner profile %[1]s is incorrect. Please run
+waypoint runner profile set -name=%[2]s -plugin-config=<path_to_config_file>
+with the corrected plugin configuration to fix this.
```
briancain, line 614:
```suggestion
The plugin config for runner profile %[1]s failed to correctly be parsed.
Plugin Config Err: %[3]s

Please run the following with the corrected plugin configuration to fix this.

waypoint runner profile set -name=%[2]s -plugin-config=<path_to_config_file>
```

We should include the error here too, which I think will show the "validated" reason for why the config isn't correct. Something like this might come from the parser.... `"'Requested' was not expected...."`

We don't have to do it in this PR but if we can it would be great to fix these docs too, they don't have the right type and need to be updated to include the nested block fields 😭 That might make this easier for people to figure out how to set the right config: https://tip.waypointproject.io/plugins/kubernetes#required-parameters-4

internal/cli/server_upgrade.go
```diff
@@ -586,5 +620,14 @@ Please run the following commands if you wish to unset these runner profiles
...
 %[2]s
+`)
+
+	runnerProfileUpgradeConfigError = strings.TrimSpace(`
+The plugin config for runner profile %[1]s failed to correctly be parsed. 
+Plugin Config Error: %[3]s
+
+Please run the following with the corrected plugin configuration to fix this.
+
+waypoint runner profile set -name=%[2]s -plugin-config=<path_to_config_file>
```
catsby, line 631: Should our error message tell people that the format changed starting with `0.9`, which is why it's not parsing correctly?

briancain: Ah yeah, we should be more specific. The issue was our old server install CLI laid down a invalid on-demand runner profile config for kubernetes, and prior to this release, we never validated Task plugin configs prior to a task being invoked and threw away the config validation error, so if you upgrade with a bad config all future tasks will fail to be launched after upgrades.

internal/cli/server_upgrade.go
```diff
@@ -446,6 +448,38 @@ func (c *ServerUpgradeCommand) upgradeRunner(
...
+					switch c.platform {
+					case "kubernetes":
+						// attempt to parse the runner profile config into the correct task launcher config struct
+						var result *k8s.TaskLauncherConfig
+						err = json.Unmarshal(cfg.PluginConfig, result)
+						if err != nil {
+							c.ui.Output(runnerProfileUpgradeConfigError, cfg.Name,
+								cfg.Name, clierrors.Humanize(err), terminal.WithWarningStyle())
+							var content map[string]interface{}
+							err = json.Unmarshal(cfg.PluginConfig, &content)
```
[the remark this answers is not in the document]
briancain: I think so, this shouldn't happen, but it's possible the json config is malformed json, in which case this would error. So we should return an error for sure!

internal/cli/server_upgrade.go
```diff
@@ -446,6 +448,38 @@ func (c *ServerUpgradeCommand) upgradeRunner(
...
 					runnerUnsetStr = append(runnerUnsetStr, fmt.Sprintf(runnerUnsetDefault, cfg.PluginType, cfg.Name))
 				}
+
+				// Checking if the default runner profile for a platform, pre-0.9, has the correct task launcher configs
+				if c.platform == cfg.Name {
+					switch c.platform {
+					case "kubernetes":
+						// attempt to parse the runner profile config into the correct task launcher config struct
+						var result *k8s.TaskLauncherConfig
+						err = json.Unmarshal(cfg.PluginConfig, result)
```
briancain, line 458: I think we also have to check if `cfg.ConfigFormat == pb.Hcl_JSON` in case their default runner profile was saved as HCL instead of json.

krantzinator (approved):

PR #3516, opened 2022-07-01
docs: update go-bindata install cmd

cicoyle:
The original `go get` command didn't work. The only command that worked to get go-bindata installed was the brew command. I figured it might be easier for others to not go thru the same install headache and just change it to one that works on the first try.


izaaklauer (approved):
Agreed, the `go get` command never worked for me either. I think `go install` might have worked at some point for me, but I can't easily replicate it.

At this point, most waypoint devs are on MacOS, and brew will be the lowest friction way for them. The linux and windows minority can click through to the project link and read the installation guide.

PR #3521, opened 2022-07-01
Skip service account creation for k8s `runner install` if exists

paladin-devops:
The name of the runner & ODR service accounts was hard-coded previously in order to support ODRs after an initial `waypoint server install` to k8s. The problem is that Helm fails to install additional releases for subsequent `waypoint runner install` runs, because the service account whose name is hard-coded already exists. 

```
$ waypoint runner install -server-addr=waypoint-server:9701 -server-tls-skip-verify -platform=kubernetes
✓ Finished connecting to: waypoint-server:9701
❌ Installing runner...
❌ Installing Waypoint Helm chart with runner options: waypoint
! Error installing runner: rendered manifests contain a resource that already exists. Unable to continue
  with install: ServiceAccount "waypoint-runner-odr" in namespace "default" exists
  and cannot be imported into the current release: invalid ownership metadata;
  annotation validation error: key "meta.helm.sh/release-name" must equal
  "waypoint-01g6xqhcdvm7wc7xbzyj3b2qtr": current value is
  "waypoint-01g6xqh06ya9xh5vxf5t6eby7q"
```

This update checks the k8s cluster to verify if the service account exists and will skip creating it if it does.

### DIFF
--- a/internal/runnerinstall/k8s.go
+++ b/internal/runnerinstall/k8s.go
@@ -165,6 +165,8 @@ func (i *K8sRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
 				return err
 			}
 		} else {
+			opts.UI.Output("Waypoint runner service account already exists - a new service account will not be created",
+				terminal.WithInfoStyle())
 			i.Config.CreateServiceAccount = false
 		}
 	}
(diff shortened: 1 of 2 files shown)


cicoyle (approved):
Great job!

PR #3522, opened 2022-07-02
chore: add hcl `replace`

thiskevinwang:
# Description

This adds the `replace` function from https://github.com/zclconf/go-cty

This achieves parity with https://www.terraform.io/language/functions/replace#replace-function

### DIFF
--- a/pkg/config/funcs/doc.go
+++ b/pkg/config/funcs/doc.go
@@ -50,7 +50,7 @@ func Docs() map[string]string {
 		"range":           "generate a list of numbers",
 		"regex":           "match the given string against the given regular expression pattern, returning captures if defined",
 		"regexall":        "same as regex but look for all matches of the given pattern rather than just the first",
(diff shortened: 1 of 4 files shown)


pkg/config/funcs/doc.go
```diff
@@ -50,6 +50,7 @@ func Docs() map[string]string {
 		"range":           "generate a list of numbers",
 		"regex":           "match the given string against the given regular expression pattern, returning captures if defined",
 		"regexall":        "same as regex but look for all matches of the given pattern rather than just the first",
+		"replace":         "`replace` searches a given string for another given substring, and replaces each occurrence with a given replacement string.",
```
briancain, line 53: Small nitpick (sorry!) but I think this reads clearer and matches the other doc func strings.

```suggestion
		"replace":         "searches a given string for another given substring, and replaces each occurrence with a given replacement string.",
```

briancain (changes requested):
Looks good, just a small comment about the doc string!

briancain (approved):

PR #3526, opened 2022-07-05
Add flags to Nomad server upgrade command for the bootstrap runner

paladin-devops:
The bootstrap runner's persistent volume is required and configurable with the `waypoint install` command in 0.9. Flags to configure the volume are added to the `waypoint server upgrade` command to ensure the runner's upgrade alongside the server is successful.


krantzinator (approved):
This looks good! I wonder if there's a better long-term way we can let people know what their original install config was and select to use it or provide new values; or allow install configs to be supplied via a file so that can be more IAC. 🤔 Just Tuesday morning thoughts. I think this comes up in conversation every so often.

PR #3538, opened 2022-07-06
Fix fix typo

krantzinator:
Fixes the typo fixed in #3532 that broke the `make gen/website-mdx` comparisons.
This is grammatically incorrect but what our tool enforces. If we leave the `.` inside the `**` then the website gen checker gets mad that there is a line that doesn't end with a `.`.


krantzinator:
@ansgarm no worries at all, tooling is what it is 😄 Sneaky buggers.

paladin-devops (approved):

PR #3546, opened 2022-07-06
core/install: Add --nomad-csi-volume-name flag to install

demophoon:
This commit allows users to specify their own volume names within their
CSI provider. Some cases where the upstream provider is incompatible
with the default name because of limitations on their end make it so
that waypoint is unable to be used within those providers. In my testing
the DigitalOcean CSI provider doesn't handle capitalization of the
volume's name specifically within the runner ID. I suspect other 
providers may have different requirements which we should be able to 
work around since there is not an explicit requirement in the CSI spec 
for the names of volumes within a provider.

### DIFF
--- a/internal/runnerinstall/nomad.go
+++ b/internal/runnerinstall/nomad.go
@@ -292,7 +292,7 @@ func (i *NomadRunnerInstaller) InstallFlags(set *flag.Set) {
 	set.StringVar(&flag.StringVar{
 		Name:   "nomad-csi-volume",
 		Target: &i.Config.CsiVolume,
-		Usage:  fmt.Sprintf("The name of the volume to initialize within the CSI provider. The default is %s.", defaultRunnerName("<runner_id>")),
+		Usage:  fmt.Sprintf("The name of the volume to initialize within the CSI provider. The default is %s.", defaultRunnerName("[runner_id]")),
 	})
 }
 
--- a/internal/runnerinstall/nomad.go
+++ b/internal/runnerinstall/nomad.go
@@ -43,6 +43,7 @@ type NomadConfig struct {
 	CsiExternalId        string            `hcl:"nomad_csi_external_id,optional"`
 	CsiPluginId          string            `hcl:"nomad_csi_plugin_id,required"`
 	CsiSecrets           map[string]string `hcl:"nomad_csi_secrets,optional"`
+	CsiVolumeName        string            `hcl:"nomad_csi_volume_name,optional"`
 
 	NomadHost string `hcl:"nomad_host,optional"`
 }
@@ -73,13 +74,16 @@ func (i *NomadRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) e
 		if i.Config.HostVolume != "" {
 			return fmt.Errorf("choose either CSI or host volume, not both")
 		}
+		if i.Config.CsiVolumeName == "" {
+			i.Config.CsiVolumeName = "waypoint-" + opts.Id + "-runner"
+		}
 
 		s = sg.Add("Creating persistent volume")
 		err = nomadutil.CreatePersistentVolume(
 			ctx,
 			client,
 			"waypoint-"+opts.Id+"-runner",
-			"waypoint-"+opts.Id+"-runner",
+			i.Config.CsiVolumeName,
 			i.Config.CsiPluginId,
 			i.Config.CsiVolumeProvider,
 			i.Config.CsiFS,
@@ -283,6 +287,12 @@ func (i *NomadRunnerInstaller) InstallFlags(set *flag.Set) {
 		Name:   "nomad-csi-secrets",
 		Target: &i.Config.CsiSecrets,
 		Usage:  "Credentials for publishing volume for Waypoint runner.",
+	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "nomad-csi-volume-name",
+		Target: &i.Config.CsiVolumeName,
+		Usage:  "The name of the volume to initialize within the CSI provider.",
 	})
 }
 
--- a/internal/runnerinstall/nomad.go
+++ b/internal/runnerinstall/nomad.go
@@ -75,14 +75,14 @@ func (i *NomadRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) e
 			return fmt.Errorf("choose either CSI or host volume, not both")
 		}
 		if i.Config.CsiVolumeName == "" {
-			i.Config.CsiVolumeName = "waypoint-" + opts.Id + "-runner"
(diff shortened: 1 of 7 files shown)


internal/runnerinstall/nomad.go
```diff
@@ -73,13 +74,16 @@ func (i *NomadRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) e
 		if i.Config.HostVolume != "" {
 			return fmt.Errorf("choose either CSI or host volume, not both")
 		}
+		if i.Config.CsiVolumeName == "" {
+			i.Config.CsiVolumeName = "waypoint-" + opts.Id + "-runner"
```
paladin-devops, line 78: I think we could use the new function `defaultRunnerName` here, in `runnerinstall.go`.

internal/runnerinstall/nomad.go
```diff
@@ -73,13 +74,16 @@ func (i *NomadRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) e
...
+		}
 
 		s = sg.Add("Creating persistent volume")
 		err = nomadutil.CreatePersistentVolume(
 			ctx,
 			client,
 			"waypoint-"+opts.Id+"-runner",
-			"waypoint-"+opts.Id+"-runner",
+			i.Config.CsiVolumeName,
```
paladin-devops, line 86: If the volume name is customizable, I think that the [way which we find the volume to clean up](https://github.com/hashicorp/waypoint/pull/3546/files#diff-d3925a67c54177f1d703959aa6e1bdce099cfdd2768b9155b621873b26e3781dR371) during `runner uninstall` might need to be updated as well. Currently for the uninstall, the volume name is tied to the Nomad job name.

paladin-devops (changes requested):
Nice job considering the flexible requirements of CSI implementations! Just had one suggestion to use a helpful function, but I think we'll need to revise the way the CSI volume is "uninstalled", when we have a customizable name.

internal/runnerinstall/nomad.go
demophoon: That's a great call out, and I also had worried about that and didn't know how to put the context into the commit but it looks like we only ever use the name when we setup the volume during creation. During uninstallation we look up volumes using the volume ID we provided the CSI provider at install time which has not been changed.

See https://github.com/hashicorp/waypoint/blob/main/internal/runnerinstall/nomad.go#L356 which adds the `prefix` url parameter when querying Nomad for a list of volumes. It doesn't look like its explicitly documented on Nomad's docs website but the `prefix` query parameter lists all volumes starting with the ID given. (also see it in nomad land: https://github.com/hashicorp/nomad/blob/main/nomad/csi_endpoint.go#L134-L135)

internal/runnerinstall/nomad.go
demophoon: Resolved in `68a0515c2`

internal/runnerinstall/nomad.go
```diff
@@ -43,6 +43,7 @@ type NomadConfig struct {
 	CsiExternalId        string            `hcl:"nomad_csi_external_id,optional"`
 	CsiPluginId          string            `hcl:"nomad_csi_plugin_id,required"`
 	CsiSecrets           map[string]string `hcl:"nomad_csi_secrets,optional"`
+	CsiVolumeName        string            `hcl:"nomad_csi_volume_name,optional"`
```
xiaolin-ninja, line 46: nit, let's group this with the other `csi_volume_` options. & maybe just `CsiVolume` without the `Name`? Since `HostVolume` isn't `HostVolumeName`

internal/serverinstall/nomad.go
```diff
@@ -67,6 +67,7 @@ type nomadConfig struct {
 	csiTopologies        map[string]string `hcl:"nomad_csi_topologies,optional"`
 	csiSecrets           map[string]string `hcl:"nomad_csi_secrets,optional"`
 	csiParams            map[string]string `hcl:"csi_parameters,optional"`
+	csiVolumeName        string            `hcl:"nomad_csi_volume_name,optional"`
```
xiaolin-ninja, line 70: same comment here as in the runner install file.

internal/runnerinstall/nomad.go
demophoon: Resolved in https://github.com/hashicorp/waypoint/pull/3546/commits/82a595eb404e8bd86ba9d7b6276ba4e894ac9137

internal/serverinstall/nomad.go
demophoon: Resolved in https://github.com/hashicorp/waypoint/pull/3546/commits/82a595eb404e8bd86ba9d7b6276ba4e894ac9137

internal/runnerinstall/nomad.go
```diff
@@ -284,6 +288,12 @@ func (i *NomadRunnerInstaller) InstallFlags(set *flag.Set) {
 		Target: &i.Config.CsiSecrets,
 		Usage:  "Credentials for publishing volume for Waypoint runner.",
 	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "nomad-csi-volume-name",
+		Target: &i.Config.CsiVolumeName,
+		Usage:  "The name of the volume to initialize within the CSI provider.",
```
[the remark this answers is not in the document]
demophoon: Resolved in https://github.com/hashicorp/waypoint/pull/3546/commits/f0a1c47f86605fd958ff23c6d81ca26f45efdf2b

internal/runnerinstall/nomad.go
```diff
@@ -284,6 +288,12 @@ func (i *NomadRunnerInstaller) InstallFlags(set *flag.Set) {
 		Target: &i.Config.CsiSecrets,
 		Usage:  "Credentials for publishing volume for Waypoint runner.",
 	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "nomad-csi-volume",
+		Target: &i.Config.CsiVolume,
+		Usage:  fmt.Sprintf("The name of the volume to initialize within the CSI provider. The default is %s.", defaultRunnerName("<runner_id>")),
```
briancain, line 295: When I tried to include docs with `<string>` it messed up the website generation because it thought `<string>` was an HTML tag. This is probably true here too, so maybe we could have it without the `<>`?

briancain (changes requested):

internal/runnerinstall/nomad.go
demophoon: Updated in https://github.com/hashicorp/waypoint/pull/3546/commits/819d3acdb202055ddd1c42e51e3b4403b63c6af1

briancain (approved):

Issue #3527, opened 2022-07-05
ECS upgrades fail on runner install


krantzinator:
**Describe the bug**
Per debugging by the Waypoint team, we have hit the below issue on attempting to upgrade from the 0.8.1 install:
```
:x: Installing runner...
:x: Creating IAM role:
! Error installing the runner: InvalidParameter: 1 validation error(s) found.
  - minimum field size of 1, CreateRoleInput.RoleName.
```

**Steps to Reproduce**
See Slack for additional context.

**Expected behavior**
ECS server installs could upgrade using `waypoint upgrade`

catsby:
This should be resolved in #3564

Issue #3513, opened 2022-06-30
Improvements to ECS installer


krantzinator:
**Is your feature request related to a problem? Please describe.**
The ECS Installer keeps track of all the underlying AWS resources it relies on through tags that are currently hardcoded as `waypoint-server` and `waypoint-runner`. This causes collisions if you install waypoint servers in the same region. This likely won't be impactful in an average working scenario, but to assist with debugging and also just to be more resilient, we can improve how we tag or keep track of resources.

**Describe the solution you'd like**
Two options that I see:
- update our tags to be unique per server install, but still something Waypoint can easily know the value of when `uninstall` or `upgrade` is called
- use Terraform somehow to build the infrastructure Waypoint needs to install/run

**Explain any additional use-cases**
The Terraform bit is one that keeps coming up, so that might be a good thing to investigate in the scope of this work

PR #3564, opened 2022-07-12
ECS Server Upgrades: stop task and reduce degregistration time

catsby:
Proactively stop and running server task in ECS and decrease the degregistration delay in ELB Target Groups to enable upgrades and failover to complete.

This PR addresses an issue related to the single-process nature of the waypoint server, which is effectively a race/lock/hold on the internal BoltDB database

Summary of the problem(s): 
- Waypoint Server ECS Task (Task A) has a hold on the BoltDB database
- During upgrade, a new task (Task B) is launched before the old one shuts down
- New task starts but blocks on opening the database
- Original task "deactivates", and while it doesn't show in the running tasks view of ECS tasks, it is still running and still holds the lock
- ELB starts draining connections from the target group that is forwarding traffic to Task A, but the [default time to drain is 300 seconds (deregistration delay)](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#deregistration-delay). Task A does not get sent a full "stop" command until after that 300 seconds.
- ELB starts sending health checks to Task B
- Task B can't respond because it's blocked by Task A holding the lock (basically race condition)
- Task B is never healthy, so ECS launches Task C. 
- ELB deregisters Task B and stops sending it health checks
- Task B is "deactivated" (but still running)
- More race things:
  - Task B and Task C are both trying to open the DB
  - Task A eventually shuts down and frees the lock
  - If Task B opens the DB before it actually shuts down
  - Task C eventually does launch, but it can't open the DB because Task B now has it locked, even though Task B is going to be shut down soon

There exists a possibility that one or more of these events happens quickly enough to enable Task B to actually open the DB and pass health checks, but in my experience it is very rare. 

This PR does two main things:
- Modifies the target groups to have a very short deregistration delay. I assumed this was "okay" (while not ideal) because the server itself is not doing long-running work itself, instead it's passing streams and those can be reconnected 
- In the event of upgrading, proactively stops the server task after the service definition has been updated. This speeds up the process of Task A stopping and freeing the lock for Task B. **This proactively introduces downtime in the server process**, however I believe with Waypoint server being single-process at this time, downtime is unavoidable regardless. Other platforms may simply be faster and the downtime less noticeable. 

Note that there are more remaining [issue(s)](https://github.com/hashicorp/waypoint/issues/3527) and [task(s)](https://github.com/hashicorp/waypoint/issues/3513) related to upgrades, but the issues addressed in this PR are likely unrelated as well as blockers to upgrading Waypoint servers in ECS.

### DIFF
--- a/internal/serverinstall/ecs.go
+++ b/internal/serverinstall/ecs.go
@@ -951,13 +951,13 @@ func (i *ECSInstaller) HasRunner(
 		installutil.DefaultRunnerName("static"),
 	}
 	ecsSvc := ecs.New(sess)
-	foundService, err := awsinstallutil.FindServices(serviceNames, ecsSvc, i.config.Cluster)
+	services, err := awsinstallutil.FindServices(serviceNames, ecsSvc, i.config.Cluster)
 	if err != nil {
 		opts.UI.Output("Could not get list of ECS services: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 		return false, err
 	}
 
-	return foundService != nil, nil
+	return len(services) > 0, nil
 }
 
 func (i *ECSInstaller) InstallFlags(set *flag.Set) {
--- a/internal/serverinstall/ecs.go
+++ b/internal/serverinstall/ecs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	json "encoding/json"
 	"fmt"
+	"github.com/hashicorp/waypoint/internal/clierrors"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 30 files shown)


internal/serverinstall/ecs.go
```diff
@@ -495,6 +494,24 @@ func (i *ECSInstaller) Upgrade(
...
+	// we need to lower. Note: "serverSvc.LoadBalancers" below is a slice of
+	// load balancer + target group pairs. We expect two target groups attached
+	// to the same load balancer.
+	// See
+	// https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeServices.html
+	for _, lb := range serverSvc.LoadBalancers {
+		if lb.TargetGroupArn != nil {
+			s.Update("Updating deregistration delay and termination for Target Groups")
+			if err := modifyTargetGroups(elbv2.New(sess), *lb.TargetGroupArn); err != nil {
+				s.Update("Error updating Target Group %s: %w", *lb, err)
```
briancain, line 508: Do we need to return the `err` here? There's a couple of other places where we don't, is that intentional? And should we preserve the error in a `UI.Output` or maybe a `log.Error`? Otherwise the next stepgroup will wipe over this output.

PR #3571, opened 2022-07-12
Ecs upgrade runner

paladin-devops:
This PR updates the `server upgrade` command to set the runner ID to `static`, which is the ID of the runner installed with the Waypoint server (using `waypoint install`) starting in 0.9. The ECS upgrade process requires this to be set in order to properly search the name of the ECS service to uninstall for the static runner as part of the upgrade.

A small refactor has also been done for querying services in ECS by creating a new helper function in `internal/installutil/aws`.

### DIFF
--- a/internal/installutil/aws/ecs.go
+++ b/internal/installutil/aws/ecs.go
@@ -25,6 +25,7 @@ const (
 	defaultServerTagName     = "waypoint-server"
 	defaultServerTagValue    = "server-component"
 	ServerName               = "waypoint-server"
+	DefaultStaticRunnerName  = "waypoint-static-runner"
 	DefaultSecurityGroupName = "waypoint-server-security-group"
 	RolePolicy               = `{
   "Version": "2012-10-17",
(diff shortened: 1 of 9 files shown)


internal/installutil/aws/ecs.go
```diff
@@ -25,6 +25,7 @@ const (
 	defaultServerTagName     = "waypoint-server"
 	defaultServerTagValue    = "server-component"
 	ServerName               = "waypoint-server"
+	DefaultStaticRunnerName  = "waypoint-static-runner"
```
demophoon, line 28: Nit: Should we use `defaultRunnerName('static')` here? Definitely understand if we want to keep the constants simple.

https://github.com/hashicorp/waypoint/blob/main/internal/runnerinstall/runnerinstall.go#L64

paladin-devops: I think that's a good idea! Hopefully not, but if we change the way runners are named in the future, and thus the default naming convention, it'd be good to already have that in place, instead of the constant.

briancain (approved):

PR #3578, opened 2022-07-15
Fixes operation log flushing

izaaklauer:
All credit to @jgwhite and pairing buddies for finding and diagnosing this.

We're deferring the context cancellation (line 449) _and_ this flush method. If we pass that context into flush, flush will see the context cancel before it tries to do any work, and will likely exit early.

We really want to flush the logs, regardless of what happens to the parent context.


jgwhite (approved):
Excellent!

cicoyle (approved):
Good catch!

briancain (approved):

PR #3581, opened 2022-07-15
Move token encoding logic to server state

xiaolin-ninja:
Moving the token encoding logic over to the server from Auth since it will be handled differently in HCP.

Context:

To prevent users from tampering with tokens, we include a secure hash in every token we create. We create an HMAC key for the server and store it in the database, and use that to create and validate tokens. In HCP, we outsource this cryptographic security to Vault. Therefore, we need to move the hashing logic out of the shared server code. This will kill two birds with one stone, and solve a problem with encrypting/decrypting HMAC keys in HCP.

### DIFF
--- a/pkg/server/singleprocess/auth_test.go
+++ b/pkg/server/singleprocess/auth_test.go
@@ -279,7 +279,7 @@ func TestServiceAuth(t *testing.T) {
 		require.NoError(err)
 
 		var buf bytes.Buffer
-		buf.WriteString(tokenMagic)
+		buf.WriteString("wp24")
 		buf.Write(ttData)
 
 		rogue := base58.Encode(buf.Bytes())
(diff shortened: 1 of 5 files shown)


pkg/server/singleprocess/auth_test.go
```diff
@@ -279,7 +279,7 @@ func TestServiceAuth(t *testing.T) {
 		require.NoError(err)
 
 		var buf bytes.Buffer
-		buf.WriteString(tokenMagic)
+		buf.WriteString("wp24")
```
briancain, line 282: Maybe this should be `boltdbstate.tokenMagic`?

briancain (approved):
Looks like a pretty straight forward copy/paste. I had one suggestion, and I didn't end up testing it, only read through the code!

PR #3584, opened 2022-07-18
{config,server}: Handle parsing, storing, and building embedded pipeline graphs

### LINKED ISSUE #3715: Auto config sync pipelines

Currently while pipelines was being worked on, I decided to make syncing pipeline configs a manual process mainly to avoid any regressions in the regular Waypoint flow while the feature was in early development. So today, prior to running pipelines, you have to run `waypoint config sync` to parse a projects `waypoint.hcl` and store the pipelines as protobufs in the server.

Similar to how other runner operations work, we should config sync pipelines more streamlined when a project is initialized and resync when a pipeline is run. There might be some UX design work here to figure out the least path of resistance to make this a natural flow.

Continue having syncing pipeline configs manual.

Not sure!

**Additional context**

briancain:
This pull request introduces support to parse an embedded pipeline, where that pipeline could either be a Pipeline reference by Owner, or a nested in-line defined pipeline within the step stanza. Currently this PR supports a single embedded pipeline for parsing HCL, however the HCL to Proto func supports infinitely nested pipelines in the case that we might support that within an HCL config. For now, we'll artificially allow for a single nested pipeline until we can get a better user story for Many embedded pipelines within a single Step.

When a runner or user goes to sync pipeline configs, the PipelineProto func will now extract any nested/embedded defined Pipelines and store them as separate proto messages. The step where that embedded pipeline was defined then turns into a Pipeline reference step, which can be looked up at execution time.

This PR updates RunPipeline to build an entire pipeline graph including any nested/embedded pipeline graphs and its dependencies. It handles embedded pipeline reference cycles when queueing a pipeline from RunPipeline.

I've been using this project to test this pull request feature: https://github.com/briancain/waypoint-tetris


briancain:
I decided to continue with part 2 on the same branch for now, I tried to leave TODOs in places where I'm still working. The config parsing and proto generation is ready for review though.

briancain:
FYI: My `serivce_pipeline` cycle test seems to randomly fail locally so I'm going to investigate that.

internal/config/pipeline.go
```diff
@@ -99,8 +104,48 @@ func (c *Config) Pipeline(id string, ctx *hcl.EvalContext) (*Pipeline, error) {
...
+				Config: c,
+			}
+			if s.Pipeline.Config != nil {
+				s.Pipeline.Config.ctx = ctx
+			}
+
+			// Parse all the steps
+			var embSteps []*Step
+			for _, embedStepRaw := range embedPipeline.StepRaw {
+				// TODO(briancain): Do we add an error here if we detect another level
```
paladin-devops, line 129: In case you're looking for an answer for this PR, then I think yes we should, so long as embedded pipelines won't work right now - which I think is the case until this function is recursive (as you mentioned in your earlier comment).

internal/config/pipeline.go
```diff
@@ -118,148 +166,244 @@ func (c *Config) PipelineProtos() ([]*pb.Pipeline, error) {
...
+		}
 
-				if diag := gohcl.DecodeBody(step.Use.Body, finalizeContext(c.ctx), &deployBody); diag.HasErrors() {
-					return nil, diag
+		// We have an embeded pipeline for this step. This can either be an hclPipeline
+		// defined directly in the step, or a pipeline reference to another pipeline
+		// defined else where. If this is a ref, the raw hcl for the pipeline should
+		// be empty.
+		if step.PipelineRaw != nil {
+			// TODO(briancain): It shouldn't be valid to have a "project/pipeline"
```
paladin-devops, line 213: Does this mean that a nested pipeline should not be able to refer to a pipeline in another project?

internal/config/validate.go
```diff
@@ -327,9 +327,15 @@ func (c *Pipeline) Validate() error {
 	var result error
 
 	for _, stepRaw := range c.StepRaw {
-		if stepRaw == nil || stepRaw.Use == nil || stepRaw.Use.Type == "" {
+		if stepRaw == nil && stepRaw.PipelineRaw == nil {
 			result = multierror.Append(result, fmt.Errorf(
 				"step stage with a default 'use' stanza is required"))
```
paladin-devops, line 332:
```suggestion
				"step stage with a default 'use' stanza or a 'pipeline' stanza is required"))
```

internal/config/validate_test.go
```diff
@@ -65,6 +65,10 @@ func TestConfigValidatePipelines(t *testing.T) {
 			"pipeline_multi_step.hcl",
 			"",
 		},
+		{
+			"pipeline_nested_pipes.hcl",
```
paladin-devops, line 69: Should the test files `pipeline_nested.hcl` and `pipeline_nested_refs.hcl` also be included?

internal/config/pipeline.go
briancain: Ah, so this comment is no longer valid! This was still around when I had external pipeline references as

```hcl
step "external-pipeline" {
  pipeline "project/pipeline" { }
  // This shouldn't be valid to have this too
  pipeline "test" {
    // config
  }
}
```

But got some feedback that the pipeline ref was too magical, so now you reference a pipeline with a step:

```hcl
step "external-pipeline-ref" {
  use "pipeline" {
    project = "test"
    name    = "my-pipeline"  
  }
}
```

So yeah, I'll remove this comment!

internal/config/pipeline.go
briancain: Done!

briancain:
Rebased off main to fix merge conflicts

briancain:
Rebased off main which includes pipeline run tracking

internal/config/pipeline.go
```diff
@@ -99,8 +104,68 @@ func (c *Config) Pipeline(id string, ctx *hcl.EvalContext) (*Pipeline, error) {
...
+			}
+			if s.Pipeline.Config != nil {
+				s.Pipeline.Config.ctx = ctx
+			}
+
+			// Parse all the steps
+			var embSteps []*Step
+			for _, embedStepRaw := range embedPipeline.StepRaw {
+				if embedStepRaw.PipelineRaw != nil {
+					// NOTE(briancain): For now, we artificially don't allow for more than
```
paladin-devops, line 130: Thank you for explicitly writing this comment!

paladin-devops: And the status message! ✨

internal/server/boltdbstate/job.go
```diff
@@ -1321,6 +1321,11 @@ func (s *State) jobCreate(dbTxn *bolt.Tx, memTxn *memdb.Txn, jobpb *pb.Job) erro
 
 		// Go through and ensure that each exists.
 		for _, id := range jobpb.DependsOn {
+			if id == "" {
+				return status.Errorf(codes.FailedPrecondition,
+					"job %q has an empty string id for depends_on", jobpb.Id)
```
paladin-devops, line 1326: Good catch! I'm wondering how this could happen... I can't recall when I've seen a job without an ID 🤔

paladin-devops (approved):

izaaklauer (approved):
I haven't tested this, but the approach and the code look pretty good to me! Also looks well-tested 👍

internal/server/boltdbstate/job.go
briancain: This came about from a mistake in my own code! I was adding empty string to the `DependsOn` slice in the job template generator (this was a bug) and figured we could use some extra validation here that this shouldn't be allowed.

internal/config/validate.go
```diff
@@ -327,9 +327,15 @@ func (c *Pipeline) Validate() error {
 	var result error
 
 	for _, stepRaw := range c.StepRaw {
-		if stepRaw == nil || stepRaw.Use == nil || stepRaw.Use.Type == "" {
+		if stepRaw == nil && stepRaw.PipelineRaw == nil {
```
xiaolin-ninja, line 330: The second part of this conditional would never get hit, since if `stepRaw == nil` then it would never evaluate `stepRaw.PipelineRaw`.
```suggestion
		if stepRaw == nil || stepRaw.PipelineRaw == nil {
```

internal/config/validate.go
```diff
@@ -327,9 +327,15 @@ func (c *Pipeline) Validate() error {
...
 			result = multierror.Append(result, fmt.Errorf(
-				"step stage with a default 'use' stanza is required"))
+				"step stage with a default 'use' stanza or a 'pipeline' stanza is required"))
+		} else if stepRaw.Use != nil && stepRaw.PipelineRaw != nil {
+			result = multierror.Append(result, fmt.Errorf(
+				"step stage with a 'use' stanza and pipeline stanza is not valid"))
```
xiaolin-ninja, line 335: This message could be clearer to express that you can't "use" a plugin _and_ have a nested "pipeline" stanza. (At least that's how I understood this code?)

internal/config/validate.go
```diff
@@ -327,9 +327,15 @@ func (c *Pipeline) Validate() error {
...
+		} else if stepRaw.PipelineRaw == nil && (stepRaw.Use == nil || stepRaw.Use.Type == "") {
+			result = multierror.Append(result, fmt.Errorf(
+				"step stage %q is required to define a 'use' stanza and label", stepRaw.Name))
```
xiaolin-ninja, line 338: Same here, I'm understanding the check as "if no nested pipeline or plugin is defined for the step", but the message doesn't express that.

pkg/server/proto/server.proto
```diff
@@ -4705,6 +4705,9 @@ message Pipeline {
       // Built-in up operation
       Up up = 8;
 
+      // A reference to a different pipeline
```
xiaolin-ninja, line 4708:
```suggestion
      // A reference to an embedded pipeline
```
But also, why do we include the entire pipeline object here instead of just a ref? It would make GetPipeline endpoints return unnecessarily heavy content, no?

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+	log hclog.Logger,
+	req *pb.RunPipelineRequest,
+	visitedPipelines map[string]interface{},
+	nodeIdMap map[string]*pb.Ref_PipelineStep,
+	parentDep map[string][]string,
+	pipeline *pb.Pipeline,
+	pipelineRun *pb.PipelineRun,
+) ([]*pb.QueueJobRequest, *pb.PipelineRun, map[string]string, error) {
+	if len(visitedPipelines) != 0 {
+		// Determine if we've already visisted this pipeline and included its jobs.
```
xiaolin-ninja, line 194:
```suggestion
		// Determine if we've already visited this pipeline and included its jobs.
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
-	// single root so the root is always the first step.
-	stepGraph, err := serverptypes.PipelineGraph(pipeline)
-	if err != nil {
-		return nil, err
+	return stepJobs, pipelineRun, stepIds, nil
+}
+
+// pipelineGraphFull takes a pipeline, and optionally accepts an existing Graph
+// and parent step, and attempts to build a full graph for a given Pipeline including
+// any nested pipeline steps. It keeps track of visisted steps so that we don't
```
xiaolin-ninja, line 404:
```suggestion
// any nested pipeline steps. It keeps track of visited steps so that we don't
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
+// get stuck in a loop building the graph. This means step names must be unique
+// across pipelines.
+func (s *Service) pipelineGraphFull(
+	ctx context.Context,
+	log hclog.Logger,
+	g *graph.Graph,
+	parentStep string,
+	visistedNodes map[string]string,
```
xiaolin-ninja, line 412:
```suggestion
	visitedNodes map[string]string,
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
-		jobId := stepIds[v.(string)]
-		jobIds = append(jobIds, jobId)
-		jobMap[jobId] = &pb.Ref_PipelineStep{
-			Pipeline:    pipeline.Id,
-			Step:        v.(string),
-			RunSequence: run.Sequence,
+	// Note that pipeline.Steps is not an ordered list of steps. It's a map of key val
+	// steps so the order will not match the order steps are defined in a waypoint.hcl.
+	for _, step := range pipeline.Steps {
+		if len(visistedNodes) != 0 {
```
xiaolin-ninja, line 427:
```suggestion
		if len(visitedNodes) != 0 {
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
+			if pipeName, ok := visistedNodes[step.Name]; ok && pipeName == pipeline.Name {
+				log.Trace("we've cycled to a node we've already visisted!", "pipeline", pipeName, "step", step.Name)
+				return nil, nil, status.Error(codes.FailedPrecondition,
+					"we've already visisted this node, that means we've got a cycle")
```
xiaolin-ninja, line 431:
```suggestion
					"we've already visited this node, that means we've got a cycle")
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
-		jobIds = append(jobIds, jobId)
-		jobMap[jobId] = &pb.Ref_PipelineStep{
-			Pipeline:    pipeline.Id,
-			Step:        v.(string),
-			RunSequence: run.Sequence,
+	// Note that pipeline.Steps is not an ordered list of steps. It's a map of key val
+	// steps so the order will not match the order steps are defined in a waypoint.hcl.
+	for _, step := range pipeline.Steps {
+		if len(visistedNodes) != 0 {
+			if pipeName, ok := visistedNodes[step.Name]; ok && pipeName == pipeline.Name {
```
xiaolin-ninja, line 428:
```suggestion
			if pipeName, ok := visitedNodes[step.Name]; ok && pipeName == pipeline.Name {
```
lol Brian. After seeing this word so many times even I'm like, "wait, how do you spell visited? Is visisted the actual word?"

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
+				log.Trace("we've cycled to a node we've already visisted!", "pipeline", pipeName, "step", step.Name)
```
xiaolin-ninja, line 429:
```suggestion
				log.Trace("we've cycled to a node we've already visited!", "pipeline", pipeName, "step", step.Name)
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
+			Pipeline: pipeline.Id,
+			Step:     step.Name,
+		}
+
+		// Add our step to the graph as a vertex
+		stepGraph.Add(nodeId)
+
+		// Keep track of the fact that we've visited this step node in this pipeline
+		// to prevent infinite cycles as we build embedded pipeline graphs
+		visistedNodes[step.Name] = pipeline.Name
```
xiaolin-ninja, line 457:
```suggestion
		visitedNodes[step.Name] = pipeline.Name
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
+			// look it up here to get the actual steps.
+			embeddedPipeline, err := s.state(ctx).PipelineGet(embedRef.Pipeline.Ref)
+			if err != nil {
+				return nil, nil, err
+			}
+
+			// Build the nested pipelines graph
+			parentStep := nodeId
+			embeddedGraph, n, err := s.pipelineGraphFull(ctx, log, stepGraph,
+				parentStep, visistedNodes, nodeIdMap, embeddedPipeline)
```
xiaolin-ninja, line 521:
```suggestion
				parentStep, visitedNodes, nodeIdMap, embeddedPipeline)
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
 		return nil, err
 	}
-	run, err := s.state(ctx).PipelineRunGetLatest(pipeline.Id)
+
+	pipelineRun, err := s.state(ctx).PipelineRunGetLatest(pipeline.Id)
+	if err != nil {
+		return nil, err
+	}
+
+	// Build out all of the queued job requests for running this pipelines steps
```
xiaolin-ninja, line 126:
```suggestion
	// Build out all of the queued job requests for running this pipeline's steps
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -213,6 +352,36 @@ func (s *Service) RunPipeline(
...
 				},
 			}
+		case *pb.Pipeline_Step_Pipeline_:
+			embeddedPipeline, err := s.state(ctx).PipelineGet(o.Pipeline.Ref)
+			if err != nil {
+				return nil, nil, nil, err
+			}
+
+			// Pass through *this* steps job dependency IDs so that the child
+			// steps aren't scheduled prior to any dependencies they are waiting for.
```
xiaolin-ninja, line 362:
```suggestion
			// Pass through *this* step's job dependency IDs, so that the child
			// step's jobs aren't scheduled prior to any dependencies.
```
Is this what you mean?

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -213,6 +352,36 @@ func (s *Service) RunPipeline(
...
+			parentStepDep := map[string][]string{pipeline.Id: job.DependsOn}
+
+			embedJobs, embedRun, embedStepIds, err := s.buildStepJobs(ctx, log, req,
```
xiaolin-ninja, line 365: A RECURSIVE FUNCTION?! 🤯

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+	nodeIdMap map[string]*pb.Ref_PipelineStep,
+	parentDep map[string][]string,
+	pipeline *pb.Pipeline,
+	pipelineRun *pb.PipelineRun,
+) ([]*pb.QueueJobRequest, *pb.PipelineRun, map[string]string, error) {
+	if len(visitedPipelines) != 0 {
+		// Determine if we've already visisted this pipeline and included its jobs.
+		// Otherwise we'll get stuck in a cycle. This only really works because
+		// pipeline names are unique for a project. If we ever start allowing for
+		// pipelines across projects we'll need to namespace this value.
```
xiaolin-ninja, line 197:
```suggestion
		// Determine if we've already visited this pipeline and included its jobs.
		// Otherwise, we'll get stuck in a cycle. This only works because
		// pipeline names are unique for a project. If we ever allow
		// pipelines across projects, we'll need to namespace this value.
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+		if _, ok := visitedPipelines[pipeline.Name]; ok {
+			return nil, nil, nil, nil
+		}
+	}
+
+	// Mark that we've visited this pipeline already
+	visitedPipelines[pipeline.Name] = struct{}{}
```
xiaolin-ninja, line 204: I would use a boolean instead of an empty struct for more clarity and space efficiency, but it doesn't actually matter.

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+				"node id %q returned a nil pipeline step ref", nodeId)
+		}
+
+		// get the generated queued job request
+		jobId, ok := stepIds[nodeId]
+		if !ok {
+			// NOTE(briancain):
+			// This could be better. It's basically here because we want to keep track
+			// of an embedded pipeline step ref within the step graph, but it doesn't have
+			// a "job id" because the root of the embedded pipeline is the actual ID where
```
xiaolin-ninja, line 154:
```suggestion
			// a "job id" because the root of the embedded pipeline is the actual ID whereas
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+		if _, ok := visitedPipelines[pipeline.Name]; ok {
+			return nil, nil, nil, nil
+		}
+	}
+
+	// Mark that we've visited this pipeline already
+	visitedPipelines[pipeline.Name] = struct{}{}
+
+	// Generate job IDs for each of the steps. We need to know the IDs in
+	// advance to setup the dependency chain.
```
xiaolin-ninja, line 207:
```suggestion
	// advance to set up the dependency chain.
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+				// Don't include a pipeline step ref because it won't actually queue a job.
+				log.Info("No step id found for nodeDepId", "pipeline", pipeline.Name,
+					"step", step.Name, "dep_id", nodeDepId)
+				continue
+			}
+
+			dependsOn = append(dependsOn, d)
+		}
+
+		// Depend on all JOB IDs from parent step. If we were given a parent pipeline
```
xiaolin-ninja, line 255:
```suggestion
		// Depend on all job IDs from parent step. If we were given a parent pipeline
```

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+		JobMap:    jobMap,
+	}, nil
+}
+
+func (s *Service) buildStepJobs(
+	ctx context.Context,
+	log hclog.Logger,
+	req *pb.RunPipelineRequest,
+	visitedPipelines map[string]interface{},
+	nodeIdMap map[string]*pb.Ref_PipelineStep,
```
xiaolin-ninja, line 188: Idea, what if we switched this to:

`map[*pb.Ref_PipelineStep]string`? Then instead of adding another nested loop in `stepToNodeId` to loop through the whole map for a value match, we can do an O(1) look up and grab the `nodeId`?

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -118,26 +117,163 @@ func (s *Service) RunPipeline(
...
+
+	// Mark that we've visited this pipeline already
+	visitedPipelines[pipeline.Name] = struct{}{}
+
+	// Generate job IDs for each of the steps. We need to know the IDs in
+	// advance to setup the dependency chain.
+	stepIds := map[string]string{}
+	for name, step := range pipeline.Steps {
+		if _, ok := step.Kind.(*pb.Pipeline_Step_Pipeline_); !ok {
+			nodeId, ok := s.stepToNodeId(ctx, log, pipeline.Id, name, nodeIdMap)
```
xiaolin-ninja, line 211: This `s.stepToNodeId` look up is another loop, which makes this triple nested loop inside a recursive function, which feels _really_ heavy to me. What if we restructure NodeId map to use `Ref_PipelineStep` as the key, and the `nodeId` as value, which will save a lot of processing power? (Which could really matter, especially with HCP or a large nesting pipeline.)

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
+	nodeIdMap map[string]*pb.Ref_PipelineStep,
+	pipeline *pb.Pipeline,
+) (*graph.Graph, map[string]*pb.Ref_PipelineStep, error) {
+	var stepGraph *graph.Graph
+	if g != nil {
+		// We're handling an embedded pipeline graph
+		stepGraph = g
+	} else if stepGraph == nil {
+		stepGraph = &graph.Graph{}
 	}
```
xiaolin-ninja, line 422: What is the point of this `else` if `stepGraph` is declared as nil on line 416 and remains nil unless the conditional is hit?

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -221,45 +390,166 @@ func (s *Service) RunPipeline(
...
-	// Queue all the jobs atomically
-	if _, err := s.queueJobMulti(ctx, stepJobs); err != nil {
-		return nil, err
+	return stepGraph, nodeIdMap, nil
+}
+
+func (s *Service) stepToNodeId(
+	ctx context.Context,
+	log hclog.Logger,
+	pipelineName string,
```
xiaolin-ninja, line 543:
```suggestion
	pipelineId string,
```

`stepRef.Pipeline` is the ID of the pipeline, but the variable is `pipelineName`, but this method is always called with the input variable `pipeline.Id`. So we should rename this variable to be clearer.

_OR_ if we want to always reference the pipeline by Name and not ID, then we should refactor `Ref_PiplineStep`.

xiaolin-ninja (changes requested):
Wow, this was mind-bending to read and clearly took a ton of work! Amazing job. 🥇 

I'm a bit concerned about the triple nested loops inside of a recursive function, and have a suggestion to mitigate that a bit. There also seems to be a bit of mix-up with referencing pipeline by ID vs. name, only with variable naming... all the actual code references are by ID but we should keep that crystal clear to avoid confusion in the future.

The logic looks super solid otherwise!

pkg/server/singleprocess/service_pipeline.go
briancain: Yep!

internal/config/validate.go
briancain: I think I'll want to split up the if-statement then because if `stepRaw` is nil, `stepRaw.PipelineRaw` will panic. But makes sense!

internal/config/validate.go
briancain: That's correct!

pkg/server/singleprocess/service_pipeline.go
briancain: I believe I initially attempted this structure, where a protobuf was the `key` for the map, but you can't safely compare protobuf messages as values I _don't think_. I know Golang does allow for regular Go structs to be used as keys for a map key, but as far as a Protobuf message goes it's recommended to use [proto.Equal](https://developers.google.com/protocol-buffers/docs/reference/go/faq#deepequal) for comparison and I'm not sure how that behaves with maps, thus the structure here.

```golang
stepRef := &pb.Ref_PipelineStep {
  Name: "test",
}
mymap := map[*pb.Ref_PipelineStep]string{stepRef: "123abc"}

// Not sure about this one:
fmt.Println(mymap[stepRef])
```

If this is safe and reliable to do though it would make this code simpler for sure and would solve the issue you mentioned later in this PR about the node ID lookup loop iterating over this map.

pkg/server/singleprocess/service_pipeline.go
briancain: The main point here is initializing `stepGraph` from nil into an empty `*graph.Graph{}` struct in the base case or first iteration of this func that I'm handling. I figured it might be worth being explicit assigning `stepGraph` when it's nil rather than an empty struct, but might not matter?

pkg/server/singleprocess/service_pipeline.go
briancain: Makes sense to me to rename this local var `pipelineId`!

pkg/server/singleprocess/service_pipeline.go
briancain: This is actually an established pattern already with our graph package libraries:

https://github.com/hashicorp/waypoint/blob/4b93417dda83c54762d80a5d996e3916b2a8fe03/internal/pkg/graph/dijkstra.go#L40-L45

https://github.com/hashicorp/waypoint/blob/4b93417dda83c54762d80a5d996e3916b2a8fe03/internal/pkg/graph/dfs.go#L18-L20

I like that it's a nonsense type mostly because I could imagine a `bool` or other var type being a bit confusing for a future reader given that we don't care about the value in this map, we're using a map for quick lookups.

pkg/server/singleprocess/service_pipeline.go
briancain: Ok, @evanphx and I figured out a decent solution to this! Solved via [d524338](https://github.com/hashicorp/waypoint/pull/3584/commits/d5243388e6c1979bd0d49ca1ee71bea3b52368a7)

pkg/server/singleprocess/service_pipeline.go
briancain: Solved via [d524338](https://github.com/hashicorp/waypoint/pull/3584/commits/d5243388e6c1979bd0d49ca1ee71bea3b52368a7)!

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -180,12 +180,32 @@ func (s *Service) RunPipeline(
...
+	pipeline, step string
+}
+
 func (s *Service) buildStepJobs(
 	ctx context.Context,
 	log hclog.Logger,
 	req *pb.RunPipelineRequest,
 	visitedPipelines map[string]interface{},
-	nodeIdMap map[string]*pb.Ref_PipelineStep,
+	nodeStepRef *nodeToStepRef,
```
xiaolin-ninja, line 209: YAS BRIAN.

pkg/server/singleprocess/service_pipeline.go
xiaolin-ninja:
```suggestion
	stepGraph := &graph.Graph{}
	if g != nil {
		// We're handling an embedded pipeline graph
		stepGraph = g
```
Yea, the way the code is written now, it's superfluous. I would declare it as the empty graph if that's the desired state if `g != nil`.

xiaolin-ninja (approved):
LGTM!

pkg/server/singleprocess/service_pipeline.go
briancain: Yeah makes sense! I'll update it 👍🏻

PR #3599, opened 2022-07-25
Pipeline Tracking Database and Server Piece

xiaolin-ninja:
RFC WP-132.

This implementation is broken up in three stages and two separate PRs. This PR contains the database and server state changes. The changelog will go out with the second PR, which will contain CLI and API changes.


pkg/server/proto/server.proto
```diff
@@ -4728,6 +4734,28 @@ message Pipeline {
...
 
+message PipelineRun {
+  string id = 1;
+  // The sequence number for this pipeline run.
+  uint64 sequence = 2;
+
+  Ref.Pipeline pipeline = 3;
+
+  Ref.Job root_job = 4;
+  Ref.Job last_job = 5;
```
briancain, line 4745: Thinking about this after we paired, maybe it makes sense to have a map of step names to job refs? Then we can see the status of every step job for a given pipeline run.

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,147 @@ func (s *State) taskComplete(jobId string) error {
...
+			},
+		},
+	}, job.Pipeline.RunSequence)
+	if err != nil || len(run.Jobs) < 1 {
+		s.log.Error("failed to retrieve pipeline to complete", "job", job.Id, "pipeline", job.Pipeline.Pipeline, "run", job.Pipeline.RunSequence)
+		return err
+	}
+
+	// If job Id matches last job queued by pipeline
+	if job.Id == run.Jobs[len(run.Jobs)-1].Id {
```
briancain, line 1753: I could see this being an issue potentially in the future when pipelines can run jobs concurrently rather than sequentially. I don't have a good idea though at the moment on how to know when the "last" job of a pipeline has completed 🤔 We can know the root job from Kahn sort, I'm not sure if that gives is the final job(s) either.

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,147 @@ func (s *State) taskComplete(jobId string) error {
...
+		s.log.Error("failed to retrieve pipeline to ack", "job", job.Id, "pipeline id", job.Pipeline.Pipeline, "run", job.Pipeline.RunSequence)
+		return err
+	}
+
+	for i, j := range run.Jobs {
+		// If job ID matches the job queued by the pipeline run
+		if j.Id == job.Id {
+			// first job is the start job
+			if i == 0 {
+				run.Status = pb.PipelineRun_STARTING
```
briancain, line 1799: I would set the `STARTING` status in `serivce_pipeline` when `RunPipeline` is invoked and its jobs are queued. That would simplify this a lot! Then I think this func might simply be setting the pipeline state to `RUNNING` and we won't need this loop.

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,147 @@ func (s *State) taskComplete(jobId string) error {
...
+		return err
+	}
+
+	// mark pipeline run state as cancelled
+	for _, j := range run.Jobs {
+		if j.Id == job.Id {
+			if job.State == pb.Job_ERROR || job.State == pb.Job_SUCCESS {
+				return nil
+			} else {
+				run.Status = pb.PipelineRun_CANCELLED
```
briancain, line 1851: We shouldn't need to loop over any of these jobs, I'd mark the whole run as cancelled if any job in a pipeline run gets cancelled.

briancain (commented):
This is looking really good! I know its in draft so I didn't give a full review, but wanted to leave a few comments related to the pipeline run state progression!

internal/server/boltdbstate/job.go
briancain: Ahh, the job system will cancel any dependent jobs when a job is cancelled I believe. I don't think we have a "continue on failure" in the job system that I know of!

xiaolin-ninja: https://github.com/hashicorp/waypoint/blob/main/internal/server/boltdbstate/job.go#L927

xiaolin-ninja: & the test suggests the behavior is if the job is already complete, cancel does nothing and doesn't affect dependents
https://github.com/hashicorp/waypoint/blob/main/pkg/serverstate/statetest/test_job.go#L2113

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,125 @@ func (s *State) taskComplete(jobId string) error {
...
+			},
+		},
+	}, job.Pipeline.RunSequence)
+	if err != nil || len(run.Jobs) < 1 {
+		s.log.Error("failed to retrieve pipeline to cancel", "job", job.Id, "pipeline", job.Pipeline.Pipeline, "run", job.Pipeline.RunSequence)
+		return err
+	}
+
+	if job.State == pb.Job_SUCCESS {
+		return nil
```
paladin-devops, line 1832: 🌫 It might be worth logging here that we tried to cancel the job, but it already successfully ran.

internal/server/boltdbstate/job.go
xiaolin-ninja: This is resolved per in-person discussion :)
But brings up another problem... in our current job system, we change job status to `ERROR` whether it errored or was manually cancelled. The timeout cancellation (error) calls the same 'JobCancel' method as the manual cancel flow, which then calls `pipelineCancel` after `jobCancel` is successful. So there's no easy way to set pipeline state = error based on job state or how the cancellation flow is called. We could put `pipelineCancel` before `jobCancel` and filter for error state jobs? But I'm leaning toward just having a "cancelled" state rather than separate "error" and "cancel".

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,125 @@ func (s *State) taskComplete(jobId string) error {
...
+
+	// grab the pipeline run
+	run, err := s.PipelineRunGet(&pb.Ref_Pipeline{
+		Ref: &pb.Ref_Pipeline_Id{
+			Id: &pb.Ref_PipelineId{
+				Id: job.Pipeline.Pipeline,
+			},
+		},
+	}, job.Pipeline.RunSequence)
+	if err != nil || len(run.Jobs) < 1 {
```
paladin-devops, line 1826: ⏳ Is it possible for there to be 0 jobs & `err` to be nil? If so, and if that is the case when this code run, should L1828 return `err` (which would be nil), or should we return a new error if there are 0 jobs?

internal/server/boltdbstate/job.go
xiaolin-ninja: we already do so in the job cancellation method.

internal/server/boltdbstate/job.go
xiaolin-ninja: It is possible for there to be 0 jobs and error is nil, if the pipelinerun was initiated without jobs, since that is a valid structure. Jobs are only added after successful queuing.

But this is a good point, we want to cancel a pipeline in any state so this check isn't necessary here, & I updated the other spots to throw an appropriate error. https://github.com/hashicorp/waypoint/pull/3599/commits/e4b4ec8d02724a5376945a1bae02da25bf0d1ea9

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,135 @@ func (s *State) taskComplete(jobId string) error {
...
+	}
+
+	// grab the pipeline run
+	run, err := s.PipelineRunGet(&pb.Ref_Pipeline{
+		Ref: &pb.Ref_Pipeline_Id{
+			Id: &pb.Ref_PipelineId{
+				Id: job.Pipeline.Pipeline,
+			},
+		},
+	}, job.Pipeline.RunSequence)
```
briancain, line 1746: It looks like we do this pattern (grab a job by id, then get the associated pipeline run) for Complete, Ack, and Cancel. I think it makes sense to create a specific boltdb transaction func that does this for us. Perhaps inside `boltdbstate/pipeline_run.go`? Since these funcs might be called often it makes sense I think to do these queries together. Maybe something like `GetPipelineRunByJobId`?

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,135 @@ func (s *State) taskComplete(jobId string) error {
...
+	}
+	if err != nil {
+		s.log.Error("failed to retrieve pipeline to complete", "job", job.Id, "pipeline", job.Pipeline.Pipeline, "run", job.Pipeline.RunSequence)
+		return err
+	}
+
+	if job.State == pb.Job_ERROR {
+		run.Status = pb.PipelineRun_FAILED
+	} else if job.State == pb.Job_SUCCESS {
+		// If job Id matches last job queued by pipeline
```
briancain, line 1760: I would add a note here that there should be some future considerations made in the event that pipeline steps run in parallel!

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,135 @@ func (s *State) taskComplete(jobId string) error {
...
+		if job.Id == run.Jobs[len(run.Jobs)-1].Id {
+			run.Status = pb.PipelineRun_COMPLETED
```
briancain, line 1762: 🤔 I wonder, should we update a run state enum from `COMPLETED` to instead be `SUCCESS`?

internal/server/boltdbstate/job.go
```diff
@@ -1709,6 +1724,135 @@ func (s *State) taskComplete(jobId string) error {
...
+		)
+	}
+	if err != nil {
+		s.log.Error("failed to retrieve pipeline to ack", "job", job.Id, "pipeline id", job.Pipeline.Pipeline, "run", job.Pipeline.RunSequence)
+		return err
+	}
+	// Update the new pipeline run state
+	run.Status = pb.PipelineRun_RUNNING
+	s.log.Trace("pipeline is running", "job", job.Id, "pipeline", job.Pipeline.Pipeline, "run", run.Sequence)
+	if err := s.PipelineRunPut(run); err != nil {
```
briancain, line 1808: This could be updated to be a no-op (i.e. skip the put all together) if the existing run we got from 1789 is already `RUNNNING` to save us a database write.

internal/server/boltdbstate/pipeline_run.go
```diff
@@ -0,0 +1,366 @@
...
+		if pr.Id == "" {
+			id, err := ulid()
+			if err != nil {
+				return err
+			}
+			pr.Id = id
+		}
+
+		// only alter sequence if this is a new pipeline run
+		if pr.Status == pb.PipelineRun_PENDING {
```
briancain, line 47: I wonder if we could simplify this and instead rely on if no `Id` was passed in, it's a new pipeline run and thus a new sequence id? Maybe this is more specific though so I understand this way too

briancain (changes requested):
Really really awesome! ❤️ Mostly a few questions from me and some minor feedback but otherwise this looks great! I gave it a test too and things are working as expected. I think that's all the feedback I have. There might be more that comes up once the remaining follow-up work happens with the CLI but we can address it then 🎉

internal/server/boltdbstate/job.go
xiaolin-ninja: Ok, good idea. At first I thought you meant we should look up from the database by JobId, and didn't want to do that since that'd be expensive... and we would need to look through the jobs field on each pipeline and check if the job ID is in it, and I don't think you can index by an array that constantly changes. Buuuut if you meant just a function that wraps the two, that works. I feel awkward putting it in boltdbstate though, since it's more of a mediator function... but also can't think of a better place to put it.

internal/server/boltdbstate/job.go
xiaolin-ninja: This is addressed in https://github.com/hashicorp/waypoint/pull/3599/commits/ba543a95b1cb53fbb22822a4850d2ea972523ef2, but Github didn't pick it up as resolved

paladin-devops (approved):
🖲️

pkg/server/proto/server.proto
```diff
@@ -4749,6 +4759,28 @@ message Pipeline {
...
+message PipelineRun {
+  string id = 1;
+  // The sequence number for this pipeline run.
+  uint64 sequence = 2;
+
+  Ref.Pipeline pipeline = 3;
+
+  repeated Ref.Job jobs = 4;
+
+  State status = 5;
```
jgwhite, line 4771: I wonder if we should call this field `state`, mainly to be consistent with the Job message.

xiaolin-ninja: Good call

pkg/server/proto/server.proto
```diff
@@ -4749,6 +4759,28 @@ message Pipeline {
...
 
+message PipelineRun {
+  string id = 1;
+  // The sequence number for this pipeline run.
+  uint64 sequence = 2;
+
+  Ref.Pipeline pipeline = 3;
+
+  repeated Ref.Job jobs = 4;
+
```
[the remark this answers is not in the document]
jgwhite: > Yes, I'd like to put this in the CLI PR coming though, since there'll be other changes with the PipelineBundle, I want to think through it all together.

That sounds like the best approach for sure 👍 

> Top of mind, there is a problem down the line with concurrent pipeline steps, and two concurrent jobs... which one do we pick as the complete time?

Oh wow, yeah that’s a great point, and more reason for the server to determine that kind of info.

briancain (approved):
Looks great! Nice Job! 🚀 🏃🏻

Issue #533, opened 2020-10-16
"waypoint destroy" doesn't work on incomplete deployments


akin-ozer:
**Describe the bug**
"waypoint destroy" command doesn't destroy resources created by "waypoint up" if waypoint up didn't successfully completed.
**Steps to Reproduce**
Create new deployment and interrupt after kubernetes deployment. Then use "waypoint destroy", this command returns okay but deployment is still there.

**Expected behavior**
"waypoint destroy" should clear resources whether deployment is successful or not.

mitchellh:
Thanks yes this is something we know about and need to resolve for 0.2.

krantzinator:
Some logs from the `deployment destroy` command on unsuccessful deployments
```
$ waypoint-dev deployment destroy -vvv
...
2021-02-05T13:17:35.147-0500 [DEBUG] waypoint: will operate on app: name=example-nodejs
2021-02-05T13:17:35.563-0500 [DEBUG] waypoint: no release found to exclude any deployments

» 3 deployments will be destroyed.
2021-02-05T13:17:35.564-0500 [TRACE] waypoint: stopping signal listeners and cancelling the context
```

acaloiaro:
Hi @mitchellh since this was reported in 2020, I'm curious if this issue has seen any progress. 

The issue affects more than  "interrupted" deploys. It also spills over to, e.g. periodic Nomad Jobspecs. 

Example output from `waypoint destroy -vvv`:

```
[...]

Issue #3601, opened 2022-07-26
Deprecation warning is a hard failure


krantzinator:
**Describe the bug**
The deprecation error for multi-app configs should be a warning, not a full error.

**Steps to Reproduce**
Run `waypoint up` with >1 app in a waypoint.hcl

**Expected behavior**
The warning should allow current multi-app behavior to continue.

* Waypoint CLI Version: 0.9.0

**Additional context**
[Discuss conversation](https://discuss.hashicorp.com/t/deprecating-projects-or-how-i-learned-to-love-apps/40888/8)

briancain:
If I remember right, the current issue is that the config `Validate()` function returns that `Warning`/`Error` struct, but every where that calls `Validate()` just treats warnings and errors only as an `error`. We could technically go into each place in Waypoint that validates a config and look to see if there are any Warnings and print those, otherwise raise an error. It would be more ideal if we could treat the Warning printing in the config validate class itself in a single place, so all upstream callers would get it for free.

PR #3602, opened 2022-07-26
Destroy Failed Deployments

paladin-devops:
Update `waypoint deployment destroy` to continue with deleting the specified deployment from the database, while also warning the user that the resources created from the failed deployment may not be fully destroyed.

Addresses #533.

### DIFF
--- a/.changelog/3602.txt
+++ b/.changelog/3602.txt
@@ -1,4 +1,4 @@
 ```release-note:bug
-cli: Update deployment destroy to proceed with deleting resources for failed
-deployments
+cli: `deployment destroy` will now attempt to destroy any known resources for failed
+deployments. Previously, `destroy` would skip unsuccessful deployments.
 ```
--- a/.changelog/3602.txt
+++ b/.changelog/3602.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+cli: Update deployment destroy to proceed with deleting resources for failed
+deployments
+```
--- a/internal/cli/deployment_destroy.go
+++ b/internal/cli/deployment_destroy.go
@@ -63,7 +63,8 @@ func (c *DeploymentDestroyCommand) Run(args []string) int {
 		for _, deployment := range deployments {
 			// Can't destroy a deployment that was not successful
 			if deployment.Status.GetState() != pb.Status_SUCCESS {
-				continue
+				c.ui.Output("The deployment was not successful - destroy may not completely destroy "+
+					"all resources", terminal.WithWarningStyle())
 			}
 
 			// Get our app client
--- a/internal/cli/deployment_destroy.go
+++ b/internal/cli/deployment_destroy.go
@@ -76,9 +76,11 @@ func (c *DeploymentDestroyCommand) Run(args []string) int {
 					Deployment: deployment,
 				},
 			}); err != nil {
-				c.ui.Output("Error destroying the deployment: %s", err.Error(), terminal.WithErrorStyle())
-				return ErrSentinel
+				c.ui.Output("Error destroying deployment %d: %s", deployment.Sequence, err.Error(), terminal.WithErrorStyle())
 			}
+		}
+		if err != nil {
+			return ErrSentinel
 		}
 		return nil
 	})
--- a/internal/cli/deployment_destroy.go
+++ b/internal/cli/deployment_destroy.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


.changelog/3602.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:bug
+cli: Update deployment destroy to proceed with deleting resources for failed
+deployments
```
catsby, line 3:
```suggestion
cli: `deployment destroy` will now attempt to destroy any known resources for failed
deployments. Previously, `destroy` would skip unsuccessful deployments.
```

I think it's important to convey that we will *attempt* to destroy the resources. Since it failed to deploy in the first place, we may not be successful, or even have all the resources tracked

catsby (commented):
I can't comment on line 79/80 b/c it didn't change in this PR, but should we change this:

```go
  }); err != nil {
    c.ui.Output("Error destroying the deployment: %s", err.Error(), terminal.WithErrorStyle())
    return ErrSentinel
  }
```

to simply print the error and continue destroying other deployments, instead of returning?

internal/cli/deployment_destroy.go
```diff
@@ -63,7 +63,8 @@ func (c *DeploymentDestroyCommand) Run(args []string) int {
 		for _, deployment := range deployments {
 			// Can't destroy a deployment that was not successful
 			if deployment.Status.GetState() != pb.Status_SUCCESS {
-				continue
+				c.ui.Output("Deployment %d was not successful - destroy may not completely destroy "+
+					"all resources", deployment.Sequence, terminal.WithWarningStyle())
```
catsby, line 67: Alternatively to my other comment about continuing instead of returning the sentinel error, maybe we should move this `GetState` error handling logic into the error handling logic below. 

In my opinion the only time it should really matter that we output to the user that the deployment in question was not in a successful state should be if we attempt to destroy it and that attempt failed. If it was in a non-successful state but `Destroy()` ran successfully, then I don't think we need to output anything at all.

catsby (commented):
See my comment/suggestion on when we should check `GetState` and output the UI message

catsby (changes requested):

internal/cli/deployment_destroy.go
```diff
@@ -75,10 +76,12 @@ func (c *DeploymentDestroyCommand) Run(args []string) int {
 					Deployment: deployment,
 				},
 			}); err != nil {
-				c.ui.Output("Error destroying the deployment: %s", err.Error(), terminal.WithErrorStyle())
-				return ErrSentinel
+				c.ui.Output("Error destroying deployment %d: %s", deployment.Sequence, err.Error(), terminal.WithErrorStyle())
 			}
 		}
+		if err != nil {
```
catsby, line 82: This `err` is now the `err` from line 44 and not any error returned from `app.Destroy`, I don't believe that is what we want. 

If we plan to continue destroy attempts we probably need to collect the errors in a slice? I'm no longer sure what the intended behavior is

catsby (changes requested):

internal/cli/deployment_destroy.go
paladin-devops: @catsby I think it'd be best to remove the error check/return here. You're right that this is actually the `err` from line 44, and we already catch/return that earlier (as we should). 

I think collecting the errors in a slice is a good idea. After the loop is finished, we can check if there's > 0 errors in the slice, and if there are, return an error saying that one or more destroy-ments failed. The output message will already have warned the user of which ones failed so no need to parse through the slice for anything more specific.

internal/cli/deployment_destroy.go
```diff
@@ -75,10 +78,13 @@ func (c *DeploymentDestroyCommand) Run(args []string) int {
...
 				},
 			}); err != nil {
-				c.ui.Output("Error destroying the deployment: %s", err.Error(), terminal.WithErrorStyle())
-				return ErrSentinel
+				c.ui.Output("Error destroying deployment %d: %s", deployment.Sequence, err.Error(), terminal.WithErrorStyle())
+				destroymentErrors = append(destroymentErrors, err)
 			}
 		}
+		if len(destroymentErrors) > 0 {
+			return errors.New("one or more deployments failed to be destroyed")
```
briancain, line 86: This is the CLI so I think we should be returning an exit code and printing the error to the user, right? 🤔

briancain (approved):

catsby (approved):

PR #3607, opened 2022-07-27
Fix integration by pinning to an older k3s version

evanphx:


briancain (approved):
I have a good feeling about this one fixing CI 👍🏻

PR #3609, opened 2022-07-28
Add additional OAuth2 hooks and handling

evanphx:
The CEB was not perform the OAuth2 exchange, so that's fixed here.

Also, the hook allows HCP Waypoint to inject OAuth2 to a token before it's encoded.

### DIFF
--- a/internal/ceb/server.go
+++ b/internal/ceb/server.go
@@ -3,19 +3,23 @@ package ceb
 import (
 	"context"
 	"crypto/tls"
+	"net/url"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
+	"golang.org/x/oauth2/clientcredentials"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/oauth"
 	"google.golang.org/grpc/status"
 	empty "google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/hashicorp/waypoint/pkg/inlinekeepalive"
 	"github.com/hashicorp/waypoint/pkg/protocolversion"
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"github.com/hashicorp/waypoint/pkg/serverclient"
 )
 
 // client returns the Waypoint client or blocks until it is set or the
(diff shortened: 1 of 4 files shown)


internal/ceb/server.go
```diff
@@ -153,8 +159,40 @@ func (ceb *CEB) dialServer(ctx context.Context, cfg *config, isRetry bool) error
 			return err
 		}
 
+		token := resp.Token
+
+		var perRPC credentials.PerRPCCredentials = staticToken(token)
```
briancain, line 164: Huh! Never seen this before...any reason why this could be simplified to `perRPC := staticToken(token)`?

pkg/server/singleprocess/service_job.go
```diff
@@ -395,6 +395,7 @@ func (s *Service) onDemandRunnerStartJob(
 	// the user of the token to be the user that queued the original job, which is
 	// the correct behavior.
 	token, err := s.newToken(ctx, 60*time.Minute, DefaultKeyId, nil, &pb.Token{
+		// TODO(emp) should this be a Token_Runner_?
```
briancain, line 398: This is a good question 🤔

briancain (approved):

internal/ceb/server.go
evanphx: actually no! The reason is I want `perRPC` to be the interface type, and if I just do `perRPC := staticToken(token)` then it will be `staticToken` type. The reason being I want to reassign it to a different implementor of `PerRPCCredentials` in the code below this.

pkg/server/singleprocess/service.go
```diff
@@ -36,6 +36,11 @@ type Service struct {
 	// with EncodeId), and returns only the waypoint-relevant ID.
 	decodeId func(encodedId string) (id string, err error)
 
+	// processToken allows an implementation the ability to alter a to be
+	// generated token before it's generated. This can be used to alter semantics
+	// about the token, such as adding labels, metadata, or additional info
```
xiaolin-ninja, line 41: Looks like a typo here?
```suggestion
	// processToken allows an implementation the ability to alter a token to be
```
I'm also not sure what it means to alter a token to be a generated token before it's generated? Like structurally?

evanphx: Yeah, as in right before the token is returned for something to use, processToken is handed the token and can alter it.

PR #3618, opened 2022-07-29
Update check for existing docker server install

catsby:
The length of a slice will never be negative. I believe the intent was to check for `0` here


briancain (approved):

briancain:
I think we're safe to ignore `percy` here and this should be good to merge

PR #3621, opened 2022-07-30
Prevent ALBs from timing out idle grpc streams

izaaklauer:
This is designed to work around a very specific problem: AWS ALBs (and perhaps other load balancers) do not respect http2 pings (https://stackoverflow.com/questions/66818645/http2-ping-frames-over-aws-alb-grpc-keepalive-ping). As a result, if a stream sits "idle" for more than the LB timeout (usually 60 seconds), the
load balancer times out that specific http2 sub-connection. This is bad for waypoint, as it's not uncommon to front the waypoint server with an ALB, we have streams that we expect to sit open and idle for long-ish periods of time (i.e. the GetLogStream stream that backs `waypoint logs`), and those streams don't yet have transparent reconnect/resume functionality. It's a bad experience for users for a log stream with no messages to time out after 60 seconds.

This PR introduces GRPC server and client keepalive interceptors. One interceptor will send "keepalive" protobuf messages that use a high-order field, and the other interceptor will detect and intercept those messages without exposing them to the underlying handlers. In this method, we keep some traffic on all of our streams without needing to be aware of it in every RPC.

It's important that neither party (client or server) receives these messages unexpectedly - if they don't have the interceptor to handle them, they will get passed onto the underlying handlers and cause unexpected behavior (likely panics). Clients are expected to add metadata to their outbound contexts indicating they can receive inline keepalives, and servers are expected to advertise via GetVersionInfo features to clients that they can receive them.

Long-term, I expect load balancers will more universally support http2 pings. I also expect that we will improve our streaming methods so that they can be resumed transparently to users if they are unexpectedly disconnected. When either of those conditions are true, we can stop using inline keepalives.

## Diagrams

These diagrams demonstrate the intended usage of inline keepalives within Waypoint.

### ClientStream

```mermaid
sequenceDiagram
autonumber

actor Waypoint Runner
participant Client
participant ClientInterceptor
participant ALB
participant ServerInterceptor
participant RunnerConfig Server Handler
participant Server


Waypoint Runner ->> +Client: RunnerConfig

%% Unnecessary detail
%% Client ->> ServerInterceptor: Open connection for RunnerConfig
%% ServerInterceptor -x ClientInterceptor: Not a ServerStream - DOES NOT send keepalives

Client ->> ClientInterceptor: Create the GRPC client handler

ClientInterceptor ->> +Server: GetVersionInfo

Server -->> -ClientInterceptor: Has feature inlinekeepalives

loop Async send inline keepalives for duration of RPC
    activate ClientInterceptor
    ClientInterceptor -) +ServerInterceptor: SendMsg(InlineKeepalive)

    %% Critical
    ServerInterceptor -x RunnerConfig Server Handler: Recognizes InlineKeepalive. DOES NOT forward.
   
    deactivate ServerInterceptor
    deactivate ClientInterceptor
end

Waypoint Runner ->> Client: Send(event)
Client ->> ClientInterceptor: Send(event) to interceptor
ClientInterceptor ->> ServerInterceptor: Passthrough event to server

%% Critical
ServerInterceptor ->> RunnerConfig Server Handler: Not a keepalive, passthrough

deactivate Client

```

### ServerStream

```mermaid
sequenceDiagram
autonumber

actor User
participant Client
participant ClientInterceptor
participant ALB
participant ServerInterceptor
participant GetLogStream Server Handler
actor CEB


User ->> +Client: GetLogStream

Client ->> +ServerInterceptor: Open connection for RunnerConfig, including inline-keepalive GRPC metadata

loop Async send inline keepalives for duration of RPC
    ServerInterceptor -) +ClientInterceptor: SendMsg(InlineKeepalive)
    
    %% Critical
    ClientInterceptor -x Client: Recognizes InlineKeepalive. DOES NOT forward.
    
    deactivate ClientInterceptor
end

Client ->> ClientInterceptor: Create the GRPC client handler

CEB ->> GetLogStream Server Handler: Send(Log)
GetLogStream Server Handler ->> ServerInterceptor: Send(Log)
ServerInterceptor ->> ClientInterceptor: Send(event) to interceptor

%% Critical
ClientInterceptor ->> Client: Not a keepalive, passthrough

Client ->> User: 

deactivate Client
```

## Manual Tests

Tested `waypoint up`, `waypoint logs`, `waypoint exec`, and `waypoint runner agent` for the following permutations:

|            | old server | new server |
|------------|------------|------------|
| old CLI/CEB | n/a        | works      |
| new CLI/CEB | works      | works      |


pkg/server/proto/server.proto
```diff
@@ -537,6 +542,10 @@ message UI {
 
 message GetVersionInfoResponse {
   VersionInfo info = 1;
+
+  // Represents additional features that the current server supports.
+  // Useful for ensuring compatibility between clients and servers.
+  repeated string features = 4;
```
izaaklauer, line 548: I kept this generic - I can imagine us using this again in the future to detect feature compatibility. 

I could have just used semver checking, but I think we'll want to disable this at some point in the future.

pkg/inlinekeepalive/server_test.go
```diff
@@ -0,0 +1,49 @@
+package inlinekeepalive
```
izaaklauer, line 1: This is a tricky package to test. This covers some of it, and demonstrates the package's functionality, but if people have ideas for more tests let me know!

pkg/inlinekeepalive/client.go
```diff
@@ -0,0 +1,119 @@
...
+		log := hclog.FromContext(ctx).With("method", method)
+
+		ctx = metadata.AppendToOutgoingContext(ctx, HeaderSendKeepalives, "true")
+
+		handler, err := streamer(ctx, desc, cc, method, opts...)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get the client handler when setting up the inlinekeepalive interceptor on method %q", method)
+		}
+
+		serverCompatible, err := isServerCompatible(ctx, cc)
```
evanphx, line 106: Is the intercepter called on every method? If so, this code is checking the version on every RPC rather than just once per clientConn.

evanphx (changes requested):
I think the code needs to clarify under what situations the keepalive goroutine is trigger. Is it every streaming RPC call gets paired with it's own goroutine?

pkg/inlinekeepalive/client.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			log.Warn("Failed to determine if server is capatible with inline keepalives - will not send them.", "err", err)
+		}
+
+		// Only send keepalives if this is a client stream - not allowed otherwise.
+		if desc.ClientStreams && serverCompatible {
+			// Send keepalives for as long as the handler has the connection open
+			go ServeKeepalives(ctx, log, handler)
```
evanphx, line 114: We need to check the behavior of grpc when calling SendMsg from 2 different goroutines.

pkg/inlinekeepalive/client.go
izaaklauer: Yup, it's calling GetVersionInfo once per client-streaming RPC. I think that's the safest thing to do - i'd rather not assume that each conn is pointing at the same waypoint server, and I also don't have a great spot to put the fact that we've already checked the server's features. I added this call to the goroutine, so it won't block main body of streaming logic.

Added a bundle of new comments, added to doc.go, and added a README with diagrams spelling this out.

Let me know if there's anywhere else you think I should add comments.

pkg/inlinekeepalive/client.go
izaaklauer: Good catch - SendMsg is not safe to call concurrently.

I added mutexes to the interceptor to prevent concurrent use.

pkg/inlinekeepalive/server_test.go
izaaklauer: I added a suite of new very thorough tests.

pkg/inlinekeepalive/keepalive_test.go
```diff
@@ -0,0 +1,231 @@
+package inlinekeepalive
```
izaaklauer, line 1: I'm quite happy with how these tests turned out.

pkg/inlinekeepalive/keepalive.go
```diff
@@ -0,0 +1,84 @@
...
+	err := proto.Unmarshal(unknownFields, &keepAlive)
+	if err != nil {
+		// couldn't marshal to the keepalive message, continue as normal
+		return false
+	}
+
+	if keepAlive.Signature != KeepaliveProtoSignature {
+		// We had some other protobuf message with an unknown field whose number matched the high-order
+		// keepalive message field number. This seems very unlikely.
+		log.Warn("Unexpectedly received proto message with an unknown field matching the inlinekeepalive reserved field number")
```
briancain, line 41: Can we include the field as a key=>val in the logger? Would be nice to see what it is for extra debug info if we ever happen to reach this

pkg/inlinekeepalive/keepalive.go
```diff
@@ -0,0 +1,84 @@
...
+) {
+	log.Trace("Starting a inlinekeepalive interceptor for request")
+
+	intervalTicker := time.NewTicker(sendInterval)
+	for {
+		sendMx.Lock()
+		err := stream.SendMsg(&pb.InlineKeepalive{Signature: KeepaliveProtoSignature})
+		sendMx.Unlock()
+		if err != nil {
+			log.Warn("Failed sending inlinekeepalive", "err", err)
```
briancain, line 73: Should this be `log.Error`?

pkg/server/proto/server.proto
```diff
@@ -9,6 +9,11 @@ import "google/protobuf/timestamp.proto";
 import "google/rpc/status.proto";
 import "any.proto";
 
+/********************************************************************
+* PERMANENTLY RESERVED FIELD NUMBERS
+* 100000000 - inline keepalives
```
briancain, line 14: Can we actually reserve this number? I know for the job message we do:

https://github.com/hashicorp/waypoint/blob/5dba36dd4f0eb165905e46e0063709c8fdba63ee/pkg/server/proto/server.proto#L1621

But this is scoped to a single message. I don't think there's a way in protobuf to reserve a field number across every single message so I'm not entirely sure what to do with this here. I think the protobuf should be able to know the message type so it shouldn't treat a field in `message Foo { field = XX; }` the same as `message Bar { field = XX; }` right?

briancain (commented):
🔍 Haven't tested this at all, just read the code!

pkg/inlinekeepalive/keepalive.go
izaaklauer: Error is better, you're right. I was initially thinking warn because one missed keepalive is very non-critical, but error is probably more appropriate. 
```suggestion
			log.Error("Failed sending inlinekeepalive", "err", err)
```

pkg/inlinekeepalive/keepalive.go
izaaklauer: I did a quick pass, and I didn't see a way to get at the field number from here. I can hard-code it, which doesn't feel too bad, as it's hard-coded into server.proto. Do you think that would help?

pkg/server/proto/server.proto
izaaklauer: Agree - I don't think it's possible to reserve it globally.

The problem is that, if someone added a field 100000000 to a streaming message with type `bytes`, we would fail here: https://github.com/hashicorp/waypoint/pull/3621/files#diff-6aec7bf0685492b402d82d82a0388113f3937282a2ebf463f3674c419850e3b5R56

The proto unmarshaller only looks at the field numbers. It will attempt to unmarshal our keepalive message into the real proto, and because they both use field number 100000000, it won't just end up in "unknown fields", it'll really to unmarshal that field. That will fail, because the types don't match, and we'll get a generic unmarshal error out. 

I've looked, and there's no way for us distinguish that error (keepalive unmarshalled into something else) from any kind of unmarshal error. The only way to guard against it is to not use field 100000000 in any streaming message. For simplicity, my verbage reserves it globally.

I'm not too worried - I think it's likely that we'll be able to retire this code before we ever hit field 100000000 😆.

izaaklauer:
The diff is huge because the current mockery version is wrong in all the mocks. If we merge this first: https://github.com/hashicorp/waypoint/pull/3635 the diff will be sane once again.

pkg/inlinekeepalive/keepalive.go
briancain: Oh I see. I was mostly thinking if the signature we're receiving is variable it would be nice to know but if it's always the same then no biggie!

pkg/inlinekeepalive/keepalive.go
```diff
@@ -0,0 +1,84 @@
...
+	KeepaliveProtoSignature   = "inline_keepalive"
+	HeaderSendKeepalivesKey   = "wp-inline-keepalives"
+	HeaderSendKeepalivesValue = "true"
+	FeatureName               = "inline-keepalives"
+)
+
+// IsInlineKeepalive determines if a given proto message is
+// an inline keepalive.
+func IsInlineKeepalive(log hclog.Logger, m protoreflect.ProtoMessage) bool {
+	unknownFields := m.ProtoReflect().GetUnknown()
```
briancain, line 25: Is there a reason we can't use the proto message directly here? If the keep alive proto message is always of a certain type we should be able to peek at it and determine if it's a keep alive.

```
keepAliveMsg, ok := m.(pb.InlineKeepalive)
if !ok { ... }
```

I think this would solve the field number reservation question I had earlier too.

izaaklauer: at that point, `m` is always the message type that the stream expects, like a `RunnerConfigResponse ` or an `EntrypointExecRequest`. The proto unmarshaller will have run by this point, and it will only ever return the expected type.

internal/ceb/server.go
```diff
@@ -79,7 +80,17 @@ func (ceb *CEB) dialServer(ctx context.Context, cfg *config, isRetry bool) error
...
+		grpc.WithChainStreamInterceptor(
+			protocolversion.StreamClientInterceptor(protocolversion.Current()),
+
+			// Send and receive keepalive messages along grpc streams.
+			// Some loadbalancers (ALBs) don't respect http2 pings.
+			// (https://stackoverflow.com/questions/66818645/http2-ping-frames-over-aws-alb-grpc-keepalive-ping)
+			// This interceptor keeps low-traffic streams active and not timed out.
+			// NOTE(izaak): long-term, we should ensure that all of our
+			// streaming endpoints are robust to disconnect/resume.
+			inlinekeepalive.KeepaliveClientStreamInterceptor(time.Duration(5)*time.Second),
```
johanbrandhorst, line 92: Nit: I don't know if this is intentional, but you don't need to type convert constants
```suggestion
			inlinekeepalive.KeepaliveClientStreamInterceptor(5*time.Second),
```

pkg/inlinekeepalive/keepalive.go
johanbrandhorst: I'm also confused about this - this logic seems like it's trying to handle the case where the server doesn't know about the message type, but surely it does?

pkg/inlinekeepalive/keepalive.go
```diff
@@ -0,0 +1,84 @@
...
+	return true
+}
+
+// GrpcStream can be either a grpc.ClientStream or a grpc.ServerStream
+type GrpcStream interface {
+	SendMsg(m interface{}) error
+}
+
+// ServeKeepalives sends keepalive messages along the provided grpc stream
+// at a rate of one every five seconds.
```
johanbrandhorst, line 54:
```suggestion
// at the rate specified by sendInterval.
```

pkg/server/proto/server.proto
johanbrandhorst: Should this not be an enum?

pkg/server/proto/server.proto
johanbrandhorst: Why would RecvMsg fail if there's another message with the same field number? Or is that link linking to the wrong line?

johanbrandhorst (commented):
This is neat! Kinda worried about intercepting the server stream and client stream in general, so make sure to thoroughly test this before rolling it out in a release.

internal/ceb/server.go
izaaklauer: Ah yup, I started by having this value fed from config, but felt like it wasn't necessary at this point. I think we might get there eventually though, and this will save our newer engineers from a bit of confusion. As far as I know, it's equivalent post-compilation for constants.

pkg/server/proto/server.proto
izaaklauer: Good idea! Done.

izaaklauer:
The question came up a few times: "Why can't we just do a type assertion to check if a message is an inline keepalive vs. any other message?"

I added a bunch of comments around those relevant parts. I also filmed a short video walking through why with the goland debugger:

https://drive.google.com/file/d/1rXnaIkJFAPxGxacIXujJIlZULtCiwgn9/view?usp=sharing

Apologies for the terrible audio - didn't realize the wind was blowing into my mic.

pkg/inlinekeepalive/keepalive.go
izaaklauer: It is weird. I added a bunch of code comments, and a new github comment with a walkthrough video - if you still have questions, respond there!

pkg/server/proto/server.proto
```diff
@@ -564,6 +569,20 @@ message UI {
...
+}
+
+// Represents features that the server supports.
+// Useful for ensuring compatibility between clients and servers.
+message ServerFeatures {
+  repeated feature features = 1;
+
+  enum feature {
+    // Advertises that this server is capable of receiving inline keepalive messages
+    INLINE_KEEPALIVES = 0;
```
johanbrandhorst, line 584: For protobuf enums, [it is conventional](https://docs.buf.build/lint/configuration#enum_zero_value_suffix) to make the zero value "unspecified". It's [also convention](https://docs.buf.build/lint/rules#enum_value_prefix) to prefix the name of each value with the name of the enum. I don't know what the convention is within waypoint specifically, so if there is already an established convention, ignore me.
```suggestion
    FEATURE_UNSPECIFIED = 0
    // Advertises that this server is capable of receiving inline keepalive messages
    FEATURE_INLINE_KEEPALIVES = 1;
```

johanbrandhorst:
Thanks for the video, that was great. I think I understand why this is happening now - it's because we're hijacking the streams SendMsg and sending a `InlineKeepalive` message on it, even though the gRPC server expects a message of the type corresponding to the RCP that is being used. I understand that we're doing this so that we can transparently send this keepalive to any stream on any service, but we can also see the cost of this flexibility - having to "reserve" a high field number and operating on unknown fields. I was under the impression that these keepalives would be well typed parts of the request/response messages (part of a oneof of some sort) which would avoid this sort of issue entirely. Was there a reason we didn't go down that path?

pkg/server/proto/server.proto
izaaklauer: Done! We don't have conventions around this as a team (that I'm aware of at least), and these conventions make sense to me. Yet another reason for us to eventually add `buf` to our process. Thank you!

evanphx (approved):

johanbrandhorst (approved):
Sounds reasonable to me, lets hope we can remove this sooner rather than later :)

PR #3626, opened 2022-08-01
`project destroy`

### LINKED ISSUE #1418: feature-request: Add `waypoint project destroy`

### LINKED ISSUE #3824: urls generated using auto_hostname are not being deleted when workspace is destroyed
From #3551 @rastakajakwanna:
>urls generated using auto_hostname are not being deleted when workspace is destroyed or when new deployment succeeds and there is a new url generated for it, there has to be another extra pipeline step in order to delete them
```
WHOSTS=`waypoint hostname list -plain -workspace=$CI_COMMIT_REF_SLUG | tail -n +3 | awk '{print $1}'`
for hostname in ${WHOSTS}; do waypoint hostname delete ${hostname}; done
```

applicable:

* Waypoint CLI Version: 0.8.2
* Waypoint Server Platform and Version: 0.8.2 (like `docker`, `nomad`, `kubernetes`)

paladin-devops:
Adds a new command to Waypoint for destroying a project. By default, this command will destroy all resources created for apps within a project, by initiating a destroy operation for each app in a project. After all resources are destroyed, the project is deleted from the database. The user may opt-out of destroying resources with the `-skip-destroy-resources` flag. Confirmation with either the `-auto-approve` flag, or interactive input is required in order to run the command.

Closes #1418.
Closes #3824.


internal/cli/project_destroy.go
```diff
@@ -0,0 +1,110 @@
...
+			Project: c.project.Ref(),
+		})
+		if err != nil {
+			return err
+		}
+
+		// Confirmation is required for destroying a project &/or its resources
+		if !c.confirm {
+			proceed, err := c.ui.Input(&terminal.Input{
+				Prompt: "Do you really want to destroy this project and its resources? Only 'yes' will be accepted to approve: ",
```
briancain, line 41: It'd be nice if we formatted this to say the specific project like

> "Do you really want to destroy project "my-project" and its resources? ....

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,110 @@
...
+				Secret: false,
+			})
+			if err != nil {
+				c.ui.Output(
+					"Error getting input: %s",
+					clierrors.Humanize(err),
+					terminal.WithErrorStyle(),
+				)
+			} else if strings.ToLower(proceed) != "yes" {
+				app.UI.Output("Destroying project and resources requires confirmation.", terminal.WithWarningStyle())
```
briancain, line 52: Maybe something like

> Destroying project %q and resources....

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,110 @@
...
+		f.BoolVar(&flag.BoolVar{
+			Name:    "skip-destroy-resources",
+			Usage:   "Skips destroying resources created for the Waypoint project.",
+			Default: false,
+			Target:  &c.skipDestroyResources,
+		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:    "auto-approve",
+			Usage:   "Auto-approve destroying the project & all resources. If unset, confirmation will be requested.",
```
briancain, line 87: > "Destroy the project and all of its resources without confirmation"

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,110 @@
...
+			Default: false,
+			Target:  &c.confirm,
+		})
+	})
+}
+
+func (c *ProjectDestroyCommand) Synopsis() string {
+	return "Delete the specified project."
```
briancain, line 95:
```suggestion
	return "Delete the specified project and optionally destroy its resources."
```

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,110 @@
...
+func (c *ProjectDestroyCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint project destroy [options]
+
+  Delete the project and all resources created for all apps within the project.
+
+  You can optionally skip destroying the resources by setting
+  -skip-destroy-resources to true.
+
+  The workspace can continue to be used after this call, this just deletes
```
briancain, line 107: I'm curious what it means to continue to use the workspace after? Is there another way to say this, I'm not sure what workspace we're referring to.

internal/cli/project_destroy.go
briancain, line 41: It'd be nice if we said the project here just for extra confirmation:

> "Do you really want to destroy project "my-project" and its resources? ...

internal/runner/operation_project_destroy.go
```diff
@@ -0,0 +1,71 @@
...
+	client := project.Client()
+	destroyProjectOp, ok := job.Operation.(*pb.Job_DestroyProject)
+	if !ok {
+		// this shouldn't happen since the call to this function is gated
+		// on the above type match.
+		panic("operation not expected type")
+	}
+
+	// Update the project state to indicate that it's being destroyed
+	// TODO: Do we need to do something with the response here?
```
briancain, line 23: Probably not, UpsertProject will return the same Project proto message that was within the request i.e. the value you set within `&pb.UpsertProjectRequest{Project: ... }`

https://github.com/hashicorp/waypoint/blob/d0b40daa74d483a3a8838e43f93bfada5736270d/pkg/server/singleprocess/service_project.go#L20-L42

internal/runner/operation_project_destroy.go
```diff
@@ -0,0 +1,71 @@
...
+			},
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	if !destroyProjectOp.DestroyProject.SkipDestroyResources {
+		for _, app := range project.Apps() {
+			// TODO: Do we need to do something with the response here?
+			_, err := r.executeDestroyOp(ctx,
```
briancain, line 38: I don't think so, looks like this returns an empty `pb.Job_Result{}`

https://github.com/hashicorp/waypoint/blob/25b2130a43e4594e67c897884f6934aa0f19d9b9/internal/runner/operation_destroy.go#L42

pkg/server/proto/server.proto
```diff
@@ -2280,6 +2290,11 @@ message Job {
     Pipeline.Step step = 1;
   }
 
+  message DestroyProjectOp {
+    Project project = 1;
+    bool skip_destroy_resources = 2;
```
briancain, line 2295: I recommend adding some doc comments here, our job operations are pretty well documented.

pkg/server/singleprocess/service_project.go
```diff
@@ -79,6 +79,23 @@ func (s *Service) ListProjects(
...
+		return nil, err
+	}
+
+	err := s.state(ctx).ProjectDelete(req.Project)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Do we need to return something other than empty?
+	return &empty.Empty{}, nil
```
briancain, line 96: In this case, maybe not? Where do we queue the DestroyOperation? In that case we should return the job ID associated with destroying the resources. But since it looks like this simply deletes Project from the DB, it's safe to return empty.

briancain (commented):
Looking pretty solid @paladin-devops ! I know it's a draft but I had some answers for TODOs and some feedback on user-facing strings that I figured I'd review. Excited for this one!! 👏🏻

internal/cli/project_destroy.go
paladin-devops: I'm removing this for now, we want the operation to destroy all the resources for apps in a project in all workspaces where they're deployed.

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (c *ProjectDestroyCommand) Synopsis() string {
+	return "Delete the specified project and optionally destroy its resources."
+}
+
+func (c *ProjectDestroyCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint project destroy [options]
+
+  Delete the project and all resources created for all apps within the project.
```
xiaolin-ninja, line 95: We could be more descriptive about what these resources are. e.g. "local and network resources" to be extra clear for users.

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			c.ui.Output(
+				"Error getting input: %s",
+				clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+		} else if strings.ToLower(proceed) != "yes" {
+			c.ui.Output("Destroying project %q and resources requires confirmation.", project.Project.Name, terminal.WithWarningStyle())
+			return 1
+		}
```
catsby, line 51: If I read this correctly, if error is not `nil` then we output an error message but continue on and attempt to destroy things regardless if the user input `yes` or anything

xiaolin-ninja (commented):
Great work! I wanna throw in an existential discussion though; do we want to hard-delete projects? How do we handle orphaned resources?

internal/cli/project_destroy.go
xiaolin-ninja, line 95: I would expand on what type of resources these are to be extra clear for the user. Just Waypoint resources? Plugin resources as well?

pkg/server/proto/server.proto
```diff
@@ -93,6 +93,10 @@ service Waypoint {
   // can use GetProject to get more information about the project.
   rpc ListProjects(google.protobuf.Empty) returns (ListProjectsResponse) {}
 
+  // DestroyProject deletes a project from the database as well as (optionally)
+  // the destruction of all resources created within a project
```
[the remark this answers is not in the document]
paladin-devops: @xiaolin-ninja to add some context here, while testing I noticed that if you destroy a project, then if you re-init the project, then the resources, deployments, builds, etc. are readily visible and "attached" to the project again, as if `project destroy` had never been run.

internal/runner/operation_project_destroy.go
```diff
@@ -0,0 +1,97 @@
...
+	client := project.Client()
+	destroyProjectOp, ok := job.Operation.(*pb.Job_DestroyProject)
+	if !ok {
+		// this shouldn't happen since the call to this function is gated
+		// on the above type match.
+		panic("operation not expected type")
+	}
+
+	// Update the project state to indicate that it's being destroyed
+	_, err := client.UpsertProject(ctx,
```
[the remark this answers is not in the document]
paladin-devops: @xiaolin-ninja your mention of polling made me more curious about this. Do you mean that while a project destroy operation is running, that a poll operation could simultaneously be run & re-upsert the project in an ACTIVE state?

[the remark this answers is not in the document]
xiaolin-ninja: Yup that's what I'm thinking.

pkg/server/proto/server.proto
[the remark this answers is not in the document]
xiaolin-ninja: Sure, but what happens if an endpoint gets or puts the resources while the project is gone, and validation fails because the project doesn't exist?

internal/runner/operation_project_destroy.go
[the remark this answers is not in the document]
briancain: Polling shouldn't run on a non-existent project if you delete a project from the db. Project messages have a `DataSourcePoll.Enabled` field that will switch the poller on and off that it checks for when it goes to look at all current projects the poller should queue a message on.

internal/cli/project_destroy.go
catsby: Was this resolved by mistake or am I missing something? The problem appears to still be present. Am I mistaken?

pkg/serverhandler/handlertest/test_service_project.go
```diff
@@ -90,6 +90,24 @@ func TestServiceProject(t *testing.T, factory Factory) {
...
+
+		// Destroys the specified project
+		{
+			_, err := client.DestroyProject(ctx, &pb.DestroyProjectRequest{
+				Project: &pb.Ref_Project{Project: "example"},
+			})
+			require.NoError(err)
+
+			resp, err := client.GetProject(ctx, &pb.GetProjectRequest{
+				Project: &pb.Ref_Project{Project: "not-found"},
```
xiaolin-ninja: We should be getting the project with ID "example" to test that project "example" was successfully destroyed.

We should also show two different tests, one that shows resources are not deleted with the destroy resources flag, and one that shows they are successfully deleted without the flag.

pkg/server/proto/server.proto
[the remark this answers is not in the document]
paladin-devops: [#3673](https://github.com/hashicorp/waypoint/pull/3673) updates this branch with cascading deletion of projects.

pkg/serverhandler/handlertest/test_service_project.go
paladin-devops: @xiaolin-ninja added! I had to implement it in the `internal/runner` package as the option to skip destroying resources is available when creating the project destroy job only, not here in `service_project`.

.changelog/3626.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+core: Add ability for projects to be deleted from the database with their resources
```
catsby, line 2:
```suggestion
core: Introduce a new CLI command `project destroy` to delete projects in Waypoint and destroy their associated resources
```

go.mod
```diff
@@ -66,7 +66,7 @@ require (
 	github.com/hashicorp/waypoint-hzn v0.0.0-20201008221232-97cd4d9120b9
 	github.com/imdario/mergo v0.3.12
 	github.com/improbable-eng/grpc-web v0.13.0
-	github.com/kevinburke/go-bindata v3.22.0+incompatible
+	github.com/kevinburke/go-bindata v3.23.0+incompatible
```
catsby, line 69: Is there a reason for this update?

internal/cli/job_inspect.go
```diff
@@ -112,6 +112,8 @@ func (c *JobInspectCommand) Run(args []string) int {
 		op = "Init"
 	case *pb.Job_PipelineStep:
 		op = "PipelineStep"
+	case *pb.Job_DestroyProject:
+		op = "DestroyProject"
```
catsby, line 116: I believe similar changes need to be made in [`cli/task_list.go`](https://github.com/hashicorp/waypoint/blob/420f95a7f08178ed88b5ce549fb04b45b443e1fd/internal/cli/task_list.go#L102) and [`cli/task_inspect.go`](https://github.com/hashicorp/waypoint/blob/420f95a7f08178ed88b5ce549fb04b45b443e1fd/internal/cli/task_inspect.go#L169), unfortunately :/

internal/cli/project_destroy.go
```diff
@@ -0,0 +1,102 @@
...
+func (c *ProjectDestroyCommand) Synopsis() string {
+	return "Delete the specified project and optionally destroy its resources."
+}
+
+func (c *ProjectDestroyCommand) Help() string {
+	return formatHelp(`
+Usage: waypoint project destroy [options]
+
+  Delete the project and all resources created for all apps in the project, within
+  each app's plugin platform.
```
catsby, line 97:
```suggestion
  Delete the project and all resources created, for all applications in the project
```

I'm not totally sure what "each app's plugin platform" means

internal/runner/operation_project_destroy.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+
+	// Delete the project from the database
+	_, err = client.DestroyProject(ctx, &pb.DestroyProjectRequest{
+		Project: &pb.Ref_Project{
+			Project: destroyProjectOp.DestroyProject.Project.Name,
+		},
+	})
+	// If there is an error destroying a project, reset it's state to ACTIVE
```
catsby, line 88:
```suggestion
	// If there is an error destroying a project, reset its state to ACTIVE
```

catsby (changes requested):
👍 I think the case checking needs to be added in task list/inspect, the rest is minor feedback

go.mod
paladin-devops: Whilst setting up the env for OSS Waypoint, I advised @cicoyle to run `go get github.com/kevinburke/go-bindata`, which resulted in the latest version being pulled. It looks like there were a few inconsequential updates between 3.22 and 3.23, so this could be rolled back if updating leads to issues.

internal/cli/project_destroy.go
paladin-devops: Updated to make it a bit more clear! I removed the mention of plugin, and just noted that the resources created in each platform are deleted.

briancain (changes requested):
Since there are dependent PRs that haven't been merged up into this branch I'll hold off on my full testing, but I caught these while testing this branch.

internal/client/operation.go
```diff
@@ -28,6 +28,21 @@ func (c *Project) Validate(ctx context.Context, op *pb.Job_ValidateOp) (*pb.Job_
...
+	}
+	// Build our job
+	job := c.job()
+	job.Operation = &pb.Job_DestroyProject{
+		DestroyProject: op,
+	}
+
+	// Execute it
+	result, err := c.doJob(ctx, job, c.UI)
+	return result.ProjectDestroy, err
```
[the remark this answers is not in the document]
briancain: I think this means there was also an issue with the `ProjectDestroy` job. It likely was returning an error from the operation but the job result is nil so it panic'd

go.mod
```diff
@@ -334,7 +334,7 @@ require (
 	github.com/evanphx/grpc-gateway v1.16.1-0.20220211183845-48e5be386c15
 	github.com/hashicorp/go-grpc-net-conn v0.0.0-20220321172933-7ab38178cb90
 	github.com/hashicorp/opaqueany v0.0.0-20220321170339-a5c6ff5bb0ec
-	github.com/hashicorp/waypoint-plugin-sdk v0.0.0-20220502215818-69a3eeb201d8
+	github.com/hashicorp/waypoint-plugin-sdk v0.0.0-20220815171738-0c7040f14e61
```
briancain, line 337: Don't forget to update this once the sdk PR is merged!

pkg/server/proto/server.proto
```diff
@@ -3500,6 +3547,8 @@ message Deployment {
   // Resources that this deployment has created or manages.
   repeated DeclaredResource declared_resources = 19;
 
+  repeated DestroyedResource destroyed_resources = 21;
```
briancain, line 3550: This could use a comment too

briancain (changes requested):
Mostly minor comments from me, but looks great! Lots of work went into this one, nice job! 👏🏻

internal/core/app_deploy_destroy.go
```diff
@@ -114,11 +118,30 @@ func (a *App) destroyDeploy(
...
-	_, _, err = a.doOperation(ctx, a.logger.Named("deploy"), &deployDestroyOperation{
+	_, destroyment, err := a.doOperation(ctx, a.logger.Named("deploy"), &deployDestroyOperation{
 		Component:  c,
 		Deployment: d,
 	})
-	return err
+
+	destroyProto, ok := destroyment.(*pb.Deployment)
+	if !ok {
+		return err
```
briancain, line 128: I think we have to write our own err here

internal/core/app_deploy_destroy.go
```diff
@@ -114,11 +118,30 @@ func (a *App) destroyDeploy(
...
+		for _, resource := range destroyProto.DeclaredResources {
+			a.UI.Output("- "+resource.Name, terminal.WithWarningStyle())
+		}
+		if len(destroyProto.DestroyedResources) > 0 {
+			a.UI.Output("These resources were destroyed for app %q:", a.ref.Application, terminal.WithSuccessStyle())
+			for _, resource := range destroyProto.DestroyedResources {
+				a.UI.Output("- "+resource.Name, terminal.WithSuccessStyle())
+			}
+		}
+	}
```
briancain, line 142: It might be nice to build up these strings in the loops and do a single `a.UI.Output` once rather than for each entry

internal/server/boltdbstate/project.go
```diff
@@ -71,15 +71,188 @@ func (s *State) ProjectGet(ref *pb.Ref_Project) (*pb.Project, error) {
...
+		}
+
+		// delete triggers for project
+		for _, trigger := range triggers {
+			if err = s.triggerDelete(dbTxn, memTxn, &pb.Ref_Trigger{Id: trigger.Id}); err != nil {
+				return err
+			}
+		}
+
+		// delete pipelines for project
```
xiaolin-ninja, line 237: We should delete pipeline runs too!

internal/server/boltdbstate/project_test.go
```diff
@@ -0,0 +1,650 @@
...
+		require.NoError(s.ReleasePut(false, &pb.Release{
+			Id: "testRelease",
+			Application: &pb.Ref_Application{
+				Application: appName,
+				Project:     projectName,
+			},
+			Workspace: &pb.Ref_Workspace{Workspace: "default"},
+		}))
+
+		// Verify that the app operation sequence is 1, since it is the 1st operation for the new (albeit
```
briancain, line 632: Maybe add 2 more deployment sequences so we can test more than 1 sequence?

pkg/server/proto/server.proto
```diff
@@ -2311,6 +2346,14 @@ message Job {
     Pipeline.Step step = 1;
   }
 
+  // DestroyProjectOp triggers the deletion of a project from the database
+  // as well as (optionally) the destruction of all resources created within
+  // a project
+  message DestroyProjectOp {
+    Project project = 1;
```
briancain, line 2353: This should probably be a `Ref.Project`

internal/runner/operation_project_destroy.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+			})
+			if err != nil {
+				return nil, err
+			}
+			// Destroy the resources in the workspaces for the app
+			for _, workspace := range workspaces.Workspaces {
+				log.Debug("Destroying resources in workspace %s", workspace.Name)
+				// TODO: The destroy operation doesn't currently respect the Workspace
+				// being set on the Job - this should be updated
```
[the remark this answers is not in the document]
paladin-devops: It would be nice to fix this! The behavior seems to be rooted in [runner/operation.go](https://github.com/hashicorp/waypoint/blob/main/internal/runner/operation.go#L292-L303). The workspace being used in the destroyment is derived from the `core.Project` that is set here:

```go
	project, err := core.NewProject(ctx,
		core.WithLogger(log),
		core.WithUI(ui),
		core.WithComponents(factories),
		core.WithClient(r.client),
		core.WithConfig(cfg),
		core.WithDataDir(projDir),
		core.WithLabels(job.Labels),
		core.WithVariables(inputVars),
		core.WithWorkspace(job.Workspace.Workspace),  <--- here
		core.WithJobInfo(jobInfo),
	)
// some other code
	case *pb.Job_DestroyProject:
		return r.executeDestroyProjectOp(ctx, log, job, project)
```

So in a way, my comment isn't accurate - the destroyment ONLY respects the workspace that is set on the job and no others... but in our case that is the project destroy job, which happens to need to loop through all workspaces and delete everything. What this means is that if you ran `project destroy -w=dev`, then all of the deployments in a `dev` workspace _would_ be deleted, but no others. But that top-level workspace flag controls which workspace's deployments are going to be destroyed... need to think more on this one but I think this defines the issue.

internal/server/boltdbstate/project_test.go
```diff
@@ -0,0 +1,650 @@
+package boltdbstate
+
+import (
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestProject(t *testing.T) {
```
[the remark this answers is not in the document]
paladin-devops: Alt Text

internal/core/project.go
```diff
@@ -245,6 +245,61 @@ func (p *Project) Ref() *pb.Ref_Project {
 	return &pb.Ref_Project{Project: p.name}
 }
 
+// Copy creates a copy of the project, for a different workspace.
+// The project's config is required to be passed in because the Config
+// option is not set on a project, so we can't reference it directly.
+// Getters for other project fields are used here to limit their exposure.
+func (p *Project) Copy(ctx context.Context, workspace string, projConfig *config.Config) (*Project, error) {
```
evanphx, line 252: I'd probably call this `InWorkspace` rather than `Copy` to convey better why it would be used.

.changelog/3626.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+core: Introduce a new CLI command `project destroy` to delete projects in Waypoint and destroy their associated resources
```
krantzinator, line 2: This is important enough that you can highlight the changelog item like how Brian did the Pipelines changelog here https://github.com/hashicorp/waypoint/pull/3777/files

go.mod
krantzinator: Interestingly, we don't appear to specify a version when pulling this for our release build https://github.com/hashicorp/waypoint-releases/blob/70d8ac07d4838b8b90d81bf5583ef5ad988e2648/.circleci/config.yml#L330

krantzinator (approved):
If no one else can look at this in the next 20 minutes, go ahead and merge! This has gone through multiple feedback rounds from almost everyone on the team, and has been an extremely thorough process. Because it's definitely going in our release, if it causes regressions those will be found in our regression testing over the next few days and should be handled separately at this point.
💯 🔥 
Edit: per convo in Slack, ensure this error is fixed:
```
Unlinked pages found in the ../content/commands directory.
```

briancain (approved):

PR #3627, opened 2022-08-02
fix: struct value syntax for WatchTaskStartupTimeoutSeconds

azazeal:
This PR fixes the tag value syntax of `WatchTaskStartupTimeoutSeconds`.


briancain (approved):
Nice catch, thank you!

PR #3634, opened 2022-08-02
Throttle unallowed runners by delaying before denying them access

evanphx:


izaaklauer (approved):
Good idea with the quick patch. There's a runner bug in there somewhere to fix, but this should keep us from getting overwhelmed while we look for it.

pkg/server/singleprocess/auth.go
```diff
@@ -237,6 +237,9 @@ func (s *Service) authRunner(
 		(r.AdoptionState != pb.Runner_ADOPTED &&
 			r.AdoptionState != pb.Runner_PREADOPTED)
 	if notAdopted {
+		// We sleep here to tarpit any runaway runners that are going to thrashing
+		// trying to connect over and over and over even though they're not allowed in.
+		time.Sleep(5 * time.Second)
```
briancain, line 242: I know this is a temp fix but would we want some kind of rolling backoff for the time we sleep/retry over time?

briancain (approved):
🏃🏻 💤

pkg/server/singleprocess/auth.go
evanphx: there isn't really the ability to do that, because we don't have a stable criteria to use to change the sleep time.

internal/runner/accept.go
```diff
@@ -62,6 +63,12 @@ func (r *Runner) AcceptMany(ctx context.Context) {
 				// the context being closed first, in which case we honor that as a valid
 				// reason to stop accepting jobs.
 				return
+			case codes.PermissionDenied:
+				// This means the runner was deregistered and we must exit.
+				// This won't be fixed unless the runner is closed and restarted.
+				r.logger.Error("runner unexpectedly deregistered, exiting")
+				time.Sleep(5 * time.Second)
```
izaaklauer, line 70:
```suggestion
				time.Sleep(time.Duration(3+rand.Intn(3))
```
wdyt?

izaaklauer (approved):

briancain (approved):

paladin-devops (approved):

PR #3636, opened 2022-08-02
Fix VCS Git Dirty Err

cicoyle:
Fix the VCS git dirty err seen when inputing the waypoint `-git-url` as type ssh, but without the git@ prepended to the url. Updated the corresponding tests to test such a case. Updated some var names to be clearer. Added a `README.md` to the `/internal/pkd/gitdirty/testdata` to better explain `from-DOTgit.sh` & `to-DOTgit.sh`.


internal/pkg/gitdirty/gitdirty.go
```diff
@@ -188,65 +212,78 @@ func remoteConvertHTTPStoSSH(httpsRemote string) (string, error) {
...
+	return strings.TrimRight(trimmedRemoteUrl, ".git")
 }
 
 // getRemoteName queries the repo at GitDirty.path for all remotes, and then
 // searches for the remote that matches the provided url, returning an error if
 // no remote url is found.
 // It will also attempt to match against different protocols - if an https protocol is
 // specified, if it can't find an exact match, it will look for an ssh-style match (and vice-versa)
-func getRemoteName(log hclog.Logger, repoPath string, remoteUrl string) (name string, err error) {
+func getRemoteName(log hclog.Logger, repoPath string, wpRemoteUrl string) (name string, err error) {
```
izaaklauer, line 251: nice new names here 👍🏼

internal/pkg/gitdirty/testdata/README.md
```diff
@@ -0,0 +1,26 @@
+# Fixture Test Data
```
izaaklauer, line 1: Thank you, this is excellent!

izaaklauer (changes requested):
Looks great!

Could you add a test case to `Test_remoteConvertSSHtoHTTPS` to convert the "no git@" case?

izaaklauer (approved):
This looks well-tested and good to me. I don't love that we have to do this much hueristics+guessing to match repo names, but this logic really all goes to feed a warning, and won't be a disaster if we discover that it doesn't work under new circumstances we haven't yet considered.

Great work on this!

evanphx (approved):
Looks great!

.changelog/3636.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+internal/pkg/gitdirty: Fix vcs git dirty err from showing when wpUrl is inputted without git@ prepended to url
```
catsby, line 2: As a user of Waypoint, this release note might not be that informative. `internal/pkg/gitdirty` is as said internal and users might not know exactly what it does. 

Could we re-word this to better explain what change the users could expect? From reading the PR it seems like something to the effect of "fix an issue detecting dirty local source code when using `git@` remotes" or something?

internal/pkg/gitdirty/gitdirty.go
```diff
@@ -188,65 +212,78 @@ func remoteConvertHTTPStoSSH(httpsRemote string) (string, error) {
...
-	httpsRemote := githubStyleSshRemoteRegexp.ReplaceAllString(sshRemote, "https://$1/$2")
+	var httpsRemote string
+	// Check if it has git@
+	if githubStyleSshRemoteRegexp.MatchString(sshRemote) {
+		httpsRemote = githubStyleSshRemoteRegexp.ReplaceAllString(sshRemote, "https://$1/$2")
+	} else {
+		// Doesn't have the git@ at the front of the url
+		httpsRemote = githubStyleSshRemoteNoAtRegexp.ReplaceAllString(sshRemote, "https://$1/$2")
+	}
+
```
catsby, line 230: Style nit: if it's going to have a "default" that's only changed in a certain condition, this format could make that simpler/clearer:

```suggestion
	httpsRemote := githubStyleSshRemoteNoAtRegexp.ReplaceAllString(sshRemote, "https://$1/$2")
	// Check if it has git@
	if githubStyleSshRemoteRegexp.MatchString(sshRemote) {
		httpsRemote = githubStyleSshRemoteRegexp.ReplaceAllString(sshRemote, "https://$1/$2")
	}
```

The current `if/else` is fine though so take it or leave it

internal/pkg/gitdirty/gitdirty.go
```diff
@@ -188,65 +212,78 @@ func remoteConvertHTTPStoSSH(httpsRemote string) (string, error) {
...
 	})
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to open git repo at path %q", repoPath)
 	}
 
-	remotes, err := repo.Remotes()
+	localRepoRemotes, err := repo.Remotes()
 	if err != nil {
-		return "", errors.Wrap(err, "failed to list remotes")
+		return "", errors.Wrap(err, "failed to list localRepoRemotes")
```
catsby, line 261: I think we should keep the errors here to say things like "remote" and not the specific variable name. As a user/operator of Waypoint that receives logs or errors, they wont know what `localRepoRemotes` necessarily means but `remotes` is more intuitive as to where to look.

internal/pkg/gitdirty/gitdirty.go
```diff
@@ -188,65 +212,78 @@ func remoteConvertHTTPStoSSH(httpsRemote string) (string, error) {
...
-			if normalizeRemote(thisRemoteUrl) == normalizeRemote(remoteUrl) {
+		for _, localRemoteUrl := range localRepoRemoteConfig.URLs {
+			if normalizeRemote(localRemoteUrl) == normalizeRemote(wpRemoteUrl) {
 				if exactMatchRemoteName != "" {
-					// NOTE(izaak): I can't think of a dev setup where you'd get multiple remotes with the same url.
+					// NOTE(izaak): I can't think of a dev setup where you'd get multiple localRepoRemotes with the same url.
 					// If it does though, I think it's likely that any remote will work for us for diffing purposes,
 					// wo we'll warn and continue.
-					log.Warn("Found multiple remotes with the target url. Will choose remote-1.", "url", thisRemoteUrl, "remote-1", exactMatchRemoteName, "remote-2", remoteConfig.Name)
+					log.Warn("Found multiple localRepoRemotes with the target url. Will choose remote-1.", "url", localRemoteUrl, "remote-1", exactMatchRemoteName, "remote-2", localRepoRemoteConfig.Name)
```
catsby, line 284: Another example of where "remotes" makes more sense as an error message rather than "localRepoRemotes"

catsby (commented):
I'd like us to re-word the changelog entry so Waypoint users better understand what's changing without having to know the codebase. 

The rest of my comments are nitpick things, or just hoping to clear up error/log messages

internal/pkg/gitdirty/gitdirty.go
cicoyle: I left this explicit in the if/else bc then it acts in an incorrect manner by having a default val w/o the if/else. For ex: in the case where we have the `sshRemote = git@git.test:testorg/testrepo.git`, then by having a default it would make the `httpsRemote = https://git@git.test/testorg/testrepo.git` bc the regex is for `githubStyleSshRemoteNoAtRegexp` in the default case and the git@ is not accounted for properly in that regex. Then it would overwrite the value in the `if` correctly, but I didn't like that it would have the wrong value in the first place. Maybe there is another edge case that I didn't account for that could return the incorrect `httpsRemote` in the future. I wrote out the explicit if/else bc I couldn't in the `githubStyleSshRemoteNoAtRegexp` regex put ^(?!git@) which would help to rectify the above mentioned case in that we want to specifically use the `githubStyleSshRemoteRegexp` in cases where we have the git@ and use the `githubStyleSshRemoteNoAtRegexp` in cases where it is still of type ssh, but w/o the git@ explicitly.

internal/pkg/gitdirty/gitdirty.go
cicoyle: Good catch. I agree and will update it accordingly.

internal/pkg/gitdirty/gitdirty.go
catsby: Ah yeah that makes sense 👍

catsby (approved):

PR #3637, opened 2022-08-03
Hide dynamic Mirage content from Percy diffs

amyrlam:
Also, flesh out Percy docs
- I believe no changelog needed: not a user-facing change. Fixes flaky diffs for from Percy visual regression testing for dev team.
- (Aside: We could component-ize icon with adjacent text, and Percy it separately, but it's not super important right now)

Resolves https://github.com/hashicorp/waypoint/issues/3631

I went ahead and added folks from @hashicorp/waypoint-frontend, @hashicorp/waypoint-ecosystem and @hashicorp/waypoint-backend to the HashiCorp Percy account, so that you can approve Percy diffs on PRs if you do indeed change the `ui/` at all. After this PR is merged to `main`, the dynamic content will be hidden and other open PRs will need to rebase, so as not to bring in "incorrect" snapshots to `main`, that will cause more flaky Percy diffs.

***

- [x] For the reviewer: Please +1 the Percy also as part of code review

I personally like to review with this option clicked, but YMMV:

image

PR #3641, opened 2022-08-04
use fixed status on release and build mock data

gregone:
The data used for the percy visual acceptance test was using randomized `status` on the build and release objects, so those were not consistent with the baseline from test to test, This assigns a fixed status to both. Once this is merged, we should approve it as the baseline.


sdav9375 (approved):
LGTM

amyrlam:
thanks! relates to https://github.com/hashicorp/waypoint/pull/3637

amyrlam:
Approved Percy, I could remove `hide-in-percy`, when not random, in a follow-on PR

PR #3646, opened 2022-08-05
Add OIDC Callback Server Port personalization

### LINKED ISSUE #3645: waypoint login uses a random callback listener port - invalid redirect_uri
**Describe the bug**

When OIDC is being used, the OIDC provider has to be configured with a `redirect_uri` that is usually static.  
Unfortunately the CLI is using a dynamic uri (because it uses a random port that gets assigned when the web server is created), and thus this causes issues for not-so-flexible OIDC providers.

**Steps to Reproduce**

Assume your OIDC provider can only use "static" URLs like:

`http://127.0.0.1:8085/oidc/callback`

Try to use `waypoint login -v`

**Expected behavior**
The CLI uses always the same port, so that the URL is recognized by the OIDC Server as a valid redirect_uri

applicable:

* Waypoint CLI Version: 0.9.0
* Waypoint Server Platform and Version: N/A
* Waypoint Plugin: N/A

**Additional context**
This causes also problems when setting up the OIDC provider with the `-allowed-redirect-uri` parameter, as random URLs aren't supported by Waypoint.

denysvitali:
Fixes #3645

This PR switches from using a dynamic TCP port for the HTTP server that listens for the OIDC Callback in the CLI to using a static port (`8087`) or a custom one provided by the user via the `-oidc-callback-server-port` flag.

Example:

```
$ waypoint login -oidc-callback-server-port=8100
Complete the authentication by visiting your authentication provider.
Opening your browser window now. If the browser window does not launch,
please visit the URL below:
https://sso-provider.example.com/oauth/authorize?client_id=XYZ&nonce=my-nonce&redirect_uri=http%3A%2F%2F127.0.0.1%3A8100%2Foidc%2Fcallback&response_type=code&scope=openid+roles&state=st_dkdue3832
```

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -28,12 +28,13 @@ import (
 type LoginCommand struct {
 	*baseCommand
 
-	flagAuthMethod     string
-	flagToken          string
-	flagK8S            bool
-	flagK8SService     string
-	flagK8STokenSecret string
-	flagK8SNamespace   string
+	flagAuthMethod             string
+	flagToken                  string
+	flagK8S                    bool
+	flagK8SService             string
+	flagK8STokenSecret         string
+	flagK8SNamespace           string
+	flagOidcCallbackServerPort uint
 }
 
 func (c *LoginCommand) Run(args []string) int {
@@ -260,8 +261,12 @@ func (c *LoginCommand) loginOIDC(ctx context.Context) (string, int) {
 	}
 	refAM := &pb.Ref_AuthMethod{Name: c.flagAuthMethod}
 
+	if c.flagOidcCallbackServerPort == 0 {
+		c.flagOidcCallbackServerPort = 8087 // Default port
+	}
+
 	// Start our callback server
-	callbackSrv, err := wpoidc.NewCallbackServer()
+	callbackSrv, err := wpoidc.NewCallbackServer(c.flagOidcCallbackServerPort)
 	if err != nil {
 		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 		return "", 1
@@ -463,6 +468,12 @@ func (c *LoginCommand) Flags() *flag.Sets {
 			Target: &c.flagK8SNamespace,
 			Usage: "The name of the Kubernetes namespace that has the Waypoint token " +
 				"when using the -from-kubernetes flag.",
+		})
+
+		f.UintVar(&flag.UintVar{
+			Name:   "oidc-callback-server-port",
+			Target: &c.flagOidcCallbackServerPort,
+			Usage:  "The port to be used when listening for an OIDC callback",
 		})
 	})
 }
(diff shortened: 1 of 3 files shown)


internal/cli/login.go
```diff
@@ -260,8 +261,12 @@ func (c *LoginCommand) loginOIDC(ctx context.Context) (string, int) {
 	}
 	refAM := &pb.Ref_AuthMethod{Name: c.flagAuthMethod}
 
+	if c.flagOidcCallbackServerPort == 0 {
+		c.flagOidcCallbackServerPort = 8087 // Default port
```
evanphx, line 265: Why was 8087 chosen? Is that a standard? This PR looks great, trying to understand why we shouldn't use an ephemeral port as the default.

evanphx (changes requested):

internal/cli/login.go
denysvitali: > Why was 8087 chosen?

Because 8080 is where usually you might have other apps running, 8087 is less likely to be used

> Is that a standard?
Nope

>  trying to understand why we shouldn't use an ephemeral port as the default.

Having a dynamic port doesn't work well with some OIDC Providers, you can't just say "allow everything that matches localhost" for some of them, and thus you have to specify a somewhat static URL.

On top of that, Waypoint itself doesn't seem to support them via `-allowed-redirect-uri=` - but I might be wrong on this one.

denysvitali: On the other hand, you have the issue where people must specify at each iteration what is the needed port (`0` for dynamic, or a custom port).

I don't think it makes sense to have to ask the users to do something like `waypoint login -oidc-server-callback=8888` every time.

Can this be maybe moved to the server instead?

evanphx: If you set 8087 as the default in the flag, they won't have to specify it, it will default to 8087.

The reason it's on login is it's so specific to the temporary server that login has to spin up to finish the OIDC process.

paladin-devops: Hello! I'm also in favor of having this default to a random port, namely because OIDC providers _should_ allow any port to be specified for the callback when the loopback address (localhost) is being used, as per [this RFC](https://www.rfc-editor.org/rfc/rfc8252#section-7.3) from the IETF. With that said though, I think that allowing a custom port to be specified is helpful when working with OIDC providers that don't allow random ports.

This is an [example implementation](https://github.com/hashicorp/vault/pull/13871) of this requirement on Vault as the OIDC provider.

briancain: > maybe you want to "force" your clients to use a static port to work around the providers which might not be complaint with RFC8252.

At this time we see value in having the default port be dynamic. We don't want to force users to use a static port by default, but we do agree that having the option to configure the port to be static is something that we want 😄 Allowing people to do this will help those who are using non-compliant clients to still be able to use their OIDC provider.

denysvitali:
I agree with @evanphx. This has to be fixed on the OIDC provider as it's a violation of the RFC.

Let me close this issue, I'm anyways tracking the resolution of our issue on the OIDC provider side:
https://github.com/cloudfoundry/uaa/issues/2030

PR #3647, opened 2022-08-05
Improve queue init job error

izaaklauer:
@sdav9375  and I hit this error in practice (I think), and it wasn't clear where it came from! This should help in the future.


sdav9375 (approved):
A friendlier error 😄

briancain (approved):

Issue #3597, opened 2022-07-25
Add e2e tests to CI


krantzinator:
We have some e2e tests in `test-e2e`, but they don't run automatically on merges to `main` at the moment. We should do that.

PR #3648, opened 2022-08-05
CI: add e2e tests for Docker platform

krantzinator:
This fixes up a few things in our bash setup for e2e to run in CI, and adds support initially for the Docker platform.
Other platforms will be tracked in #3597. This base test case would have caught some of the edge cases we've missed in recent pre-release testing, so it's good to get this in as a starting point.

Since I've now set these to run only on `main` after an image is built off of `main`, look at [the second-to-last workflow](https://app.circleci.com/pipelines/github/hashicorp/waypoint/17223/workflows/b069f1e5-7c28-434f-b5aa-1a3c4d6f6b48) for the example run.

### DIFF
--- a/test-e2e/run-test.sh
+++ b/test-e2e/run-test.sh
@@ -26,11 +26,6 @@ if [ "${E2E_PLATFORM}" != "Docker" ] && [ "${E2E_PLATFORM}" != "Kubernetes" ] &&
   exit 1
 fi
 
-if [ -z "$WP_SERVERIMAGE_UPGRADE" ]; then
-  echo "Env var 'WP_SERVERIMAGE_UPGRADE' must be set (e.g. 'hashicorp/waypoint:latest')"
-  exit 1
-fi
-
 # For running script outside of `test-e2e` folder
 TESTDIR="${WP_TESTE2E_DIR:-$(pwd)}"
 
--- a/test-e2e/run-test.sh
+++ b/test-e2e/run-test.sh
@@ -20,6 +20,11 @@ spin()
 echo "Beginning Waypoint end-to-end tests..."
 echo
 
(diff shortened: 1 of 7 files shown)


test-e2e/run-test.sh
```diff
@@ -81,21 +95,31 @@ echo
...
+if [ -z "$CI_ENV" ]; then
   spin &
   SPIN_PID=$!
-  trap "kill -9 $SPIN_PID" `seq 0 15`
+  trap 'kill -9 $SPIN_PID' $(seq 0 15)
 fi
 
-go test -v "github.com/hashicorp/waypoint/test-e2e"
+# Run Docker tests
+go test -v "github.com/hashicorp/waypoint/test-e2e" -run "$E2E_PLATFORM"
```
paladin-devops, line 105: Does the env var `E2E_PLATFORM` need to be set before running this script? If so, might be helpful to document that 😄

paladin-devops (approved):

test-e2e/run-test.sh
briancain: We could update the script to check if it's set and throw and error if it's empty too saying what's expected

krantzinator: This is in the bash file
```
if [ "${E2E_PLATFORM}" != "Docker" ] && [ "${E2E_PLATFORM}" != "Kubernetes" ] && [ "${E2E_PLATFORM}" != "Ecs" ] && [ "${E2E_PLATFORM}" != "Nomad" ]; then
   echo "Environment variable 'E2E_PLATFORM' must be one of: 'Docker', 'Kubernetes', 'Ecs', 'Nomad'"
   exit 1
 fi
```

What other documentation would you like to see? A top-level comment with instructions?

briancain: That seems pretty legit to me personally! 👍🏻

PR #3652, opened 2022-08-06
(potential) bug fix: properly queue ODR source job in a task

xiaolin-ninja:
Currently, we queue the source job, but not the modified source job that contains the task information, and the taskSource (renamed sourceJob) variable is constructed, but never actually used.

Currently, if `job.Task` is called on the queued source job, Task related functions would silently exit, thinking the job is not a part of a Task.


briancain (commented):
It'd be good to update our tests in `serivce_job_test` to check that the ODR wrapped job for sourceJob has the proper task info set on it to match the fix this PR is making. Otherwise makes sense to me! I'll be sure to test this manually as well

briancain (approved):
🚀 looks good!

PR #3658, opened 2022-08-08
Explicitly returning ODR config when setting it in state

izaaklauer:
For better or worse, currently, we don't require that a user set an ODR config's name. If they call OnDemandRunnerConfigPut with no name OR id, it's OK, we'll generate an ID and use it as the name, then we return that ODR profile with the name and ID set.

The problem is, our state function wasn't returning an ODR profile with the name set - it was mutating the input, and relying on the server handler to know about it. If someone else were implementing a new state backend, the would have no way of knowing it was important to set the name on the input, and likely wouldn't. This bit us in HCP.

Now, state returns and explicit OnDemandRunnerConfig. A much more clear API contract.

### DIFF
--- a/pkg/serverstate/statetest/test_runner_ondemand.go
+++ b/pkg/serverstate/statetest/test_runner_ondemand.go
@@ -27,12 +27,33 @@ func TestOnDemandRunnerConfig(t *testing.T, factory Factory, restartF RestartFac
 		defer s.Close()
 
 		// Set
-		ret, err := s.OnDemandRunnerConfigGet(&pb.Ref_OnDemandRunnerConfig{
+		_, err := s.OnDemandRunnerConfigGet(&pb.Ref_OnDemandRunnerConfig{
 			Id: "foo",
 		})
 		require.Error(err)
 		require.Equal(codes.NotFound, status.Code(err))
-		require.Equal(ret.Name, "foo")
+	})
+
+	t.Run("Set with no name", func(t *testing.T) {
+		require := require.New(t)
+
+		s := factory(t)
+		defer s.Close()
+
+		// Set
+		rec := serverptypes.TestOnDemandRunnerConfig(t, &pb.OnDemandRunnerConfig{
+			OciUrl:               "h/w:s",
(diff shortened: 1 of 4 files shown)


pkg/serverstate/statetest/test_runner_ondemand.go
```diff
@@ -34,6 +34,28 @@ func TestOnDemandRunnerConfig(t *testing.T, factory Factory, restartF RestartFac
 		require.Equal(codes.NotFound, status.Code(err))
 	})
 
+	t.Run("Set with no name", func(t *testing.T) {
```
izaaklauer, line 37: Added an explicit test for auto-name-generation

pkg/serverstate/statetest/test_runner_ondemand.go
```diff
@@ -34,6 +34,28 @@ func TestOnDemandRunnerConfig(t *testing.T, factory Factory, restartF RestartFac
...
+			TargetRunner:         &pb.Ref_Runner{Target: &pb.Ref_Runner_Any{Any: &pb.Ref_RunnerAny{}}},
+			PluginConfig:         []byte(`{"foo":"bar"}`),
+			ConfigFormat:         pb.Hcl_JSON,
+			Default:              true,
+		})
+
+		result, err := s.OnDemandRunnerConfigPut(rec)
+		require.NoError(err)
+		require.NotEmpty(result.Id)
+		require.NotEmpty(result.Name)
```
briancain, line 56: Maybe one more for extra assurance:

```golang
require.Equal(result.Name, result.Id)
```

briancain (approved):
That makes sense to me, it should have gone there anyway I think! 👍🏻

pkg/serverstate/statetest/test_runner_ondemand.go
izaaklauer: I thought about this, but it feels like testing an implementation detail. We don't really care what the name is - just that we gave it _some_ name.

I honestly don't love using the ulid as the name. It's very long, and it's hard for human eyes to distinguish from other ulids. I feel like i'd prefer us to use pet names, or 6 random letters k8s-pod-hash-style, or something.

PR #3661, opened 2022-08-09
Sanitize server errs

cicoyle:
Utilizing the new server package hcerr to sanitize raw server err messages.


evanphx (approved):
Perfection.

pkg/server/singleprocess/service_deploy.go
```diff
@@ -127,7 +128,7 @@ func (s *Service) GetDeployment(
 
 	d, err := s.state(ctx).DeploymentGet(req.Ref)
 	if err != nil {
-		return nil, err
+		return nil, hcerr.Externalize(hclog.FromContext(ctx), err, "failed to get deployment", "id", req.Ref.Target)
```
izaaklauer, line 131: `req.Ref.Target` is a bit of a weird one. It's possible to get deployments either by ID or by sequence number, so Target can be either of those: https://github.com/hashicorp/waypoint/blob/a0d9c3d198617858909e816233cb367d7652883a/pkg/server/proto/server.proto#L942-L947

That said, I gave this a test, and the logger actually formats it ok!

```
$ waypoint deploy

» Deploying web...
  Performing operation locally
✓ Running deploy v1
✓ Setting up network...
✓ Starting container
✓ App deployed as container: web-01GA3SSGPMANZ1TQ7662R1HTDM
✓ Docker image "waypoint.local/web:latest" up to date!
! failed to get deployment
  "id": "&{01GA3SSGA608XDMQ0KQWCJRYN2}"
```

It's not beautiful, but it's better than omitting the context I think. The one change i'd make:

```suggestion
		return nil, hcerr.Externalize(hclog.FromContext(ctx), err, "failed to get deployment", "target", req.Ref.Target)
```

pkg/server/singleprocess/service_runner.go
```diff
@@ -82,15 +87,19 @@ func (s *Service) AdoptRunner(
...
+			return &empty.Empty{}, hcerr.Externalize(log, err, "failed to adopt runner", "id", req.RunnerId)
+		}
 	} else {
-		err = s.state(ctx).RunnerReject(req.RunnerId)
+		if err = s.state(ctx).RunnerReject(req.RunnerId); err != nil {
+			return &empty.Empty{}, hcerr.Externalize(log, err, "failed to reject runner", "id", req.RunnerId)
+		}
 	}
 
 	return &empty.Empty{}, err
```
izaaklauer, line 105: Totally unrelated to you, just a thing I noticed. Err has to be nil here, so it'd be nice to be explicit about that.
```suggestion
	return &empty.Empty{}, nil
```

izaaklauer (changes requested):
🎉 Great work! Just a few fiddly things, but this is a huge improvement. I am so excited to never again see "err: record not found" with no additional context

pkg/server/singleprocess/service_runner.go
cicoyle: I did see this, but thought it was fine since default value for err would be nil anyways. I can change it to be explicit if that is preferred.

izaaklauer (approved):

izaaklauer:
Weird that the frontend tests are failing - that's usual, and worth looking into before merging.

PR #3662, opened 2022-08-09
Fixing inlinekeepalive early stream exit warning

izaaklauer:
Prior to this, if a stream were to exit very quickly, we would fail our call to the server's GetVersionInfo due to context cancelled, and log a warning. This is actually fine - we can just stop setting up the inlinekeepalive logic, because there's no stream anymore!


evanphx (approved):

briancain (approved):
👍🏻 fixed the extra warning issue for me!

PR #3664, opened 2022-08-10
e2e updates

krantzinator:
Ignore the branch name; there was no bug with the tests -- they found a real bug!

But this does fix the fact that when it found the bug, we weren't seeing the error message from `stderr`. Plus updating the `CI` env var so that we don't see the spinner output `|\-/|\-/|\-/|\-/---` prefixing the test results.

As these only run on main; see [this run](https://app.circleci.com/pipelines/github/hashicorp/waypoint/17378/workflows/557a7efd-a414-4c1b-8d73-d765526f2a93/jobs/186773) for the results of this PR.

### DIFF
--- a/test-e2e/README.md
+++ b/test-e2e/README.md
@@ -50,7 +50,7 @@ func TestWaypointAvailable(t *testing.T) {
 	}
 
 	if stderr != "" {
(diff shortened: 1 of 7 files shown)


catsby (approved):

briancain (approved):
🚀

PR #3669, opened 2022-08-11
When auth'ing via oauth2, pass the waypoint token as well via grpc

evanphx:
The waypoint token is used as an identification document as well, so we
need the server to see it as well, such that we can validate the
identity even after oauth2 has authenticated the user.

Effectively, oauth2 is authentication, the waypoint token is used for
authorization.


catsby (approved):
LGTM

pkg/tokenutil/grpc.go
```diff
@@ -43,14 +43,34 @@ func (t ContextToken) GetRequestMetadata(ctx context.Context, uri ...string) (ma
 	}
 
 	return map[string]string{
+		// TODO remove passing the token in authorization after the next major version.
 		"authorization": finalT,
+		MetadataKey:     finalT,
```
izaaklauer, line 48: Do you know this might push us close to a size limit? I did some quick looking, and it looks grpc metadata winds up in http/2 headers, which are compressed but also maybe have a 4k byte limit for most load balancers?

evanphx: because they're compressed, the second value is basically "free"

pkg/server/auth.go
```diff
@@ -63,8 +64,14 @@ func AuthUnaryInterceptor(checker AuthChecker) grpc.UnaryServerInterceptor {
...
-			token = authHeader[0]
+		// First, let's see if the token is in our dedicated key.
+		if tokenHeader, ok := md[tokenutil.MetadataKey]; ok {
+			token = tokenHeader[0]
+		} else {
+			// Otherwise, see if it's been stuffed into Authorization.
+			if authHeader, ok := md["authorization"]; ok {
+				token = authHeader[0]
+			}
 		}
```
catsby, line 76: In `AuthStreamInterceptor()` we check if `token == ""` do we need to do that here as well?

catsby (approved):
LGTM just one question I wasn't sure on

pkg/server/auth.go
evanphx: We have some anonymous RPC endpoints that aren't streaming, that's why the non-streaming side let's it go through. The authChecker that is called here then performs another check to see if the endpoint is anonymous or not.

So basically, we have no anonymous streaming endpoints, and thus we just terminate it early.

pkg/tokenutil/grpc.go
izaaklauer: AH good point. And I don't think this will show up in the headers the UI sends either.

pkg/server/proto/server.proto
```diff
@@ -4860,6 +4860,30 @@ message TokenTransport {
   // Any configuration style metadata that can be passed along with the token
   // without invalidating the token body itself.
   map<string, string> metadata = 4;
+
+  // The external creds are stored outside of the authenticated token body
+  // because we want clients that see the creds in the transport to use
+  // them BUT then strip them out before transmitting the token, so that
+  // the creds aren't be transfer constantly.
```
briancain, line 4867:
```suggestion
  // the creds aren't transfered constantly.
```

PR #3670, opened 2022-08-12
Move token signing logic into serverstate

izaaklauer:
This refactors token signature generation and verification into the state package, allowing different state implementations to generate token signatures however they chose (i.e. not through state-stored HMAC keys).

This also adds the ability to set a server-global "keyId", which is encoded into tokens we generate, and can be used to ensure that we always verify a signature using the same key that we used to generate it.

Builds upon and closes https://github.com/hashicorp/waypoint/pull/3581


evanphx (approved):
Looks good! This mostly just refactors the hmac code we had in server into state, which is great.

PR #3671, opened 2022-08-12
Backport of (potential) bug fix: properly queue ODR source job in a task into release/0.9.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #3652 to be assessed for backporting due to the inclusion of the label backport/0.9.x.

---

Currently, we queue the source job, but not the modified source job that contains the task information, and the taskSource (renamed sourceJob) variable is constructed, but never actually used.

Currently, if `job.Task` is called on the queued source job, Task related functions would silently exit, thinking the job is not a part of a Task.


xiaolin-ninja:
This is actually just an internal fix and a potential edge case bug that doesn't need to be backported.

PR #3673, opened 2022-08-12
Cascading deletion of project-associated records

paladin-devops:
Prior to this PR, `project destroy` _did_ hard-delete the project record from the database. However, it _did not_ delete records from the database associated with a project. This PR does just that, and updates the deletion of a project in Bolt DB to also include deleting:

- builds
- artifacts
- deployments
- releases
- status reports
- triggers
- pipelines
- workspaces
- config (at the app scope of apps within a project, and at the project scope)

Builds, artifacts, deployments, releases and status reports are now able to be deleted with the implementation of the `Delete` function for the struct `appOperation`.


internal/server/boltdbstate/project.go
```diff
@@ -69,10 +69,172 @@ func (s *State) ProjectGet(ref *pb.Ref_Project) (*pb.Project, error) {
...
+					return err
+				} else {
+					triggers = append(triggers, triggerList...)
+				}
+				if workspaceDetail, err := s.WorkspaceGet(workspace.Workspace.Workspace); err != nil {
+					return err
+				} else {
+					// If the project we're deleting is the only project in the workspace, we delete the workspace
+					// We don't delete the default workspace
+					if len(workspaceDetail.Projects) == 1 && workspace.Workspace.Workspace != "default" {
```
paladin-devops, line 150: Regarding workspaces, one behavior that this PR does not address: 
- there is a workspace with multiple projects
- one of the projects in this workspace is the one we want to delete
- goal: remove the project from the workspace

This isn't yet possible as `func (s *State) WorkspacePut(workspace *pb.Workspace) error` notes:
```go
// WorkspacePut creates or updates the given Workspace.
//
// Project changes will be ignored
```

Prior to learning of this note, locally I was testing getting all the workspaces with multiple projects, removing the project we're deleting from the slice of projects, and then re-putting the workspace. However, until this limitation is addressed, workspaces will continue to display that a deleted project is amongst its projects.

The deletion of workspaces where the project we're deleting is the only project in the workspace is working, though.

internal/server/boltdbstate/app_operation_test.go
```diff
@@ -63,25 +63,31 @@ func TestAppOperation(t *testing.T) {
...
+		require.Equal("A", b.Id)
+		require.Equal(uint64(1), b.Sequence)
+
+		// Delete it by ID
+		err = op.Delete(s, &pb.Ref_Operation{
+			Target: &pb.Ref_Operation_Id{
+				Id: b.Id,
+			},
+		})
+		require.Nil(err)
```
briancain, line 90: A test or two around deleting something with a non-existent operation id would be good too! Maybe in its own `Run` block.

internal/server/boltdbstate/project.go
```diff
@@ -69,10 +69,172 @@ func (s *State) ProjectGet(ref *pb.Ref_Project) (*pb.Project, error) {
...
+	// We perform all of our reads before our write to avoid the deadlocked state
+	var builds []*pb.Build
+	var artifacts []*pb.PushedArtifact
+	var deployments []*pb.Deployment
+	var releases []*pb.Release
+	var statusReports []*pb.StatusReport
+	var workspaces []*pb.Workspace_Project
+	var triggers []*pb.Trigger
+	var pipelines []*pb.Pipeline
+	var configVars []*pb.ConfigVar
```
briancain, line 90: For readability, I recommend using a `var` block like:

```golang
var (
  builds    []*pb.Build
  artifacts []*pb.PushedArtifact
  // ... etc ...
)

File: internal/server/boltdbstate/project_test.go [hunk 4]
```diff
@@ -0,0 +1,241 @@
...
+		// Create a project with one app
+		require.NoError(s.ProjectPut(&pb.Project{
+			Name: projectName,
+			Applications: []*pb.Application{
+				{
+					Project: &pb.Ref_Project{Project: projectName},
+					Name:    appName,
+				},
+			},
+		}))
```
briancain, line 27: We could use a multi-app test too! And in a similar vein, it would be nice if build/deploy/release etc messages had more than one. Maybe 2-3 each just to fully test the full flow beyond a single message per resource?

internal/server/boltdbstate/project_test.go
```diff
@@ -0,0 +1,241 @@
...
+		require.NoError(s.TriggerPut(&pb.Trigger{
+			Id:        "testTrigger",
+			Name:      "testTrigger",
+			Project:   &pb.Ref_Project{Project: projectName},
+			Workspace: &pb.Ref_Workspace{Workspace: "testWorkspace"},
+		}))
+
+		// Delete the project (this should also delete the build)
+		err = s.ProjectDelete(&pb.Ref_Project{Project: projectName})
+		require.NoError(err)
```
briancain, line 140: Small nit but you can combine a require.NoError/require.Error and a function so it's all on a single line:

```golang
require.NoError(s.ProjectDelete(&pb.Ref_Project{Project: projectName}))
```

briancain (changes requested):
Looking pretty good! Mostly did a quick code review and left some feedback, but this is coming together nicely 👍🏻

go.sum
```diff
@@ -1245,8 +1245,6 @@ github.com/julienschmidt/httprouter v1.3.0/go.mod h1:JR6WtHb+2LUe8TCKY3cZOxFyyO8
 github.com/kardianos/osext v0.0.0-20190222173326-2bc1f35cddc0/go.mod h1:1NbS8ALrpOvjt0rHPNLyCIeMtbizbir8U//inJ+zuB8=
 github.com/karrick/godirwalk v1.15.8 h1:7+rWAZPn9zuRxaIqqT8Ohs2Q2Ac0msBqwRdxNCr2VVs=
 github.com/karrick/godirwalk v1.15.8/go.mod h1:j4mkqPuvaLI8mp1DroR3P6ad7cyYd4c1qeJ3RV7ULlk=
-github.com/kevinburke/go-bindata v3.22.0+incompatible h1:/JmqEhIWQ7GRScV0WjX/0tqBrC5D21ALg0H0U/KZ/ts=
-github.com/kevinburke/go-bindata v3.22.0+incompatible/go.mod h1:/pEEZ72flUW2p0yi30bslSp9YqD9pysLxunQDdb2CPM=
```
catsby, line 1249: I feel like this shouldn't change unless `go.mod` did

internal/server/boltdbstate/project.go
```diff
@@ -69,10 +69,185 @@ func (s *State) ProjectGet(ref *pb.Ref_Project) (*pb.Project, error) {
...
+
+	// Builds, artifacts, deployments, releases, status reports, pipelines, triggers,
+	// workspaces and config are deleted with a project
+	// Jobs and tasks will NOT be deleted along with a project
+	// Instances are expected to be deleted before ProjectDelete, via the destroy op
+	for _, build := range builds {
+		if err = s.BuildDelete(&pb.Ref_Operation{Target: &pb.Ref_Operation_Id{Id: build.Id}}); err != nil {
+			return err
+		}
+		s.log.Debug("deleted build " + build.Id)
```
catsby, line 187: Should we emit a debug log for all the other things we delete? This is the only one

internal/server/boltdbstate/project.go
```diff
@@ -69,10 +69,185 @@ func (s *State) ProjectGet(ref *pb.Ref_Project) (*pb.Project, error) {
...
+	// Get our project to delete
+	project, err := s.ProjectGet(ref)
+	if err != nil {
+		if status.Code(err) == codes.NotFound {
+			return nil
+		}
+		return err
+	}
+
+	// We perform all of our reads before our write to avoid the deadlocked state
```
catsby, line 81: This seems lossy; we don't lock any of these tables or operations on projects, so in theory operations could be recorded for a project _after_ we preform these reads. 

Examples I can think of would be app polling, and if say someone was in the middle of running a `build` operation while this project delete operation happened.

catsby (changes requested):
Please see my comments/thoughts/concerns on the lossy-ness here. Please let me know if I'm mistaken or if we're covered here.

internal/server/boltdbstate/app_operation_test.go
```diff
@@ -511,6 +517,23 @@ func TestAppOperation(t *testing.T) {
 		// Observe that the watch fires
 		require.False(ws.Watch(time.After(1 * time.Second)))
 	})
+
+	t.Run("attempt deletion of non-existest operation id", func(t *testing.T) {
```
xiaolin-ninja, line 521:
```suggestion
	t.Run("attempt deletion of non-existent operation id", func(t *testing.T) {
```

internal/server/boltdbstate/project.go
```diff
@@ -69,10 +69,185 @@ func (s *State) ProjectGet(ref *pb.Ref_Project) (*pb.Project, error) {
...
+	for _, pipeline := range pipelines {
+		if err = s.PipelineDelete(&pb.Ref_Pipeline{Ref: &pb.Ref_Pipeline_Id{
+			Id: &pb.Ref_PipelineId{Id: pipeline.Id},
+		},
+		}); err != nil {
+			return err
+		}
+	}
+
 	memTxn := s.inmem.Txn(true)
```
xiaolin-ninja, line 247: I'm concerned that if half way through the cascading delete something fails, e.g. we delete all the configs, but fail to delete the pipelines, there is no rollback and the project is in a half-deleted state. Since it doesn't look like the cascading deletes are wrapped in a transaction. I also didn't look too closely at the ordering, but looks like workspace is deleted before project is deleted, but shouldn't that happen after?

paladin-devops: @xiaolin-ninja I think your comment addressed each concern that I had as I was attempting to implement this 😄 I agree it's not ideal that it's seemingly unable to be wrapped in a single transaction which can be rolled back. Fortunately though, the project is the last thing that is deleted, so if a failure happens before the project gets deleted, another attempt can be made to delete the project, and the remaining records will still be able to be referenced by the project.

Regarding the comment about `workspace`, I think it's okay to delete the workspace before the project, because if the project is destroyed but we fail to delete the workspace after that, then we won't be able to later look up the workspace(s) by project, since the project will be gone.

go.sum
paladin-devops: This change is related to [your comment](https://github.com/hashicorp/waypoint/pull/3626#discussion_r949294235) on. The vendor check on that branch is currently failing, and this change addresses the version update. Regarding if this upgrade is necessary, we'll address that on the other PR 😄

internal/server/boltdbstate/project.go
paladin-devops: I included this during testing the deletion of multiple builds for an app in a project. I think it's worth adding a debug log for other deleted things, so that we can see what was successfully deleted if desired.

internal/server/boltdbstate/project.go
paladin-devops: @catsby I agree that we could lose the deletion of records which are added after the deletion of a project begins. After having researched this topic, I learned that having transactions depend upon each other, like we're doing here by listing all of our transactions to delete before we delete them, doesn't work with Bolt's architecture (see [here](https://pkg.go.dev/go.etcd.io/bbolt@v1.3.6#readme-transactions) and [here](https://github.com/boltdb/bolt/issues/240#issuecomment-52409844).

With that said, I'm not sure how else we could retrieve the IDs of all of the records we need to delete during the deletion of a project _without_ first listing them, and thus the write transactions depending upon the reads. I previously considered making the calls to the state package from the  `runner` package. After discussing where to perform the necessary deletions for project-related records, we decided that in the `boltdbstate` package (specifically in the `ProjectDelete` function) would be best, but that was before digging in this far. It might be worth revisiting the idea of calling them from the runner, but then we're back to how to lock specific tables (or "buckets" in Bolt), which I don't think can be done in Bolt. 🤔 

cc/ @cicoyle @briancain

briancain: @paladin-devops - You want to do these listings inside the `db.Update()` func, which _does_ act as a lock for writes. That also means you'll want to use the lower case state versions, which are meant to be called inside a transaction.

Then in the case @catsby mentioned, it wouldn't be possible to do a write during this operation because this transaction holds onto the table lock.

internal/server/boltdbstate/project.go
briancain: All of these should be called in a single `db.Update()` with the lower-case state versions used (meant to called inside a transaction, i.e. `triggerDelete` instead of `TriggerDelete`.). Then if _any_ of them fail, the entire transaction is rolled back. Doing all of this in a single transaction means we shouldn't have any issues with race conditions or locking as far as I know.

internal/server/boltdbstate/project.go
paladin-devops: @briancain thanks for the tip!! If I recall, when I was initially attempting to perform the reads and writes from `ProjectDelete`, I was attempting to use the unexported (lowercased) state functions. However, upon looking at that once again now, I've noticed that for `appOperation`, the delete functions don't have unexported versions of themselves - i.e., when they are called, new transactions are unavoidably started (the transaction isn't being passed in to the function, it's being set inside the function):
```go
// Delete deletes an operation record by reference
func (op *appOperation) Delete(s *State, ref *pb.Ref_Operation) error {
	memTxn := s.inmem.Txn(true)
	defer memTxn.Abort()

	target, ok := ref.Target.(*pb.Ref_Operation_Id)
	if !ok {
		return status.Errorf(codes.FailedPrecondition,
			"unknown operation reference type: %T", ref.Target)
	}
	id := target.Id

	err := s.db.Update(func(dbTxn *bolt.Tx) error {
		return op.dbDelete(dbTxn, []byte(id))
	})
	if err == nil {
		memTxn.Commit()
	}

	return err
}
```
Unlike one of the other functions such as `triggerDelete`:
```go
func (s *State) triggerDelete(
	dbTxn *bolt.Tx,
	memTxn *memdb.Txn,
	ref *pb.Ref_Trigger,
) error {
	// Get the trigger. If it doesn't exist then we're successful.
	_, err := s.triggerGet(dbTxn, memTxn, ref)
	if err != nil {
			return nil
		}

		return err
	}

	// Delete from bolt
	id := s.triggerIdByRef(ref)
	if err := dbTxn.Bucket(triggerBucket).Delete(id); err != nil {
		return err
	}

	// Delete from memdb
	if err := memTxn.Delete(triggerIndexTableName, &triggerIndexRecord{Id: string(id)}); err != nil {
		return err
	}

	return nil
}
```

To address this, I'm considering adding unexported versions of the delete endpoints for each app operation, with additional input params to allow the Bolt and Mem DB transactions to be passed in to the functions, rather than being created during the `Delete` function, leading to the deadlocked state.

briancain: Yep, you'll need to add an app operation version for a `delete` that can be called in a transaction, makes sense! 👍🏻

paladin-devops: Updated! 😎 Thank you @catsby & @briancain !

internal/server/boltdbstate/project.go
paladin-devops: Updated! 😎

briancain (changes requested):
I gave this a shot (haven't had a chance to read the code yet again for review, fyi!) and I think there's still some things we have to remove.

It looks like we're still holding onto the Horizon URL hostname between destroys. And I also noticed we still use the old sequence numbers between destroys (i.e. the destroyed project was on operation sequence 5, destroy the project, then the _first_ sequence in the new project is 6). We should fix those up!

PR #3684, opened 2022-08-16
Use the active key id for signing adopted runner tokens

izaaklauer:
Missed this as part of https://github.com/hashicorp/waypoint/pull/3670, I think because this is a multi-line funciton call.

### DIFF
--- a/pkg/server/singleprocess/service_runner.go
+++ b/pkg/server/singleprocess/service_runner.go
@@ -2,10 +2,11 @@ package singleprocess
 
 import (
 	"context"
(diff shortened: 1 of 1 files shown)


catsby (approved):
LGTM assuming `activeAuthKeyId` is always set. I see it gets a default value in `singleprocess/service.go`, I assume that code path always hits before this one?

evanphx (approved):

izaaklauer:
> LGTM assuming `activeAuthKeyId` is always set. I see it gets a default value in `singleprocess/service.go`, I assume that code path always hits before this one?

Yup, it is! That's in the server constructor, so it'll always be set.

PR #3693, opened 2022-08-16
test-e2e/docker: Include `stderr` output on CLI failures

briancain:
If the CLI fails to execute during one of the end to end tests, we don't get too much information as to why we received an `err` and we fail early before printing `stdout` and `stderr`. This is a quick fix to include `stderr` if we get an `err` from executing the Waypoint CLI during the e2e tests.

Before this PR, we'd get:

>     docker_smoke_test.go:81: unexpected error deploying waypoint project: exit status 1


briancain:
Maybe a future enhancement is to include a special "Error" func for e2e tests that accepts `error`, `stdout`, `stderr` as parameters and we print the test failure all at once rather than 3 special if-statements per test

PR #3704, opened 2022-08-17
CLI HCL generator tool

HenryEstberg:
The complete CLI based HCL generator tool. This includes embedding the JSON files in the binary, and relies on #3579 to generate and test those JSON files.

To test, run waypoint init without an .hcl file in your project directory, then answer 'yes' when it asks if you want to start the interactive generator and follow the prompts. 

`docs/gen` was renamed to `embedJson/gen` due to some embedding syntax limitations that require the .go file with the embedding code to be located in a parent directory of the files that we want to embed. As a result of that the directory was renamed to clarify its use as a package. 

Example generated HCL file:

```project = "test"
app "testapp" {
  build {
     use "files" {
     }
    registry {
      use "aws-ecr" {
        tag = "foo"
      }
    }
  }
  deploy {
    use "nomad-jobspec" {
      vault_token  = "foo"
      consul_token = "foo"
      jobspec      = "foo"
    }
  }
  release {
    use "nomad-jobspec-canary" {
    }
  }
}

### CODE REVIEW

**Review by briancain — CHANGES_REQUESTED** [review]
This looks awesome @HenryEstberg !! Well done. I did a first pass on code review, but overall looks great. Nice job! 🎉

File: internal/cli/init.go [hunk 1]
```diff
@@ -157,8 +163,32 @@ func (c *InitCommand) Run(args []string) int {
...
 	// If we have no config, initialize a new one.
 	if path == "" {
-		if !c.initNew() {
-			return 1
+		proceed, err := c.ui.Input(&terminal.Input{
+			Prompt: "Do you want help generating a waypoint.hcl file? Type 'yes' to initialize the interactive generator or 'no' to generate a template waypoint.hcl file: ",
+			Style:  "",
+			Secret: false,
+		})
+		if err != nil {
```
paladin-devops, line 171: Should we return an error here?

internal/cli/init.go
```diff
@@ -157,8 +163,32 @@ func (c *InitCommand) Run(args []string) int {
...
+				"Error getting input: %s",
+				clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
+		} else if strings.ToLower(proceed) == "yes" || strings.ToLower(proceed) == "y" {
+			c.ui.Output("Starting interactive .hcl generator.\n")
+			if !c.hclGen() {
+				return 1
+			}
+		} else if strings.ToLower(proceed) == "no" {
```
paladin-devops, line 182: Since we are allowing y/Y for "yes", can we allow n/N for "no" here?

paladin-devops: Actually, is this if/else necessary, if our else does the same thing (generate a template file)?

internal/cli/init.go
```diff
@@ -233,6 +263,614 @@ validate the configuration and initialize your project.
...
+		Optional bool   `json:"Optional"`
+		Category bool   `json:"Category"`
+		EnvVar   string `json:"EnvVar"`
+	} `json:"requiredFields"`
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Type        string `json:"type"`
+}
+
+func (c *InitCommand) hclGen() bool {
```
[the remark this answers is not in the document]
HenryEstberg: Sounds good! My plan is to put all the HCL generation code into its own directory in waypoint/internal, does that sound like a good spot for it?

internal/cli/init.go
```diff
@@ -233,6 +263,614 @@ validate the configuration and initialize your project.
...
+	return nil
+}
+
+func (c *InitCommand) genIndent(outstanding int) string {
+	spaces := ""
+	for i := outstanding; i > 0; i-- {
+		spaces = spaces + "    "
+	}
+	return spaces
+}
```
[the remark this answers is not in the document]
HenryEstberg: I think you're totally right that the function isn't needed in the case that format runs correctly, but I've encountered a case where bad plugin documentation causes the generator to spit out bad HCL syntax that results in the formatter failing. If/when the formatter fails due to bad documentation I think its nice to have a mostly formatted HCL file that you can look at to diagnose the problem, and this is lightweight enough that just having it run every time isn't that much of a drawback.

internal/cli/init.go
HenryEstberg: Good point, I've gotten rid of the extraneous else if.

internal/cli/init.go
HenryEstberg: Yes I think we do want to return an error there.

hclgen/hcl_gen.go
```diff
@@ -0,0 +1,659 @@
+package hclgen
```
xiaolin-ninja, line 1: For organization, this folder should live in the `internal/CLI` folder, rather than in its own top level folder. The file name should also be `hclgen.go`, no underscore. Like `internal/cli/datagen.go`

internal/cli/init.go
```diff
@@ -17,6 +17,7 @@ import (
 
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 
+	hclpkg "github.com/hashicorp/waypoint/hclgen"
```
xiaolin-ninja, line 20: yup, this should be `... waypoint/internal/cli/hclgen` to follow existing patterns, like datagen.

internal/cli/init.go
```diff
@@ -157,8 +158,28 @@ func (c *InitCommand) Run(args []string) int {
...
+		})
+		if err != nil {
+			c.ui.Output(
+				"Error getting input: %s",
+				clierrors.Humanize(err),
+				terminal.WithErrorStyle(),
+			)
 			return 1
+		} else if strings.ToLower(proceed) == "yes" || strings.ToLower(proceed) == "y" {
+			c.ui.Output("Starting interactive .hcl generator.\n")
```
xiaolin-ninja, line 174:
```suggestion
			c.ui.Output("Starting interactive HCL generator.\n")
```

embedJson/embed_json.go
```diff
@@ -0,0 +1,8 @@
+package embedJson
```
xiaolin-ninja, line 1: Why did we pull this into its own package instead of just importing and assigning `embed.FS` in `cli/init`, since that's the only file where it is used?

.circleci/config.yml
```diff
@@ -146,7 +146,7 @@ jobs:
       - run:
           command: git status
       - run: |
-          if ! git diff --exit-code docs/gen > /dev/null; then
+          if ! git diff --exit-code embedJson/gen > /dev/null; then
```
xiaolin-ninja, line 149: Why did we move all the documentation generation out of `docs/gen` and into `embedJson/gen`?

hclgen/hcl_gen.go
```diff
@@ -0,0 +1,659 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+
+	embedJson "github.com/hashicorp/waypoint/embedJson"
```
xiaolin-ninja, line 16: Since this import is only used in this file, I don't think we need to turn it into a package, especially since we also import "embed" on line 4.

xiaolin-ninja (changes requested):
Great work Henry! Thanks for making this happen for Waypoint. :) 
I just have some file placement comments to keep the code base organized.

embedJson/embed_json.go
HenryEstberg: Because of how embedding files works, the embedded variable needs to be in a parent directory of the files that we want to embed, which is why I renamed the parent directory that we store the JSON files in as well.

internal/cli/hclgen/hclgen.go
```diff
@@ -0,0 +1,659 @@
...
+					return m, nil, true
+				}
+				m[field.Field] = fieldInfo{contents: cont, isParent: false}
+			}
+		}
+	}
+	return m, nil, false
+}
+
+func populateField(name string, fType string, ui terminal.UI) (string, error, bool) {
```
briancain, line 375: I think typically in Go `err` is the last thing returned like:

> (string, bool, error)

briancain: We should document the return type here (i.e. what the `bool` is for!)

internal/cli/hclgen/hclgen.go
```diff
@@ -0,0 +1,659 @@
...
+			} else {
+				ui.Output("%s rejected\n", strings.Title(name))
+			}
+		}
+	}
+	return "", nil, true
+}
+
+// <plug> indicates the plugin that the user needs to select. 1: Builder, 2: Deployer/Platform, 3: Releaser, 4: Registry
+func selectPlugin(plug int, fList []string, fSystem embed.FS, ui terminal.UI) (PlugDocs, error, bool) {
```
briancain, line 447: `plug int` could be simplified to be `plug string` and then the case statements would read

`case "builder":` ... `case "releaser":`

embedJson/embed_json.go
```diff
@@ -0,0 +1,8 @@
...
+
+import (
+	"embed"
+)
+
```
briancain, line 6: This file could use a code block comment for what it's for!

briancain (changes requested):

PR #3711, opened 2022-08-19
Docs - Custom plugins in remote ops

paladin-devops:
Update the docs for how to customize the ODR image with a custom plugin, for using custom plugins in remote operations.


demophoon (approved):
Very much needed, thanks for the update

briancain (approved):
Sick!

PR #3720, opened 2022-08-22
make gen/ts

jgwhite:
It’s been a while since we ran this so I can’t quite make sense of the changes. Treating this PR as something of a smoke test.


amyrlam (approved):
Haven't run make gen/ts, but seems a-ok from the README https://github.com/hashicorp/waypoint/tree/main/ui#generating-type-definitions-after-making-api-changes

PR #3730, opened 2022-08-23
Prune old releases

### LINKED ISSUE #2131: Releases should be destroyed when their corresponding deployments are destroyed
When we prune deployments, we do not subsequently prune the attached releases.
So we end up like this:

```
$ waypoint release list
     | ID | DEPLOYMENT ID | PLATFORM |  DETAILS   |    STARTED     |   COMPLETED    | HEALTH  
-----+----+---------------+----------+------------+----------------+----------------+---------
  🚀 |  5 |             5 | aws-ecs  | artifact:4 | 3 seconds ago  | 3 seconds ago  | n/a     
  🚀 |  4 |             4 | aws-ecs  | artifact:3 | 42 seconds ago | 42 seconds ago | n/a     
  🚀 |  3 |             3 | aws-ecs  | artifact:2 | 4 minutes ago  | 4 minutes ago  | n/a     
  🚀 |  2 |             2 | aws-ecs  | artifact:1 | 4 minutes ago  | 4 minutes ago  | n/a     

$ waypoint deployment list
     | ID | PLATFORM |  DETAILS   |    STARTED    |   COMPLETED   |      HEALTH        
-----+----+----------+------------+---------------+---------------+--------------------
  🚀 |  5 | aws-ecs  | artifact:4 | 1 minute ago  | 1 minute ago  | ● - 1 minute ago   
  ✔  |  4 | aws-ecs  | artifact:3 | 2 minutes ago | 2 minutes ago | ● - 2 minutes ago  

```

paladin-devops:
Closes #2131.

### DIFF
--- a/internal/runner/operation_release.go
+++ b/internal/runner/operation_release.go
@@ -155,15 +155,12 @@ func (r *Runner) executeReleaseOp(
 			Application: app.Ref(),
 			Workspace:   project.WorkspaceRef(),
 		})
-		// Set up a slice of releases, whose max length is the # of releases
-		// we have
-		rs := make([]*pb.Release, 0, len(rl.Releases))
-		var rc int
+
+		var rs []*pb.Release
 		for _, release := range rl.Releases {
 			for _, d := range ds {
 				if release.DeploymentId == d.Id {
 					rs = append(rs, release)
-					rc++
 				}
 			}
 
@@ -171,8 +168,8 @@ func (r *Runner) executeReleaseOp(
 
 		log.Info("will prune deploys", "len", len(ds))
 		pruneDeploys = ds
-		log.Info("will prune releases", "len", len(rs[:rc]))
-		pruneReleases = rs[:rc]
+		log.Info("will prune releases", "len", len(rs))
+		pruneReleases = rs
 	}
 
 	// Do the release
--- a/internal/runner/operation_release.go
+++ b/internal/runner/operation_release.go
@@ -171,6 +171,7 @@ func (r *Runner) executeReleaseOp(
 
 		log.Info("will prune deploys", "len", len(ds))
 		pruneDeploys = ds
+		log.Info("will prune releases", "len", len(rs[:rc]))
 		pruneReleases = rs[:rc]
 	}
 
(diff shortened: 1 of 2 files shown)


internal/runner/operation_release.go
```diff
@@ -148,9 +151,27 @@ func (r *Runner) executeReleaseOp(
...
+					rs = append(rs, release)
+					rc++
+				}
+			}
+
+		}
 
 		log.Info("will prune deploys", "len", len(ds))
 		pruneDeploys = ds
+		pruneReleases = rs[:rc]
```
briancain, line 174: Why not set this to use `rs` directly instead? That would match how we build up the deployments slice too. Maybe also add a `log.Info` to match the prune deploy log.

internal/runner/operation_release.go
```diff
@@ -148,9 +151,27 @@ func (r *Runner) executeReleaseOp(
...
+		})
+		// Set up a slice of releases, whose max length is the # of releases
+		// we have
+		rs := make([]*pb.Release, 0, len(rl.Releases))
+		var rc int
+		for _, release := range rl.Releases {
+			for _, d := range ds {
+				if release.DeploymentId == d.Id {
+					rs = append(rs, release)
+					rc++
```
briancain, line 166: We could always use `len(rs)` later rather than capturing its length with this intermediary var!

briancain (changes requested):
Makes sense to me! Blocking on my question for some code simplification

internal/runner/operation_release.go
paladin-devops: I did this because when I `make` the slice `rs`, it's the length of the total # of releases. Since the slice of releases we delete is a subset of our full list of releases, I set `pruneReleases` equal to the subset of the original slice, so that way we don't have empty elements in the `pruneReleases` slice.

The deployments list does something kind of similar, but further up in this file. However in this case, it is assumed that we're only pruning deployments older than 2 deployments ago. 
```go
// If we have less than the prune amount, then we do nothing. Otherwise
// we prune away the ones we're definitely keeping.
if len(resp.Deployments) <= retain {
	log.Debug("less than the limit deployments exists, no pruning")
	resp.Deployments = nil
} else {
	resp.Deployments = resp.Deployments[retain:]
}
```

evanphx (changes requested):

evanphx (approved):

PR #3746, opened 2022-08-25
Runner install - extra arguments

paladin-devops:
Allow extra arguments to be supplied to the `runner agent` command, run by the container installed with `runner install`.

Set the extra arguments on the k8s w/Helm, AWS ECS, Docker and Nomad runner installs.

### DIFF
--- a/.changelog/3746.txt
+++ b/.changelog/3746.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+install/runner: Allow additional arguments for `waypoint runner agent` command
+to be supplied to `waypoint runner install` CLI
+```
--- a/.changelog/3746.txt
+++ b/.changelog/3746.txt
@@ -2,3 +2,7 @@
 install/runner: Allow additional arguments for `waypoint runner agent` command
 to be supplied to `waypoint runner install` CLI
 ```
+```release-note:improvement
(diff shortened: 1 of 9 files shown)


internal/runnerinstall/runnerinstall.go
```diff
@@ -52,6 +52,9 @@ type InstallOpts struct {
 
 	// Unique ID for the runner.
 	Id string
+
+	// Flags which will be supplied to the `waypoint runner agent` command
+	RunnerRunFlags []string
```
krantzinator, line 57: nit: could call this `RunnerAgentFlags` (since `ServerRun` is the command the flags are for in the case of `ServerRunFlags`)

krantzinator (approved):
Works great! 🎉

.changelog/3746.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:improvement
+install/runner: Allow additional arguments for `waypoint runner agent` command
+to be supplied to `waypoint runner install` CLI
+```
```
krantzinator, line 4: Oh you know what, want to add a quick changelog about how we don't show that internal-only label to users in the CLI output?

krantzinator: The changelog is going to be longer than the PR 😂

PR #3754, opened 2022-08-29
builtin/pack: Only show warning if on unsupported arch

briancain:
Prior to this commit, we would warn if *any* architecture didn't match
`amd64`. This is a bit confusing, because the arch reported could be
`x86_64` which is technically supported.

Instead, this commit updates the pack builder to warn if arm
architectures are detected, which are currently unsupported by the
heroku pack builder.

### DIFF
--- a/builtin/pack/builder.go
+++ b/builtin/pack/builder.go
@@ -352,7 +352,7 @@ func (b *Builder) Build(
 	if err != nil {
 		return nil, err
 	}
-	if serverInfo.Architecture != "amd64" {
+	if serverInfo.Architecture != "amd64" && serverInfo.Architecture != "x86_64" {
 		ui.Output(
 			"Warning! Buildpacks are known to have issues on architectures "+
 				"other than amd64. The architecure being reported by the Docker "+


catsby (approved):

builtin/pack/builder.go
```diff
@@ -352,7 +352,7 @@ func (b *Builder) Build(
 	if err != nil {
 		return nil, err
 	}
-	if serverInfo.Architecture != "amd64" {
+	if serverInfo.Architecture == "aarch64" || serverInfo.Architecture == "arm64" {
```
evanphx, line 355: Probably ok unless someone is running buildpack on MIPS? RISCV? Then they won't get the warning.

briancain: Hrm that's fair. Maybe it should be `... != "amd64" || ... != "x86_64"`?

briancain: Updated as such!

evanphx (approved):
Wrote this but then you fixed it:

```suggestion
	if serverInfo.Architecture != "amd64" && serverInfo.Architecture != "x86_64" {
```

De Morgan says it should be and. Confirmed experimenting with: https://go.dev/play/p/FHWlPptEVhZ

PR #3756, opened 2022-08-29
core/runner: Fix server panic when runner is forgotten

### LINKED ISSUE #3448: Waypoint server panics when runner is forgotten before stopped
**Describe the bug**
When a Waypoint runner is forgotten before it is stopped, when that runner is stopped the Waypoint server panics.

I believe this happens because when the runner is forgotten, the runner record is removed from boltdb. Once the Waypoint runner is stopped the server notices and attempts to mark the runner as offline. During that process, server attempts to fetch the runner from BoltDB.

https://github.com/hashicorp/waypoint/blob/822d3ca4252fb4fbf93b29514f3c54ee05605ef3/internal/server/boltdbstate/runner.go#L304-L312

If either `r` is set to `nil` or `codes.NotFound` is the error that is returned, `r` is `nil` by the time we attempt to determine what `Kind` of runner we are dealing with.

https://github.com/hashicorp/waypoint/blob/822d3ca4252fb4fbf93b29514f3c54ee05605ef3/internal/server/boltdbstate/runner.go#L322

In this `r = nil` case the server panics and exits.

**Steps to Reproduce**
Steps to reproduce the behavior.

1. Start a Waypoint server
2. Start a preadopted Waypoint runner
  ```sh
  export WAYPOINT_SERVER_ADDR=localhost:9701
  export WAYPOINT_SERVER_TLS=true
  export WAYPOINT_SERVER_TLS_SKIP_VERIFY=true
  export WAYPOINT_SERVER_TOKEN=$(waypoint user token) # May need to be two commands
  waypoint runner agent -id=testrunner
  ```
3. Forget the Waypoint runner in a new terminal
  ```sh
  waypoint runner forget testrunner
  ```
4. Stop the Waypoint runner

At this point the server will have panicked.

**Expected behavior**
The server should continue as if the runner had always been forgotten in the first place.

* Waypoint CLI Version: v0.8.1
* Waypoint Server Platform and Version: docker v0.8.2, but non-platform specific

demophoon:
Before this commit when Waypoint was determining whether or not to
remove a Runner from boltdb it was possible for runner to be nil at the
time we attempted to determine what type of Runner the Runner was. This
caused the server to panic as soon as the runner became unavailable.

This commit fixes the panic by avoiding the runner from being set to nil
by instead initializing an empty runner variable so that if a runner is
not found the type can still be determined and the runner cleaned up.

Fixes #3448

### DIFF
--- a/internal/server/boltdbstate/runner.go
+++ b/internal/server/boltdbstate/runner.go
@@ -304,8 +304,7 @@ func (s *State) runnerSetAdoptionState(
 func (s *State) runnerOffline(dbTxn *bolt.Tx, memTxn *memdb.Txn, id string) error {
 	r, err := s.runnerById(dbTxn, id)
 	if status.Code(err) == codes.NotFound {
-		r = nil
-		err = nil
+		return nil
 	}
 	if err != nil {
 		return err
--- a/.changelog/3756.txt
+++ b/.changelog/3756.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+core/runner: Server no longer panics when a runner stopped after it is forgotten.
+```


paladin-devops (approved):
🐛 💥

internal/server/boltdbstate/runner.go
```diff
@@ -302,9 +302,10 @@ func (s *State) runnerSetAdoptionState(
 }
 
 func (s *State) runnerOffline(dbTxn *bolt.Tx, memTxn *memdb.Txn, id string) error {
+	var r *pb.Runner
+
 	r, err := s.runnerById(dbTxn, id)
 	if status.Code(err) == codes.NotFound {
-		r = nil
```
briancain, line 307: In the `NotFound` case when r is nil, won't the switch case on 323 still panic because `r.Kind`? Should we be returning here?

demophoon: You are right, I didn't double check to make sure that r.Kind was seeded with a default when initializing the struct and sure enough it doesn't.

Instead of initializing a blank runner struct i've short-circuited and returned nil in `83f900d76` as even if we get a NotFound error there isn't a runner we would need to delete, thus we can skip the rest of the function.

briancain (approved):
Nice! 👍🏻 🎉

PR #3765, opened 2022-08-30
include building the image in `make docker/gen/server`

krantzinator:


briancain (approved):
Looks good!

Not sure if this matters but there's an extra file addition likely from running `make docker/server`.

PR #3772, opened 2022-08-31
`runner install` fixes

### LINKED ISSUE #3681: Runner uninstall hangs w/o region specified
**Describe the bug**
Waypoint runner uninstall hangs infinitely and I cant cancel it if no region is specified.

```
$ waypoint runner uninstall -platform=ecs -id=054890e209a79dc64 
⠴ Uninstalling runner...
❌ Could not get list of ECS services
! Error uninstalling runner: MissingRegion: could not find region configuration^C^C^C^C^C


^C^C^C^C

```

Also, ran into the same err when region was specified. Looks like this is always a problem when the cluster does not exist.
```
$waypoint runner uninstall -platform=ecs -id=054890e209a79dc64 -ecs-region=us-west-2
⠇ Uninstalling runner...
❌ Could not get list of ECS services
! Error uninstalling runner: ClusterNotFoundException: 
```

**Expected behavior**
I would expect an err message saying to specify a region, rather than just hang.

applicable:

* Waypoint CLI Version:
CLI: v0.9.1 (9562f453c)

krantzinator:
Fixes https://github.com/hashicorp/waypoint/issues/3681

This is largely a code cleanup PR. The main issue was that if I installed a runner on Cluster A, then switched my kube context to Cluster B, and then tried to uninstall that runner, I would get the waypoint error message "No runners installed" because the code path would go through the `k8sUninstall` function.
This improvement first checks if the runner is on the currently-active cluster before choosing the uninstall method, and returning the appropriate error message.

Per the comments in this PR:
```
	// Our checks here follow the logic of:
 	// Up until v0.8.2, we installed runners with the k8s client,
 	// and the Label was "app=waypoint-runner" and the Name "waypoint-runner-random-id"
 	// As of 0.9.0, we install runners with helm, with a Label following the
 	// pattern ("app.kubernetes.io/instance=waypoint-%s", runnerId)
 	// and the Name ("waypoint-"+strings.ToLower(runnerId))
 	//
 	// Therefore we need to ascertain A) if the runner exists on the cluster at
 	// all (it might not be if the user is auth'd to the wrong cluster), and then B)
 	// if the name/label matches the Helm pattern or the k8s client pattern so
 	// we know how to uninstall it
```

Previous to this update, if the runner wasn't found on the cluster, we'd still remove the runner from Waypoint and then tell the user that no runners were installed:
```
❯ wd runner uninstall -platform=kubernetes -id=01GBD70R1TYVKH5T41HHYE9D49
✓ Runner "01GBD70R1TYVKH5T41HHYE9D49" uninstalled successfully
✓ Runner "01GBD70R1TYVKH5T41HHYE9D49" forgotten on server
✓ No runners installed.
```

After this update, the error messages are more specific and timely (we don't list the specific context because if the user hasn't set it with the `-k8s-context` flag then we don't have it; but we include this note just as a troubleshooting hint for the user):
```
❯ wd runner uninstall -platform=kubernetes -id=01GBTCZCF74K88V3DBB1MGXR25
❌ Unable to uninstall runner
! runner with ID "01GBTCZCF74K88V3DBB1MGXR25" not found in namespace "default"
  with current context
```

### DIFF
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -137,7 +137,7 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 		Id:         c.id,
 	})
 	if err != nil {
-		s.Update("Unable to uninstall runner", terminal.StatusError)
+		s.Update("Unable to uninstall runner")
 		s.Abort()
 		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 		return 1
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -128,6 +128,8 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 	defer sg.Wait()
 
 	s := sg.Add("Uninstalling runner...")
+	defer func() { s.Abort() }()
+
 	err := p.Uninstall(ctx, &runnerinstall.InstallOpts{
 		Log:        log,
 		UI:         c.ui,
@@ -135,9 +137,10 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 		Id:         c.id,
 	})
 	if err != nil {
-		c.ui.Output("Error uninstalling runner: %s", clierrors.Humanize(err),
-			terminal.WithErrorStyle(),
-		)
+		s.Update("Error uninstalling runner")
+		s.Status(terminal.StatusError)
+		s.Done()
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 		return 1
 	}
 	s.Update("Runner %q uninstalled successfully", c.id)
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -137,6 +137,8 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 		Id:         c.id,
 	})
 	if err != nil {
+		s.Update("Unable to uninstall runner", terminal.StatusError)
+		s.Abort()
 		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 		return 1
 	}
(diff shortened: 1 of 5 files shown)


internal/cli/runner_uninstall.go
```diff
@@ -128,16 +128,18 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
...
 		Log:        log,
 		UI:         c.ui,
 		ServerAddr: c.serverUrl,
 		Id:         c.id,
 	})
 	if err != nil {
-		c.ui.Output("Error uninstalling runner: %s", clierrors.Humanize(err),
-			terminal.WithErrorStyle(),
-		)
+		s.Update("Unable to uninstall runner")
```
paladin-devops, line 140: I think this will fix #3681 !

internal/runnerinstall/k8s.go
```diff
@@ -287,39 +288,64 @@ func (i *K8sRunnerInstaller) InstallFlags(set *flag.Set) {
...
 
-	for _, release := range releases {
-		opts.Log.Debug(fmt.Sprintf("Releases %+v", release.Name))
-		if release.Name == "waypoint-"+strings.ToLower(opts.Id) {
-			isHelm = true
-			break
-		}
+	// Search for runner with 0.9+ tag format, installed with helm
+	podClient := clientset.CoreV1().Pods(i.Config.Namespace)
+	listHelmClient, err := podClient.List(ctx, metav1.ListOptions{
```
paladin-devops, line 322: I was wondering why this is named `listHelmClient`, but I realize now that this is looking for pods which were installed via Helm. It made me think that this was somehow used for a programmatic version of `helm list` 😄

paladin-devops (approved):
🏃🏻‍♀️

internal/runnerinstall/k8s.go
krantzinator: Ahhh yes my brain was a victim to precedence. The original variable here was `list` so then I made `listK8s` and `listHelm`. Those are confusing names! I can update them to `helmClientList` and `k8sClientList`.

PR #3785, opened 2022-09-01
config/pipeline: Auto-prune deployments on release by default

### LINKED ISSUE #3779: pipelines: Update the `release` built-in step to prune by default like `waypoint up` or `waypoint release`.

Currently, if you run a release built-in step, it's not configured to auto-prune old deployments in the same way `waypoint up` is. This is likely a behavior we should preserve.

Update the built-in release step to auto-prune by default, with the option to disable on request.

**Additional context**

briancain:
This commit updates a built-in step Release to auto-prune deployments by
default. This matches the behavior of our CLIs `waypoint deploy` and
`waypoint release`.

Fixes #3779

### DIFF
--- a/internal/config/pipeline.go
+++ b/internal/config/pipeline.go
@@ -305,7 +305,7 @@ func (c *Config) buildPipelineProto(pl *hclPipeline) ([]*pb.Pipeline, error) {
 		case "release":
 			var releaseBody struct {
 				DeploymentRef       uint64 `hcl:"deployment_ref,optional"` // 0 or "unset" means latest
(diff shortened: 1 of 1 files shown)


krantzinator (approved):
Looks simple enough. For my own eddification, curious why we need all the pointers?

briancain:
@krantzinator - the `*bool` means if users don't specify prune, it will show up parsed as `nil` which means we can interpret it as unset and use a default which is true. If it was a straight `bool` it means unset parses as `false`, which we can't differentiate between them setting it to `prune = false` or not setting it. Mostly a work around for this!

Issue #3782, opened 2022-09-01
UI freezes occasionally due to missing `resource_id`


krantzinator:
**Describe the bug**
Not sure if this bug is more in the UI, or more in the plugin handling of its status reports. Maybe a bit of both. And maybe the UI could be more resilient around handling missing status reports?

~Basically, every now and then my latest deployment goes from `Available` to `Missing`, causing my UI to completely freeze up.~ Actually, now it is frozen even with `Available` status so idk 😢 

https://user-images.githubusercontent.com/8461333/187955212-5cff4195-9547-46ec-9140-5a1f385f0d4a.mov

After a few minutes, the status resolved to `Available` and then I could navigate the UI again. But then a few minutes later, we were back to `Missing` and frozen yet again.

**=> Update:**
Better error message output now from the UI:
```
Uncaught (in promise) Object
Could not pass stateJson for resource Object
Error: You must provide a param `resource_id`.
```

**Steps to Reproduce**
- install server on AWS EKS
- install runner
- deploy app remotely via gitops
- access the UI

**Expected behavior**
The UI not to freeze up. Also, the status report to not say "Missing," maybe?

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:
CLI: v0.9.0-592-ge534c347f (e534c347f+CHANGES)
Server: v0.9.1 `kubernetes` AWS EKS
* Waypoint Plugin: `pack`, `kubernetes`
* Browser: Safari Version 15.6.1 (17613.3.9.1.16)

**Additional context**
```
variable "registry_username" {
  type = string
  default = ""
  env = ["REGISTRY_USERNAME"]
}

variable "registry_password" {
  type = string
  sensitive = true
  default = ""
  env = ["REGISTRY_PASSWORD"]
}

variable "registry_imagename" {
  type = string
  default = ""
  env = ["REGISTRY_IMAGENAME"]
}

project = "nodejs-example"

app "nodejs-example" {
  runner {
    profile = "01GBWN6ZBQP6GH4R4GGWRG9W8M"
  }
  labels = {
    "service" = "example-nodejs",
    "env"     = "dev"
  }

  build {
    use "pack" {}
    registry {
      use "docker" {
        image = "${var.registry_username}/${var.registry_imagename}"
        tag   = "${gitrefpretty()}"
        username = var.registry_username
        password = var.registry_password
        local = false
      }
    }
  }

  deploy {
    use "kubernetes" {
      probe_path = "/"
    }
[...]

PR #3786, opened 2022-09-01
ui: make `<ContainerImageTag>` JSON parsing more resilient

jgwhite:
This is a guess at a fix. A deeper fix would involve inspect the state JSON that caused the UI crash. This change will at least log the rogue resource data out in the console.

Addresses #3782


krantzinator (approved):
lgtm

Issue #3791, opened 2022-09-01
pipeline list lists pipelines for a different project


xiaolin-ninja:
**Describe the bug**
```
❯ waypoint-dev status
Current project statuses in server context "127.0.0.1:9731"

PROJECT     	WORKSPACE	DEPLOYMENT STATUSES	RELEASE STATUSES 
example-java	default  	1 READY            	N/A             	
k8s-tetris  	default  	N/A                	N/A             	
```

```
❯ waypoint-dev pipeline list -project=k8s-tetris

» Waypoint Pipelines for k8s-tetris
              ID             | NAME |    OWNER     | CURRENT STEPS | LAST RUN STARTED | LAST RUN COMPLETED |  STATE  | TOTAL RUNS  
-----------------------------+------+--------------+---------------+------------------+--------------------+---------+-------------
  01GBBBC63TR5DYQXQVGJPS8JGZ | test | example-java |             2 | 3 hours ago      | 3 hours ago        | success |          2  
```

**Steps to Reproduce**
1. Make two projects.
2. `waypoint pipeline list -project=A`
3. Observe.

**Expected behavior**
`-project` flag should filter and only list pipelines for the specified project.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version:
* Waypoint Server Platform and Version: (like `docker`, `nomad`, `kubernetes`)
* Waypoint Plugin: (like `aws/ecs`, `pack`, `azure`)

**Additional context**

briancain:
Fixed via #3792 !

PR #3792, opened 2022-09-01
Fix pipeline list project scoping

xiaolin-ninja:
https://github.com/hashicorp/waypoint/issues/3791

Fixes issue where `waypoint pipeline list -project` wasn't actually scoping to the specified project.

### DIFF
--- a/internal/server/boltdbstate/pipeline.go
+++ b/internal/server/boltdbstate/pipeline.go
@@ -213,6 +213,11 @@ func (s *State) PipelineList(pRef *pb.Ref_Project) ([]*pb.Pipeline, error) {
 			if err != nil {
 				return err
 			}
+			if pRef != nil {
+				if val.Owner.(*pb.Pipeline_Project).Project.Project != pRef.Project {
+					continue
+				}
+			}
 
 			out = append(out, val)
 		}
--- a/internal/server/boltdbstate/pipeline.go
+++ b/internal/server/boltdbstate/pipeline.go
@@ -213,12 +213,6 @@ func (s *State) PipelineList(pRef *pb.Ref_Project) ([]*pb.Pipeline, error) {
 			if err != nil {
 				return err
 			}
-			if pRef != nil {
-				if val.Owner.(*pb.Pipeline_Project).Project.Project != pRef.Project {
-					continue
-				}
-			}
-
 			out = append(out, val)
 		}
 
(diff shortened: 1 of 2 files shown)


internal/server/boltdbstate/pipeline.go
```diff
@@ -213,6 +213,11 @@ func (s *State) PipelineList(pRef *pb.Ref_Project) ([]*pb.Pipeline, error) {
 			if err != nil {
 				return err
 			}
+			if pRef != nil {
+				if val.Owner.(*pb.Pipeline_Project).Project.Project != pRef.Project {
+					continue
+				}
+			}
```
briancain, line 220: This makes sense, but should we instead make `project` an index on pipelines (it might be already) and fix the query in lower case `pipelineList` so it'll only return a list of pipelines scoped to the `Ref_Project`? 🤔

briancain (commented):
Nice! A lil' suggestion, and we should include some unit tests to validate this fix

internal/server/boltdbstate/pipeline.go
xiaolin-ninja: This pattern actually matches the pattern in JobList, where we just grab all the jobs and then filter out the ones we don't need.

xiaolin-ninja: So we definitely could add a `listPipelineByProject` method, but I wouldn't want to add filtering into the db method since I think those should be kept as clean and single-purpose as possible. Also keeping things consistent in the codebase is good :)

briancain: Oh yeah, I was mostly meaning make `PipelineList` _only_ ever scoped to a project since it's already accepting a project ref in its parameters. We don't list pipelines across all projects right now so I was assuming that's how it should work! I wouldn't add another method though at this time. The index query would be faster than a loop too. We do already index on Project it looks like too so should be an easy add! 

https://github.com/hashicorp/waypoint/blob/7a1f832cf8a6cc8e3eac869ce9a135dbd4bef1cc/internal/server/boltdbstate/pipeline.go#L302-L310

xiaolin-ninja: Oh I see what you mean, and we could add a `PipelineListAll` in the future if we ever want cross-project pipeline listing.

internal/server/boltdbstate/pipeline.go
```diff
@@ -230,7 +229,7 @@ func (s *State) pipelineList(
 	memTxn *memdb.Txn,
 	ref *pb.Ref_Project,
 ) ([]*pb.Ref_Pipeline, error) {
-	iter, err := memTxn.Get(pipelineIndexTableName, pipelineIndexId+"_prefix", "")
+	iter, err := memTxn.Get(pipelineIndexTableName, pipelineIndexProjectId, ref.Project)
```
briancain, line 232: Yasss, exactly! Perfect

briancain (approved):
Look great!! 👏🏻

internal/server/boltdbstate/pipeline.go
briancain: Yep, exactly!

PR #3796, opened 2022-09-01
Fix installing runner without specifying labels

izaaklauer:
This fixes a bug where if you tried a runner install with no targeting, it would attempt to target by labels but not include any actual labels, resulting in an invalid runner profile.

This fixes the cli, and also introduces a new validator on the server for the condition.


jgwhite (approved):
Looks good to me!

PR #3797, opened 2022-09-01
`WatchTask` for Nomad Task Launcher

### LINKED ISSUE #3714: Implement WatchTask for Nomad
See parent issue for more information!

paladin-devops:
Closes #3714.

The task launcher Nomad plugin will use the Nomad API to stream logs from the Nomad allocation which is running the Waypoint on-demand runner container. These logs will be available in the log output of the static runner. This means that long after the ODR job is purged from Nomad, the ODR's logs are consumable from the static runner, which runs indefinitely, making troubleshooting of remote operations easier since previously, the ODR logs were unavailable (unless the operator quickly got the ODR's logs with `nomad logs` or in the Nomad UI before it was purged).

### DIFF
--- a/.changelog/3797.txt
+++ b/.changelog/3797.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/nomad: Store Nomad on-demand runner logs in the job system
+```
--- a/.changelog/3797.txt
+++ b/.changelog/3797.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-plugin/nomad: Store Nomad on-demand runner logs in the job system
+plugin/nomad: Implement WatchTask plugin for Nomad task launcher. Store Nomad on-demand runner logs in the job system.
 ```
--- a/builtin/nomad/task.go
+++ b/builtin/nomad/task.go
@@ -317,18 +317,12 @@ func (p *TaskLauncher) WatchTask(
 				}
 				task := tg.Tasks[0]
 
-				var state string
-				allocTask, ok := alloc.TaskStates[task.Name]
-				if !ok {
-					return nil, errors.New("ODR task not in alloc")
-				}
-				state = allocTask.State
-
 				// We'll give the ODR 5 minutes to start up
 				// TODO: Make this configurable
 				ctx, cancel := context.WithTimeout(ctx, time.Minute*time.Duration(5))
 				defer cancel()
 				ticker := time.NewTicker(5 * time.Second)
+				state := "pending"
 				for state == "pending" {
 					select {
 					case <-ticker.C:
@@ -339,7 +333,11 @@ func (p *TaskLauncher) WatchTask(
 						log.Error("Failed to get info for alloc "+allocs[0].ID+". Error: %s", err.Error())
 						return nil, err
 					} else {
-						state = alloc.TaskStates[task.Name].State
+						allocTask, ok := alloc.TaskStates[task.Name]
+						if !ok {
+							return nil, errors.New("ODR task not in alloc")
+						}
+						state = allocTask.State
 					}
 				}
 
@@ -352,15 +350,21 @@ func (p *TaskLauncher) WatchTask(
 				log.Debug("Getting logs for alloc: " + alloc.Name + ", task: " + task.Name)
 				ch := make(chan struct{})
 				logStream, errChan := client.AllocFS().Logs(alloc, follow, task.Name, "stderr", "", 0, ch, queryOpts)
+			READ_LOGS:
+				for {
+					select {
+					case data := <-logStream:
+						if data == nil {
+							break READ_LOGS
+						}
+						message := string(data.Data)
+						log.Info(message)
+						ui.Output(message)
+					case err := <-errChan:
+						log.Error("Error reading logs from alloc: %q", err.Error())
+						return nil, err
 
-				select {
-				case err := <-errChan:
-					log.Error("Error reading logs from alloc: %q", err.Error())
-					return nil, err
-				case data := <-logStream:
-					message := string(data.Data)
-					log.Info(message)
-					ui.Output(message)
+					}
 				}
 			}
 		}
--- a/builtin/nomad/task.go
+++ b/builtin/nomad/task.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"strings"
 	"time"
 
@@ -298,12 +300,15 @@ func (p *TaskLauncher) WatchTask(
 		return nil, err
 	} else {
 		if allocs, _, err := client.Jobs().Allocations(ti.Id, true, queryOpts); err != nil {
+			log.Error("Failed to get allocations for ODR job: %s", ti.Id)
 			return nil, err
 		} else {
 			if len(allocs) != 1 {
+				log.Error("Invalid # of allocs for ODR job.")
 				return nil, errors.New("there should be one allocation in the job")
 			}
 			if alloc, _, err := client.Allocations().Info(allocs[0].ID, queryOpts); err != nil {
+				log.Error("Failed to get info for alloc "+allocs[0].ID+". Error: %s", err.Error())
 				return nil, err
 			} else {
 				tg := alloc.GetTaskGroup()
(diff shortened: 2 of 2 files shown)


builtin/nomad/task.go
```diff
@@ -286,9 +288,90 @@ func (p *TaskLauncher) StartTask(
...
+				return nil, err
+			} else {
+				tg := alloc.GetTaskGroup()
+				if len(tg.Tasks) != 1 {
+					return nil, errors.New("there should be one task in the allocation")
+				}
+				task := tg.Tasks[0]
+
+				// We'll give the ODR 5 minutes to start up
+				// TODO: Make this configurable
```
paladin-devops, line 322: I hadn't planned on including this in the PR, but will give it a try!

builtin/nomad/task.go
```diff
@@ -286,9 +288,90 @@ func (p *TaskLauncher) StartTask(
...
-	return nil, status.Errorf(codes.Unimplemented, "WatchTask not implemented")
+	// We'll query for the allocation in the namespace of our task launcher
+	queryOpts := &api.QueryOptions{Namespace: p.config.Namespace}
+
+	// Accumulate our result on this
+	var result component.TaskResult
+
+	if client, err := getNomadClient(); err != nil {
+		return nil, err
+	} else {
```
demophoon, line 302: We might not need to nest if statements since it looks like we are returning on failure in many of them.

.changelog/3797.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+plugin/nomad: Store Nomad on-demand runner logs in the job system
```
briancain, line 2:
```suggestion
plugin/nomad: Implement WatchTask plugin for Nomad task launcher. Store Nomad on-demand runner logs in the job system.
```

builtin/nomad/task.go
briancain: Agreed - this should be refactored to remove the elses. It'll simplify this function by a lot and remove the extra nesting.

builtin/nomad/task.go
```diff
@@ -286,9 +288,90 @@ func (p *TaskLauncher) StartTask(
...
+				log.Error("Invalid # of allocs for ODR job.")
+				return nil, errors.New("there should be one allocation in the job")
+			}
+			if alloc, _, err := client.Allocations().Info(allocs[0].ID, queryOpts); err != nil {
+				log.Error("Failed to get info for alloc "+allocs[0].ID+". Error: %s", err.Error())
+				return nil, err
+			} else {
+				tg := alloc.GetTaskGroup()
+				if len(tg.Tasks) != 1 {
+					return nil, errors.New("there should be one task in the allocation")
```
briancain, line 317: Let's make these `status.Error` errors to be consistent with our plugins!

builtin/nomad/task.go
```diff
@@ -286,9 +288,90 @@ func (p *TaskLauncher) StartTask(
...
+
+				log.Debug("Getting logs for alloc: " + alloc.Name + ", task: " + task.Name)
+				ch := make(chan struct{})
+				logStream, errChan := client.AllocFS().Logs(alloc, follow, task.Name, "stderr", "", 0, ch, queryOpts)
+			READ_LOGS:
+				for {
+					select {
+					case data := <-logStream:
+						if data == nil {
+							break READ_LOGS
```
briancain, line 359: I think in this case we would just return? Should we also be checking if the watched resource has exited? It's possible I imagine for the stream to return nil but there might be more logs and we'd accidentally end early with this right now?

briancain (changes requested):

builtin/nomad/task.go
paladin-devops: @briancain as the code is written right now, it appears that the `WatchTask` job is stuck in the running state. This is likely because with this goto statement, at the conclusion of reading the logs, it's re-entering the for loop, hitting the goto again, and never returning. I'm going to update this to query the Nomad API to check if the task's state is not "running", and return if that is the case.

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+		return nil, err
+	}
+	allocs, _, err := client.Jobs().Allocations(ti.Id, true, queryOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(allocs) != 1 {
+		log.Error("Invalid # of allocs for ODR job.")
+		return nil, status.Error(codes.Internal, "there should be one allocation in the job")
```
briancain, line 309: I'd include the length here too for extra debug information in both the log an error:

> log.Error("Invalid ...", "total_allocs", len(allocs))

For example

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+	}
+	alloc, _, err := client.Allocations().Info(allocs[0].ID, queryOpts)
+	if err != nil {
+		log.Error("Failed to get info for alloc "+allocs[0].ID+". Error: %s", err.Error())
```
briancain, line 313:
```suggestion
		log.Error("Failed to get info for alloc.", "alloc_id", allocs[0].ID, "err", err.Error())
```

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+	state := "pending"
+	for state == "pending" {
+		select {
+		case <-ticker.C:
+		case <-ctx.Done(): // cancelled
+			return nil, status.Errorf(codes.Aborted, "Context cancelled from timeout waiting for ODR task to start %s", ctx.Err())
+		}
+		alloc, _, err := client.Allocations().Info(allocs[0].ID, queryOpts)
+		if err != nil {
+			log.Error("Failed to get info for alloc "+allocs[0].ID+". Error: %s", err.Error())
```
briancain, line 336:
```suggestion
		log.Error("Failed to get info for alloc waiting for task to start", "alloc_id", allocs[0].ID, "err", err.Error())
```

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+	// TODO: Make this configurable
+	ctx, cancel := context.WithTimeout(ctx, time.Minute*time.Duration(5))
+	defer cancel()
+	ticker := time.NewTicker(5 * time.Second)
+	state := "pending"
+	for state == "pending" {
+		select {
+		case <-ticker.C:
+		case <-ctx.Done(): // cancelled
+			return nil, status.Errorf(codes.Aborted, "Context cancelled from timeout waiting for ODR task to start %s", ctx.Err())
```
briancain, line 332:
```suggestion
			return nil, status.Errorf(codes.Aborted, "Context cancelled from timeout waiting for ODR task to start: %s", ctx.Err())
```

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+		state = allocTask.State
+	}
+
+	// Only follow the logs if our task is still alive
+	follow := true
+	if state == "dead" {
+		follow = false
+	}
+
+	log.Debug("Getting logs for alloc: " + alloc.Name + ", task: " + task.Name)
```
briancain, line 352:
```suggestion
	log.Debug("Getting logs for alloc", "alloc_name", alloc.Name, "task_name", task.Name)
```

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+	logStream, errChan := client.AllocFS().Logs(alloc, follow, task.Name, "stderr", "", 0, ch, queryOpts)
+READ_LOGS:
+	for {
+		select {
+		case data := <-logStream:
+			if data == nil {
+				// check if task is dead, if it is dead, return
+				alloc, _, err := client.Allocations().Info(allocs[0].ID, queryOpts)
+				if err != nil {
+					log.Error("Failed to get info for alloc "+allocs[0].ID+". Error: %s", err.Error())
```
briancain, line 363:
```suggestion
		log.Error("Failed to get info for alloc to stream logs", "alloc_id", allocs[0].ID, "err", err.Error())
```

builtin/nomad/task.go
```diff
@@ -286,9 +287,106 @@ func (p *TaskLauncher) StartTask(
...
+					return &result, nil
+				}
+
+				break READ_LOGS
+			}
+			message := string(data.Data)
+			log.Info(message)
+			ui.Output(message)
+		case err := <-errChan:
+			log.Error("Error reading logs from alloc: %q", err.Error())
```
briancain, line 383:
```suggestion
			log.Error("Error reading logs from alloc", "err", err.Error())
```

briancain (changes requested):
Code looks pretty good now! I've got some suggestions for making the logging fit hclog's key=>value param style, otherwise I think the code makes sense! 🎉

briancain (approved):
Looks good! 🚀

builtin/nomad/task.go
paladin-devops: Will add in future PR!

PR #3799, opened 2022-09-02
ui: bump google-protobuf to 3.21.0

jgwhite:
## Why the change?

We recently (#3720) regenerated the waypoint-pb (protobuf types) and waypoint-client (grpc-web client) libraries using protoc 3.17.3. It seems like this latest generated code requires a more up-to-date version of the google-protobuf dependency.

## How do I test this?

1. Check out the branch
2. `(cd ui && make build)`
3. `make static-assets`
4. `make bin`
5. `./waypoint server run -accept-tos`
6. Run the bootstrap command it gives you
7. `./waypoint ui -authenticate`
8. Verify you see no errors in the JS console
9. Verify the version info shows up in the footer

### DIFF
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -2437,10 +2437,10 @@
     "@types/minimatch" "*"
     "@types/node" "*"
 
-"@types/google-protobuf@^3.7.2":
-  version "3.7.2"
-  resolved "https://registry.yarnpkg.com/@types/google-protobuf/-/google-protobuf-3.7.2.tgz#cd8a360c193ce4d672575a20a79f49ba036d38d2"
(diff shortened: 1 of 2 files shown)


ui/yarn.lock
```diff
@@ -11722,7 +11724,8 @@ onetime@^5.1.0, onetime@^5.1.2:
     mimic-fn "^2.1.0"
 
 "opaqueany@link:./lib/opaqueany":
-  version "1.0.0"
+  version "0.0.0"
```
krantzinator, line 11727: 😓

krantzinator (approved):
let's try it

ui/yarn.lock
jgwhite: ah that old chestnut

gregone (approved):
LGTM

Issue #3481, opened 2022-06-22
`runner profile delete` should use `name` and not `id`


krantzinator:
Once #3379 is fixed and runner profiles are guaranteed to have unique names, we should update `runner profile delete` (to be released as of this issue creation) to require the `name` and not the `id`, as getting the id is cumbersome.

**Describe the bug**
```
❯ waypoint runner profile delete -name=docker
! flag provided but not defined: -name

 ❯ waypoint runner profile list
Runner profiles
   NAME  | PLUGIN TYPE |            OCI URL            | TARGET RUNNER | DEFAULT
---------+-------------+-------------------------------+---------------+----------
  docker | docker      | hashicorp/waypoint-odr:latest | *             | yes
  docker | docker      | hashicorp/waypoint-odr:latest | *             | yes
```

* Waypoint CLI Version: upcoming release

tharun208:
@krantzinator I am new to waypoint and I like to contribute to this

briancain:
We should update the dbstate package to accept name on delete. We can have it index on a profile name and do the lookup that way, and return an error if there are two or more profiles by the same name and ask the user to specify `id` instead.

PR #3803, opened 2022-09-02
Delete runner profiles by name

izaaklauer:
Closes https://github.com/hashicorp/waypoint/issues/3481

This allows runner profiles to be deleted by name, not ID. We don't surface IDs to users, so it's not useful to delete them by ID.

I also found that we were able to create multiple runner profiles with the same name. This PR fixes that, adds a test around it, and still allows users to delete runner profiles by `id` (via a flag) if they find themselves with duplicate-named runners as an emergency backdoor.

### DIFF
--- a/.changelog/3803.txt
+++ b/.changelog/3803.txt
@@ -1,0 +1,7 @@
+```release-note:bug
+cli: `runner profile delete` deletes by name rather than the user-invisible ID.
+```
+
+```release-note:bug
+cli: `runner profile apply` does not create entries with duplicate names. 
+```
--- a/.changelog/3803.txt
+++ b/.changelog/3803.txt
@@ -3,5 +3,5 @@ cli: `runner profile delete` deletes by name rather than the user-invisible ID.
 ```
 
 ```release-note:bug
-cli: `runner profile apply` does not create entries with duplicate names. 
+cli: `runner profile set` does not create entries with duplicate names. 
 ```
--- a/internal/cli/runner_profile_delete.go
+++ b/internal/cli/runner_profile_delete.go
@@ -1,17 +1,26 @@
 package cli
 
 import (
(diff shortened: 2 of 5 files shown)


krantzinator:
`runner list` returns ID and not Name -- should we fix that as part of this for consistency?

.changelog/3803.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
+cli: `runner profile delete` deletes by name rather than the user-invisible ID.
+```
+
+```release-note:bug
+cli: `runner profile apply` does not create entries with duplicate names.
```
paladin-devops, line 6:
```suggestion
cli: `runner profile set` does not create entries with duplicate names. 
```

internal/cli/runner_profile_delete.go
```diff
@@ -1,17 +1,26 @@
 package cli
 
 import (
+	"github.com/posener/complete"
```
paladin-devops, line 4: It was recently suggested to me to not re-order the import statements, so I'm passing that along here! 😄

paladin-devops (approved):

internal/cli/runner_profile_delete.go
izaaklauer: Ha! We talked about this once. My memory was that we decided we have three groupings:

```
stdlib

external packages

internal packages
```

I'll put this on the team discussion list.

izaaklauer:
> `runner list` returns ID and not Name -- should we fix that as part of this for consistency?

Runners are weird - they have IDs, but the ID can be determined by the user, so it's really more of a name. Maybe the right thing to do eventually would be to rename the runner's `id` field to `name`, but I don't think i'm up for that as part of this change.

PR #3804, opened 2022-09-02
Fix Nomad upgrade of runner

paladin-devops:
`server upgrade`s for Nomad succeeded in upgrading the server, but failed to detect the runner because the old name for a runner's Nomad job, "waypoint-runner" was still being used. This PR corrects that by searching for a runner with the old name from before 0.9, "waypoint-runner", or the new name from after 0.9, "waypoint-static-runner".


krantzinator (approved):
Just needs a changelog!

briancain (approved):
+1 to changelog!

PR #3805, opened 2022-09-02
WIP: workspace scoped step jobs

catsby:
This PR updates how steps are used to create jobs, by reading the new workspace reference and updating the job definition to use that workspace. 

This PR is a continuation of the work started in https://github.com/hashicorp/waypoint/pull/3802 and is currently built off of and targeting that branch.

Currently Waypoint only allows a single layer of nesting pipelines. If the `step` that embeds a pipeline has a `workspace` set, that workspace will be set on all the embedded jobs that get created.

### DIFF
--- a/pkg/server/singleprocess/service_pipeline.go
+++ b/pkg/server/singleprocess/service_pipeline.go
@@ -390,12 +390,21 @@ func (s *Service) buildStepJobs(
 			// step's jobs aren't scheduled prior to any dependencies.
 			parentStepDep := map[string][]string{pipeline.Id: job.DependsOn}
 
-			// TODO: add the parent step workspace ref. Note this only works now
-			// because we are limiting nested pipelines to 1 layer.
 			embedJobs, embedRun, embedStepIds, err := s.buildStepJobs(ctx, log, req,
 				visitedPipelines, nodeStepRef, parentStepDep, embeddedPipeline, pipelineRun)
 			if err != nil {
 				return nil, nil, nil, err
+			}
+
+			// Add the parent step workspace ref.
+			// **Note** this only works now because we are limiting nested
+			// pipelines to 1 layer.
+			if step.Workspace != nil {
+				for _, jobReq := range embedJobs {
+					jobReq.Job.Workspace = &pb.Ref_Workspace{
+						Workspace: step.Workspace.Workspace,
+					}
+				}
 			}
 
 			// add the nested jobs
--- a/pkg/server/singleprocess/service_pipeline.go
+++ b/pkg/server/singleprocess/service_pipeline.go
@@ -298,6 +298,12 @@ func (s *Service) buildStepJobs(
 			RunSequence:  pipelineRun.Sequence,
 		}
 
+		if step.Workspace != nil {
+			job.Workspace = &pb.Ref_Workspace{
+				Workspace: step.Workspace.Workspace,
+			}
+		}
+
 		// Queue the right job depending on the Step type. We will queue a Waypoint
 		// operation if the type is a reserved built in step.
 		switch o := step.Kind.(type) {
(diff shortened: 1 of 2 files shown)


izaaklauer (approved):
looks reasonable to me

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -390,6 +396,17 @@ func (s *Service) buildStepJobs(
 				return nil, nil, nil, err
 			}
 
+			// Add the parent step workspace ref.
+			// **Note** this only works now because we are limiting nested
+			// pipelines to 1 layer.
+			if step.Workspace != nil {
+				for _, jobReq := range embedJobs {
+					jobReq.Job.Workspace = &pb.Ref_Workspace{
+						Workspace: step.Workspace.Workspace,
```
briancain, line 405: This makes sense, but should we also respect any nested step workspace options too? Right now this will override any workspace options set in the embedded pipeline steps like:

```hcl
pipeline "nested" {
  step "another-one" {
    workspace = "test"
    pipeline "two" {
      step "different" {
        workspace = "different"
        // ...
      }
    }
  }
}
```

I think we should preserve that step workspace if it was set explicitly, otherwise this Step Pipeline Ref could be seen as a default for any steps not explicitly set in the embedded pipeline.

pkg/server/singleprocess/service_pipeline.go
```diff
@@ -390,6 +396,17 @@ func (s *Service) buildStepJobs(
 				return nil, nil, nil, err
 			}
 
+			// Add the parent step workspace ref.
+			// **Note** this only works now because we are limiting nested
+			// pipelines to 1 layer.
```
briancain, line 401: A quick note about this - these functions are written in such a way that embedded pipelines could be infinitely nested. The 1 layer of pipeline stanza nesting is mostly an artificial limitation of our parser, but users could still use pipeline step refs to make more complicated references that go beyond a single layer. Since this function is recursive I think this approach will work regardless of the limit, so maybe the comment could be something like:

```suggestion
			// Add the parent step workspace ref and apply it to all embedded
			// pipeline job templates if step was not configured with a
			// workspace ref.
```

briancain (commented):
Approach looks solid, I added some feedback! Didn't do a test since it's in draft still but looking good so far, and the unit tests look great too!

pkg/server/singleprocess/service_pipeline.go
catsby: Sounds reasonable, thanks

pkg/server/singleprocess/service_pipeline.go
catsby: Agreed, this is the odd "bug" I mentioned in chat earlier. I've updated the logic and test code in https://github.com/hashicorp/waypoint/pull/3805/commits/ecf2e7d6259bb1d7d5f9bc04f4867522187fd476 to address this

pkg/server/singleprocess/service_pipeline_test.go
```diff
@@ -333,6 +485,167 @@ func TestServicePipeline_Run(t *testing.T) {
...
+		// We should have all the job IDs
+		require.Len(resp.JobMap, 7)
+		require.Len(resp.AllJobIds, 7)
+
+		// find our embedded jobs and verify they used the correct workspace
+		for id, jobStep := range resp.JobMap {
+			if jobStep.PipelineId == embeddedResp.Pipeline.Id {
+				// all the embedded steps should inherit the parent steps
+				// workspace, except the step named "otherws", which has itself
+				// a specified workspace
```
briancain, line 637: Great test!!

briancain (commented):
Looks good! Not sure if there was more to do on this one but I think it's good to be ready for review? 🎉

briancain (approved):

PR #3812, opened 2022-09-06
config: Exec and Pipeline Step params have required fields

briancain:
Command for Exec, and Project+Name for Pipeline ref step are actually required fields for these steps. Update the HCL parsing to ensure these are required when parsing a StepRaw when turning it into a Protobuf.

Related to the findings from https://github.com/hashicorp/waypoint/pull/3810


evanphx (approved):

izaaklauer (commented):
👍

PR #3820, opened 2022-09-07
Fix version check on server during ECS upgrade

paladin-devops:
Previously during ECS upgrades, if the current server image was hashicorp/waypoint:latest, regardless of user input, the upgrade would just force-redeploy the existing ECS service and not update the image. With this change, we check the existing image to see if it matches hashicorp/waypoint:latest, AND check if that's what the user wanted to upgrade to. If either of those is false, we update the task definition during the upgrade.


internal/serverinstall/ecs.go
```diff
@@ -536,8 +536,7 @@ func (i *ECSInstaller) Upgrade(
 	s = sg.Add("Updating task definition")
 	defer func() { s.Abort() }()
 
-	// assume upgrade to latest
-	if *containerDef.Image == installutil.DefaultServerImage {
+	if *containerDef.Image == installutil.DefaultServerImage && upgradeImg == installutil.DefaultServerImage {
```
briancain, line 539: I know this was already here but you might need a `*containerDef != nil` too to be safe?

paladin-devops: Updated!

briancain (approved):

PR #3827, opened 2022-09-07
Backport of `runner install` fixes into release/0.9.x

hc-github-team-waypoint:
## Backport

This PR is auto-generated from #3772 to be assessed for backporting due to the inclusion of the label backport/0.9.x.

---

Fixes https://github.com/hashicorp/waypoint/issues/3681

This is largely a code cleanup PR. The main issue was that if I installed a runner on Cluster A, then switched my kube context to Cluster B, and then tried to uninstall that runner, I would get the waypoint error message "No runners installed" because the code path would go through the `k8sUninstall` function.
This improvement first checks if the runner is on the currently-active cluster before choosing the uninstall method, and returning the appropriate error message.

Per the comments in this PR:
```
	// Our checks here follow the logic of:
 	// Up until v0.8.2, we installed runners with the k8s client,
 	// and the Label was "app=waypoint-runner" and the Name "waypoint-runner-random-id"
 	// As of 0.9.0, we install runners with helm, with a Label following the
 	// pattern ("app.kubernetes.io/instance=waypoint-%s", runnerId)
 	// and the Name ("waypoint-"+strings.ToLower(runnerId))
 	//
 	// Therefore we need to ascertain A) if the runner exists on the cluster at
 	// all (it might not be if the user is auth'd to the wrong cluster), and then B)
 	// if the name/label matches the Helm pattern or the k8s client pattern so
 	// we know how to uninstall it
```

Previous to this update, if the runner wasn't found on the cluster, we'd still remove the runner from Waypoint and then tell the user that no runners were installed:
```
❯ wd runner uninstall -platform=kubernetes -id=01GBD70R1TYVKH5T41HHYE9D49
✓ Runner "01GBD70R1TYVKH5T41HHYE9D49" uninstalled successfully
✓ Runner "01GBD70R1TYVKH5T41HHYE9D49" forgotten on server
✓ No runners installed.
```

After this update, the error messages are more specific and timely (we don't list the specific context because if the user hasn't set it with the `-k8s-context` flag then we don't have it; but we include this note just as a troubleshooting hint for the user):
```
❯ wd runner uninstall -platform=kubernetes -id=01GBTCZCF74K88V3DBB1MGXR25
❌ Unable to uninstall runner
! runner with ID "01GBTCZCF74K88V3DBB1MGXR25" not found in namespace "default"
  with current context
```


krantzinator:
Since we don't have a planned 0.9.x release, I'm not going to back port this.

PR #3835, opened 2022-09-08
ui: allow dependabot to update all dependencies

jgwhite:
## Why the change?

Dependabot does a good job with our security and design system updates, I thought it might be worth giving it a spin with all ui dependencies.


amyrlam (approved):
LGTM let's try it out

Issue #3832, opened 2022-09-07
`waypoint install` now sets up two runner profiles, ignores requested ODR/Server image


briancain:
**Describe the bug**

A fresh `waypoint install` with the k8s platform ends up with two runner profiles.

```
brian@ubuntu:waypoint-tetris λ waypoint install -accept-tos -platform=kubernetes -k8s-server-image="localhost:5000/waypoint:dev"
⚠️ Kind cluster detected!

Installing Waypoint to a Kind cluster requires that the cluster has
LoadBalancer capabilities (such as metallb). If Kind isn't configured
in this way, then the install may hang. If this happens, please delete
all the Waypoint resources and try again.
✓ Service "waypoint" is ready
✓ Server bootstrap complete!
✓ Waypoint server installed with Helm!
✓ Configured server connection
✓ Successfully connected to Waypoint server in Kubernetes!
✓ Server installed and configured!
✓ Runner "static" installed
✓ Registered ondemand runner!
✓ Waypoint runner installed with Helm!
✓ Runner "static" adopted successfully.
Waypoint server successfully installed and configured!

The CLI has been configured to connect to the server automatically. This
connection information is saved in the CLI context named "install-1662590313".
Use the "waypoint context" CLI to manage CLI contexts.

The server has been configured to advertise the following address for
entrypoint communications. This must be a reachable address for all your
deployments. If this is incorrect, manually set it using the CLI command
"waypoint server config-set".

To launch and authenticate into the Web UI, run:
waypoint ui -authenticate

Advertise Address: 172.20.0.20:9701
Web UI Address: https://172.20.0.20:9702
brian@ubuntu:waypoint-tetris λ waypoint runner profile list                                                                 ±[main]
Runner profiles
              NAME             | PLUGIN TYPE |                OCI URL                 | TARGET RUNNER | DEFAULT
-------------------------------+-------------+----------------------------------------+---------------+----------
  01GCD1NR5MGSSX2584BYHYFEX5   | kubernetes  | docker.io/hashicorp/waypoint-odr:0.9.1 | *             | yes
  kubernetes-bootstrap-profile | kubernetes  | localhost:5000/waypoint-odr:dev        | *             | yes
```

We should only have the `kubernetes-bootstrap-profile`.

**Steps to Reproduce**

`waypoint install -platform=kubernetes`

**Expected behavior**

Only 1 profile should exist, `kubernetes-bootstrap-profile`.

**Waypoint Platform Versions**
[...]

PR #3837, opened 2022-09-08
serverinstall/k8s: Actually set on-demand runner image with Helm

briancain:
Prior to this commit, Helm would never properly set the right on-demand runner image specified by the CLI install flags. This commit fixes that by ensuring those values are properly set with the Helm chart yaml prior to the install.

I don't know if this requires a changelog because the bug might of been introduced after the last release.

### DIFF
--- a/internal/serverinstall/k8s.go
+++ b/internal/serverinstall/k8s.go
@@ -194,7 +194,13 @@ func (i *K8sInstaller) Install(
 
 	imageRef, err := dockerparser.Parse(i.Config.ServerImage)
 	if err != nil {
-		ui.Output("Error parsing image ref: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+		ui.Output("Error parsing server image ref: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


briancain:
Note that this only fixes one part of https://github.com/hashicorp/waypoint/issues/3832, it fixes the ODR image being properly set but does not yet fix the double bootstrap issue. That will happen in a different PR.

briancain:
With this fix now:
```
brian@ubuntu:waypoint-tetris λ waypoint runner profile list                                                                         ±[main]
Runner profiles
              NAME             | PLUGIN TYPE |             OCI URL             | TARGET RUNNER | DEFAULT
-------------------------------+-------------+---------------------------------+---------------+----------
  01GCEW5DRG12FS9KG70PRT14ZJ   | kubernetes  | localhost:5000/waypoint-odr:dev | *             | yes
  kubernetes-bootstrap-profile | kubernetes  | localhost:5000/waypoint-odr:dev | *             | yes

```

xiaolin-ninja (approved):
LGTM, can you also add in the install CLI command you used?

krantzinator (approved):

izaaklauer (commented):
👍

PR #3800, opened 2022-09-02
k8s runner profile should use odr img, not server

krantzinator:
My runner profile created via running `waypoint runner install` did not use the ODR image by default.

Before:
```
$ wd runner install -k8s-context=secondary -server-tls-skip-verify -- -label=environment=secondary
...

$ wd runner profile list
Runner profiles
                  NAME                  | PLUGIN TYPE |                OCI URL                 |            TARGET RUNNER            | DEFAULT
----------------------------------------+-------------+----------------------------------------+-------------------------------------+----------
  01GBWMDG7YRX1A7HRPQKWNSX2S            | kubernetes  | docker.io/hashicorp/waypoint-odr:0.9.1 | *                                   | yes
  kubernetes-bootstrap-profile          | kubernetes  | hashicorp/waypoint-odr:latest          | *                                   | yes
  kubernetes-01GBWN6ZBQP6GH4R4GGWRG9W8M | kubernetes  | hashicorp/waypoint:latest              | labels: {"environment":"secondary"} |
```

After:
```
$ wd runner install -k8s-context=secondary -server-tls-skip-verify -- -label=environment=secondary2
...

$ wd runner profile list
Runner profiles
                  NAME                  | PLUGIN TYPE |                OCI URL                 |            TARGET RUNNER             | DEFAULT
----------------------------------------+-------------+----------------------------------------+--------------------------------------+----------
  01GBWMDG7YRX1A7HRPQKWNSX2S            | kubernetes  | docker.io/hashicorp/waypoint-odr:0.9.1 | *                                    | yes
  kubernetes-bootstrap-profile          | kubernetes  | hashicorp/waypoint-odr:latest          | *                                    | yes
  kubernetes-01GBWN6ZBQP6GH4R4GGWRG9W8M | kubernetes  | hashicorp/waypoint:latest              | labels: {"environment":"secondary"}  |
  kubernetes-01GBZ9Q7ZPMHJKVR5XJJHQX96S | kubernetes  | hashicorp/waypoint-odr:latest          | labels: {"environment":"secondary2"} |
```

PR #3838, opened 2022-09-08
runner img should be server and other fixups

krantzinator:
1. changed the name of the var `const DefaultRunnerImage = "hashicorp/waypoint-odr:latest"` to `DefaultOdrImage`
2. updated the k8s server install `odr-image` flag description to match the other platforms
3. reverted back the change made in https://github.com/hashicorp/waypoint/pull/3800/files that changed the default flag value for RunnerInstall `k8s-runner-image` (it has been `hashicorp/waypoint`; I changed it in PR-3800 to be `hashicorp/waypoint-odr`; I have now realized that is wrong based on the current structure of our code and in this PR revert it back to `hashicorp/waypoint` for now; will open an issue for a larger code refactor)

### DIFF
--- a/internal/runnerinstall/k8s.go
+++ b/internal/runnerinstall/k8s.go
@@ -117,7 +117,7 @@ func (i *K8sRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
 
 	odrImage := i.Config.OdrImage
 	if odrImage == "" {
-		odrImage, err = installutil.DefaultODRImage(i.Config.ServerImage)
+		odrImage, err = installutil.DefaultODRImage(i.Config.RunnerImage)
 		if err != nil {
 			opts.UI.Output("Error getting default ODR image: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 			return err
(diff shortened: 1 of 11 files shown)


briancain (commented):
I'd like to wait on merging this until https://github.com/hashicorp/waypoint/pull/3840 goes through. Other than removing the `c.flagRunner` commenting out changes, I think the rest of this makes sense!

internal/runnerinstall/k8s.go
```diff
@@ -238,7 +238,7 @@ func (i *K8sRunnerInstaller) InstallFlags(set *flag.Set) {
 	set.StringVar(&flag.StringVar{
 		Name:    "k8s-runner-image",
 		Target:  &i.Config.RunnerImage,
-		Default: installutil.DefaultRunnerImage,
+		Default: defaultRunnerImage,
```
briancain, line 241: Should this be `defaultODRImage`? It looks like the website docs are showing this as the server image rather than the `-odr` prepended image.

krantzinator: This matches what all the other platforms do, e.g.:
https://github.com/hashicorp/waypoint/blob/305d27d0242571dc04e9cfeb192439898dff070f/internal/runnerinstall/ecs.go#L255-L260

So, you might be right, but that would be a bug PR, maybe related to #3839 . This PR I'm just trying to make it consistent with what it was/currently is on other platforms.

krantzinator: 😖

krantzinator:
I think this has been handled in subsequent bug fix PRs for 0.10.1

PR #3840, opened 2022-09-08
serverinstall/k8s: Prevent double runner profiles with Helm install

### LINKED ISSUE #3832: `waypoint install` now sets up two runner profiles, ignores requested ODR/Server image

briancain:
Prior to this commit, the CLI installer would set up two runner profiles: The first one coming from Helm, and the second one coming from our runnerInstall function setting up the bootstrap profile.

Because Helm already sets up a profile, we use that runner profile instead of the bootstrap profile so that we don't install with multiple default runnerp profiles.

Fixes #3832

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -96,7 +96,11 @@ func (c *InstallCommand) Run(args []string) int {
 		}
 	}
 
-	result, err := p.Install(ctx, &serverinstall.InstallOpts{
+	// NOTE(briancain): Currently only the K8s Helm installer returns a bootstrap
+	// token. Helm also installs a runner, so we use this fact to determine if
+	// we should not only bootstrap a token but also install a runner. We could
+	// make this better in the future.
+	result, bootstrapToken, err := p.Install(ctx, &serverinstall.InstallOpts{
 		Log:            log,
 		UI:             c.ui,
 		ServerRunFlags: secondaryArgs,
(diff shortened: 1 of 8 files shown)


briancain:
There's still some issues with this...it seems....


```
brian@ubuntu:waypoint-tetris λ waypoint pipeline run example                                                                        ±[main]

» Running pipeline "example" for application "tetris"
:x: Syncing pipeline configs...
! ondemand runner config not found
! error sentinel
brian@ubuntu:waypoint-tetris λ waypoint runner profile list                                                                         ±[main]
Runner profiles
             NAME            | PLUGIN TYPE |             OCI URL             | TARGET RUNNER | DEFAULT
-----------------------------+-------------+---------------------------------+---------------+----------
  01GCF18ZTG0B2WEEBR7CWZ2EAX | kubernetes  | localhost:5000/waypoint-odr:dev | *             | yes
```

internal/cli/install.go
```diff
@@ -375,9 +367,17 @@ func (c *InstallCommand) Run(args []string) int {
 	s.Update("Server installed and configured!")
 	s.Done()
 
+	// NOTE(briancain): Right now we rely on the fact that only K8S returns a bootstrap
+	// token from the installation interface helper. Since we know helm sets up
+	// a bootstrap job AS WELL AS a runner, we don't need to install another runner
+	// or another runner profile.
 	if c.flagRunner {
+		existingODRConfigSetup := false
+		if bootstrapToken != "" {
```
briancain, line 376: I don't love this conditional being based on the `bootstrapToken` value being returned from the installer interface. We could add a fourth return value which is a bool that says whether or not it already set up a runner profile? I was trying to keep this change small so went with this instead but if people feel strongly that relying on `bootstrapToken` being set is bad I can add another return bool param.

xiaolin-ninja, line 376: Is this supposed to be `token != ""` ?

briancain: No because `bootstrapToken` is the variable that gets returned from install, where as `token` could of been setup by the CLI bootstrap. See my other comment on the PR about that variable.

xiaolin-ninja (approved):
Makes sense to me!

PR #3846, opened 2022-09-08
Fix deployment Pruning fail on k8s

### LINKED ISSUE #3842: Bug: Deployment Pruning fails due to missing expected components
**Describe the bug**

Deployment pruning fails due to missing expected components that were introduced in the #3626 pull request

```
 Release: 01GCF2QWZAT10KG7694S89VB7F (v1)
✓ Running release destroy v1
! 1 error occurred:
        * rpc error: code = Unknown desc = There was an error while executing a
  Waypoint plugin for this operation!

  One or more required arguments for the plugin was not satisfied. This is usually
  due to a missing or incompatible set of plugins. For example, only certain build
  plugins are only compatible with certain registries, and so on. Please inspect
  the missing argument, the set of plugins you are using, and the documentation to
  determine if your plugin combination is valid.

  Plugin function:
  github.com/hashicorp/waypoint/builtin/k8s.(*Releaser).Destroy-fm

  ==> Missing arguments:

      - type: *component.DeclaredResourcesResp
      - type: *component.DestroyedResourcesResp

```

**Steps to Reproduce**

Run `waypoint up` more than once and see waypoint attempt to prune old deployments and fail.

**Expected behavior**

It shouldn't fail.

xiaolin-ninja:
Fixes issue https://github.com/hashicorp/waypoint/issues/3842

Wasn't sure whether to add declared and destroyed resource info to `nomad/platform.go`, since it looks like declared resources response was only added to `nomad/jobspec/platform.go`, and I lack context on recent Nomad & plugins changes to know what's going on there.


xiaolin-ninja:
https://github.com/hashicorp/waypoint/issues/3842

xiaolin-ninja:
Closes https://github.com/hashicorp/waypoint/issues/3842

briancain (approved):
Looks good! And fixes my problem.

Not sure, but do we need to fix the Nomad Canary releaser too?

briancain (changes requested):
Oops, looks like the deployment plugin was fixed, but we have to fix the releaser too for Kubernetes!

briancain (approved):
Looks good! I think you'll need to re-gen the server protos from a conflict

PR #3854, opened 2022-09-09
core/cli: Update profile set to use defaults on init only

### LINKED ISSUE #3853: Waypoint runner profile set shouldn't reset the profile for settings not given
Today, when running `waypoint runner profile set` to create a new profile the profile is setup with sane defaults for those settings not given. However when updating an existing profile if you do not specify all of these flags the values you had previously set, instead the profile options get replaced with the initial setup defaults. This can have the unintended consequence of resetting the profile to be targetable by all runners instead of the ones you had initially specified.

It would be nice to not have to re-target existing runners when you want to modify an OCI container or an environment variable within the profile.

When running `waypoint runner profile set` to update an existing runner profile, Waypoint should only update the profile with the given flags instead of resetting the runner profile.

This can be worked around by either specifying all runner profile configuration each time `waypoint runner profile set` is ran or by storing runner configs within a file and using the `-plugin-config` flag to specify the file with runner configs.

demophoon:
This commit modifies `runner profile set` to use profile defaults only when setting up an initial profile instead of resetting the profile to defaults when updating an existing profile.

Fixes #3853

### DIFF
--- a/internal/cli/runner_profile_set.go
+++ b/internal/cli/runner_profile_set.go
@@ -130,7 +130,7 @@ func (c *RunnerProfileSetCommand) Run(args []string) int {
 			},
 		}
 		if c.flagTargetRunnerAny != nil {
-			c.ui.Output("Both -target-runner-id and -target-runner-any detected, only one can be set at a time. ID takes priority.",
+			c.ui.Output("Both -target-runner-label and -target-runner-any detected, only one can be set at a time. Labels take priority.",
 				terminal.WithWarningStyle())
 		}
 	} else if od.TargetRunner == nil || (c.flagTargetRunnerAny != nil && *c.flagTargetRunnerAny) {
--- a/internal/cli/runner_profile_set.go
+++ b/internal/cli/runner_profile_set.go
@@ -133,7 +133,7 @@ func (c *RunnerProfileSetCommand) Run(args []string) int {
 			c.ui.Output("Both -target-runner-id and -target-runner-any detected, only one can be set at a time. ID takes priority.",
 				terminal.WithWarningStyle())
 		}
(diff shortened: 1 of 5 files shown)


internal/cli/runner_profile_set.go
```diff
@@ -124,7 +129,11 @@ func (c *RunnerProfileSetCommand) Run(args []string) int {
 				},
 			},
 		}
-	} else {
+		if c.flagTargetRunnerAny != nil {
+			c.ui.Output("Both -target-runner-id and -target-runner-any detected, only one can be set at a time. ID takes priority.",
```
paladin-devops, line 133: I think that this is meant to be `Both -target-runner-label and -target-runner-any detected...`

```terminal
$ ./waypoint runner profile set -name=docker-def -target-runner-label=env=production -target-runner-any -default=false
✓ Runner profile updated
Both -target-runner-id and -target-runner-any detected, only one can be set at a time. ID takes priority.
```

demophoon: oops, copy paste error on my part! Good catch

izaaklauer (approved):
I tested this myself, and it works great!

website/content/commands/runner-profile-set.mdx
```diff
@@ -44,5 +44,6 @@ lifecycle operation.
 - `-default` - Indicates that this remote runner profile should be the default for any project that doesn't otherwise specify its own remote runner.
 - `-target-runner-id=<string>` - ID of the runner to target for this remote runner profile.
 - `-target-runner-label=<key=value>` - Labels on the runner to target for this remote runner profile. e.g. `-target-runner-label=k=v`. Can be specified multiple times.
+- `-target-runner-any` - Set profile to target any available runner.
```
izaaklauer, line 47: very good catch

PR #3855, opened 2022-09-09
Add values required for the updated Destroy interface

evanphx:
The destroy plugin interface requires (Declared|Destroyed)ResourceResp now, but release was not updated to provide these values. This simply provides them but doesn't use them. This prevents calling a Release plugins Destroy method from failing to execute (which was failing because argmapper was looking for these values)


briancain (approved):

xiaolin-ninja:
Um, these aren't it, I already added them in my PR.

PR #3860, opened 2022-09-12
cli: Show noop job operation type in task and job CLIs

briancain:
This operation was initially left off the job and task CLIs because no jobs used the operation. Now, the pipelines exec step uses a `noop` job to execute commands in the spawned resource, so we should show the noop job operation type too so it doesn't show up as "Unknown" in the job and task CLIs.


gregone (approved):

cicoyle (approved):
LGTM!

demophoon (approved):

PR #3861, opened 2022-09-12
bug: Show error if inspecting pipeline with no runs

demophoon:
Before this commit waypoint would attempt to fetch the -1th Pipeline run if no pipelines had ran when inspecting a pipeline. This commit adds an error to the output instead of crashing.

### DIFF
--- a/internal/cli/pipeline_inspect.go
+++ b/internal/cli/pipeline_inspect.go
@@ -192,46 +192,55 @@ func (c *PipelineInspectCommand) Run(args []string) int {
 			},
 		}...)
 	} else {
-		lastRun := runs.PipelineRuns[len(runs.PipelineRuns)-1]
-		startJob, err := c.project.Client().GetJob(c.Ctx, &pb.GetJobRequest{
-			JobId: lastRun.Jobs[0].Id,
-		})
-		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/cli/pipeline_inspect.go
```diff
@@ -192,6 +192,10 @@ func (c *PipelineInspectCommand) Run(args []string) int {
 			},
 		}...)
 	} else {
+		if len(runs.PipelineRuns) == 0 {
+			c.ui.Output("No pipelines have ran", terminal.WithErrorStyle())
+			return 1
```
briancain, line 197: We shouldn't err in this case, because we still want to be able to inspect a pipeline if it hasn't been run before, we just can't include pipeline run info. Maybe instead of `else` on 194 we do `else if len(runs.PipelineRuns) == 0`?

briancain (changes requested):

internal/cli/pipeline_inspect.go
demophoon: That sounds reasonable to me

xiaolin-ninja (approved):
Great catch! I pushed a commit that fixes this.

PR #3863, opened 2022-09-12
server: Track embedded step refs for pipelines with noop job

briancain:
Prior to this commit we would ignore step refs of type pipeline and not
generate its own job. This caused an issue where if a embedded pipeline
steps parent step was also an embedded pipeline, we wouldn't have a job
id to wait on, which would make the step run out of order from its
parent.

This commit fixes that by tracking the step refs as a noop job, then
that job depends on all of its embedded jobs before it's complete. This
makes it easier for any consuming steps to wait on the noop job which is
essentially the last job of a pipeline.

### DIFF
--- a/pkg/server/singleprocess/service_pipeline.go
+++ b/pkg/server/singleprocess/service_pipeline.go
@@ -227,20 +227,18 @@ func (s *Service) buildStepJobs(
 	// Generate job IDs for each of the steps. We need to know the IDs in
 	// advance to set up the dependency chain.
 	stepIds := map[string]string{}
-	for name, step := range pipeline.Steps {
-		if _, ok := step.Kind.(*pb.Pipeline_Step_Pipeline_); !ok {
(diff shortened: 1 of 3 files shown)


pkg/server/singleprocess/service_pipeline.go
```diff
@@ -425,6 +424,17 @@ func (s *Service) buildStepJobs(
 						"an embedded pipeline step matches a parent step name: %s", k)
 				}
 			}
+
+			// Steps of type Pipline Refs are now noop jobs. This is currently a work around to ensure that
```
evanphx, line 428:
```suggestion
			// Steps of type Pipeline Refs are now noop jobs. This is currently a work around to ensure that
```

evanphx (approved):
It's a wonder that I found a spelling mistake.

demophoon (approved):

PR #3872, opened 2022-09-14
Update all error returns in RunnerJobStream to use hcerr.Externalize()

catsby:
Wrap error returns in [`hcerr.Externalize()`](https://github.com/hashicorp/waypoint/blob/52e13de50ac6da2ffedb9c34f17f592c76de9fe1/pkg/server/hcerr/hcerr.go#L16-L29) to log more details correctly


izaaklauer (approved):
Thanks, this looks great!

PR #3873, opened 2022-09-14
Install documentation updates

paladin-devops:
Minor updates to install docs.

### DIFF
--- a/website/content/docs/server/install.mdx
+++ b/website/content/docs/server/install.mdx
@@ -38,17 +38,15 @@ Currently, Waypoint supports running its server on the following platforms:
 
 ### Kubernetes Platform
 
-Installing Waypoint server onto Kubernetes has a couple of options: Using the
-waypoint-helm chart or using `waypoint install`.
-
-The officially supported [hashicorp/waypoint-helm](https://github.com/hashicorp/waypoint-helm)
-chart can be used to deploy Waypoint server directly onto a Kubernetes cluster with Helm charts.
-We recommend reading the README detailed on that project for more information on
-how to get going using Helm and Waypoint.
-
 Installing the Waypoint server to Kubernetes using the `waypoint install` command
-will also use the Waypoint Helm chart behind the scenes for the installation, setting sensible default
-values for the chart values.
+uses the [hashicorp/waypoint-helm](https://github.com/hashicorp/waypoint-helm)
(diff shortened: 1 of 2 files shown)


website/content/docs/server/install.mdx
```diff
@@ -38,9 +46,9 @@ chart can be used to deploy Waypoint server directly onto a Kubernetes cluster w
 We recommend reading the README detailed on that project for more information on
 how to get going using Helm and Waypoint.
 
-You can also install Waypoint server onto Kubernetes using the `waypoint install`
-command. This command sets up a Service, Statefulset, Deployment, and Replicaset
-and deploys Waypoint server and its runner as two separate pods by default.
+Installing the Waypoint server to Kubernetes using the `waypoint install` command
+will also use the Waypoint Helm chart for the installation, setting sensible default
+values for the chart values.
```
catsby, line 51: Above here we say "Installing Waypoint server onto Kubernetes has a couple of options: Using the
waypoint-helm chart or using `waypoint install`.", I think we should probably remove that part now and combine what's left with this new sentence, what do you think?

website/content/docs/server/install.mdx
```diff
@@ -19,6 +19,14 @@ a specific setup that doesn't work for you with `waypoint install`, we recommend
 looking at the [server run](/docs/server/run) documentation instead or opening
 an issue on GitHub and requesting a feature enhancement.
 
+The server which is installed by the `waypoint install` command runs a Waypoint
+server in a Docker container, which is actually running the `waypoint server run`
+command:
```
briancain, line 24:
```suggestion
The server installed by the `waypoint install` command runs Waypoint
server in a Docker container, which runs the `waypoint server run`
command:
```

briancain (changes requested):

website/content/docs/server/install.mdx
paladin-devops: @catsby do you mean removing the presentation of the option to directly install with Helm via `helm install`?

catsby: I think we should reword or restructure this part to mention/include that `waypoint install` uses the helm chart, but they could also use it directly if they wanted. Right now it's worded as either or, but `waypoint install` uses the helm chart

briancain (approved):

PR #3882, opened 2022-09-15
Makefile: Add new target `help` for printing valid Make targets

briancain:
Have you ever wondered what valid targets our `Makefile` has? Wonder no longer!

```
brian@ubuntu:waypoint λ make help 
Valid targets:
bin             bin Creates the binaries for Waypoint for the current platform
bin/cli-only    Builds only the cli with no ceb
bin/entrypoint  Create the entrypoint for the current platform
bin/linux       bin Creates the binaries for Waypoint for the linux platform
bin/windows     Create windows binaries
docker/gen/server: docker/tools
docker/odr      Builds a Waypoint on-demand runner docker image
docker/server: docker/server-only docker/odr
docker/server-only Builds a Waypoint server docker image
docker/tools    Creates a docker tools file for generating waypoint server protobuf files
format          Format all go code in project
gen/changelog   Generates the changelog for Waypoint
gen/doc         generates the server proto docs
gen/plugins     Generates plugin protobuf Go files
gen/server      Generates server protobuf Go files from server.proto
gen/ts          Generates frontend typescript files
gen/website-mdx Generates the website markdown files
help            Print valid Make targets
install         Build and copy binaries to $GOPATH/bin/waypoint
static-assets   Generates the UI static assets
test/boltdbstate Runs the boltdbstate tests
test            Run tests
test/service    Runs the server API function tests. Requires a local postgresql and horizon via docker-compose
tools           Install dependencies and tools required to build
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,12 @@ WP_SERVER_PLATFORM?="linux/amd64"
 LAST_RELEASE?=$$(git describe --tags $$(git rev-list --tags --max-count=1))
 THIS_RELEASE?=$$(git rev-parse --abbrev-ref HEAD)
 
+.PHONY: help
+help: # print valid Make targets
+	@echo "Valid targets:"
+	@grep --extended-regexp --no-filename '^[a-zA-Z/_-]+:' Makefile | sort | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
+
+
 .PHONY: bin
 bin: # bin creates the binaries for Waypoint for the current platform
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./internal/assets/ceb/ceb ./cmd/waypoint-entrypoint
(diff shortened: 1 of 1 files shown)


paladin-devops (approved):
🆘

Makefile
```diff
@@ -25,58 +31,58 @@ bin: # bin creates the binaries for Waypoint for the current platform
...
 	mkdir -p $(GOPATH)/bin
 	cp ./waypoint $(GOPATH)/bin/waypoint
 
 .PHONY: format
-format: # format go code
+format: # Format all go code in project
 	gofmt -s -w ./
 
 .PHONY: docker/server
 docker/server: docker/server-only docker/odr
```
demophoon, line 68: Do we care about these targets not formatting like the rest of the targets?

demophoon: e.g. When looking at `docker/gen/server` it also calls `docker/tools` and the output looks like this as a header.
```
docker/gen/server: docker/tools
```

We can exclude them by modifying your regex above to be
```
^[a-zA-Z/_-]+:\s+#' Makefile | sort | awk 'BEGIN {FS = ":.*?# "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}
```

briancain: I think it's ok since they're like a "meta" target that runs other targets. Maybe in a follow-up PR if someone is motivated?

demophoon: Works for me

briancain: I'll steal your regex!

demophoon (approved):
🔨 🗄️ 📈

PR #3883, opened 2022-09-15
Nomad Plugin bug fixes

paladin-devops:
1. Fixed a bug in the jobspec status check where a deployment that had canary allocs would always report as partially healthy, even if it was healthy
2. Fixed the Nomad task launcher plugin to deploy the Nomad ODRs to the region and namespace specified in the runner profile plugin config


briancain (approved):
🐛 🐦

demophoon (approved):
Looks great

PR #3888, opened 2022-09-15
Handle Remote datasources properly

evanphx:
We introduced a Remote data source type, that goes along with Git and Local. The difference with Remote is that in indicates that only the Waypoint server knows how to perform the data source fetching.

The server already has the ability to translate a project with a Remote data source into job with a Git data source.

This updates the CLI to handle the presence of Remote properly (rather than assuming if it's not Git, it's local).


briancain (approved):
🏃🏻

PR #3893, opened 2022-09-16
bolt: Cancel WatchJob for TaskCancel

briancain:
This commit updates TaskCancel to also cancel the WatchTask job if it exists. Not all task launcher plugins implement WatchTask so there's an extra check that the job ref was set prior to cancellation.


izaaklauer (approved):

paladin-devops (approved):
❌  Code looks good!

PR #3894, opened 2022-09-16
Setting the k8s runner install odr service account.

izaaklauer:
Helm creates a service account, but we don't use it unless it's specified in the runner profile. Otherwise, you get an error like this:

```
$ wp up -workspace=dev
locally with the -local flag.

» Operation is queued waiting for job "01GD3CJWM5ABKYZ9YRH0VMCNCZ". Waiting for runner assignment...
  If you interrupt this command, the job will still run in the background.
  Performing operation on "kubernetes" with runner profile "kubernetes-DEV"

» Cloning data from Git
  Ref: go-multicluster

» Downloading from Git
  Git Commit: 741ce4489f554981e15fcc703ffd4b8cde5b93de
   Timestamp: 1970-01-01 00:00:00 +0000 UTC
     Message:✓ All services available.

... <truncated for brevity> ...

✓ Set ECR Repository name to 'acmecorp'

» Deploying backend...
✓ Running deploy v2
✓ Kubernetes client connected to https://10.100.0.1:443 with namespace default
❌ Preparing deployment...
✓ Kubernetes client connected to https://10.100.0.1:443 with namespace default
❌ Deleting deployment...
! 2 errors occurred:
  	* deployments.apps "backend-v2" is forbidden: User
  "system:serviceaccount:default:default" cannot get resource "deployments" in API
  group "apps" in the namespace "default"
  	* Error during rollback: resource name may not be empty
```

The default service account doesn't have permissions (like list deployments) that the ODR needs for deploy ops.

## Workaround

You can get around this by editing the runner profile and specifying your own k8s tasklauncher config. Create a file like this:

```
cat /tmp/k8s-plugin-config.hcl

image_pull_policy = "Always"
namespace = "default"
service_account = "waypoint-runner-odr"
```

And then apply it to your runner profile:

```
waypoint runner profile set -name=<your runner profile name> -plugin-type=kubernetes -plugin-config=/tmp/k8s-plugin-config.hcl
```

You can run `waypoint runner profile inspect <your profile name>` to see your current plugin config, in case you want to preserve any of those settings.

### DIFF
--- a/.changelog/3894.txt
+++ b/.changelog/3894.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+runner-install/k8s: Use the helm-creates service account in the runner profile, rather than omitting it and getting the default service account with no permissions.
+```
--- a/internal/runnerinstall/k8s.go
+++ b/internal/runnerinstall/k8s.go
@@ -32,8 +32,9 @@ type K8sRunnerInstaller struct {
 }
 
 const (
-	defaultRunnerMemory = "256Mi"
-	defaultRunnerCPU    = "250m"
+	defaultRunnerMemory          = "256Mi"
+	defaultRunnerCPU             = "250m"
+	defaultOdrServiceAccountName = "waypoint-runner-odr"
 )
 
 type InstalledRunnerConfig struct {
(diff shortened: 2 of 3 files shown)


paladin-devops (approved):
🐕‍🦺

briancain (approved):

PR #3903, opened 2022-09-19
feat(plugins/k8s): add `security_context` option for on-demand runners

denysvitali:
Certain locked down Kubernetes clusters require the users to run containers in an unprivileged mode, and enforce this by checking that the Pod SecurityContext settings do include stuff like `runAsNonRoot: true`.

This commit adds support for custom SecurityContexts, so that unprivileged runners can be dynamically spawned.


krantzinator:
Finishing this up over on https://github.com/hashicorp/waypoint/pull/4346 with the generated files!

PR #3912, opened 2022-09-21
feat(plugins/docker): add `docker-ref` plugin

denysvitali:
The `docker-pull` plugin really pulls the image in the case of On Demand Runners. Since we do require there a `docker.AccessInfo`, we can't avoid pulling the image (because HasRegistry automatically becomes true, and not providing a registry causes an error).

To avoid patching `docker-pull` further, a `docker-ref` is created. This is only used in the very specific case where a `docker-pull` can't be used because all of the following conditions are met:

- Access to the registry is not available
- On-Demand Runner (ODR) is being used

As a solution, this plugin allows you to be a sort of "noop" that adds a reference to a pre-built
Docker image (e.g: in an external CI) in the `build` stage.

---

Example usage:

```hcl
app "whoami" {
  build {
    use "docker-ref" {
      image = "traefik/whoami"
      tag   = "latest"
    }
  }

  deploy {
    use "kubernetes" {
      // ...
    }
  }
}
```

### DIFF
--- a/.changelog/3912.txt
+++ b/.changelog/3912.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugins/docker: Add new `docker-ref` for a build noop for referencing a Docker Image
+```
--- a/builtin/docker/ref/builder.go
+++ b/builtin/docker/ref/builder.go
@@ -1,0 +1,109 @@
+package dockerref
+
+import (
+	"github.com/hashicorp/go-argmapper"
+	"github.com/hashicorp/waypoint-plugin-sdk/docs"
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	wpdocker "github.com/hashicorp/waypoint/builtin/docker"
+	empty "google.golang.org/protobuf/types/known/emptypb"
+)
+
+// Builder uses `docker build` to build a Docker iamge.
+type Builder struct {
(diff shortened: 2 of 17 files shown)


paladin-devops (commented):
This is looking awesome! The website docs should definitely include the documentation for this plugin as well. 😄 

I just tried this out locally with Docker and it worked great! Giving it a shot with ODRs next.

paladin-devops (changes requested):
@denysvitali I think the last thing we need to do here is get the documentation generated! As mentioned elsewhere, `make gen/website-mdx` will take care of this for you. I should also add that since this is a new plugin, you won't see the files in the diff, and you'll need to `git add` each of the `.mdx` files generated in `website/content/partials/components` for `docker-ref`.

Once they're added to the repo, the other required change to get the documentation to appear on the website is in `website/content/plugins/docker.mdx`. There, you should add `@include "components/builder-docker-ref.mdx"`, so the content will be displayed for the builder plugin!

https://github.com/swisscom/waypoint/blob/3b2731519929713eb52a632aa856b9d52647d4ed/website/content/plugins/docker.mdx?plain=1#L12-L20

paladin-devops (approved):
@denysvitali  thanks for adding the docs, this is a great plugin! 😄 🐋 💰

PR #3161, opened 2022-03-30
website: Stage content for Dev Portal

topfunky:
[🔍  Preview content changes](https://developer.hashi-mktg.com/waypoint)

This PR modifies content to be rendered in the Developer Portal. This is a long lived draft PR until we are ready to launch.

Modifications include:

- Remove mention of HashiCorp Learn (reference tutorials instead)
- Fix images and code snippets to fit within the new layout
- Rearrange content and menus where necessary

PR #3915, opened 2022-09-21
website: adds redirects for moved paths

brkalow:
Add redirect rules for paths which were moved as part of #3161


website/redirects.js
```diff
@@ -5,4 +5,29 @@ module.exports = [
...
+  {
+    source: '/docs/kubernetes/:path*',
+    destination: '/docs/platforms/kubernetes/:path*',
+    permanent: true,
+  },
+  {
+    source: '/docs/glossary',
+    destination: '/docs/resources/glossary',
+    permanent: true,
+  },
```
thiskevinwang, line 17: 🤔 I'm not following the redirects here:

I see https://www.waypointproject.io/docs/glossary as valid, but we want to redirect to `/docs/resources/glossary`?

BRKalow: The changes are not yet live, but they are on main. See https://github.com/hashicorp/waypoint/pull/3161

thiskevinwang: Oh <kbd>derp</kbd> I completely forgot about `stable-website` 

I see that here https://tip.waypointproject.io/docs/resources/glossary

thiskevinwang (approved):

izaaklauer (approved):

PR #3922, opened 2022-09-22
Runner install default profile

paladin-devops:
During `runner install`, if no existing runner profiles are detected, then we set the new runner profile as the default.


jgwhite (approved):
Code looks good and works great

CleanShot 2022-09-26 at 21 23 45@2x

Issue #3928, opened 2022-09-23
Missing resource ID in `aws-ecs` status report


jgwhite:
**Describe the bug**

When deploying an app using the aws-ecs plugin, one of the resources in the initially-generated status report is missing an ID. Unfortunately, the UI assumes all resources have IDs, so this causes some confusion. Is it expected that the resource is missing an ID?

<details>
<summary><b>Click to see example status report as JSON</b></summary>

<br>

Some fields removed for brevity…

```json
{
  "health": {
    "healthStatus": "PARTIAL",
    "healthMessage": "1 cluster READY, 1 service READY, 1 task MISSING",
    "deprecatedName": "",
    "deprecatedId": ""
  },
  "resourcesList": [
    {
      "id": "arn:aws:ecs:us-east-1:797645259670:cluster/waypoint",
      "declaredResource": {
        "name": "cluster"
      },
      "parentResourceId": "",
      "name": "waypoint",
      "platform": "aws-ecs",
      "type": "cluster",
      "platformUrl": "",
      "categoryDisplayHint": 0,
      "stateJson": "{\"ActiveServicesCount\":16,\"Attachments\":null,\"AttachmentsStatus\":null,\"CapacityProviders\":[],\"ClusterArn\":\"arn:aws:ecs:us-east-1:797645259670:cluster/waypoint\",\"ClusterName\":\"waypoint\",\"Configuration\":null,\"DefaultCapacityProviderStrategy\":[],\"PendingTasksCount\":7,\"RegisteredContainerInstancesCount\":0,\"RunningTasksCount\":7,\"Settings\":[],\"Statistics\":[],\"Status\":\"ACTIVE\",\"Tags\":[]}",
      "health": 2,
      "healthMessage": "ACTIVE"
    },
    {
      "id": "arn:aws:ecs:us-east-1:797645259670:service/waypoint/jgwhite-aws-ecs-web-01GDNB42ZDKB",
      "declaredResource": {
        "name": "service"
      },
      "parentResourceId": "",
      "name": "jgwhite-aws-ecs-web-01GDNB42ZDKB",
      "platform": "aws-ecs",
      "type": "service",
      "platformUrl": "https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/waypoint/services/jgwhite-aws-ecs-web-01GDNB42ZDKB",
      "categoryDisplayHint": 3,
      "createdTime": {
        "seconds": 1663942405,
        "nanos": 867000000
      },
[...]

PR #3929, opened 2022-09-23
ui: make `<ResourceTable>` resilient to resources without IDs

jgwhite:
## Why the change?

See #3928 for some background. In short, this stops the app from crashing hard when it encounters a resource without an ID.

### Before

CleanShot 2022-09-23 at 16 46 41@2x

### After

CleanShot 2022-09-23 at 16 47 54@2x

## How do I test it?

1. Check out the branch
2. Run Waypoint Server locally
3. `cd ui && yarn start local`
4. Deploy an app to aws-ecs (without having remote runner or gitops configured)
5. Visit the deployment page in the UI
6. Verify it doesn’t crash

### DIFF
--- a/.changelog/3929.txt
+++ b/.changelog/3929.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: don’t crash when encountering resources without IDs
+```
--- a/ui/app/components/resources-table.hbs
+++ b/ui/app/components/resources-table.hbs
@@ -24,13 +24,18 @@
     {{#each @resources key="id" as |resource|}}
       <tr>
         <th scope="row">
-          <LinkTo
-            title={{resource.name}}
-            @route={{@route}}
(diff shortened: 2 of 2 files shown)


ui/app/styles/utilities/_index.scss
```diff
@@ -1,2 +1,3 @@
 @import 'animations';
 @import 'focus-ring';
+@import 'sr-only';
```
gregone, line 3: [dust] tbh that `sr-only` class/file-name is not very descriptive. I'm still wondering what it refers to 🤔

gregone (approved):

ui/app/styles/utilities/_index.scss
jgwhite: Good call. I’ve switched over to `pds--visuallyHidden` (which I couldn’t remember how to invoke in the moment).

PR #3941, opened 2022-09-28
AWS App Runner builder+releaser

### LINKED ISSUE #1503: Add AWS App Runner
No. Today, AWS released a new service [App Runner](https://docs.aws.amazon.com/apprunner/latest/dg/what-is-apprunner.html) and would be great if waypoint can integrate with it.

Add a plugin similar to the one for google cloud run but for aws app runner.

doing it via cli.

**Additional context**

thiskevinwang:
# Description

This introduces the `aws-apprunner` **builder** & **release-manager** components

example:

```hcl
#...
  deploy {
    use "aws-apprunner" {
      name   =  "my-service-name"
      region = var.region
      port   = 8080
      static_environment = {
        "PORT" = "8080"
      }
    }
  }

  release {
    use "aws-apprunner" {
    }
  }
# ...
```

### Screenshots

CleanShot 2023-01-30 at 20 04 52@2x

### TODO

- [x] pass Waypoint Labels to App Runner Tags
- [ ] ~`waypoint exec` will not be supported as you cannot _shell into an app runner replica_~
- [x] static env vars get updated on subsequent `waypoint up` - 5b7a79d
- [x] document recommended `-prune=false`, similar to `builtin/nomad/canary` - 9e75fea

Fixes https://github.com/hashicorp/waypoint/issues/1503


thiskevinwang:
@briancain @padmaksekar 👋 — do you have any guidance for making a component (`platform`, `releaser`, etc.) opt out of Waypoint's default pruning behavior?

I'd like to have only `waypoint destroy` invoke the Destroyer, and have `waypoint up` **not** prune older deployments/releases. AppRunner really only keeps track of 1 "service", so there's nothing to prune.

thiskevinwang:
@briancain @paladin-devops 👋 — do you have any guidance for making a component (`platform`, `releaser`, etc.) opt out of Waypoint's default pruning behavior?

I'd like to have only `waypoint destroy` invoke the Destroyer, and have `waypoint up` **not** prune older deployments/releases. AppRunner really only keeps track of 1 "service", so there's nothing to prune.

briancain:
@thiskevinwang - Unfortunately I don't think this is something that's configurable via a plugin. Pruning is handled at the job level and as far as I can think I don't know of a clean way for you to disable it automatically.

thiskevinwang:
@paladin-devops, ok, sounds good to me! (Though an accidental `waypoint up` without `-prune=false` seems very scary 😅 )

briancain:
@thiskevinwang @paladin-devops - since this has come up a couple of times with plugins, might be worth opening a feature request and we can figure out how to support something like configuring pruning via plugins!

ashleemboyer:
Hello @thiskevinwang! 😊 This is a quick comment to let you know I'm updating this PR with main so it's up to date with the changes made in [#4277](https://github.com/hashicorp/waypoint/pull/4277).

builtin/aws/apprunner/destroyer.go
```diff
@@ -0,0 +1,113 @@
...
+	})
+	if err != nil {
+		step.Update(fmt.Sprintf("Failed to get session: %s", err))
+		return err
+	}
+
+	arSvc := apprunner.New(sess)
+
+	step.Update("Deleting service: %s", release.ServiceName)
+	dso, err := arSvc.DeleteService(&apprunner.DeleteServiceInput{
```
paladin-devops, line 53: We should check this error.

builtin/aws/apprunner/destroyer.go
```diff
@@ -0,0 +1,113 @@
...
+		ServiceArn: aws.String(release.ServiceArn),
+	})
+	step.Done()
+
+	d := time.Now().Add(DEFAULT_TIMEOUT)
```
paladin-devops, line 58: 10 minutes seems like a while! In testing, has it taken that long to delete the service?

builtin/aws/apprunner/platform.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	envVars := make(map[string]*string)
+	for k, v := range p.config.StaticEnvVars {
+		envVars[k] = aws.String(v)
+	}
+
+	step.Done()
+
+	step = sg.Add("Deploying App Runner service %q", p.config.Name)
```
paladin-devops, line 131: Is it necessary to add a new step here, since it's `Done()` in the next line?

builtin/aws/apprunner/platform.go
```diff
@@ -0,0 +1,445 @@
...
+
+		if err != nil {
+			step.Update("Failed to update service: %s", err)
+			if aerr, ok := err.(awserr.Error); ok {
+				if aerr.Code() == apprunner.ErrCodeInvalidStateException {
+					// If state is CREATE_FAILED, the service likely should be manually deleted,
+					// and then re-created.
+
+					// A substring check for "CREATE_FAILED" is our best effort to detect this
+					if strings.Contains(aerr.Message(), "CREATE_FAILED") {
```
paladin-devops, line 190: If it ends up in this state, could `waypoint destroy` handle this instead of manual deletion?

paladin-devops (commented):
Hey @thiskevinwang, this is looking great! I had a few minor comments, but left this unapproved for now since as I understand it, you're working on implementing Resource Manager for this.

thiskevinwang:
@paladin-devops Thanks for the initial review!
I'm going to move this back into **draft** while I go about implementing resource-manager

builtin/aws/apprunner/destroyer.go
thiskevinwang: I've seem some operations take >7 minutes, so 10 minutes seemed like a reasonable value with _some buffer_ :|

builtin/aws/apprunner/platform.go
thiskevinwang: Hmm I guess not — this mostly stems from my lack of understanding of the exact intended usage of Steps/StepGroups 🤔

builtin/aws/apprunner/destroyer.go
thiskevinwang: Added a missing check for this error

builtin/aws/apprunner/platform.go
thiskevinwang: @paladin-devops I'm not certain if `waypoint destroy` _can_ but maybe you can help confirm/deny? (Context below 👇 )

(I've since moved the `Destroy` functionality from release-manager to the platform — I don't think this changes too much.)

Thinking out load + I have loose understanding here, so take this with a grain of salt: The `Destroy` function depends on a `Deployment` protobuf so that it can access the `.ServiceArn` field to know which service to delete.
If `Deploy` fails,would `Destroy` have access to a `deployment *Deployment` protobuf? 
- I think this state is only expected to ever occur on a _very first_ `waypoint up` for an AppRunner project.

briancain:
Hey there @thiskevinwang - what's the status on this PR? If there's still work to do, it might make sense to keep your branch active but close this PR to save you the trouble of fixing rebases while it goes stale.

PR #3943, opened 2022-09-28
Make UserFromContext public.

teresamychu:
Make UserFromContext a public function to be able to grab the user info from the context.


briancain (approved):
This makes sense I think. It should be safe to make this "public"?

teresamychu:
> This makes sense I think. It should be safe to make this "public"?

I think so? It was actually a bit odd where `UserWithContext` which is essentially the setter is exported, but `UserFromContext` (getter) was not exported. 

lol apologies for the private/public lingo, i hail from java land. 😛

izaaklauer (approved):

Issue #3677, opened 2022-08-15
There is no way to destroy all the aws resources that get created with a runner install if the runner install fails


cicoyle:
**Describe the bug**
Runner install failed, but created some resources. I wanted to run runner uninstall to clean up the resources that were created, but am unable to.

```
$ waypoint runner install \             
  -platform=ecs \
  -server-addr=api.hashicorp.cloud:443 \
  -ecs-runner-image=ghcr.io/hashicorp/waypoint/alpha \
  -ecs-cluster=cassie-cluster \
  -ecs-region=us-west-2
✓ Finished connecting to: api.hashicorp.cloud:443
❌ Installing runner...
✓ Networking setup
✓ EFS ready
✓ Found existing IAM role to use: waypoint-runner-execution-role
✓ Found existing IAM role to use: waypoint-runner
✓ Using existing log group
❌ Using existing security group: waypoint-server-security-group
! Error installing runner: ClusterNotFoundException: Cluster not found.

$ waypoint runner list
No runners found

$ waypoint runner uninstall -platform=ecs
! The -id flag is required

```

**Steps to Reproduce**
Start w/ no cluster existing, so it fails on runner install.

**Expected behavior**
On a runner uninstall, I expect all resources to be cleaned up as well. In my case aws ecs, but in other plugin installs as well.

paladin-devops:
This is a great idea for an improvement to the runner (and server) install processes. Currently if either fails, the corresponding uninstall command won't work. It would be great to update our uninstall commands to have the ability to look up the resources in a given platform and uninstall them.

Currently we do this by runner ID as well as with configuration flags for a given platform. If we can still provide the would-be ID of the runner to the user if the install fails, and we allow them to specify all of the configuration options necessary to look up each of the resources created during the install, I think this is doable!

krantzinator:
Fixed with #3944

PR #3944, opened 2022-09-28
Runner uninstall enhancements

paladin-devops:
This PR makes a few updates to `runner uninstall`, some general, and some specific to ECS. This addresses #3677. 

Regarding the general use of `runner uninstall`, if the install operation fails, a warning message is now output to let the user know that they can _attempt_ to uninstall the runner (which failed to be installed) with `runner uninstall`:

```terminal
$ waypoint runner install -platform=ecs -server-addr=<redacted>:9701 -server-tls-skip-verify -ecs-cluster=paladin-devops-cluster -ecs-region=ca-central-1
✓ Finished connecting to: <redacted>:9701
❌ Installing runner...
✓ Networking setup
✓ EFS ready
✓ Found existing IAM role to use: waypoint-runner-execution-role
✓ Found existing IAM role to use: waypoint-runner
✓ Using existing log group
❌ Using existing security group: waypoint-server-security-group
! Error installing runner: ClusterNotFoundException: Cluster not found.
Please run the following to clean up the resources from the unsuccessful runner installation,
specifying additional platform flags as needed:

waypoint runner uninstall -platform=ecs -id=01GE2MABHDJ7JPXGE5JDPTH86R <additional_platform_flags>
```

The other change being made for the general usage of `runner uninstall` is that we will check if a runner is registered to the server before attempting to forget it.

For ECS specifically, this PR adds a tag to the EFS volume which is created for the runner, and the tag contains the ID of the runner. When we uninstall the runner now, the EFS volume will also be deleted.

The log group, security group, IAM task role, and other resources are still **NOT** deleted, because they may be shared by other runners.

Tested scenarios:
- [x] runner install with existing server in ECS
- [x] runner uninstall with existing server in ECS
- [x] server install w/static runner to ECS
- [x] server uninstall w/static runner from ECS
- [ ] server upgrade w/static runner, from v0.10.1 in ECS

### DIFF
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -268,6 +268,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		c.ui.Output("Error installing runner: %s", clierrors.Humanize(err),
 			terminal.WithErrorStyle(),
 		)
+		c.ui.Output(runnerInstallFailed, c.platform[0], id, terminal.WithWarningStyle())
 		return 1
 	}
 	s.Update("Runner %q installed successfully to %s", id, platform[0])
@@ -332,5 +333,12 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 var (
 	runnerInstalledButNotYetAdopted = strings.TrimSpace(`The installed runner must be adopted.
 Please run "waypoint runner adopt" before the runner can start accepting jobs.
+`)
+
+	runnerInstallFailed = strings.TrimSpace(`
+Please run the following to clean up the resources from the unsuccessful runner installation,
+specifying additional platform flags as needed:
+
+waypoint runner uninstall -platform=%[1]s -id=%[2]s <additional_platform_flags>
 `)
 )
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -4,12 +4,13 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/posener/complete"
+
 	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/internal/clierrors"
 	"github.com/hashicorp/waypoint/internal/pkg/flag"
 	"github.com/hashicorp/waypoint/internal/runnerinstall"
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
-	"github.com/posener/complete"
 )
 
 type RunnerUninstallCommand struct {
@@ -164,11 +165,13 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 			if err != nil {
 				s.Update("Couldn't forget runner: %s", clierrors.Humanize(err))
 				s.Status(terminal.StatusWarn)
+				return 1
 			} else {
 				s.Update("Runner %q forgotten on server", c.id)
 				s.Status(terminal.StatusOK)
 			}
 			s.Done()
+			break
 		}
 	}
 
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -170,10 +170,10 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 				s.Update("Runner %q forgotten on server", c.id)
 				s.Status(terminal.StatusOK)
 			}
-			s.Done()
 			break
 		}
 	}
+	s.Done()
 
 	// TODO: Remove runner profiles associated solely with this runner
 
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -137,8 +137,6 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 		Id:         c.id,
 	})
 	if err != nil {
-		s.Update("Unable to uninstall runner")
-		s.Abort()
 		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 		return 1
 	}
--- a/internal/cli/runner_uninstall.go
+++ b/internal/cli/runner_uninstall.go
@@ -149,17 +149,30 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
 	s = sg.Add("Forgetting runner %q on server...", c.id)
 	defer func() { s.Abort() }()
 
-	_, err = c.project.Client().ForgetRunner(ctx, &pb.ForgetRunnerRequest{
-		RunnerId: c.id,
-	})
+	runners, err := c.project.Client().ListRunners(ctx, &pb.ListRunnersRequest{})
 	if err != nil {
-		s.Update("Couldn't forget runner: %s", clierrors.Humanize(err))
-		s.Status(terminal.StatusWarn)
-	} else {
-		s.Update("Runner %q forgotten on server", c.id)
-		s.Status(terminal.StatusOK)
+		c.ui.Output("Error while listing runners: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
 	}
-	s.Done()
+
+	// We loop and check for the runner in the list of runners, so we don't
+	// fail if it doesn't exist. We allow this so that users can "uninstall"
+	// runners which failed to be installed, and never registered with the server
+	for _, runner := range runners.Runners {
+		if runner.Id == c.id {
+			_, err = c.project.Client().ForgetRunner(ctx, &pb.ForgetRunnerRequest{
+				RunnerId: c.id,
+			})
+			if err != nil {
+				s.Update("Couldn't forget runner: %s", clierrors.Humanize(err))
+				s.Status(terminal.StatusWarn)
+			} else {
+				s.Update("Runner %q forgotten on server", c.id)
+				s.Status(terminal.StatusOK)
+			}
+			s.Done()
+		}
+	}
 
 	// TODO: Remove runner profiles associated solely with this runner
 
--- a/internal/installutil/aws/ecs.go
+++ b/internal/installutil/aws/ecs.go
@@ -124,6 +124,7 @@ func SetupEFS(
 	ui terminal.UI,
 	sess *session.Session,
 	netInfo *NetworkInformation,
+	efsTags []*efs.Tag,
 ) (*EfsInformation, error) {
 	sg := ui.StepGroup()
 	defer sg.Wait()
@@ -134,16 +135,10 @@ func SetupEFS(
 	efsSvc := efs.New(sess)
 	ulid, _ := component.Id()
 
-	// TODO: Use different tags for runner volume
 	fsd, err := efsSvc.CreateFileSystem(&efs.CreateFileSystemInput{
 		CreationToken: aws.String(ulid),
 		Encrypted:     aws.Bool(true),
-		Tags: []*efs.Tag{
-			{
-				Key:   aws.String(defaultServerTagName),
-				Value: aws.String(defaultServerTagValue),
-			},
-		},
+		Tags:          efsTags,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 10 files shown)


internal/cli/runner_install.go
```diff
@@ -268,6 +268,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		c.ui.Output("Error installing runner: %s", clierrors.Humanize(err),
 			terminal.WithErrorStyle(),
 		)
+		c.ui.Output(runnerInstallFailed, c.platform[0], id, terminal.WithWarningStyle())
```
briancain, line 271: Pretty sure it should be this? Otherwise I think it's the same platform?

```suggestion
		c.ui.Output(runnerInstallFailed, platform, id, terminal.WithWarningStyle())
```

Or hrm, why is `c.platform` a slice string? Does runner install allow for installing to multiple platforms?

internal/cli/runner_uninstall.go
```diff
@@ -149,17 +147,30 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
...
+	// fail if it doesn't exist. We allow this so that users can "uninstall"
+	// runners which failed to be installed, and never registered with the server
+	for _, runner := range runners.Runners {
+		if runner.Id == c.id {
+			_, err = c.project.Client().ForgetRunner(ctx, &pb.ForgetRunnerRequest{
+				RunnerId: c.id,
+			})
+			if err != nil {
+				s.Update("Couldn't forget runner: %s", clierrors.Humanize(err))
+				s.Status(terminal.StatusWarn)
```
briancain, line 166: This should probably be a ui.Output. Otherwise the next iteration, for either case, it'll get wiped due to how StepGroups work and the err will be lost.

internal/cli/runner_uninstall.go
```diff
@@ -149,17 +147,30 @@ func (c *RunnerUninstallCommand) Run(args []string) int {
...
+			} else {
+				s.Update("Runner %q forgotten on server", c.id)
+				s.Status(terminal.StatusOK)
+			}
+			s.Done()
+		}
 	}
-	s.Done()
```
briancain, line 162: We might still need this done, especially in the case where the length of `runners.Runners` is 0?

internal/cli/runner_uninstall.go
catsby: @briancain wouldn't that next iteration only happen if 2 runners have the same ID?

briancain: Ah true, should we `break` out of this loop them for the `if runner.Id == c.id` case?

internal/serverinstall/ecs.go
```diff
@@ -158,7 +158,14 @@ func (i *ECSInstaller) Install(
 				return err
 			}
 
-			if netInfo, err = awsinstallutil.SetupNetworking(ctx, ui, sess, i.config.Subnets); err != nil {
+			grpcPort, _ := strconv.Atoi(serverconfig.DefaultGRPCPort)
+			httpPort, _ := strconv.Atoi(serverconfig.DefaultHTTPPort)
+			ports := []*int64{
+				aws.Int64(int64(grpcPort)),
+				aws.Int64(int64(httpPort)), // TODO: Not needed for runner install
```
catsby, line 165: What is `TODO` about this? Just curious. The Runner install doesn't doesn't seem to include it or the `grpcPort`

catsby (changes requested):
Questions and suggestions specifically with efs deleting

internal/cli/runner_install.go
paladin-devops: @briancain it's a slice because this var is an `EnumVar` 😄 

```go
                f.EnumVar(&flag.EnumVar{
			Name:   "platform",
			Usage:  "Platform to install the Waypoint runner into. If unset, uses the platform of the local context.",
			Values: sortedPlatformNames,
			Target: &c.platform,
		})
```

internal/cli/runner_uninstall.go
paladin-devops: +1 to return 1, and I added a `break` at the conclusion of the `if` statement, since we don't need to continue looping if we found and successfully forgot the runner.

internal/runnerinstall/ecs.go
```diff
@@ -330,14 +345,76 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
 	}
 	s.Done()
+
+	// TODO: Still attempt to delete the EFS volume if the ECS service
+	// uninstall fails
+	efsSvc := efs.New(sess)
+	fileSystemsResp, err := efsSvc.DescribeFileSystems(&efs.DescribeFileSystemsInput{
+		CreationToken: nil,
+		FileSystemId:  nil,
+		Marker:        nil,
```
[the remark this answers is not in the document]
paladin-devops: Fixed here! https://github.com/hashicorp/waypoint/pull/3944/commits/0f91f3e67be1ea8a70cf9c95cd4f087635d85f0e

internal/runnerinstall/ecs.go
```diff
@@ -330,14 +345,76 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
+		for _, tag := range fileSystem.Tags {
+			if *tag.Key == "runner-id" && *tag.Value == opts.Id {
+				describeAccessPointsResp, err := efsSvc.DescribeAccessPoints(&efs.DescribeAccessPointsInput{
+					FileSystemId: fileSystem.FileSystemId,
+				})
+				if err != nil {
+					return err
+				}
+				for _, accessPoint := range describeAccessPointsResp.AccessPoints {
+					_, err = efsSvc.DeleteAccessPoint(&efs.DeleteAccessPointInput{AccessPointId: accessPoint.AccessPointId})
```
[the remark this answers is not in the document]
paladin-devops: As I understand it, it is required to delete the access point prior to the mount points.

The purpose for deleting the file system for the runner and not the server is that the server file system contains the server database. If we delete that via the CLI command, then having a chance to recover the database becomes much more difficult. The file system for the runner contains only a token and the runner ID, so it can be deleted with a low risk to impacting the user as they more likely won't need to "recover" a token or runner ID.

internal/serverinstall/ecs.go
paladin-devops: The ports set here ultimately result in a security group being created that is shared by both the server and the runner. The server needs the HTTP port but the runner has no use for it. The TODO is getting at that we might want to split them up somehow, just to tighten the SG a bit.

internal/runnerinstall/ecs.go
```diff
@@ -330,14 +345,76 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
+		Marker:        nil,
+		MaxItems:      nil,
+	})
+	if err != nil {
+		return err
+	}
+	for _, fileSystem := range fileSystemsResp.FileSystems {
+		// Check if tags match ID, if so then delete things
+		for _, tag := range fileSystem.Tags {
+			if *tag.Key == "runner-id" && *tag.Value == opts.Id {
```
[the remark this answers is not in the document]
paladin-devops: > use the marker and handle pagination
Updated this to use `Marker` before reading this comment based on your earlier comment.

I'm okay with refactoring the deletion out of the loop, and saving the ID once we have it to be used then. Will update shortly.

internal/runnerinstall/ecs.go
```diff
@@ -330,14 +345,76 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
+
+				for {
+					ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+					defer cancel()
+					select {
+					case <-ctx.Done():
+						return errors.New("after 5 minutes, the file system could" +
+							"not be deleted, because the mount targets weren't deleted")
+					default:
+						_, err = efsSvc.DeleteFileSystem(&efs.DeleteFileSystemInput{FileSystemId: fileSystem.FileSystemId})
```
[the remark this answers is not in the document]
paladin-devops: I agree, will add polling or a sleep here!

internal/runnerinstall/ecs.go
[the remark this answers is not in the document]
catsby: I'm not sure I follow; when we uninstall the server, we do in fact delete the file system, correct? I'm just pointing out the differences here and the server uninstall path. If this code path works correctly and everything is cleaned up then I suppose there's no issue.

internal/serverinstall/ecs.go
catsby: 👍

internal/runnerinstall/ecs.go
```diff
@@ -160,12 +158,26 @@ func (i *ECSRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
...
 			i.netInfo = netInfo
 
-			if efsInfo, err = awsinstallutil.SetupEFS(ctx, ui, sess, netInfo); err != nil {
+			efsTags := []*efs.Tag{
+				{
+					Key:   aws.String(DefaultRunnerTagName),
+					Value: aws.String(defaultRunnerTagValue),
+				},
+				{
+					Key:   aws.String("runner-id"),
```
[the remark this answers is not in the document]
paladin-devops: > why is DefaultRunnerTagName exported but defaultRunnerTagValue is not?
Perhaps at one point it may have been used in the `installutil` or `serverinstall` packages, but it appears that if that was the case, it no longer is. Will change to unxported.

Also, I've updated "runner-id" to be a const called "defaultRunnerIdTagName".

internal/cli/runner_install.go
briancain: Strange! Yeah I see EnumVar's target is a slice, but it's only ever valid for 1 flag haha.

internal/runnerinstall/ecs.go
```diff
@@ -160,12 +160,26 @@ func (i *ECSRunnerInstaller) Install(ctx context.Context, opts *InstallOpts) err
 				return err
 			}
 
-			if netInfo, err = awsinstallutil.SetupNetworking(ctx, ui, sess, i.Config.Subnets); err != nil {
+			// TODO: Add a port here if the user specified the -liveness-tcp-addr flag
```
briancain, line 163: Does this need to be turned into a GH issue to be tracked?

internal/runnerinstall/ecs.go
```diff
@@ -330,15 +347,91 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
+		// Check if tags match ID, if so then delete things
+		for _, tag := range fileSystem.Tags {
+			if *tag.Key == "runner-id" && *tag.Value == opts.Id {
+				fileSystemId = fileSystem.FileSystemId
+				goto DeleteFileSystem
+			}
+		}
+	}
+
+DeleteFileSystem:
```
briancain, line 390: Delete this label? It doesn't look like it's used

briancain (approved):

internal/runnerinstall/ecs.go
paladin-devops: Sure, I'll add one!

internal/runnerinstall/ecs.go
paladin-devops: It is used [here](https://github.com/hashicorp/waypoint/pull/3944/files#diff-af12170bdcf18b294209fa40b9af88dc690c261e30f331fc7f520b1798b5a385R385)!

internal/runnerinstall/ecs.go
paladin-devops: https://github.com/hashicorp/waypoint/issues/4254

xiaolin-ninja (approved):

internal/installutil/aws/ecs.go
```diff
@@ -298,6 +288,7 @@ func CreateService(serviceInput *ecs.CreateServiceInput, ecsSvc *ecs.ECS) (*ecs.
 }
 
 // TODO: Add runner ID as tag
+// SetupExecutionRole
```
xiaolin-ninja, line 291: Is this comment half-finished with the explanation of the method?

PR #3947, opened 2022-09-28
Externalize auth method

catsby:
Wrap error returns in [hcerr.Externalize()](https://github.com/hashicorp/waypoint/blob/52e13de50ac6da2ffedb9c34f17f592c76de9fe1/pkg/server/hcerr/hcerr.go#L16-L29) to log more details correctly, and keep some details in logs as opposed to surfacing them to users. 

This PR builds off of #3940 and should only be merged afterwards


briancain (approved):
At first glance this looks OK to me! 👍🏻

PR #3949, opened 2022-09-28
Detected deleted projects in long-running EntrypointConfig

evanphx:
When a project is deleted, the EntrypointConfig could still access that project. This makes it such that the EntrypointConfig won't refresh config when the project has been deleted.

### DIFF
--- a/pkg/server/singleprocess/service_entrypoint.go
+++ b/pkg/server/singleprocess/service_entrypoint.go
@@ -111,13 +111,13 @@ func (s *Service) EntrypointConfig(
 		// Refresh the application record in case it's been deleted
 		_, err = s.state(ctx).AppGet(deployment.Application)
 		if err != nil {
-			log.Warn("detecting application has been removing in entrypoint",
+			log.Warn("detected removed application in entrypoint",
 				"project", deployment.Application.Project,
 				"application", deployment.Application.Application,
 				"lookup-error", err,
 			)
 
-			log.Warn("exitting EntrypointConfig because project was deleted")
+			log.Warn("exiting EntrypointConfig because project was deleted")
 			return status.Error(codes.Unavailable, "project has been deleted")
 		}
 
--- a/pkg/server/singleprocess/service_entrypoint.go
+++ b/pkg/server/singleprocess/service_entrypoint.go
@@ -108,6 +108,19 @@ func (s *Service) EntrypointConfig(
 			}
 		}
 
+		// Refresh the application record in case it's been deleted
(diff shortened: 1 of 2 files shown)


paladin-devops (approved):
Good catch! Think we need a changelog since `project destroy` has been released already, but otherwise approved ✅

pkg/server/singleprocess/service_entrypoint.go
```diff
@@ -108,6 +108,19 @@ func (s *Service) EntrypointConfig(
...
+		// Refresh the application record in case it's been deleted
+		_, err = s.state(ctx).AppGet(deployment.Application)
+		if err != nil {
+			log.Warn("detecting application has been removing in entrypoint",
+				"project", deployment.Application.Project,
+				"application", deployment.Application.Application,
+				"lookup-error", err,
+			)
+
+			log.Warn("exitting EntrypointConfig because project was deleted")
```
catsby, line 120:
```suggestion
			log.Warn("exiting EntrypointConfig because project was deleted")
```

pkg/server/singleprocess/service_entrypoint.go
```diff
@@ -108,6 +108,19 @@ func (s *Service) EntrypointConfig(
 			}
 		}
 
+		// Refresh the application record in case it's been deleted
+		_, err = s.state(ctx).AppGet(deployment.Application)
+		if err != nil {
+			log.Warn("detecting application has been removing in entrypoint",
```
catsby, line 114:
```suggestion
			log.Warn("detected removed application in entrypoint",
```

pkg/server/singleprocess/service_entrypoint.go
```diff
@@ -108,6 +108,19 @@ func (s *Service) EntrypointConfig(
...
+				"project", deployment.Application.Project,
+				"application", deployment.Application.Application,
+				"lookup-error", err,
+			)
+
+			log.Warn("exitting EntrypointConfig because project was deleted")
+			return status.Error(codes.Unavailable, "project has been deleted")
```
catsby, line 121:
```suggestion
			return hcerr.Externalize(log, status.Error(codes.Unavailable, "project has been deleted"))
```

catsby (approved):
👍 minor suggestions

pkg/server/singleprocess/service_entrypoint.go
evanphx: The status.Error is already externalized, no need to call hcerr.Externalize here.

briancain (approved):

PR #3950, opened 2022-09-29
Add more Helm options

denysvitali:
This PR adds support for the following new Helm options:

- `create_namespace`
- `skip_crds`

### DIFF
--- a/builtin/k8s/helm/platform.go
+++ b/builtin/k8s/helm/platform.go
@@ -117,12 +117,12 @@ func (p *Platform) Deploy(
 		client.NameTemplate = ""
 		client.OutputDir = ""
 		client.Atomic = false
-		client.SkipCRDs = false
+		client.SkipCRDs = p.config.SkipCRDs
 		client.SubNotes = true
 		client.DisableOpenAPIValidation = false
 		client.Replace = false
 		client.Description = ""
-		client.CreateNamespace = true
+		client.CreateNamespace = p.config.CreateNamespace
 
 		s.Update("Installing Chart...")
 		rel, err := client.Run(c, values)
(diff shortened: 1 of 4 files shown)


builtin/k8s/helm/platform.go
```diff
@@ -376,11 +378,27 @@ deploy {
...
+		"create_namespace",
+		"Create Namespace if it doesn't exist.",
+		docs.Summary(
+			"This option will instruct Helm to create a namespace if it doesn't exist.",
+		),
+	)
+
+	doc.SetField(
+		"skip_crds",
+		"Do not create CRDss",
```
paladin-devops, line 396:
```suggestion
		"Do not create CRDs",
```

builtin/k8s/helm/platform.go
```diff
@@ -117,12 +117,12 @@ func (p *Platform) Deploy(
 		client.NameTemplate = ""
 		client.OutputDir = ""
 		client.Atomic = false
-		client.SkipCRDs = false
+		client.SkipCRDs = p.config.SkipCRDs
```
paladin-devops, line 120: We should probably also make this configurable for _upgrading_ Helm charts. In the snippet below, it's hard-coded to false as well. Making this configurable here is only for the 1st installation of the chart.

https://github.com/hashicorp/waypoint/blob/ded15e17ddd6c2bcb896665efd6ec7de559d8508/builtin/k8s/helm/platform.go#L152

paladin-devops (commented):
Looking great @denysvitali! In addition to my feedback, I think we also need to generate the documentation for this PR's plugin changes. With our Makefile, you can run `make gen/website-mdx` to do that automatically. 😄 I've run into issues with generating the docs in the past myself as a contributor, but it works like a charm now! Let us know if you need a hand.

builtin/k8s/helm/platform.go
denysvitali: Oh, I missed that, thanks

builtin/k8s/helm/platform.go
denysvitali: Missed this part too, good point!

paladin-devops (approved):
⛑️ 🎉  thanks @denysvitali!

paladin-devops (approved):

PR #3952, opened 2022-09-29
Externalize errors in service_auth_oidc

catsby:
Like https://github.com/hashicorp/waypoint/pull/3947, but for `singleprocess/service_auth_oidc.go`


briancain (approved):
At a quick glance, looks OK to me! 👍🏻

PR #3960, opened 2022-09-29
core/cli: Emit warning a week before used token expires

demophoon:
This commit adds a locally run check before creating a server client which commands use to talk to Waypoint server. This check emits a warning when a token is set to expire within 7 days so users can reauthenticate with Waypoint.

### DIFF
--- a/internal/cli/base_init.go
+++ b/internal/cli/base_init.go
@@ -4,11 +4,16 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"time"
 
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
 	"github.com/hashicorp/waypoint/internal/clicontext"
 	clientpkg "github.com/hashicorp/waypoint/internal/client"
 	configpkg "github.com/hashicorp/waypoint/internal/config"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
 	"github.com/hashicorp/waypoint/pkg/serverclient"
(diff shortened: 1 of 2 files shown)


internal/cli/base_init.go
```diff
@@ -63,6 +68,44 @@ func (c *baseCommand) initConfigLoad(path string) (*configpkg.Config, configpkg.
...
+// Given a duration, check the clientContext's auth token to determine if the
+// token expires within the duration.
+// Emits a warning on the cli if token expires within the duration.
+// Returns any errors which may have occurred parsing the token.
+//
+// This function provides no guarantee that the token itself is valid as
+// validation checks can only happen on the server.
+func (c *baseCommand) checkTokenExpiry(duration time.Duration) error {
+	token := c.clientContext.Server.AuthToken
+	tokenMagic := "wp24"
```
briancain, line 80: This might be stored as a const, maybe we can use this instead? I think it's this:

https://github.com/hashicorp/waypoint/blob/4ff589ab89363e83fbdf031e13382b7b42861268/pkg/tokenutil/decode.go#L17-L22

Actually maybe we could just use that whole function instead and use the return values to handle the `body.ValidUntil`?

https://github.com/hashicorp/waypoint/blob/4ff589ab89363e83fbdf031e13382b7b42861268/pkg/tokenutil/decode.go#L24-L27

briancain:
What's the status of this one @demophoon ?

PR #3962, opened 2022-09-30
WAYP-404 Log context canceled errors in trace level

teresamychu:
Logs were being polluted by an error caused by context cancellation. 

The solution is a little more widespread. We basically catch any error that is a `context cancellation` error and log it at trace level. Any other errors will be logged through the normal flow.

This PR essentially assumes that most (if not all) context cancellation errors are expected errors and are mostly user produced when a user kills a build or cancels an action.

### DIFF
--- a/pkg/server/hcerr/hcerr.go
+++ b/pkg/server/hcerr/hcerr.go
@@ -1,6 +1,7 @@
 package hcerr
 
 import (
+	"context"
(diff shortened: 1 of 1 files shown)


briancain (approved):

izaaklauer (commented):
👍

PR #3967, opened 2022-09-30
Externalize multiple services

catsby:
More externalizing errors like https://github.com/hashicorp/waypoint/pull/3956

### DIFF
--- a/pkg/server/singleprocess/service_entrypoint.go
+++ b/pkg/server/singleprocess/service_entrypoint.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/hashicorp/waypoint/internal/server/boltdbstate"
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"github.com/hashicorp/waypoint/pkg/server/hcerr"
 	"github.com/hashicorp/waypoint/pkg/server/logbuffer"
 	"github.com/hashicorp/waypoint/pkg/server/ptypes"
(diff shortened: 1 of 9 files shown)


pkg/server/singleprocess/service_entrypoint.go
```diff
@@ -315,32 +368,52 @@ func (s *Service) EntrypointExecStream(
...
 	}
 	log = log.With("instance_id", exec.InstanceId, "index", open.Open.Index)
 
 	// Mark we're connected
 	if !atomic.CompareAndSwapUint32(&exec.Connected, 0, 1) {
-		return status.Errorf(codes.FailedPrecondition,
-			"exec session is already open for this index")
+		return hcerr.Externalize(
+			log,
+			status.Errorf(codes.FailedPrecondition,
```
izaaklauer, line 413: I think it's ok to return the status.Errorf here because we're not wrapping any error from deeper in the codebase - we already have total control over what it looks like, so we don't really need the `externalize` here.

But i'm ok with adding externalize for consistency.

izaaklauer (approved):

pkg/server/singleprocess/service_entrypoint.go
catsby: 👍 I think I'll leave it for now but that is something I noticed later on (not wrapping errors) and I've started to not convert those for the reason(s) you point out.

PR #3973, opened 2022-09-30
Typo in example code.

SunSparc:


website/content/docs/extending-waypoint/creating-plugins/build-interface.mdx
```diff
@@ -95,14 +95,14 @@ plugin. To do this, you can use `terminal.UI`; this allows you to write output t
 all Waypoint plugins.
 
 The method `ui.Status()` requests a live updating status; you use this to write updates to the Waypoint terminal output.
-When finished with `Status` you should always close a Status, using`st.Close()`; this ensures that the status is updated
+When finished with `Status` you should always close a Status, using`u.Close()`; this ensures that the status is updated
```
briancain, line 98: Since `u` doesn't really exist until the example on 103, can we move the `go` code block above this paragraph in between 95-97? Otherwise this seems a little confusing if you haven't read ahead as to where `u` comes from.

briancain (changes requested):
Thanks for taking the time to make a pull request to fix up some docs! I have a change request, but otherwise looks good 👍🏻

SunSparc (commented):
Good idea.

briancain (approved):
Thank you!

PR #3981, opened 2022-10-03
website: rename files per changes on stable-website

krantzinator:
The links we have at the bottom of our website docs to `View this on GitHub` are broken for all of these items, since they don't exit on `main` at the same paths as on `stable-website`. Fixing that! Which will also hopefully prevent future cherry pick failures on backports when we edit these pages again.


krantzinator:
Actually, #4050 is what we want

Issue #1469, opened 2021-05-13
internal/core: StepGroups get cut off or ignored if func executes quick


briancain:
**Describe the bug**

As a plugin author, if a func is quick to execute and uses StepGroups for iterative output, the channel or stream gets cancelled early, skipping much of the output for that given func. If one adds a `sleep`, the remaining output will show up, which seems to indicate a stream is getting cancelled the moment the func returns before any remaining StepGroup output has time to finish printing.

**Steps to Reproduce**

Create or modify a func in one of the platform plugins and attempt to print some iterative output using StepGroups or any UI Output. The func shouldn't take too long, and the output gets cut off when the func returns.

**Expected behavior**

The output should be able to finish printing to the terminal instead of getting cut off or skipped entirely

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: any
* Waypoint Server Platform and Version: any
* Waypoint Plugin: any

**Additional context**

This might be a blocker for status reports, as these funcs end up being pretty quick and only a single API call to the requested platform. The UI in this case ends up getting skipped entirely.

briancain:
Potentially fixed by https://github.com/hashicorp/waypoint/pull/1480, I think.

PR #3985, opened 2022-10-04
Null plugin

izaaklauer:
Adds a new "null" plugin. The first intended usecase is loadtesting, but I imagine this will be useful for the core waypoint team for other kinds of testing.

### DIFF
--- a/builtin/null/platform.go
+++ b/builtin/null/platform.go
@@ -1,0 +1,90 @@
+package null
+
+import (
+	"time"
+
+	sdk "github.com/hashicorp/waypoint-plugin-sdk/proto/gen"
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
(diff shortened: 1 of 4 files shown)


builtin/null/platform.go
```diff
@@ -0,0 +1,90 @@
...
+	return &p.config, nil
+}
+
+// DeployFunc implements component.Platform
+func (p *Platform) DeployFunc() interface{} {
+	return func(ui terminal.UI) *emptypb.Empty {
+		sg := ui.StepGroup()
+		step := sg.Add("performing null deploy")
+
+		time.Sleep(time.Second * 2)
```
briancain, line 28: Heh, this must be because of https://github.com/hashicorp/waypoint/issues/1469

briancain (approved):

PR #3988, opened 2022-10-04
Vault Config Sourcer Fixes

### LINKED ISSUE #2333: Waypoint + Vault - Deployment not able to start
Attempting to use Waypoint's Vault integration for configDynamic variables results in apps not starting on GKE. Watching the logs, I can verify that Waypoint fetches the secret accordingly, entrypoint detects the config changes, but then the deployment hangs:

```
2021-09-21T15:43:32.638Z [DEBUG] entrypoint.url: discovering hubs
2021-09-21T15:43:32.638Z [DEBUG] entrypoint.url: refreshing data
2021-09-21T15:43:33.119Z [DEBUG] entrypoint.url.agent: connecting to hub: addr=34.208.154.111:222
2021-09-21T15:43:33.130Z [DEBUG] entrypoint.config.watcher.watchloop: config variables changed, calling Stop: source=vault
2021-09-21T15:43:33.134Z [DEBUG] entrypoint.config.watcher.watchloop: initializing the Vault client: source=vault
2021-09-21T15:43:33.162Z [DEBUG] entrypoint.config.watcher.watchloop: secret is renewable, starting renewer: key=password path=database/creds/my-role source=vault
2021-09-21T15:43:33.455Z [DEBUG] entrypoint.url.agent: connection latency: latency=154.600812ms
2021-09-21T15:43:33.456Z [DEBUG] entrypoint.url.agent: connected successfully: status=connected latency=154.600812ms skew=48.37914ms
2021-09-21T15:43:37.941Z [DEBUG] entrypoint.url.agent: connected to hub: addr=34.208.154.111:222
```

After this final message, the pod restarts and loops through the process again. 

```
» Deploying...
✓ Kubernetes client connected to https://35.203.6.111 with namespace app2
✓ Creating deployment...
❌ Waiting on deployment to become available: 1/1/0
✓ Kubernetes client connected to https://35.203.6.222 with namespace app2
✓ Deleting deployment...
! Deployment was not able to start pods after 10m0s
```

If there is a misconfiguration anywhere, I see entrypoint stripping those configDynamic values and the app deploys without them. The issue only occurs when I see a successful lease issued. In this set up I am using Vault k8s auth method, a database engine and I simply modified the example Go project from [the Hashi Learn portal here](https://learn.hashicorp.com/tutorials/waypoint/static-application-configuration?in=waypoint/configuration). I obtained the same result when trying with a nodejs app as well.

paladin-devops:
Fixes #2333.

This PR fixes a bug where an application deployed with the Vault config sourcer plugin using a dynamic secrets engine in Vault was inaccessible. `waypoint exec` would hang forever, the address provided by the URL service would be inaccessible, as well as any other way to access the application outside of the URL service.

A subsequent bug which is also fixed in this PR is that dynamic secrets were previously being overwritten as an empty map when the secret was renewed. This is the normal behavior of the Vault lease API, as that API does not return the data of a secret whose lease is renewed. However, now, if a lease is successfully renewed, the Vault config sourcer plugin will still retain the data from when the secret was originally read.

Little bonus in this PR - the CEB binary will now include the git commit in its version, whereas previously only the last major version was logged as the version - this is the Makefile change.


builtin/vault/config_sourcer.go
```diff
@@ -240,25 +240,31 @@ func (cs *ConfigSourcer) stop() error {
...
 		cache.Err = err
 		return
 	}
+	log.Debug("lifetime watcher created")
 
+	log.Debug("starting renewer...")
 	// Start the renewer in the background
-	renewer.Renew()
+	go renewer.Renew()
+	log.Debug("renewer started")
```
briancain, line 267: Can we move these logs to TRACE?

builtin/vault/config_sourcer.go
```diff
@@ -305,8 +314,9 @@ func (cs *ConfigSourcer) startRenewer(client *vaultapi.Client, path string, s *v
 			}
 
 			cs.cacheMu.Unlock()
+			log.Debug("secret renewed")
```
briancain, line 317: These all seem good to log at TRACE as well

builtin/vault/config_sourcer.go
```diff
@@ -240,25 +240,31 @@ func (cs *ConfigSourcer) stop() error {
...
+	log.Debug("checking if secret is in cache...")
 	cache, ok := cs.secretCache[path]
 	if !ok {
 		return
 	}
+	log.Debug("secret is in cache")
 
-	renewer, err := client.NewRenewer(&vaultapi.RenewerInput{
+	log.Debug("creating new lifetime watcher...")
+	renewer, err := client.NewLifetimeWatcher(&vaultapi.LifetimeWatcherInput{
```
briancain, line 255: I think this is the main change right? It'd be nice if the commit message says what this changes and why it was broken before. I'm not entirely sure what this changes versus the `NewRenewer`

edit: Just peeped the Vault client source. Looks like `NewRenewer` was depreciated and called `NewLifetimeWatcher` anyways, so that seems like a good enough thing to include in the commit message

briancain: Should this function also now be renamed `startLifetimeWatcher`?

briancain (commented):
I think you still have some outstanding TODOs, but I got some review here anyways! ✨

builtin/vault/config_sourcer.go
paladin-devops: done and done!

izaaklauer (approved):

briancain (approved):
🖖🏻

PR #4021, opened 2022-10-10
jobstream: Fix log message with missing value

briancain:
This commit fixes the log message to properly use the key,val syntax for hclog rather than string interpolation.


izaaklauer (approved):
The classic blunder

PR #4042, opened 2022-10-12
Feature: Reattach to a pipeline run by latest or sequence id

### LINKED ISSUE #4010: cli: Provide a way to re-attach to a pipeline run

If the jobstream client fails for some reason, or if a user ctrl-c's during a pipeline run, they will loose access to watching the jobstream easily. They could manually attach to each stream with `waypoint job get-stream <jobid>` but that would be pretty tedious.

Provide a way for the CLI to re-attach to a running pipeline to continue following the job stream.

Nothing really, other than manually getting each jobstream for the pipeline.

**Additional context**

briancain:
This pull request introduces a new CLI flag for `pipeline run` to enable the CLI to reattach to an existing pipeline run, either by the latest known run, or by a specific sequence id. It will attempt to reply the job ids in the order defined in the pipeline run message.

Fixes https://github.com/hashicorp/waypoint/issues/4010


internal/cli/pipeline_get_run.go
```diff
@@ -0,0 +1,232 @@
...
+			if _, ok := job.Operation.(*pb.Job_Noop_); ok {
+				continue
+			}
+
+			ws := "default"
+			if job.Workspace != nil {
+				ws = job.Workspace.Workspace
+			}
+			// TODO(briancain): fix me
+			//app.UI.Output("Executing Step %q in workspace: %q", resp.JobMap[jobId].Step, ws, terminal.WithHeaderStyle())
```
briancain, line 152: We might want to handle this in a follow-up PR prior to the next release, but ideally if we want to display step names next to job streams like we do for `waypoint pipeline run` we'll need to modify the Run API endpoints to calculate the JobMap in the same way `RunPipeline` does.

briancain: Oh wait this is easier than I thought. We store this directly on the job, so we can grab it there...

xiaolin-ninja (commented):
This feels like it's repeating code that's in the `pipeline run` command and introduces more maintenance surface and not very dry.

If the key is so the user knows that happens to the run, there is existing `waypoint job list -pipeline=<pipeline> -run=<seq>` to monitor the state of the job, or `waypoint pipeline inspect -run=<seq>` for the state of previous runs.

So they could just run `job get-stream` on the latest job... though I guess if the point is to re-attach them to the stream in real-time... maybe we can figure out how to extract this code so it's not repeated, or add a `-reattach` command to the existing `pipeline run`.

briancain:
@xiaolin-ninja so the main thing this provides is being able to replay or attach to an existing pipeline run (maybe that run was invoked in the UI or automatically by the server and a CLI client wants to follow along, for example).

Using `waypoint job get-stream` isn't a great user experience. They'd have to figure out what jobs are associated with pipeline runs, and then manually run the `get-stream` command for every job id. The appeal of something like this PR is only having to provide a pipeline name or run sequence number and replaying or reattaching to a running jobstream for all jobs in execution order.

I'm good with adding a `-reattach` flag to `waypoint pipeline run`. The majority of the code in the CLI command on this PR is building up protobufs for requests and sending them over the API so I don't think the duplication is a huge deal, but adding a `-reattach` to `waypoint pipeline run` makes sense.

briancain:
Done via [c4d5c28](https://github.com/hashicorp/waypoint/pull/4042/commits/c4d5c28bf251871bf57b3384f6f3ef7582b1d0e0) @xiaolin-ninja !

evanphx (approved):

xiaolin-ninja (approved):
Tested & looks great!

PR #4044, opened 2022-10-12
Adding SetConfig to the grpc gateway config

izaaklauer:
We missed it earlier. This also adds an endpoint for getting config by application. I'm not sure how we'd get globally scoped config vars, so I omitted that here - it's not a pressing need.


teresamychu (approved):
LGTM

cicoyle (approved):

PR #4045, opened 2022-10-12
Consul Config Sourcer Plugin

paladin-devops:
This PR adds a new config sourcer plugin to support dynamic configuration from the Consul KV data store.

Example config source settings:
`waypoint config source-set -type=consul -config=address=$CONSUL_HTTP_ADDR -config=token=$CONSUL_HTTP_TOKEN`

Example `waypoint.hcl` snippet using this plugin:
```hcl
  config {
    env = {
      "TEST" = dynamic("consul", {
        key  = "paladin"
      })
    }
  }
```

This plugin was authored by @mkeeler! 🎉


xiaolin-ninja (approved):

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+	CertFile      string `hcl:"cert_file,optional"`
+	CertPEM       []byte `hcl:"cert_pem,optional"`
+	KeyFile       string `hcl:"key_file,optional"`
+	KeyPEM        []byte `hcl:"key_pem,optional"`
+	InsecureHTTPs bool   `hcl:"insecure_https,optional"`
+}
+
+func (t *TLSConfig) toAPITLSConfig() *api.TLSConfig {
+	return &api.TLSConfig{
+		Address:            t.ServerName,
```
catsby, line 91: I assume the default golang value of `""` is acceptable input here?

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+	apiConfig := api.Config{
+		Address:    conf.Address,
+		Scheme:     conf.Scheme,
+		Datacenter: conf.Datacenter,
+		HttpAuth:   conf.HTTPAuth.toApiAuth(),
+		Token:      conf.Token,
+		TokenFile:  conf.TokenFile,
+		Namespace:  conf.Namespace,
+		Partition:  conf.Partition,
+		TLSConfig:  *conf.TLSConfig.toAPITLSConfig(),
```
catsby, line 71: nit: `toAPITLSConfig()` only seems to be used in a single place, it seems strange that it returns a pointer to a struct to only then be dereferenced because the 1 place that uses it expects a struct and not a pointer. Is there a reason it's designed this way?

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+		CAPem:              t.CAPem,
+		CertFile:           t.CertFile,
+		CertPEM:            t.CertPEM,
+		KeyFile:            t.KeyFile,
+		KeyPEM:             t.KeyPEM,
+		InsecureSkipVerify: t.InsecureHTTPs,
+	}
+}
+
+type ConsulHTTPAuth struct {
```
catsby, line 103: Does this need to be exported? Do we anticipate using it outside of the package?

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+func (cs *ConfigSourcer) Config() (interface{}, error) {
+	return &cs.config, nil
+}
+
+// Implement ConfigurableNotify
+func (cs *ConfigSourcer) ConfigSet(config interface{}) error {
+	conf, ok := config.(*ConfigSourcerConfig)
+	if !ok {
+		// The Waypoint SDK should ensure this never gets hit
+		return fmt.Errorf("Expected *ConfigSourcerConfig as parameter")
```
catsby, line 137: nit: in general, error strings should not be capitalized

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+		result.Result = &pb.ConfigSource_Value_Value{
+			Value: string(cacheVal.kvPair.Value),
+		}
+	}
+
+	return results, nil
+}
+
+func (cs *ConfigSourcer) startBlockingQuery(ctx context.Context, logger hclog.Logger, kvReq reqConfig, lastIndex uint64, opts *api.QueryOptions) {
+	go func() {
```
catsby, line 279: I think we should rename `startBlockingQuery` to something else, and perhaps a more idiomatic golang way of writing this would be to move the `go` goroutine call to the invocation, so instead of `cs.startBlockingQuery(refreshCtx, log, kvReq, meta.LastIndex, opts)` you have `go cs.startBlockingQuery(refreshCtx, log, kvReq, meta.LastIndex, opts)` and then remove the `go func(){}()` here. The reason being is that it's more intuitive to the reader that calling `startBlockingQuery` is supposed to be operating in a goroutine. 

For example, reading these lines:

```
cs.cache[kvReq.cacheKey()] = cacheVal
cs.startBlockingQuery(refreshCtx, log, kvReq, meta.LastIndex, opts)
```

a reader might think that `cs.startBlockingQuery()` in fact blocks here, and we wait until it returns, because after all the method name has `Blocking` in it... when in reality `startBlockingQuery()` doesn't block at all... it launches a goroutine and returns, regardless of whatever might block in that goroutine.

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+		// namespaces and partitions except with the global client defaults and thus isn't suitable for this usage.
+
+		for {
+			logger.Trace("Issuing blocking query", "wait-index", lastIndex)
+			// set the wait index to use for the query
+			opts.WaitIndex = lastIndex
+			pair, meta, err := cs.client.Get(kvReq.Key, opts)
+
+			// check if we are being stopped
+			select {
```
catsby, line 292: nit: this `select` should be the first thing in the `for` loop.

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			// KV entry not updated - do nothing
+			if meta != nil && meta.LastIndex == lastIndex {
+				logger.Trace("KV value unchanged")
```
catsby, line 300: should we sleep here for `retryInterval` duration?

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+				if retry > maxBackoffTime {
+					retry = maxBackoffTime
+				}
+				logger.Error("KV Get errored", "error", err, "retry", retry.String())
+				select {
+				case <-time.After(retry):
+				case <-ctx.Done():
+					return
+				}
+			}
```
catsby, line 336: Nit: reduce `if err == nil {...} else {...}` to just `if err != nil {...}`, less mental stack to keep track of when reading the code. Just a suggestion 

```suggestion
			if err != nil {
				// reset the last index to 0 to do a non-blocking query
				lastIndex = 0
				// Set up for exponental backoff
				failures++

				retry := retryInterval * time.Duration(failures*failures)
				if retry > maxBackoffTime {
					retry = maxBackoffTime
				}
				logger.Error("KV Get errored", "error", err, "retry", retry.String())
				select {
				case <-time.After(retry):
				        continue
				case <-ctx.Done():
					return
				}
			}
			lastIndex = meta.LastIndex
			failures = 0
```

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+	}()
+}
+
+func (cs *ConfigSourcer) stop() error {
```
catsby, line 341: Do we anticipate ever returning an `error` here?

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,410 @@
...
+			cs.mu.Unlock()
+
+			// now determine the next wait index
+			if err == nil {
+				lastIndex = meta.LastIndex
+				failures = 0
+			} else {
+				// reset the last index to 0 to do a non-blocking query
+				lastIndex = 0
+				// Set up for exponental backoff
```
catsby, line 323:
```suggestion
				// Set up for exponential backoff
```

catsby (commented):
A lot of nit-pick things, most of which are inconsequential but a few I'd like your thoughts on before 👍

builtin/consul/config_sourcer.go
catsby: I do realize that even in Effective Go there is an example of this style where a function or method immediately launches a goroutine, I just feel it's clearer to the reader to do `go functionName()`. If the method didn't have `Blocking` in the name I may not have even brought this up 🤷

builtin/consul/config_sourcer.go
paladin-devops: According to the Consul API documentation, this field is optional, so `""` should be acceptable.

from `github.com/hashicorp/consul/api/api.go`:
```
	// Address is the optional address of the Consul server. The port, if any
	// will be removed from here and this will be set to the ServerName of the
	// resulting config.
	Address string
```

builtin/consul/config_sourcer.go
briancain: I agree, I think it makes more sense to start a go routine on this function rather than making the whole body of this function wrapped in a `go func()`.

paladin-devops: @catsby I'm updating this for the function to be called using a goroutine, rather than the entire contents of the function being in an anonymous function, called as a goroutine.

Also, I think that the nomenclature might be adding to the confusion. Consul has "[Blocking Queries](https://developer.hashicorp.com/consul/api-docs/features/blocking)", which are "used to wait for a potential change using long polling." I'll add information about this to the godoc for `startBlockingQuery`, which I may rename to `startConsulBlockingQuery`.

builtin/consul/config_sourcer.go
paladin-devops: The select block at the link below, which is at the end of the for loop, should take care of waiting for us:

https://github.com/paladin-devops/waypoint/blob/5b671f4ee5809a42f2a3a68ed0aafeeccacbe8db/builtin/consul/config_sourcer.go#L333-L337

builtin/consul/config_sourcer.go
paladin-devops: No, nothing in this function really warrants returning an error. However, in all other config sourcer implementations I've seen, including in tests, the return type for `stop`, which is called by `StopFunc`, which satisfies the config sourcer interface from the SDK, is indeed `error`, and I elected not to stray from the path this time around. 😄

builtin/consul/config_sourcer.go
paladin-devops: Not particularly, I've updated this to:
```go
		TLSConfig: api.TLSConfig{
			Address:            conf.TLSConfig.ServerName,
			CAFile:             conf.TLSConfig.CAFile,
			CAPath:             conf.TLSConfig.CAPath,
			CAPem:              conf.TLSConfig.CAPem,
			CertFile:           conf.TLSConfig.CertFile,
			CertPEM:            conf.TLSConfig.CertPEM,
			KeyFile:            conf.TLSConfig.KeyFile,
			KeyPEM:             conf.TLSConfig.KeyPEM,
			InsecureSkipVerify: conf.TLSConfig.InsecureHTTPs,
		},
```

builtin/consul/config_sourcer.go
```diff
@@ -0,0 +1,412 @@
...
+
+type cachedVal struct {
+	kvPair *api.KVPair
+	cancel context.CancelFunc
+	err    error
+}
+
+// ConfigSourcerConfig is used to configure where to talk to Consul, and from
+// where the KV data is to be retrieved
+type ConfigSourcerConfig struct {
```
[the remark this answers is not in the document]
paladin-devops: They're not required to be set as configurations here, because the Consul API will attempt to retrieve the configuration from environment variables as well (not unlike Nomad), and will set reasonable defaults if they're not present. 

https://github.com/hashicorp/consul/blob/main/api/api.go#L439-L441

briancain (approved):
Sweet!

builtin/consul/config_sourcer.go
catsby: @paladin-devops that's good context to know about Consul queries, thanks

builtin/consul/config_sourcer.go
catsby: The `continue` statement immediately below this `logger.Trace()` call shoots us back up to the top of the `for` loop.

paladin-devops: [Updated this to sleep for 5 seconds](https://github.com/hashicorp/waypoint/pull/4045/commits/22a277f59dd452fc5946c331babb08bcaea4f2cd) before hitting the API again!

builtin/consul/config_sourcer.go
catsby: 👍 having actually looked at the sdk and interface now, this makes senes

catsby (approved):
👍 LGTM

PR #4050, opened 2022-10-13
[website] Sync stable-website with main

krantzinator:
Per conversation, we need to bring `stable-website` into sync with the ordering of these pages on `main`.


website/data/docs-nav-data.json
```diff
@@ -615,36 +615,53 @@
...
+      },
+      {
+        "title": "Community forum",
+        "href": "https://discuss.hashicorp.com/c/waypoint/"
       },
       {
-        "title": "Architecture",
-        "path": "internals/architecture"
+        "title": "Glossary",
+        "path": "resources/glossary"
```
catsby, line 651: does the path changing require any redirects or is that done somewhere else already?

BRKalow: Yeah good call! Can we also pull over these redirect definitions too?

https://github.com/hashicorp/waypoint/blob/9d06d53e4b4da92fcae1994b6e623fde867095be/website/redirects.js#L7-L26

krantzinator: @BRKalow should I just copy/paste the whole nav doc? There's a lot of extra stuff in there that we don't have on stable-website.

krantzinator: I'll push it up so y'all can see the diff, and then we can drop the commit if it doesn't look good.

BRKalow: @krantzinator sure we can do everything here! All of the k8s pages got shifted around. Let me know if I can help!

brkalow (approved):
@krantzinator I updated this so it should be in-sync with main! The deploy preview looks correct to me.

krantzinator:
Fantastic, thank you @BRKalow!

PR #4053, opened 2022-10-14
Bug: Fix pipeline run complete state machine

### LINKED ISSUE #4032: pipelines: Pipeline Run state machine isn't updating on completion
**Describe the bug**

For both the success and error case, it doesn't seem like the `PipelineRun` state machine is properly updating itself when a pipeline completes.

```
$ waypoint pipeline run simple-nested
...
...
✓ Running release v6
✓ Kubernetes client connected to https://10.96.0.1:443 with namespace default
✓ Updating existing service...
✓ Service is ready!
✔ Pipeline "simple-nested" (k8s-tetris) finished! 9/9 steps successfully completed.
brian@ubuntu:waypoint-tetris λ waypoint pipeline list                                                                                                ±[main]

» Waypoint Pipelines for k8s-tetris
              ID             |     NAME      |   OWNER    | CURRENT STEPS | LAST RUN STARTED | LAST RUN COMPLETED |  STATE  | TOTAL RUNS
-----------------------------+---------------+------------+---------------+------------------+--------------------+---------+-------------
  01GF4E3B17S3SF0WFMCQZYWWG5 | marathon      | k8s-tetris |             1 |                  |                    |         |
  01GF4E3B17VQ6M4JCVPJW5TNVB | up            | k8s-tetris |             3 | 10 minutes ago   | 10 minutes ago     | success |          2
  01GF4E3B18A3WZG9E099PMXTBT | deploy        | k8s-tetris |             3 |                  |                    |         |
  01GF4E3B18DF1JY13EHYDNY5NG | deploy-prod   | k8s-tetris |             3 |                  |                    |         |
  01GF4E3B19S54TVV9SQHXQ7WR3 | simple-nested | k8s-tetris |             3 | 1 minute ago     | 55 seconds ago     | running |          3
  01GF4E3B1AG2W1Q1C16JEQJF99 | test          | k8s-tetris |             4 |                  |                    |         |
  01GF4E3B1ATVFAHRFZB2MSH8SP | prod          | k8s-tetris |             4 |                  |                    |         |
  01GF4E3B1BP29E8C61X2NRWC05 | release       | k8s-tetris |             5 | 12 minutes ago   | a long while ago   | running |          3

brian@ubuntu:waypoint-tetris λ waypoint pipeline inspect -name=simple-nested                                                                         ±[main]

» Pipeline Configuration
                       ID: 01GF4E3B19S54TVV9SQHXQ7WR3
                     Name: simple-nested
                    Owner: k8s-tetris
           Root Step Name: here-we-go
              Total Steps: 3
         Last Run Started: 1 minute ago
       Last Run Completed: 1 minute ago
          Last Run Status: RUNNING
      Last Run Commit SHA: f0198730b9341e02f3c91d7d5b53694c9fb17737
  Last Run Commit Message: Add encoded_auth to docker-pull plugin

  Total Runs: 3
```

The run status should be `COMPLETE` or `ERRORED` for the two pipelines in `waypoint pipeline list` that are shown as `RUNNING`.

**Steps to Reproduce**

Run a pipeline until it completes or errors, check its status with `list` or `inspect`.

**Expected behavior**

A pipeline run state should reflect reality.

applicable:

* Waypoint CLI Version:

**Additional context**

briancain:
This pull request fixes a bug where the boltdb state for a pipeline run would never mark a finished pipeline as success and would be stuck in running forever. It fixes it by looking at all job references on a run and seeing if they are also complete before marking the pipeline as complete.

Fixes https://github.com/hashicorp/waypoint/issues/4032

### DIFF
--- a/internal/server/boltdbstate/job.go
+++ b/internal/server/boltdbstate/job.go
@@ -1772,11 +1772,25 @@ func (s *State) pipelineComplete(jobId string) error {
 	if job.State == pb.Job_ERROR {
 		run.State = pb.PipelineRun_ERROR
 	} else if job.State == pb.Job_SUCCESS {
-		// If job Id matches last job queued by pipeline.
-		// We will have to change this in the future when pipeline steps run in parallel,
-		// and the last job queued may not be the last job to complete in the pipeline
-		// TODO:XX figure out how ^
-		if job.Id == run.Jobs[len(run.Jobs)-1].Id {
+		// Look at all job ids in a run and check if any are not SUCCESS
+		runComplete := true
+		for _, j := range run.Jobs {
+			if j.Id == job.Id {
+				continue
+			}
+
+			rj, err := s.JobById(j.Id, nil)
+			if err != nil {
+				return err
+			}
+
+			if rj.State != pb.Job_SUCCESS {
+				runComplete = false
+				break
+			}
+		}
+
+		if runComplete {
 			run.State = pb.PipelineRun_SUCCESS
 			s.log.Trace("pipeline run is complete", "job", job.Id, "pipeline", job.Pipeline.PipelineId, "run", run.Sequence)
 		}
--- a/pkg/server/gen/server.swagger.json
+++ b/pkg/server/gen/server.swagger.json
@@ -9477,16 +9477,19 @@
           "description": "The sequence number for this pipeline run."
         },
         "pipeline": {
-          "$ref": "#/definitions/hashicorp.waypoint.Ref.Pipeline"
+          "$ref": "#/definitions/hashicorp.waypoint.Ref.Pipeline",
+          "description": "The pipeline this run is apart of."
         },
         "jobs": {
           "type": "array",
           "items": {
             "$ref": "#/definitions/hashicorp.waypoint.Ref.Job"
-          }
+          },
+          "description": "The full list of jobs that are associated with this run."
         },
         "state": {
-          "$ref": "#/definitions/hashicorp.waypoint.PipelineRun.State"
+          "$ref": "#/definitions/hashicorp.waypoint.PipelineRun.State",
+          "title": "The current state of this pipeline run,"
         }
       }
     },
--- a/pkg/server/gen/server.swagger.json
+++ b/pkg/server/gen/server.swagger.json
@@ -9478,7 +9478,7 @@
         },
         "pipeline": {
           "$ref": "#/definitions/hashicorp.waypoint.Ref.Pipeline",
-          "description": "The pipeline this run is apart of."
+          "description": "The pipeline associated with this run."
         },
         "jobs": {
           "type": "array",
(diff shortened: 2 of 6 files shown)


internal/server/boltdbstate/job.go
```diff
@@ -1772,11 +1772,25 @@ func (s *State) pipelineComplete(jobId string) error {
...
 	} else if job.State == pb.Job_SUCCESS {
-		// If job Id matches last job queued by pipeline.
-		// We will have to change this in the future when pipeline steps run in parallel,
-		// and the last job queued may not be the last job to complete in the pipeline
-		// TODO:XX figure out how ^
-		if job.Id == run.Jobs[len(run.Jobs)-1].Id {
+		// Look at all job ids in a run and check if any are not SUCCESS
+		runComplete := true
+		for _, j := range run.Jobs {
+			if j.Id == job.Id {
```
xiaolin-ninja, line 1778: Why are we skipping the current job?

Instinctively this feels really redundant to loop through every job in the run, every time we hit a job... but I can't think of a better solution off the top of my head. Maybe combine it with the existing logic so that it only hits this loop if job.Id is the last job in the queue, and _then_ make sure that every other job has completed as well?

pkg/server/gen/server.swagger.json
```diff
@@ -9477,16 +9477,19 @@
           "description": "The sequence number for this pipeline run."
         },
         "pipeline": {
-          "$ref": "#/definitions/hashicorp.waypoint.Ref.Pipeline"
+          "$ref": "#/definitions/hashicorp.waypoint.Ref.Pipeline",
+          "description": "The pipeline this run is apart of."
```
xiaolin-ninja, line 9481:
```suggestion
          "description": "The pipeline that is run."
```

xiaolin-ninja (commented):
Looks like we're already hitting the edge case, left my thoughts on the loop & a rewording suggestion.

internal/server/boltdbstate/job.go
briancain: We skip the current job because we already know from line 1774 that the current jobs state is SUCCESS. No need to make another db call to determine this - we already have the job.

The main reason for this fix was because the `run.Jobs` slice wasn't actually in run order, the last job wasn't the final job in the pipeline, and so the run never completes and the state is stuck as `running`. Although this adds some extra overhead to look up jobs by id, I think it's an ok tradeoff until we implement the job system improvements we are planning for the next major release which will end up making this simpler anyways.

One performance improvement we could make is a specialized job state func that looks up an array of jobs by ids in a single transaction instead of per-job.

evanphx: > The main reason for this fix was because the run.Jobs slice wasn't actually in run order

🤣

evanphx (approved):

xiaolin-ninja (approved):
Ok! Looks good to me for a temporary fix until something more efficient comes along. :)

PR #4054, opened 2022-10-14
ui: removed fix for resolved safari bug with xterm/webgl

sabrinako:
### Explanation for this change
- During HCP Waypoint beta testing, we were informed of a bug with viewing logs on Safari. The issue ended up being we had this if condition to handle a bug that previously existed with xterm rendering with Safari/WebGL which has since been fixed.
- Bug has been fixed in HCP Waypoint but not OSS yet until this PR

### DIFF
--- a/.changelog/4054.txt
+++ b/.changelog/4054.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+ui: fix safari bug with xterm/webgl rendering
+```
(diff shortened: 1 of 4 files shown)


amyrlam:
Deletion looks good 👍. Question: should it be a bug fix in the changelog?

jgwhite:
> Question: should it be a bug fix in the changelog?

Yup, this is a user-facing bugfix so no reason not to.

sabrinako:
@amyrlam sounds good! I just updated the branch and added the changelog

amyrlam (approved):
:shipit: looks good!

Not sure if the `go-tests` failure is just flaky.

PR #4062, opened 2022-10-17
update project serverState interface methods to add context

cicoyle:
Adding ctx to the project server state interface methods to be able to better access the ctx to cancel expensive queries. I am phasing this out into smaller PRs, so there will be more to come.


izaaklauer (approved):

briancain (approved):
Looks pretty straight forward to me!

Issue #3989, opened 2022-10-04
`waypoint runner install` with `-id` results in mismatched profile


jgwhite:
**Describe the bug**
When installing a runner using `waypoint runner install -id=some-memorable-id`, the resultant runner profile uses an upcased version of the ID for both profile name and target. As a result, the profile doesn’t match its target runner.

**Steps to Reproduce**
Install a runner with ID specified:
```
waypoint runner install \                       
  -platform=kubernetes \
  -id=test \
  -server-addr=127.0.0.1:9701 \
  -k8s-runner-image=hashicorp/waypoint:latest \
  -server-tls-skip-verify
```

This result in the expected runner:

```
$ waypoint runner list 
   ID  |  STATE  |  KIND  | LABELS | LAST REGISTERED  
-------+---------+--------+--------+------------------
  test | adopted | remote |        | 1 minute ago     
```

But the profile uses `TEST` instead:

```
$ waypoint runner profile list
Runner profiles
       NAME       | PLUGIN TYPE |            OCI URL            | TARGET RUNNER | DEFAULT  
------------------+-------------+-------------------------------+---------------+----------
  kubernetes-TEST | kubernetes  | hashicorp/waypoint-odr:latest | TEST          | yes      
```

**Expected behavior**
The runner profile name and target should match the case of the runner ID.

**Waypoint Platform Versions**
```
$ waypoint version
CLI: v0.10.0-304-g9771295a0 (9771295a0)
Server: v0.10.0-304-g9771295a0
```

PR #4063, opened 2022-10-17
Remove automatic uppercasing of ids, so that future runner profiles will match

amyrlam:
Resolves: https://github.com/hashicorp/waypoint/issues/3989

We shouldn't need to backport this change. 

~~@paladin-devops wondering if you can test this out on Nomad on your machine? Also down to pair. We tried bootstrapping a Nomad cluster with -dev, but it needed a host volume.~~ Update: done

### DIFF
--- a/internal/cli/runner_install.go
+++ b/internal/cli/runner_install.go
@@ -285,14 +285,14 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 
 		// Creating a new runner profile for the newly adopted runner
 		var odrConfig *pb.OnDemandRunnerConfig
-		s = sg.Add("Creating runner profile and targeting runner %s", strings.ToUpper(id))
+		s = sg.Add("Creating runner profile and targeting runner %s", id)
 		if odc, ok := p.(installutil.OnDemandRunnerConfigProvider); ok {
 			odrConfig = odc.OnDemandRunnerConfig()
-			odrConfig.Name = odrConfig.Name + "-" + strings.ToUpper(id)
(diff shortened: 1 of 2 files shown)


internal/cli/runner_install.go
```diff
@@ -308,9 +308,7 @@ func (c *RunnerInstallCommand) Run(args []string) int {
 		} else {
 			odrConfig.TargetRunner = &pb.Ref_Runner{
 				Target: &pb.Ref_Runner_Id{
-					Id: &pb.Ref_RunnerId{
-						Id: strings.ToUpper(id),
-					},
+					Id: &pb.Ref_RunnerId{Id: id},
```
amyrlam, line 311: This is the one that we need to solve the issue https://github.com/hashicorp/waypoint/issues/3989, but we changed them all for consistency / so the CLI looks nice.

briancain (commented):
This makes sense to me, I must of missed this. I imagine this might cause some confusion - and do we need to do anything with upgrades too?

amyrlam:
Not sure about upgrades. I'll ping you to discuss

amyrlam:
@briancain Talked about your question more with @jgwhite. With this change, when `id=your-inputted-id-here` is passed, the `id` will not be automatically uppercased at creation. In the UI, we can now have the user input an `id`, and then confirm that that runner with said `id` was just created/adopted.

 It shouldn't affect upgrades, which would upgrade the previously inputted `id` (which could be all uppercased or not).

briancain (approved):

PR #4082, opened 2022-10-21
Increase coverage of `TestJobUpdateRef`

jgwhite:
Previously we were just checking that the `Commit` was successfully stored. This change adds assertions for `CommitMessage` and `Timestamp`.


briancain (approved):
Looks legit to me! 🎉

PR #4100, opened 2022-10-24
Job list filtered by pipelines displays latest run if none supplied

xiaolin-ninja:
Previously, `waypoint job list -pipeline-name/id=` would display no results if no `-run` flag is supplied. 

Now output defaults to latest run.


briancain (approved):
👍🏻 looks good to me! I was able to repo the issue and this fixes it

PR #4121, opened 2022-10-26
Fix issue with docker-pull and authentication 

### LINKED ISSUE #4055: Kaniko: docker-pull fails to authenticate to container registry to pull container artifact
**Describe the bug**

Attempting to use the `docker-pull` plugin with on-demand runners, which uses Kaniko to pull containers, ends up failing to authenticate to the container registry despite being configured with a username and password via Waypoint:

```hcl
app "tetris" {
  build {
    use "docker" {
    }
    workspace "production" {
      use "docker-pull" {
        image = var.image
        tag   = var.tag
        encoded_auth = base64encode(
          jsonencode({
            username = var.registry_username,
            password = var.registry_password
          })
        )
      }
    }
    registry {
      use "docker" {
        image    = var.image
        tag      = var.tag
        username = var.registry_username
        password = var.registry_password
        local    = var.registry_local
      }
    }
  }
...
...
```

Note that `var.registry_username` and `var.registry_password` are the same input variables used to _push_ the container to that same registry, so I know at least internally those vars are set.

```
✓ Running build v3
✓ Injecting entrypoint...
⠹ Executing Kaniko...
 │ INFO[0000] Retrieving image manifest <registry>/tetris:latest
 │ INFO[0000] Retrieving image <registry>/tetris:latest from registry <registry>
 │ error building image: GET https://<registry>/token?scope=repository%3Atetris%3Ap
 │ ull&service=<registry>: : Authentication is required
! 1 error occurred:
        * rpc error: code = Unknown desc = exit status 1

```

**Steps to Reproduce**

Attempt to run a docker-pull plugin for a build step using on-demand runners via Kaniko.

**Expected behavior**

Kaniko should have the proper credentials setup to pull a container from the specified registry.

applicable:

* Waypoint CLI Version: 0.10.2
* Waypoint Server Platform and Version: n/a
* Waypoint Plugin: `docker-pull` with `kaniko`

**Additional context**

There are some closely related issues but they don't 100% line up with this one:

- https://github.com/hashicorp/waypoint/issues/4001
- https://github.com/hashicorp/waypoint/issues/3557

catsby:
Fix an issue where both `docker-pull` and subsequent `registry` blocks needed separate authentication. 

Example:

```hcl
app "some-app" {
  build {
    use "docker" {
      // Use the included Dockerfile
    }
    workspace "production" {
      use "docker-pull" {
        image = "some-private-registry/image"
        tag   = "stable"
        encoded_auth = base64encode(
          jsonencode({
            username = "some-username",
            password = "some-password"
          })
        )
      }
    }

    registry {
      use "docker" {
        image    = "somerepo/image"
        tag      = "latest"
        username = "some-username"
        password = "some-password"
      }
    }
  }
  
  [...]
}
```

Prior to these changes, Waypoint could only authenticate with the registry in the `registry` block, and did not properly authenticate with the registry that the `docker-pull` block would eventually be talking to. 

In this PR we create a 2nd [OCI Proxy Server](https://github.com/hashicorp/waypoint/blob/b0017806980a11f93d9c2411815818a123ae6ccd/internal/pkg/epinject/ociregistry/server.go#L39) to handle authentication with the registry we pull from, and modify the resulting generate `Dockerfile` to then pull from the proxy and not directly from the source. 

Along the way we ran into some issues with manifests so we had to make some changes to the proxy server itself as well. 

Fixes https://github.com/hashicorp/waypoint/issues/4055

### DIFF
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -160,7 +160,7 @@ func (b *Builder) pullWithKaniko(
 			}
 			sourceProxy.AuthConfig.Username = user
 			sourceProxy.AuthConfig.Password = pass
-		} else if *b.config.Auth != (wpdocker.Auth{}) {
+		} else if b.config.Auth != nil && *b.config.Auth != (wpdocker.Auth{}) {
 			//If EncodedAuth is not set, and Auth is, use Auth
 			sourceProxy.AuthConfig.Username = b.config.Auth.Username
 			sourceProxy.AuthConfig.Password = b.config.Auth.Password
--- a/builtin/docker/pull/kaniko.go
+++ b/builtin/docker/pull/kaniko.go
@@ -131,11 +131,80 @@ func (b *Builder) pullWithKaniko(
 
 	localRef := fmt.Sprintf("localhost:%d/%s:%s", port, refPath, ai.Tag)
 
-	dockerfileBS := []byte(fmt.Sprintf("FROM %s:%s\n", b.config.Image, b.config.Tag))
+	////// setup pull ref
+	var remoteRef string
+	var rci ociregistry.Server
+	rci.DisableEntrypoint = b.config.DisableCEB
+	rci.Logger = log
+	{
+		if b.config.EncodedAuth != "" {
+			//If EncodedAuth is set, use that
+			user, pass, err := wpdocker.CredentialsFromConfig(b.config.EncodedAuth)
+			if err != nil {
+				return nil, err
+			}
+			rci.AuthConfig.Username = user
+			rci.AuthConfig.Password = pass
+		} else if *b.config.Auth != (wpdocker.Auth{}) {
+			//If EncodedAuth is not set, and Auth is, use Auth
+			rci.AuthConfig.Username = b.config.Auth.Username
+			rci.AuthConfig.Password = b.config.Auth.Password
+			rci.AuthConfig.Auth = b.config.Auth.Auth
+			rci.AuthConfig.IdentityToken = b.config.Auth.IdentityToken
+			rci.AuthConfig.RegistryToken = b.config.Auth.RegistryToken
+		}
+
+		// Determine the host that we're setting auth for. We have to parse the
+		// image for this cause it may not contain a host. Luckily Docker has
+		// libs to normalize this all for us.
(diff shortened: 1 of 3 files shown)


builtin/docker/pull/kaniko.go
```diff
@@ -41,98 +41,177 @@ func (b *Builder) pullWithKaniko(
...
+	{
+		if b.config.EncodedAuth != "" {
+			//If EncodedAuth is set, use that
+			user, pass, err := wpdocker.CredentialsFromConfig(b.config.EncodedAuth)
+			if err != nil {
+				return nil, err
+			}
+			sourceProxy.AuthConfig.Username = user
+			sourceProxy.AuthConfig.Password = pass
+		} else if *b.config.Auth != (wpdocker.Auth{}) {
```
evanphx, line 163: If `Auth` is nil, this will panic I believe.

internal/pkg/epinject/ociregistry/server.go
```diff
@@ -265,19 +265,8 @@ func (s *Server) FetchBlob(name string, id string) ([]byte, error) {
...
-	found, err := s.probeBlob(s.entrypointRepo, s.entrypointId)
-	if err != nil {
-		return err
-	}
-
-	if found {
-		s.Logger.Debug("entrypoint layer already exists, not reuploading")
-		return nil
-	}
-
```
evanphx, line 277: We can put this back in, it will make things faster and wasn't causing an issue.

evanphx (changes requested):
A couple of minor things.

briancain (approved):
I agree with Evans feedback, but wanted to approve because I gave this a test and it worked and solves the bug! ✨ 🐛 🐳

builtin/docker/pull/kaniko.go
catsby: Added a nil check in e5b5fe97c

internal/pkg/epinject/ociregistry/server.go
catsby: Restored in 4480ed51f

catsby:
I addressed the feedback, @evanphx please take another look when you can, thanks!

catsby:
@evanphx please take another look when you can

evanphx (approved):

PR #4132, opened 2022-10-28
Feature: Update pipelines parser to evaluate input variables

### LINKED ISSUE #4122: Are variables not supported in Waypoint Pipelines?
We have about 5 microservices that I am trying to bring variables across from our Terraform deployment into, I am trying to create a pipeline that fetches the necessary information from Vault and insert it into the container. However there doesn't seem to be any way to interpolate variables into the exec command? There is no documentation showing this behaviour either.

```hcl
variable "stage" {}
variable "app_name" {}
variable "gcp_project_id" {}
variable "git_sha" {}
variable "gcp_service_account" {}
variable "vault_addr" {}
variable "vault_token" {}
variable "vault_namespace" {}

project = "tymlez-platform"

pipeline "full-up" {

  step "tymlez-config" {
    use "exec" {
      command = "curl"
      args    = ["--header \"X-Vault-Token :",var.vault_token,"\"",var.vault_addr,"/v1/kvv2/",var.app_name,"/",var.stage,"/all-app-variables", "jq -r .data.data > /tmp/variables.json"]
    }
  }
...
```

**Expected behavior**
Variables should be able to be included anywhere in a HCL file

applicable:

CLI: v0.10.2 (2c9926ef8)
Server: v0.10.2

briancain:
This pull request introduces input variables being evaluated inside `pipeline {}` stanzas. It ends up rearranging how we parse pipelines and steps to support the lazy-decoding that Waypoint does to evaluate and resolve variables later in the parsing process.

It also had to modify how `waypoint init` does its initial sync of pipeline definitions. Now instead we do a "shallow sync" of the pipeline name, owner, and a base step. This mirrors how we sync applications on `waypoint init` in that we don't sync the entire parsed hcl config, only the app name on init.

Fixes #4122

### DIFF
--- a/.changelog/4132.txt
+++ b/.changelog/4132.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+pipelines: Add ability to evaluate input variables in pipelines stanzas.
+```
--- a/internal/cli/init.go
+++ b/internal/cli/init.go
@@ -404,17 +404,29 @@ func (c *InitCommand) validateProject() bool {
 	}
 
 	sp := sg.Add("Registering all pipelines for project %q", ref.Project)
-	protoPipes, err := c.cfg.PipelineProtos()
-	if err != nil {
-		c.stepError(s, initStepPipeline, err)
-		return false
-	}
+	pipeNames := c.cfg.Pipelines()
 
-	for _, pipeline := range protoPipes {
-		sp.Update("Registering Pipeline %q with the server...", pipeline.Name)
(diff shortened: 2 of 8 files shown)


briancain:
A note to the reviewer: It doesn't seem like `waypoint pipeline run` is properly sprinkling in CLI var flag overrides _only_ defined within the app stanzas. I think I'll leave this as a follow-up since I don't think it's directly related to this PR 🤔

evanphx (approved):
:shipit:

PR #4137, opened 2022-10-31
cli/pipeline_run: Include variables in job template for pipeline run

### LINKED ISSUE #4134: pipelines: Input variables at app stanza not overriding
**Describe the bug**

When running `waypoint pipeline run` with an input var override defined on the CLI, it doesn't look like that input var gets set properly from the override:

```
brian@ubuntu:waypoint-tetris λ waypoint pipeline run -var="tag=1" marathon                                                                                 ±[main]

» Running pipeline "marathon" for application "tetris"
✓ Pipeline "marathon" has started running. Attempting to read job stream sequentially in order
✓ 1 steps detected, run sequence 2
...
...
» Building tetris...
✓ Running build v4
✓ Building Docker image with kaniko...
✓ Testing registry and uploading entrypoint layer
✓ Executing kaniko...
 │ INFO[0001] Retrieving image manifest nginx:stable
 │ INFO[0001] Returning cached image manifest
 │ INFO[0001] Executing 0 build triggers
 │ INFO[0001] Unpacking rootfs as cmd COPY ./public/ /var/www requires it.
 │ INFO[0004] COPY ./public/ /var/www
 │ INFO[0004] Taking snapshot of files...
 │ INFO[0004] COPY ./nginx.conf /etc/nginx/conf.d/default.conf
 │ INFO[0004] Taking snapshot of files...
 │ INFO[0004] Pushing image to localhost:38817/tetris:latest
 │ INFO[0005] Pushed image to 1 destinations
✓ Image pushed to '<contianer_registry>/tetris:latest'
...
...
# the variable used
 tag               | latest                                        | string    | default


```
**Steps to Reproduce**

Run a pipeline via the CLI with a CLI input var override for a variable defined inside an `app {}` stanza. See that it uses the default rather than the override.

```hcl
variable "tag" {
  default     = "latest"
  type        = string
  description = "Image tag for the image"
}
app "tetris" {
  build {
    use "docker" {
    }

    registry {
      use "docker" {
        image    = var.image
        tag      = var.tag
        username = var.registry_username
        password = var.registry_password
        local    = var.registry_local
      }
    }
  }
...
...
}
```

**Expected behavior**

The override should be respected and used for the pipeline run.

applicable:

* Waypoint CLI Version: `main` at [b7c98b1](https://github.com/hashicorp/waypoint/commit/b7c98b16f8959bc63d26ec88c313792408364fcc)
* Waypoint Server Platform and Version: n/a
* Waypoint Plugin: n/a

**Additional context**

briancain:
Prior to this commit, when a pipeline was run, the CLI variables from the job client were not included in the submitted job template that the CLI would send to the server to run pipeline jobs. This means any CLI variables would not be properly resolved when the job went to execute. This commit fixes that by including those variables in the job template.

Fixes #4134

### DIFF
--- a/internal/cli/pipeline_run.go
+++ b/internal/cli/pipeline_run.go
@@ -99,6 +99,7 @@ func (c *PipelineRunCommand) Run(args []string) int {
 			Application:  app.Ref(),
 			Workspace:    c.project.WorkspaceRef(),
 			TargetRunner: &pb.Ref_Runner{Target: &pb.Ref_Runner_Any{}},
+			Variables:    c.variables,
 		}
 
 		// build the base api request
--- a/.changelog/4137.txt
+++ b/.changelog/4137.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Fix bug where input variables were not included on pipeline run jobs.
+```


xiaolin-ninja (approved):

internal/cli/pipeline_run.go
```diff
@@ -99,6 +99,7 @@ func (c *PipelineRunCommand) Run(args []string) int {
 			Application:  app.Ref(),
 			Workspace:    c.project.WorkspaceRef(),
 			TargetRunner: &pb.Ref_Runner{Target: &pb.Ref_Runner_Any{}},
+			Variables:    c.variables,
```
xiaolin-ninja, line 102: I haven't dug deeper so it might be there, but I don't see the code where these variables are then handled?

briancain: When you initialize a CLI client, we evaluate variables and load them onto the base client to be made available to any CLI clients:

https://github.com/hashicorp/waypoint/blob/b7c98b16f8959bc63d26ec88c313792408364fcc/internal/cli/base.go#L289-L299

Configuring a basic CLI client gets these variables attached to it (i.e. `c.variables`)

https://github.com/hashicorp/waypoint/blob/b7c98b16f8959bc63d26ec88c313792408364fcc/internal/cli/base_init.go#L95-L104

CLI commands like `waypoint up`, `waypoint deploy`, etc use DoApp to run app operations directly, and when that happens, the job client gets those variables attached to the job template:

https://github.com/hashicorp/waypoint/blob/b7c98b16f8959bc63d26ec88c313792408364fcc/internal/client/job.go#L22-L42

The `waypoint pipeline run` CLI is a bit different because it sends the job template off to the server to be used in the job generater when resolving pipeline steps to queued job requests. So the fix here is including those init CLI variables on the job template!

https://github.com/hashicorp/waypoint/blob/1698c1df4e167f794dcc044e9dd0dde23f2d5a03/internal/cli/pipeline_run.go#L97-L103

PR #4139, opened 2022-10-31
UI/hide sensitive variables

### LINKED ISSUE #2138: ui: The UI should enable users to mark variables as sensitive and hide their values

See https://github.com/hashicorp/waypoint/issues/2136. This is the UI implementation for respecting that value. (sorry to be brief! 😅 ).

I know TF Cloud has a good answer for this, so perhaps something like that to keep this feeling the same?

For example you can enter in an API token or password, and once saved, the UI will omit the value of the sensitive variable.

**Note for implementors**

The scope of this enhancement is strictly as follows:

1. Allow users to set the `sensitive` flag to true/false when adding/editing input variables in the UI
2. Hide or obfuscate the values marked as sensitive

As you undertake this work, you’ll discover that input variables managed in the UI and those defined in the waypoint.hcl file are somewhat independent of each other. The UX isn’t perfect, but try to set that aside for this particular piece of work. We’ll address it more holistically further down the road.

For now, you may wish to add some inline documentation to the UI indicating that the “sensitive” setting will only affect how the value appears in the UI, and users must make the corresponding change in their waypoint.hcl for the value to be hidden everywhere.

sdav9375:
Closes [#2138](https://github.com/hashicorp/waypoint/issues/2138)

### Purpose of change

This PR adds a toggle switch on the input variables create/edit form for making an input "sensitive". If the input is set as such, in the list, there will be a badge in place of the displayed which says `Sensitive - write only`. If the user edits the variable, the value will be hidden and only the placeholder text `sensitive - write only` will be visible in the text input. The user will be unable to toggle off the sensitive switch and see the variable. If they toggle sensitive off, they will need to provide a new value in the input. This behavior matches HCP.  Please note this does not obscure the value from network requests.

Two component tests have been added.

### Demo

https://user-images.githubusercontent.com/17822084/199099172-c37d2fc6-0aed-435c-b41e-ca4e3ead4fef.mov

### To test the change

1. Navigate to a project
2. Click Manage
3. Click input variables
4. Create input variable and mark it "sensitive"
5. See that the variable value is hidden in the list
6. Edit the variable and see that the value is hidden in the form

### DIFF
--- a/ui/app/components/project-input-variables/list-item.hbs
+++ b/ui/app/components/project-input-variables/list-item.hbs
@@ -13,6 +13,7 @@
             <Pds::Input
               @type='text'
               data-test-input-variables-var-name
+              required
               id='input-variable-name'
               placeholder={{t 'form.project_variables_settings.variable_name_placeholder'}}
               value={{this.variable.name}}
@@ -28,6 +29,7 @@
                 <Pds::Input
                   @type='text'
                   data-test-input-variables-var-hcl
+                  required
                   id='input-variable-value'
                   placeholder={{if
                     this.writeOnly
@@ -42,6 +44,7 @@
                   @type='text'
                   id='input-variable-value'
                   data-test-input-variables-var-str
+                  required
                   placeholder={{if
                     this.writeOnly
                     (t 'form.project_variables_settings.variable_form_placeholder_sensitive')
--- a/ui/app/components/project-input-variables/list-item.hbs
+++ b/ui/app/components/project-input-variables/list-item.hbs
@@ -1,112 +1,112 @@
 <li data-test-input-variables-list-item>
   {{#if (or this.isEditing this.isCreating)}}
-    <form class='pds-form' data-test-input-variables-form {{on 'submit' this.saveVariable}}>
-      <div part='fields' class='card'>
-        <div class='card-header'>
-          <h4>{{t 'form.project_variables_settings.title'}}</h4>
+    <form class="pds-form" data-test-input-variables-form {{on "submit" this.saveVariable}}>
+      <div part="fields" class="card">
+        <div class="card-header">
+          <h4>{{t "form.project_variables_settings.title"}}</h4>
         </div>
-        <fieldset class='pds-formFieldSet'>
-          <div class='pds-formField'>
-            <label for='input-variable-name' class='pds-fieldName'>
-              {{t 'form.project_variables_settings.variable_name'}}
+        <fieldset class="pds-formFieldSet">
+          <div class="pds-formField">
+            <label for="input-variable-name" class="pds-fieldName">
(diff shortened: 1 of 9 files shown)


ui/app/components/project-input-variables/list-item.hbs
```diff
@@ -1,116 +1,139 @@
 <li data-test-input-variables-list-item>
   {{#if (or this.isEditing this.isCreating)}}
-    <form class="pds-form" data-test-input-variables-form {{on "submit" this.saveVariable}}>
```
amyrlam, line 3: (suggestion) Looks like we don't enforce it in linting tests, but perhaps it would be better to remove the quote changing in this file, to match the rest of the repo, which uses `"`. 

(I think `"` is the Ember default in `.hbs`, but `cloud-ui` uses `'`.) Maybe you have a VSCode extension doing an autosave?

ui/translations/en-us.yaml
```diff
@@ -150,6 +150,9 @@ form:
     variable_name_placeholder: 'var_key'
     variable_value: 'Value'
     variable_value_placeholder: 'var_value'
+    variable_set_sensitive: Set as sensitive
+    variable_value_placeholder_sensitive: 'Sensitive - write only'
+    variable_form_placeholder_sensitive: 'sensitive - write only'
```
amyrlam, line 155: (suggestion) Have the casing for both be 'Sensitive - write only' so that they match?

amyrlam (commented):
The video and code looks good so far! Left a couple of suggestions.

You'll need to address the failing tests and add a changelog too.

ui/translations/en-us.yaml
sdav9375: This is the way it is in HCP and the accompanying designs so I'm just making them match.

ui/app/components/project-input-variables/list-item.hbs
sdav9375: Added changelog.

ui/translations/en-us.yaml
amyrlam: Makes sense, but we should follow https://design-system-components-hashicorp.vercel.app/content/writing-guidelines in the future to be consistent with casing

amyrlam (approved):
👍 Looks good! 

Seems like there were some non-quote auto-prettier edits, but not super important

ui/translations/en-us.yaml
sdav9375: @joshklekamp or @ryenotbread what do you think? This difference in case is how it works in TFC which is what I modeled ours on.

joshklekamp: I think they should match and they should be sentence cased.

ryenotbread: I agree they should match and be sentence cased.

Issue #4005, opened 2022-10-07
`waypoint project destroy -p <<project name>>` throws error when no waypoint.hcl file is available


mjwhodur:
**Describe the bug**
One cannot delete empty project after it was created.
Also, there is no possibility to destroy a project from web UI.

**Steps to Reproduce**
1. Create project in the UI. 
2. List the project in the CLI
3. Issue a command: `waypoint project destroy -p projectname`

Example output:
```
michal.weglowskihodur@mhodur1m-war1 ~ % waypoint project list
go-k8s
kubernetes-go-multiapp-k8s-ingress
kubernetes-go-multiworkspace
kubernetes-nodejs
mmd
waypoint-test
michal.weglowskihodur@mhodur1m-war1 ~ % waypoint project destroy -p mmd
Do you really want to destroy project "mmd" and its resources? Only 'yes' will be accepted to approve: yes


» Performing operation locally
! Error destroying project: rpc error: code = FailedPrecondition desc = A waypoint.hcl was not found. Please either add a waypoint.hcl to the project source or in the project settings in the Waypoint UI.
```

```
michal.weglowskihodur@mhodur1m-war1 ~ % waypoint project destroy -p mmd -vvv
2022-10-07T16:07:19.302+0200 [INFO]  waypoint: waypoint version: full_string="v0.10.2 (2c9926ef8)" version=v0.10.2 prerelease="" metadata="" revision=2c9926ef8
2022-10-07T16:07:19.302+0200 [TRACE] waypoint: starting interrupt listener for context cancellation
2022-10-07T16:07:19.302+0200 [TRACE] waypoint: interrupt listener goroutine started
2022-10-07T16:07:19.303+0200 [DEBUG] waypoint: home configuration directory: path=/Users/michal.weglowskihodur/Library/Preferences/waypoint
2022-10-07T16:07:19.305+0200 [TRACE] waypoint: no API client provided, initializing connection if possible
2022-10-07T16:07:19.305+0200 [TRACE] waypoint.server: Local server may be created later - server credentials optional
2022-10-07T16:07:19.305+0200 [INFO]  waypoint.server: attempting to source credentials and connect
2022-10-07T16:07:19.308+0200 [DEBUG] waypoint.serverclient: connection information: address=34.123.34.105:9701 tls=true tls_skip_verify=true send_auth=true has_token=true
2022-10-07T16:07:19.784+0200 [DEBUG] waypoint.server: connection established with sourced credentials
[...]

PR #4141, opened 2022-10-31
Warn but do not raise error when destroying projects without source config file

xiaolin-ninja:
Resolves https://github.com/hashicorp/waypoint/issues/4005

### DIFF
--- a/internal/runner/operation.go
+++ b/internal/runner/operation.go
@@ -115,10 +115,15 @@ func (r *Runner) executeJob(
 	}
 
 	if path == "" {
-		// No waypoint.hcl file is found.
-		return nil, status.Errorf(codes.FailedPrecondition,
-			"A waypoint.hcl was not found. Please either add a waypoint.hcl to "+
-				"the project source or in the project settings in the Waypoint UI.")
+		// Only warn if operation is project destroy
+		if _, ok := job.Operation.(*pb.Job_DestroyProject); ok {
+			log.Warn("A waypoint.hcl was not found.")
+		} else {
+			// No waypoint.hcl file is found.
+			return nil, status.Errorf(codes.FailedPrecondition,
+				"A waypoint.hcl was not found. Please either add a waypoint.hcl to "+
+					"the project source or in the project settings in the Waypoint UI.")
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/runner/operation.go
```diff
@@ -115,10 +115,15 @@ func (r *Runner) executeJob(
 	}
 
 	if path == "" {
-		// No waypoint.hcl file is found.
-		return nil, status.Errorf(codes.FailedPrecondition,
-			"A waypoint.hcl was not found. Please either add a waypoint.hcl to "+
-				"the project source or in the project settings in the Waypoint UI.")
+		// Only warn if operation is project destroy
+		if _, ok := job.Operation.(*pb.Job_DestroyProject); ok {
+			log.Warn("A waypoint.hcl was not found.")
```
briancain, line 120: I don't know about this fix, we get around this error, but it only leads to more:

```
brian@ubuntu:~ λ waypoint project destroy -project=delete-me
Do you really want to destroy project "delete-me" and its resources? Only 'yes' will be accepted to approve: yes


! Error destroying project: rpc error: code = Unknown desc = <nil>: Failed to read configuration; Can't read /home/brian: read /home/brian: is a directory.
```

Or if you run it in a different Waypoint project, it fails with no data source configured because the destroy operations can't run:

```
brian@ubuntu:waypoint-tetris λ waypoint project destroy -project=delete-me                                                                                ±[main]
Warning: Currently in project directory for "k8s-tetris", but will operate
against specified project "delete-me"
Do you really want to destroy project "delete-me" and its resources? Only 'yes' will be accepted to approve: yes

! Error destroying project: rpc error: code = FailedPrecondition desc = Project delete-me does not have a data source configured. Remote jobs require a data
source such as Git to be configured with the project. Data sources can be configured via the CLI or UI. For help, see :
https://www.waypointproject.io/docs/projects/git#configuring-the-project
brian@ubuntu:waypoint-tetris λ                                            
```

I think instead we should have a flow to support deleting projects with no data source supported via the Destroy operation in the case where we have a blank project.

briancain:
Hey @xiaolin-ninja - I recommend we close this PR since it's been a while since it was opened without any forward progress.

PR #4143, opened 2022-10-31
Add operation flags to `config sync` command

paladin-devops:
This PR adds the set of operation flags (-local, for example) to the config sync command.

One benefit that is gained here is that in the (albeit unlikely) scenario that a user does not have Waypoint configured to launch on-demand runners, but they do have a static runner, `waypoint config sync -local=false` can be used to still sync dynamic configuration variables. Such behavior is not possible on a local runner.


briancain (approved):
This makes sense - don't forget to regen the website docs with the new flags!

PR #4144, opened 2022-10-31
datasource/git: Align git fields for UX

briancain:
This commit aligns the auth fields if used when displaying the git datasource information


jgwhite:
That was fast!

jgwhite (approved):
NICE

PR #4185, opened 2022-11-09
Fix pipeline `exec` steps run in Nomad

paladin-devops:
Prior to this PR, the Nomad task launcher plugin previously set the[ "command" field](https://developer.hashicorp.com/nomad/docs/drivers/docker#command) for launching tasks. Before pipelines, this was fine for ODRs. However, pipelines pass a slice of string to the `TaskLaunchInfo` struct (from the Waypoint plugin SDK) for the Entrypoint field. The Nomad Docker driver's command field accepts only string, not slice of string. Therefore, any exec steps for pipelines run by Nomad ODRs would fail to start before this PR.

The entrypoint field of the Nomad Docker driver config does accept a slice of string, and that is the config now being set in the Nomad task launcher plugin.

### DIFF
--- a/builtin/nomad/task.go
+++ b/builtin/nomad/task.go
@@ -263,9 +263,9 @@ func (p *TaskLauncher) StartTask(
 
 	// On-Demand runner specific configuration to start the task with
 	config := map[string]interface{}{
-		"image":   tli.OciUrl,
-		"args":    tli.Arguments,
-		"command": tli.Entrypoint,
+		"image":      tli.OciUrl,
+		"args":       tli.Arguments,
+		"entrypoint": tli.Entrypoint,
 	}
 
 	job.TaskGroups[0].Tasks[0].Config = config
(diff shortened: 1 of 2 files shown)


builtin/nomad/task.go
```diff
@@ -263,9 +263,9 @@ func (p *TaskLauncher) StartTask(
 
 	// On-Demand runner specific configuration to start the task with
 	config := map[string]interface{}{
-		"image":   tli.OciUrl,
-		"args":    tli.Arguments,
-		"command": tli.Entrypoint,
+		"image":      tli.OciUrl,
+		"args":       tli.Arguments,
+		"entrypoint": tli.Entrypoint,
```
briancain, line 268: Interesting, surprised we didn't catch this with other ODR jobs! Do we not need to also set the command for the launched resource too?

paladin-devops: Nope! Setting the command in addition to the entrypoint here would just add args to the command that the entrypoint runs.

briancain (approved):

PR #4186, opened 2022-11-09
Add additional logging to find race condition

demophoon:
It is possible for a job to disappear and have the event handler attempt to use job before checking if there was an error on the error stream which can cause Waypoint to panic.

### DIFF
--- a/pkg/server/singleprocess/service_runner.go
+++ b/pkg/server/singleprocess/service_runner.go
@@ -826,6 +826,14 @@ func (s *Service) RunnerJobStream(
 			return hcerr.Externalize(log, err, "err from err channel")
 
 		case req := <-eventCh:
+			if job == nil {
+				select {
+				case err := <-errCh:
+					log.Error("Job disappeared and there was an error while processing job event", "error", err, "event", req)
+					return hcerr.Externalize(log, err, "failed to process job event")
+				default:
+				}
+			}
 			if err := s.handleJobStreamRequest(log, job, server, req, logStreamWriter); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/server/singleprocess/service_runner.go
```diff
@@ -881,6 +888,10 @@ func (s *Service) handleJobStreamRequest(
 	log.Trace("event received", "event", req.Event)
 	switch event := req.Event.(type) {
 	case *pb.RunnerJobStreamRequest_Complete_:
+		if event.Complete == nil {
+			log.Error("event.Complete was not filled when we expected it to be", "id", job.id)
```
briancain, line 892: Maybe log the `req.Event` too?

pkg/server/singleprocess/service_runner.go
```diff
@@ -881,6 +888,10 @@ func (s *Service) handleJobStreamRequest(
...
+			return hcerr.Externalize(log, err, "failed to complete job", "id", job.id)
```
briancain, line 893: I think the linter is mad about this one? There's no err available on this branch I think.

pkg/server/singleprocess/service_runner.go
demophoon: Added!

pkg/server/singleprocess/service_runner.go
demophoon: Fixed, I think my vim plugin stopped working and didn't catch that `job.id` should actually be `job.Id` 😨

briancain (approved):

PR #4191, opened 2022-11-10
Scoped config sources - protobuf & Bolt DB state

paladin-devops:
This PR updates the protobuf and the Bolt DB state implementation to support scoped configuration sources. Waypoint runners, the entrypoint, as well as the CLI continue to exclusively use the global scope, which continues to be the de facto scope in use with this PR. However, the changes in this PR lay the groundwork for forthcoming changes to the aforementioned "clients" of configuration sources. Some tests were also added for the config source Bolt DB state implementation.

The `pr/no-changelog` label is on this PR because there are no user-facing changes.


internal/server/boltdbstate/config_source.go
```diff
@@ -68,11 +72,21 @@ func (s *State) configSourceSet(
...
+	idHash, err := hashstructure.Hash(map[string]interface{}{
+		"scope":     value.Scope,
+		"type":      value.Type,
+		"workspace": value.Workspace,
+	}, hashstructure.FormatV2, nil)
+	if err != nil {
+		return err
+	}
+
+	id := s.configSourceId(idHash)
```
evanphx, line 86: Note for later readers, there is a note on configSourceId that it's only used for in memory indexes, so we aren't invalidating anything by changing the hash calculate algorithm.

evanphx (approved):

PR #4196, opened 2022-11-15
Scoped config sources - CLI, runners, and entrypoint

paladin-devops:
This PR updates the CLI, runners, and the entrypoint to use the config source in the Waypoint database which most closely matches the input scope.


evanphx (approved):
Looks good! Eventually we should figure out a testing harness for the CLI code...

PR #4203, opened 2022-11-16
Add pagination

andrew-hashicorp:
This PR:
- implements the pagination protobuff discussed internally
- adds pagination support to `ProjectList` as an initial example implementation
- adds validation for `ListProjects` at the service layer

While there are stubs for pagination in the BoltDB state, the initial implementation is limited to `ProjectList` in HCP. As such, this PR should be reviewed in conjunction with its companion branch `add_list_projects_with_pagination` in `cloud-waypoint-service`. The BoltDB implementation will be added in a forthcoming PR, along with sorting.

The `TestProjectPagination` tests in `statetest/test_project.go` are meant to be run out of `cloud-waypoint-service` and are skipped for the `hashicorp/waypoint` state tests.


pkg/server/proto/server.proto
```diff
@@ -5402,3 +5407,44 @@ message InlineKeepalive {
...
+  // This token allows you to get the previous page of results for list
+  // requests. If the number of results is larger than `page_size`, use the
+  // `previous_page_token` as a value for the query parameter
+  // `previous_page_token` in the next request. The value will become empty when
+  // there are no more pages.
+  string previous_page_token = 2 [json_name="previous_page_token"];
+
+  // The total number of entities stored by the server.
+  // This field is an addition not included in the base hashicorp/cloud/common pagination proto
+  uint64 total_count = 3 [json_name="total_count"];
```
evanphx, line 5449: Is this value necessary? Computing it is, strangely enough, a real drag on databases.

jgwhite: We show these counts prominently on the “list” pages:

total-count

Is the DB problem tractable?

evanphx: I'm thinking about in generically, like when there are 10k deployments. Would it be ok if the backend decided "this is too complex to compute" and just left it 0, such that if 0 then the frontend knows it's not calculated?

jgwhite: Yeah, I think that’s reasonable. Or if the field is nullable in the proto, then we could set it to null to indicate “this can’t be computed for some reason”.

jgwhite: Sounds fine to me, though I think we should decide whether we want to keep `total_count` on the generic `PaginationResponse` message, or whether we want to move it up to the specific `ListProjectsResponse` message (and all other `ListXResponse` messages that will support total count).



My rationale for including it on `PaginationResponse` is to give us a little nudge to consider it for endpoints such as this. But if we’re not always planning to support it then maybe it will set unhelpful expectations with consumers of the API.

Alternatively, we could indicate “total count not supported” using some other mechanism, like `oneof`…

```proto
message PaginationResponse {
  // ...
  oneof total_count_result {
    uint64 total_count = 3 [json_name="total_count"];
    bool total_count_not_supported = 4 [json_name="total_count_not_supported"];
  }
}
```

But this feels a little clunky, even to me 😅

izaaklauer: My vote would be for adding `count` to ListProjects, and not the global PaginationResponse. I don't think it's something we'll want to add to most list endpoints.

If I turn out to be wrong about that, and there turn out to be compelling reasons to add it broadly, wouldn't be bad toadd it to the `PaginationResponse` proto, and either keep filling or deprecate the rpc-specific `count` fields.

jgwhite: I’m happy to proceed with that modification. I’ll add a note to the RFC.

andrew-hashicorp: Mentioned this in Slack but posting here as well:

Was pairing w/ @jgwhite and he brought up another possibility, which is to separate count into its own method at the state layer, which can then be called where needed at the service layer. This prevents the ListProjects state method from having to `return []Projects, Pagination, count, err`.

I think this makes more sense and have updated the code/PR to reflect this.

internal/server/boltdbstate/project.go
```diff
@@ -256,12 +256,16 @@ func (s *State) ProjectDelete(ctx context.Context, ref *pb.Ref_Project) error {
 	return nil
 }
 
+func (s *State) ProjectCount(ctx context.Context) (count uint64, err error) {
```
evanphx, line 259: Not a big deal, but we generally don't use named returns in our code base.

evanphx (approved):

internal/server/boltdbstate/project.go
andrew-hashicorp: Ah, good to know. I'll remove the named returns.

PR #4208, opened 2022-11-17
cli: Remove extra https in final output for app urls

briancain:
The deployUrl that comes in already will have https, so this removes the extra prepended https from the deployment url.

This matches how the Up CLI looks as well when it prints the deployment URL.

### DIFF
--- a/.changelog/4208.txt
+++ b/.changelog/4208.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Remove extra `https` from deployment URL from pipeline run final output.
+```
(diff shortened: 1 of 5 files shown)


evanphx (changes requested):
I think if the url doesn't have a scheme on it, we should add https://. The DeployURL can come from plugins so they might not add a scheme.

briancain:
@evanphx I updated `pipeline run`, `up`, `deploy`, and `release` to append a scheme if missing via [d5fd63e](https://github.com/hashicorp/waypoint/pull/4208/commits/d5fd63e8b8fa1c1dcb334ccf12f5a7d52ccb9301)

evanphx (approved):
We should probably DRY up that code eventually!

briancain:
Agreed - I opened an issue to track this for now! https://github.com/hashicorp/waypoint/issues/4209

PR #4212, opened 2022-11-17
Modify `project destroy` to require `-p` flag and delete projects created in UI without data source

xiaolin-ninja:
As the title suggests. :)

Previously, `waypoint project destroy` would ignore whatever arguments provided and attempt to destroy project in current directory. This enforces the use of the `-project` or `-p` flag regardless of where the command is called to resolve this bug, and for extra guardrails.

This PR also introduces the ability to delete projects created in the UI without a data source or `waypoint.hcl`.

### DIFF
--- a/internal/cli/project_destroy.go
+++ b/internal/cli/project_destroy.go
@@ -39,7 +39,7 @@ func (c *ProjectDestroyCommand) Run(args []string) int {
 		return 1
 	}
 
-	// Confirmation required
+	// Confirmation required without `-auto-approve` flag
 	if !c.confirm {
 		proceed, err := c.ui.Input(&terminal.Input{
 			Prompt: "Do you really want to destroy project \"" + project.Project.Name + "\" and its resources? Only 'yes' will be accepted to approve: ",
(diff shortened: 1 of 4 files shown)


internal/cli/project_destroy.go
```diff
@@ -56,12 +58,19 @@ func (c *ProjectDestroyCommand) Run(args []string) int {
...
+
+	if project.Project.DataSource == nil {
+		_, err = c.project.Client().DestroyProject(c.Ctx, &pb.DestroyProjectRequest{
+			Project: c.project.Ref(),
+		})
+	} else {
+		_, err = c.project.DestroyProject(c.Ctx, &pb.Job_DestroyProjectOp{
+			Project:              &pb.Ref_Project{Project: project.Project.Name},
+			SkipDestroyResources: c.skipDestroyResources,
+		})
```
briancain, line 70: This confused me at first - I wasn't sure what the difference was between these two `DestroyProject` funcs. Maybe a quick doc comment note about how one is a client func that queues a DestroyProject operation, and the other is an API call to the server that deletes the projects record from the db with no job?

briancain (commented):
Makes sense! I have a few suggestions but otherwise I tested it out and works as expected 👍🏻

briancain (approved):
Look great!

PR #4229, opened 2022-11-28
Initial addition of GH->Jira workflow

krantzinator:
Things this does:
- checks for a `jira` label before doing anything
- checks for a ticket in Jira that matches the format this workflow will create new ticket titles with
- if the `jira` label was just added **and** a ticket does not already exist **and** the user who added the label is a `waypoint` team member, create a new ticket in the Waypoint project
- if the issue already had a `jira` label and was updated with a comment, was closed, or was reopened, update the activity/status of the corresponding Jira ticket


evanphx (approved):
Very nice!

Issue #3781, opened 2022-09-01
Podman support


jdoss:
**Is your feature request related to a problem? Please describe.**
I want Podman support in Waypoint! 🤠 

**Describe the solution you'd like**
Podman supports a local socket that can be used with Docker compose. It would be super cool to be able to use the rootless Podman socket with Waypoint.

**Describe alternatives you've considered**

*Try to install waypoint*

```
$ waypoint install -platform=docker -accept-tos
❌ Checking for existing installation...
! Error installing server into docker: Got permission denied while trying to connect to the Docker daemon socket at
  unix:///var/run/docker.sock: Get
  "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/json?all=1&filters=%7B%22label%22%3A%7B%22waypoint-type%3Dserver%22%3Atrue%7D%7D&limit=0":
  dial unix /var/run/docker.sock: connect: permission denied
```

*Enable rootless podman socket*

```
$ systemctl --user enable podman.socket
$ systemctl --user start podman.socket
$ ls -al $XDG_RUNTIME_DIR/podman/podman.sock
srw-rw----. 1 jdoss jdoss 0 Aug 31 10:05 /run/user/1000/podman/podman.sock
```

*Tell Waypoint to use it*

```
$ export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/podman/podman.sock
$ waypoint install -platform=docker -accept-tos
✓ Pulling image: hashicorp/waypoint:latest
 │ e338015c18dd: Download complete
 │ 3a0d753041d6: Download complete
 │ 583006983478: Download complete
 │ b1709af6d764: Downloading [==================>                                ]
 │  41.75MB/114.2MBlling fs layer
 │ b1709af6d764: Downloading [====================================>              ]
 │ b1709af6d764: Download complete
 │ f098031ccdab: Download complete
 │ d5243211887c: Download complete
 │ 
✓ Installing Waypoint server to docker
✓ Server container started!
✓ Server installed and configured!
✓ Successfully connected to Waypoint server in Docker!
❌ Installing runner...
❌ Initializing Docker client...
! Error installing the runner: Error response from daemon: container create: statfs /var/run/docker.sock:
  permission denied
  
  The Waypoint runner failed to install. This error occurred after the
  Waypoint server was successfully installed. Your CLI is configured to
  use the installed server. If you want to retry, you must uninstall the
  server first.
```

[...]

PR #4246, opened 2022-11-30
feat(cli): add socket-path flag to runner install

Codelax:
Add a socket-path flag to `waypoint runner install`, this allows to override the socket that will be bound using volume when creating the runner.
The issue is that docker socket may not be `/var/run/docker.sock` on the server that will host the runner, this would be in another place when running rootless docker for example.
This also helps greatly to start runners when using podman (ref: #3781).

### DIFF
--- a/.changelog/4246.txt
+++ b/.changelog/4246.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+cli: add socket-path flag to runner install
+```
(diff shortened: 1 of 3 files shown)


paladin-devops (approved):
Looks good to me! Tried this with the flag used and without, both worked. 😄 

```shell
$ waypoint runner install -platform=docker -server-addr=<ADDR> -server-tls-skip-verify=true -server-tls=true
```

```shell
$ waypoint runner install -platform=docker -docker-socket-path=/var/run/docker.sock -server-addr=<ADDR> -server-tls-skip-verify=true -server-tls=true
```

PR #4251, opened 2022-12-01
HCP Packer Config Sourcer plugin

paladin-devops:
The HCP Packer plugin added in this PR enables users to source a machine image ID from an HCP Packer channel. The user-configured bucket is checked for the specified channel. Within the channel, if the latest machine image iteration has a build that matches the desired cloud provider and region, it is returned to Waypoint. Below is an example of this being used to source the desired Nginx base image for a ReactJS app build, where the sourced machine image is templated into the Dockerfile.

#### Configuration source settings
```shell
$ waypoint config source-set -type=packer -config=client_id=sunset -config=client_secret=sarsaparilla -config=organization_id=nuka-cola -config=project_id=quantum
```

#### waypoint.hcl
```hcl
...
variable "image" {
  default = dynamic("packer", {
    bucket  = "nginx"
    channel = "base"
    // Packer registers the region and cloud provider for a Docker image as "docker".
    // Other Packer plugins may do things differently.
    region  = "docker" 
    cloud   = "docker"
  })
  type        = string
  description = "The name of the base image to use for building app Docker images."
}

app "packer-plugin-app" {
  build {
    use "docker" {
      dockerfile = templatefile("${path.app}/Dockerfile", {
        base_image = var.image
      })
    }
  }
...
```

#### Dockerfile
```Dockerfile
# Multi-stage Docker build
FROM node:19.2-alpine as build
WORKDIR /app
ENV PATH /app/node_modules/.bin:$PATH
COPY package.json ./
COPY package-lock.json ./
RUN npm ci --silent
RUN npm install react-scripts@3.4.1 -g --silent
COPY . ./
RUN npm run build

# Below is where the Docker repository are templated for building the image
FROM ${base_image}
COPY nginx/default.conf /etc/nginx/conf.d/
COPY --from=build /app/build /usr/share/nginx/html
EXPOSE 80
CMD ["nginx", "-g", "daemon off;"]
```


builtin/packer/config_sourcer.go
```diff
@@ -0,0 +1,235 @@
...
+	OrganizationId string `hcl:"organization_id,attr"`
+	ProjectId      string `hcl:"project_id,attr"`
+}
+
+type reqConfig struct {
+	Bucket  string `hcl:"bucket,attr"`
+	Channel string `hcl:"channel,attr"`
+	Region  string `hcl:"region,attr"`
+	Cloud   string `hcl:"cloud,attr"`
+}
```
briancain, line 36: Both this and the values in `sourceConfig` could use some quick/short doc strings for each parameter!

builtin/packer/config_sourcer.go
```diff
@@ -0,0 +1,235 @@
...
+func (cs *ConfigSourcer) Documentation() (*docs.Documentation, error) {
+	doc, err := docs.New(
+		docs.FromConfig(&sourceConfig{}),
+		docs.RequestFromStruct(&reqConfig{}),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	doc.Description("Read machine image information from HCP Packer.")
```
briancain, line 142: We should probably have a bit more information here, like for example that only container images are valid, etc.

builtin/packer/config_sourcer.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		channelParams.BucketSlug = packerConfig.Bucket
+		channelParams.Slug = packerConfig.Channel
+
+		// An HCP Packer channel points to a single iteration of a bucket.
+		channel, err := hcpPackerClient.PackerServiceGetChannel(channelParams, nil)
+		if err != nil {
+			return nil, err
+		}
+		log.Debug("Retrieved HCP Packer channel.")
```
briancain, line 103: I'd log which channel was retrieved here!

builtin/packer/config_sourcer.go
```diff
@@ -0,0 +1,235 @@
...
+		iteration := channel.Payload.Channel.Iteration
+
+		// An iteration can have multiple builds, so we check for the first build
+		// with the matching cloud provider and region.
+		for _, build := range iteration.Builds {
+			if build.CloudProvider == packerConfig.Cloud {
+				log.Debug("Found build with matching cloud provider.")
+				for _, image := range build.Images {
+					if image.Region == packerConfig.Region {
+						log.Debug("Found image with matching region.")
```
briancain, line 113: I'd log which image and region were matched

builtin/packer/config_sourcer.go
```diff
@@ -0,0 +1,235 @@
...
+		// An HCP Packer channel points to a single iteration of a bucket.
+		channel, err := hcpPackerClient.PackerServiceGetChannel(channelParams, nil)
+		if err != nil {
+			return nil, err
+		}
+		log.Debug("Retrieved HCP Packer channel.")
+		iteration := channel.Payload.Channel.Iteration
+
+		// An iteration can have multiple builds, so we check for the first build
+		// with the matching cloud provider and region.
```
briancain, line 107: This is interesting - is there a chance that we could unintentionally be picking the wrong build here? Or is "first" build essentially `latest` build?

builtin/packer/config_sourcer.go
```diff
@@ -0,0 +1,235 @@
...
+		for _, build := range iteration.Builds {
+			if build.CloudProvider == packerConfig.Cloud {
+				log.Debug("Found build with matching cloud provider.")
```
briancain, line 110: Same comment here - always nice to include extra relevant info on what was found in these log statements!

briancain (changes requested):
Nice job on this one! I've got a few minor code comments, but this looks great and will be an interesting addition to the Waypoint plugin collection! ✨

builtin/packer/config_sourcer.go
paladin-devops: With one `packer build` operation, I have never tried to push more than one image of the same cloud provider and region to the same bucket in an HCP Packer registry. 🤔 I don't know for certain if it's possible or not, but maybe we could add more specific input parameters here in a follow up PR if it seems useful?

builtin/packer/config_sourcer.go
paladin-devops: I'm hesitant to point out that only container images will work, since Packer supports more than just containers. Our plugins do work almost exclusively with containers, but since plugins don't _have_ to, and since we have the EC2 plugin, I think it's okay to leave it at "image". However, I will at least add some color to this description! 😃

builtin/packer/config_sourcer.go
briancain: Sure - totally fine to fix this up in a follow-up PR! Make an issue so we don't lose track 👍🏻

PR #4258, opened 2022-12-02
add slack notification for failed backports

krantzinator:
This workflow will check if a newly opened PR is a draft, and if it has "Backport" in the title. If both of those are true (which will both be true for failed backport PRs), it sends a small message to our team's Slack channel.
Screen Shot 2022-12-02 at 4 32 56 PM

Just as a nudge, since this message will come within a minute or less of someone merging in a PR that had the `backport/*` label attached. I've updated the Slack workflow so it will now post to our team's channel, and added the necessary secrets to this repo.


gregone (approved):
☑️ I really like the use of `:fail-sucess:`

PR #4272, opened 2022-12-07
use pull_request_target instead of pull_request

krantzinator:
This workflow didn't run on the most recent failed backport. Upon inspection of the GitHub docs:
>Note: Workflows will not run on pull_request activity if the pull request has a merge conflict.

Switching to `pull_request_target` should fix this.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target


briancain (approved):
🎯

PR #4294, opened 2022-12-08
Enhance Waypoint Job Cancel Output

cicoyle:
Previously, if a user tried cancelling a `Running` job, then it would output:
image

We should give users more context if the job is in a `Running` state upon canceling.
image


internal/cli/job_cancel.go
```diff
@@ -72,7 +72,18 @@ func (c *JobCancelCommand) Run(args []string) int {
...
-		s.Update("Marked job %q for cancellation", jobId)
+		job, err := c.project.Client().GetJob(ctx, &pb.GetJobRequest{JobId: jobId})
+		if err != nil {
+			c.ui.Output("Job not found: %s", clierrors.Humanize(err),
+				terminal.WithErrorStyle())
+			return 1
+		}
+		switch job.State {
+		case pb.Job_RUNNING:
+			s.Update("Job %q is still running. To remove Waypoint's knowledge of it, add the -dangerously-force flag.", jobId)
```
briancain, line 83: I don't think we should recommend users to use this flag because it is dangerous for a reason 😅 Waypoint should be gracefully handling job cancellations so if it's not then we should fix the bug!

How about something like this:


```
brian@ubuntu:~ λ waypoint job cancel 01GKVT0JKCG172J76JBW107HZZ
✓ Marked job "01GKVT0JKCG172J76JBW107HZZ" for cancellation

Waypoint will gracefully cancel the requested job and wait for any
downstream listeners to close the connection. This could take a while.
```

Then if it's still stuck, the flag is still around but only if they look through the help?

briancain (changes requested):

evanphx (approved):

briancain (approved):

PR #4297, opened 2022-12-09
More detailed runner job stream logging

izaaklauer:
We're getting vauge error messages now, and need to figure out where they're coming from.

### DIFF
--- a/pkg/server/singleprocess/service_runner.go
+++ b/pkg/server/singleprocess/service_runner.go
@@ -564,7 +564,8 @@ func (s *Service) RunnerJobStream(
 	// the preexisting job. Otherwise, we assign a new job.
 	var job *serverstate.Job
 	reattach := false
-	if jobId := reqEvent.Request.ReattachJobId; jobId != "" {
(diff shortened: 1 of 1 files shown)


pkg/server/singleprocess/service_runner.go
```diff
@@ -743,7 +743,7 @@ func (s *Service) RunnerJobStream(
 			ws := memdb.NewWatchSet()
 			job, err = s.state(ctx).JobById(ctx, job.Id, ws)
 			if err != nil {
-				errCh <- err
+				errCh <- errors.Wrapf(err, "failed getting job by id from state %q", job.Id)
```
briancain, line 746: Since we're reassigning job on 744, is it possible it's nil here and this might cause a panic?

briancain: Maybe we should grab the Id as its own var for these err messages

briancain (approved):

demophoon (approved):

PR #4298, opened 2022-12-09
Upgrading to go 1.19

izaaklauer:
Most of these changes are the result of running the new go 1.19 formatter.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,7 +236,7 @@ jobs:
           at: .
       - run:
           command: |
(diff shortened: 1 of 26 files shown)


briancain (approved):
I haven't tried it, but if the integration tests pass, let's ship it!

paladin-devops (approved):
I was able to build the binary as well as the Docker image for the server and runner w/`make bin` and `make docker/server`. I ran a few commands as well, and attempted a remote operation - the ODR started up and tried running the job. Looks great to me! 1️⃣.1️⃣9️⃣

PR #4319, opened 2022-12-12
cli: Respect DataSourceOverrides CLI flag

briancain:
Prior to this commit, the job template submitted by the pipeline run CLI would ignore any data source overrides defined by the base CLI flag. This commit updates that to respect those overrides, if set.

### DIFF
--- a/.changelog/4319.txt
+++ b/.changelog/4319.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+cli: Respect `-remote-source` overrides for submitted job template when running
+`waypoint pipeline run`.
+```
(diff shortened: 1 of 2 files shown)


catsby (approved):

catsby (approved):
Should we add a changelog entry? It seems like a bug fix or at the least a change in behavior if users are using that flag and perhaps not noticing that it's not working as intended

briancain:
@catsby already added a changelog! I add that label when I open the PR so the commit doesn't get a PR check failure, and then remove it when I commit the changelog with the right PR number.

PR #4324, opened 2022-12-13
server: Ensure a job is not stuck in Running and is Completed on err processing job events

briancain:
This pull request attempts to add more detailed logging in the bad event where Waypoint server fails to process a job event, and loses the job id that it was originally processing events for. Additionally, it will attempt to complete the original job we were processing with the original jobId so that Running jobs aren't stuck and don't lock out the users ability to run jobs.


izaaklauer (approved):

catsby (approved):

cicoyle (approved):

paladin-devops (approved):
I like it, failing successfully! ✅

Issue #3994, opened 2022-10-05
Need a mechanism for displaying user-facing errors


izaaklauer:
## Background

We've recently updated the errors the server returns to clients so that we're not returning raw error messages - only errors defined in the highest-level service handler. This is important, as it removes the possibility of us leaking potentially sensitive server details to users through errors. Example: https://github.com/hashicorp/waypoint/pull/3986

## The problem

There are some placed, though, where we are building error messages below the top-level handler that we expect users to see. E.x: https://github.com/hashicorp/waypoint/blob/21e46cf80a8e2dbc0edc2c4b6c4085ffc579c003/pkg/server/singleprocess/service_job.go#L142-L146

This is currently being obfuscated by this externalize: https://github.com/hashicorp/waypoint/blob/21e46cf80a8e2dbc0edc2c4b6c4085ffc579c003/pkg/server/singleprocess/service_pipeline.go#L235-L240

This manifests when running pipelines:

```
$ waypoint pipeline run up

» Running pipeline "up" for application "web"
❌ Requesting to queue run of pipeline "up"...

» Performing operation locally
! 1 error occurred:
  	* rpc error: code = FailedPrecondition desc = error queueing jobs for pipeline
  
```
The waypoint server tried to tell me the problem was a missing datasource, but that error was eaten by externalization.

## Solutions

I think we need some way to indicate text that a user can see while creating an error. Externalize can then leave that text, and strip everything else.

Allan-Nava:
Exactly I have the same problem

briancain:
@Allan-Nava can you be more specific as to which problem you have? Thanks!

briancain:
For handling this, we could introduce a way to mark an err as "sensitive" so we don't accidentally leak sensitive errors to the user, but can still pass through useful errors like missing data source.

PR #4327, opened 2022-12-13
Adding mechanism for surfacing deep errors to users

izaaklauer:
Closes https://github.com/hashicorp/waypoint/issues/3994 (and recommend reading that for context)

Big thanks to @cicoyle for pairing on this! 

Will squash before merging

## What does this do?

This adds a new error type in hcerr. If a function other than the top-level serivice handler (who calls hcerr.Externalize) wants to surface user-facing detail, they can use `hcerr.UserError*` functions to create a new error that contains user-safe detail. `hcerr.Externalize` will notice if the error stack has any UserErrors on it, and will gather those details and surface them to the user in the final error.

## Demonstration

Running `waypoint pipeline run` on a project without a remote ref before:

```
$ waypoint pipeline run up

» Running pipeline "up" for application "web"
❌ Requesting to queue run of pipeline "up"...

! 1 error occurred:
  	* rpc error: code = FailedPrecondition desc = error queueing jobs for pipeline
  
```

and after

```
$ waypoint pipeline run hello

» Running pipeline "hello" for application "web"
❌ Requesting to queue run of pipeline "hello"...

! 1 error occurred:
  	* rpc error: code = FailedPrecondition desc = error queueing jobs for pipeline
  Project nginx-project-metrictest does not have a data source configured. Remote
  jobs require a data source such as Git to be configured with the project. Data
  sources can be configured via the CLI or UI. For help, see :
  https://www.waypointproject.io/docs/projects/git#configuring-the-project

```

### DIFF
--- a/pkg/server/hcerr/hcerr_test.go
+++ b/pkg/server/hcerr/hcerr_test.go
@@ -1,6 +1,7 @@
 package hcerr
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/hashicorp/go-hclog"
@@ -54,4 +55,25 @@ func TestExternalize(t *testing.T) {
 		details := statusErr.Details()
 		require.Len(details, 2)
 	})
+}
+
+func TestUserError(t *testing.T) {
+	log := hclog.New(&hclog.LoggerOptions{Level: hclog.Debug})
+
+	baseError := errors.New("base")
+
+	userError := UserErrorf(baseError, "user facing message")
+
+	finalError := Externalize(log, userError, "top-level message")
+	fmt.Println(finalError)
+
+	t.Run("Multiple levels of user error", func(t *testing.T) {
+		baseError := errors.New("base")
+		userError1 := UserErrorf(baseError, "user facing message 1")
+		userError2 := UserErrorf(userError1, "user facing message 2")
+
+		finalError := Externalize(log, userError2, "top-level message")
+		fmt.Println(finalError)
+	})
+
 }
--- a/pkg/server/hcerr/hcerr_test.go
+++ b/pkg/server/hcerr/hcerr_test.go
@@ -1,7 +1,6 @@
 package hcerr
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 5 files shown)


pkg/server/hcerr/hcerr_test.go
```diff
@@ -55,3 +55,42 @@ func TestExternalize(t *testing.T) {
...
+		require.NotContainsf(finalError.Error(), "failed to connect to db at 10.0.0.sensitive", "contains base error")
+		require.Contains(finalError.Error(), "missing required field 'foo'. Try resubmitting after specifying 'foo'")
+		require.NotContainsf(finalError.Error(), "failed to do specific internal thing", "contains middle internal error")
+		require.Contains(finalError.Error(), "invalid value 'baz' for argument 'bar'")
+		require.Contains(finalError.Error(), "failed to perform action foobarbaz")
+	})
+
+	t.Run("Supports status codes", func(t *testing.T) {
+		internalStatusErr := status.Errorf(codes.Internal, "unknown internal failure")
+		userErr := UserErrorWithCodef(codes.InvalidArgument, internalStatusErr, "it's not an internal error actually, you made a mistake o user")
```
cicoyle, line 91: Did you mean to put `o`?

pkg/server/hcerr/user_error.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// UserError is a custom error type designed to package
+// detail that will be displayed to a user, alongside
```
cicoyle, line 11:
```suggestion
// details that will be displayed to a user, alongside
```

cicoyle (approved):
This is so helpful! Thanks for letting me tackle it with you. Looks like 1 test is failing, but otherwise it looks great!!

pkg/server/hcerr/hcerr_test.go
briancain: I think it's phonetic... like _O' User_ instead of "oh" 😄

briancain (approved):
Makes sense....will there be more PRs to update the rest of the service errors?

pkg/server/hcerr/hcerr_test.go
```diff
@@ -55,3 +55,42 @@ func TestExternalize(t *testing.T) {
...
+	log := hclog.New(&hclog.LoggerOptions{Level: hclog.Debug})
+	require := require.New(t)
+
+	t.Run("One level of user errors", func(t *testing.T) {
+		baseError := errors.New("failed to connect to db at 10.0.0.sensitive")
+		userError := UserErrorf(baseError, "missing required field 'foo'. Try resubmitting after specifying 'foo'")
+
+		finalError := Externalize(log, userError, "failed to perform action foobarbaz")
+
+		require.NotContainsf(finalError.Error(), "failed to connect to db at 10.0.0.sensitive", "contains base error")
```
briancain, line 69: Looks like this is panic'ing

pkg/server/hcerr/hcerr_test.go
izaaklauer: Brian got it 😆 . I'm being a bit silly and using old english because the string in the test doesn't really matter. https://www.allgreatquotes.com/twelfth-night-quote-67/

izaaklauer:
> Makes sense....will there be more PRs to update the rest of the service errors?

I haven't done a survey for other places where it's obvious that we wanted to surface errors to users. As we find them, we can use this to fix them! I don't think there are many more.

PR #4330, opened 2022-12-14
internal/core: Gracefully handle bad proto message from destroyments

briancain:
Prior to this commit, if a plugin returned a non-Deployment proto message, we would hard fail the entire operation. Considering that we are retreiving this message to print what resources were left around or destroyed, we should not hard fail on this and instead print the error to the user that we cannot tell them about these resources and move on. It also incldues the "type" that was returned.

### DIFF
--- a/internal/core/app_deploy_destroy.go
+++ b/internal/core/app_deploy_destroy.go
@@ -138,7 +138,7 @@ func (a *App) destroyDeploy(
 	if !ok {
 		a.UI.Output("Failed to convert operation to a Deployment proto message, "+
 			"cannot detail which resources were destroyed or left around. The deployment "+
-			"message is of type %T"+
+			"message is of type %T. "+
 			"Please report this issue if it persists. Waypoint will continue on...",
 			reflect.TypeOf(destroyment),
 			terminal.WithErrorStyle())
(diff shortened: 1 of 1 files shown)


paladin-devops (approved):
I think this is a good approach! Getting stuck in a state where a deployment can't be destroyed for this reason is a bad experience for users - they'd hit a failure message at minimum after each release (when pruning runs), and every destroy attempt would fail.

cicoyle (approved):
Great idea to notify users better upon running into this issue.

internal/core/app_deploy_destroy.go
```diff
@@ -127,10 +128,21 @@ func (a *App) destroyDeploy(
...
+			"determine what old resources were destroyed or left behind", terminal.WithWarningStyle())
+		return nil
+	}
 
 	destroyProto, ok := destroyment.(*pb.Deployment)
 	if !ok {
-		return errors.New("error converting operation message into a Deployment proto message")
+		a.UI.Output("Failed to convert operation to a Deployment proto message, "+
+			"cannot detail which resources were destroyed or left around. The deployment "+
+			"message is of type %T"+
```
briancain, line 141:
```suggestion
			"message is of type %T. "+
```

PR #4338, opened 2022-12-15
Update mockery 2 15

catsby:
Our version of [vektra/mockery](https://github.com/vektra/mockery) (1.12) is out of date and no longer supported, to the point where it's no longer compatible with modern versions of Go. 

In 6dfaae2947af76bef8072ff9075cfada8c833730 we update to [v2.15.0](https://github.com/vektra/mockery/releases/tag/v2.15.0) and modify our code that consumes mockery and generates our mocks.

Commit 3fb28e7185ebf98cb739d943ca4fc212f5aad87e then is the result of running `make docker/tools` followed by `make docker/gen/server`.


evanphx (approved):

tools/tools.go
```diff
@@ -17,14 +17,14 @@ package tools
 //go:generate go install github.com/evanphx/grpc-gateway/protoc-gen-swagger
 
 //go:generate go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
-//go:generate go install "-ldflags=-s -w -X github.com/vektra/mockery/mockery.SemVer=2.12.1" github.com/vektra/mockery/cmd/mockery
+//go:generate go install "-ldflags=-s -w -X github.com/vektra/mockery/cmd/mockery.SemVer=2.12.1" github.com/vektra/mockery/v2@v2.15.0
```
jgwhite, line 20: I’ve never fully understood the `SemVer=2.12.1` bit of this line, but does it need updating too to match the new target version? (Or maybe we can drop it now?)

briancain: Hm yeah, I'm not sure either! But might be good to have the versions match??

briancain (approved):
I tried this locally, did a server gen and it fixed the bug, and did a waypoint up with the server and that worked. Should be good! 🙏🏻

tools/tools.go
evanphx: It's a thing that was added to mockery to verify its own version. It's not great but it's needed today.

catsby: Thanks @evanphx , I updated to match `2.15.0` in d337936bf

PR #4340, opened 2022-12-16
state: add `JobLatestInit`

jgwhite:
Should help us to speed up loading of the project UI against BoltDB and other state implementations.

### DIFF
--- a/internal/server/boltdbstate/job.go
+++ b/internal/server/boltdbstate/job.go
@@ -1169,6 +1169,57 @@ func (s *State) JobIsAssignable(ctx context.Context, jobpb *pb.Job) (bool, error
 	}
 }
 
+// JobLatestInit looks up the latest init job for the given project. The returned
+// Job will be a deep copy of the job so it is safe to read/write. If the job
+// can't be found, a nil result with no error is returned.
+func (s *State) JobLatestInit(ctx context.Context, proj *pb.Ref_Project) (*pb.Job, error) {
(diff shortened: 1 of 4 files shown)


internal/server/boltdbstate/job.go
```diff
@@ -1169,6 +1169,57 @@ func (s *State) JobIsAssignable(ctx context.Context, jobpb *pb.Job) (bool, error
 	}
 }
 
+// JobLatest looks up the latest init job for the given project. The returned
```
izaaklauer, line 1172:
```suggestion
// JobLatestInit looks up the latest init job for the given project. The returned
```

izaaklauer (approved):
Ideally, it would be great to get this by doing a ListJobs, filtering by the init job type, sorting by most recent, with a limit of 1.

I think we'll get there eventually, but this is nice and straightforward and will help performance a lot, so let's do it! I don't think we should build many more of these bespoke job-getting functions before enhancing the list endpoint instead though.

jgwhite:
> Ideally, it would be great to get this by doing a ListJobs, filtering by the init job type, sorting by most recent, with a limit of 1.
> 
> I think we'll get there eventually, but this is nice and straightforward and will help performance a lot, so let's do it! I don't think we should build many more of these bespoke job-getting functions before enhancing the list endpoint instead though.

It actually didn’t occur to me to modify ListJobs 😅 You’re absolutely right though — that is the better long term approach.

PR #4277, opened 2022-12-07
docs: Migrate link formats

ashleemboyer:
## Relevant links

- Preview URL: https://waypoint-git-docs-ambmigrate-link-formats-hashicorp.vercel.app/

## What

- Migrates the formats of links in Markdown content as part of the Docs Content Link Rewrites project.
  - See the one-pager for additional context on the project: [Dev Dot - Docs Content Link Rewrites](https://docs.google.com/document/d/1WaSyvoVPS-YCNiSPX0ynGpvc1gySEcpbrRoQVhimQCA/edit#heading=h.wx39rhxfetpg)
  - Most MDX files were changed by running the new [`scripts/docs-content-link-rewrites/rewrite-links`](https://github.com/hashicorp/dev-portal/blob/main/scripts/docs-content-link-rewrites/rewrite-links.ts) script in dev-portal on a local copy of the repo.
  - The `go` files changed in this PR are for updating the automatically generated documentation and associated JSON.
- Adds two new workflows, which call workflows that live in dev-portal
  - `.github/workflows/ test-link-rewrites.yml` -- this workflow is only intended for this current migration PR. The workflow it calls does a check for this, and will only run for branches named `docs/amb.migrate-link-formats`.
  - `.github/workflows/check-legacy-links-format.yml` -- this workflow is intended to be used to maintain the new link format. It does not yet check nav-data JSON files, but it will after some platform updates are made. At that time, no changes will be needed in this file since the workflow lives in dev-portal.

## Testing/Reviewing

- [ ] The following files need manual review the most:
  - [ ] `builtin/k8s/apply/platform.go`
  - [ ] `builtin/k8s/helm/platform.go`
  - [ ] `builtin/nomad/canary/releaser.go`
  - [ ] `builtin/nomad/jobspec/platform.go`
  - [ ] `builtin/vault/config_sourcer.go`
  - [ ] `internal/cli/app_docs.go`
- [ ] _**All**_ changes in this PR have automated tests associated with them, so make sure the following checks have passed:
  - [ ] All of the existing checks that were in place before this PR
  - [ ] The new [check-legacy-links-format](https://github.com/hashicorp/waypoint/actions/workflows/check-legacy-links-format.yml)
  - [ ] The new [test-link-rewrites](https://github.com/hashicorp/waypoint/actions/workflows/test-link-rewrites.yml)

PR #4356, opened 2022-12-22
core/server: only queue init ops for active projects

jgwhite:
## Why the change?

I noticed a few spurious init jobs while testing. This won’t have a massive impact, but should make things a little tidier.

### DIFF
--- a/pkg/server/singleprocess/service_project.go
+++ b/pkg/server/singleprocess/service_project.go
@@ -245,6 +245,10 @@ func queueInitOps(s *Service, ctx context.Context, project *pb.Project) error {
 }
 
 func projectNeedsRemoteInit(project *pb.Project) bool {
+	if project.State != pb.Project_ACTIVE {
+		return false
+	}
+
 	if project.DataSource == nil {
 		return false
 	}


ashleemboyer:
Hello @jgwhite! 😊 This is a quick comment to let you know I'm updating this PR with main so it's up to date with the changes made in [#4277](https://github.com/hashicorp/waypoint/pull/4277).

pkg/server/singleprocess/service_project.go
```diff
@@ -245,6 +245,10 @@ func queueInitOps(s *Service, ctx context.Context, project *pb.Project) error {
 }
 
 func projectNeedsRemoteInit(project *pb.Project) bool {
+	if project.State == pb.Project_DESTROYING {
```
briancain, line 248: Not sure if we'll ever make more Project states...but perhaps this could work if it makes sense to protect when we do?
```suggestion
	if project.State != pb.Project_ACTIVE {
```

briancain (approved):

pkg/server/singleprocess/service_project.go
jgwhite: Ah yeah, that makes a lot of sense. I think at the time I wasn’t 100% sure if projects created prior to project destroy being introduced would be marked as active.

briancain: I believe since it's an Enum in protobuf, it'd take the default state which is `Active`: https://github.com/hashicorp/waypoint/blob/af6141fa752085ec2536bfb3c56a15d60c5c0af5/pkg/server/proto/server.proto#L831-L838 So hopefully yes!

jgwhite: Perfect, I’ve committed that suggestion.

PR #4357, opened 2022-12-22
Allow complex non-string types for dynamic default variables

izaaklauer:
Depends on https://github.com/hashicorp/waypoint-plugin-sdk/pull/79
Demonstrated in https://github.com/hashicorp/waypoint-examples/pull/106

Internal video describing the motivation (thanks @andrew-hashicorp): https://drive.google.com/file/d/1aCDfFioVELb7Q9MZhN0XiRegUlNJ4LdM/view?usp=share_link

Internal video code walkthrough: https://drive.google.com/file/d/1-58UzSXTJFWnlBoE4LpoIQsg1hExubxV/view?usp=share_link

### What does this do?

Prior to this, waypoint only supported string types in variables with dynamic defaults. Configsourcer plugins could only emit strings, and if a user tried to use any type other than `string` for a variable with a dynamic default, waypoint would throw a validation error.

This updates configsourcer plugins to optionally return json, which is then interpreted internally as an hcl snippet and can be used as a complex type anywhere else in the hcl. It also allows users to specify non-string types, or use the `any` type to opt out of early type checking.

This also updates the terraform configsourcer to output json if a user picks a variable with a non-string json-representable type. It also allows the user to omit the `output` field, in which case the plugin returns all of the workspace's outputs as a map.

### Motivation

The primary motivation here is to make it easier to pull more values from terraform cloud.

See this example of using terraform to manage all the infrastructure associated with an ECS deployment in two environments: https://github.com/izaaklauer/acmeapp1/blob/main/waypoint.hcl. You'll notice that it involves 27 variables across three workspaces. It also forces users to take the list of subnets output and turn it into many individual string outputs, as waypoint can't natively consume the list.

This change brings us down to three variable stanzas, one per workspace: https://github.com/izaaklauer/acmeapp1/blob/443ab9e64734bcf834a9309d1035d98e505263d0/waypoint.varmulti.hcl#L74-L107

It also allows native use of the subnets list variable: https://github.com/izaaklauer/acmeapp1/blob/443ab9e64734bcf834a9309d1035d98e505263d0/waypoint.varmulti.hcl#L36

After this is merged, we intend to write a full public use-case.

### How do I test this?

I build a sample project in waypoint-examples: https://github.com/hashicorp/waypoint-examples/pull/106. You can use that to test the feature, and we can merge the example once this is released in core waypoint.


builtin/tfc/config_sourcer.go
```diff
@@ -54,7 +55,7 @@ type ConfigSourcer struct {
 }
 
 type cachedSecret struct {
-	Outputs map[string]string // The most recent outputs for a workspace
+	Outputs map[string]StateIncludeAttributes // The most recent outputs for a workspace
```
izaaklauer, line 58: Caching the full TFC attributes response now, because it's useful to look at the value and the type later on, and we're now allowing types other than `string.`

embedJson/gen/builder-aws-ami.json
```diff
@@ -12,6 +12,7 @@
          "Default": "",
          "EnvVar": "",
          "Category": false,
+         "Example": "",
```
izaaklauer, line 15: I'm not really sure why anything in `embedJson` changed when I regenerated website mdx, but it seems benign.

ashleemboyer:
Hello @izaaklauer! 😊 This is a quick comment to let you know I updated this PR with main so it's up to date with the changes made in [#4277](https://github.com/hashicorp/waypoint/pull/4277).

builtin/tfc/config_sourcer.go
```diff
@@ -411,15 +461,23 @@ config {
...
 			"refreshInternal, a source field.",
 		),
 	)
 
 	doc.SetRequestField(
 		"output",
-		"The name of the output to read the value of",
+		"Optional: The name of the output to read the value of. All outputs if blank.",
+		docs.Summary(
+			"The name of the output from terraform cloud to return.",
```
briancain, line 474:
```suggestion
			"The name of the output from Terraform Cloud to return.",
```

builtin/tfc/config_sourcer_test.go
```diff
@@ -58,33 +61,160 @@ func TestConfigSourcer(t *testing.T) {
...
+				case []interface{}:
+					require.ElementsMatch(v, actualV)
+				}
+			}
+		}
+
+	})
+
+	t.Run("Can read a single var with a complex type", func(t *testing.T) {
+
```
briancain, line 218: Missing test?

internal/appconfig/watcher.go
```diff
@@ -852,6 +852,40 @@ func buildAppConfig(
 					}
 				}
 
+			case *sdkpb.ConfigSource_Value_Json:
+				if req.cv.Internal {
+					// We don't yet support using non-string types as internal vals.
+					// Logging and moving along isn't great though - the user
+					// won't have much indication as to why their variable isn't working.
+					L.Warn("Complex-typed outputs cannot be used as internal values. Skipping.", "var name", req.cv.Name)
+				} else {
```
briancain, line 861: I think we should probably in-line the remaining if-statements to be `else if`? I don't think we need this else at least

briancain (commented):
Haven't had a chance to test, but have a few comments! Overall looks pretty good though 👍🏻

builtin/tfc/config_sourcer_test.go
izaaklauer: Ah, good catch! I actually have that case covered earlier, so this can be deleted.

internal/appconfig/watcher.go
izaaklauer: Updated for clarity 👍

internal/config/variables/variables.go
```diff
@@ -237,18 +240,24 @@ func decodeVariableBlock(
...
-			// have the value. For now we require v.Type to be string so that
-			// the user isn't surprised by anything. Users can use explicit
-			// type conversion such as `tonumber`.
-			if v.Type != cty.String {
+			// Dynamic types can either be strings, or values that can be
+			// represented as json (i.e. objects and maps). The only allowed
+			// primitive type is string, and users can use explicit
+			// type conversion such as `tonumber` to achieve other primitive types.
+
+			// This is intended to help catch users catch invalid types early. If
```
cicoyle, line 248:
```suggestion
			// This is intended to help users catch invalid types early. If
```

internal/config/variables/variables.go
```diff
@@ -237,18 +240,24 @@ func decodeVariableBlock(
...
+			case cty.Number, cty.Bool:
 				diags = append(diags, &hcl.Diagnostic{
 					Severity: hcl.DiagError,
 					Summary:  fmt.Sprintf("type for variable %q must be string for dynamic values", name),
 					Detail: "When using dynamically sourced configuration values, " +
-						"the variable type must be string. You may use explicit " +
-						"type conversion functions such as `tonumber` when using " +
-						"the variable.",
+						"can either be strings, or complex types. If you're unsure of " +
+						"which, consult your configsourcer documentaiton. If you want " +
```
cicoyle, line 258:
```suggestion
						"which, consult your configsourcer documentation. If you want " +
```

cicoyle (approved):
This looks great! It looks like more changes due to the `embedJson/` changes, but yeah those look benign.

builtin/tfc/config_sourcer.go
```diff
@@ -300,6 +303,14 @@ func (cs *ConfigSourcer) read(
 
 				continue
 			}
+			if resp.StatusCode != 200 {
+				L.Error("error in sending request for state version for workspace %q, unexpected response code %q", id, resp.Status)
```
briancain, line 307: Maybe we should log `resp.StatusCode` too here? Not sure if that comes included with `resp.Status`.

builtin/tfc/config_sourcer.go
```diff
@@ -333,10 +344,57 @@ func (cs *ConfigSourcer) read(
...
+				Json: allOutputsJson,
+			}
+		} else {
+			value := cachedSecretVal.Outputs[tfcReq.Output]
+
+			switch value.Type {
+			case "string":
+				stringVal, ok := value.Value.(string)
+				if !ok {
+					return nil, fmt.Errorf("variable %s is a string according to TFC, but is not actually a string type", value.Name)
```
briancain, line 371:
```suggestion
					return nil, fmt.Errorf("variable %q is a string according to TFC, but is not actually a string type", value.Name)
```

briancain: Maybe also include the type? `value.Type`

builtin/tfc/config_sourcer.go
```diff
@@ -411,15 +469,23 @@ config {
...
 		"output",
-		"The name of the output to read the value of",
+		"Optional: The name of the output to read the value of. All outputs if blank.",
+		docs.Summary(
+			"The name of the output from Terraform Cloud to return.",
+			"Only strings and objects representable as json (i.e. maps and lists)",
+			"are currently supported.",
+			"If unspecified, all outputs from the workspace will be read",
+			"and returned as a map[string]object, and can be referenced as such",
+			"in the waypoint.hcl. See <docs> for more details.",
```
briancain, line 487: Was `<docs>` intended to be a URL?

embedJson/gen/builder-aws-ami.json
briancain: Ah, I think this is because of the `waypoint-plugin-sdk` changes. It added a new `Example` field, and so the re-gen is adding these for every doc given that SDK change.

internal/config/variables/variables.go
```diff
@@ -524,13 +533,67 @@ func LoadDynamicDefaults(
...
+						// data or not. That's work though, and option 2 is a happy side effect
+						// of us not knowing, so it's OK for today.
+
+						// Make sure the plugin returned valid non-string data.
+						if err := json.Unmarshal(f.Data, &json.RawMessage{}); err != nil {
+							diags = append(diags, &hcl.Diagnostic{
+								Severity: hcl.DiagError,
+								Summary:  "Plugin output <> hcl type mismatch",
+								Detail: fmt.Sprintf(
+									"Variable %s is declared as non-string type %s, \n"+
```
briancain, line 565:
```suggestion
									"Variable %q is declared as non-string type %q, \n"+
```

demophoon (approved):

internal/appconfig/watcher_test.go
```diff
@@ -486,11 +486,72 @@ func TestWatcher_variableEscape(t *testing.T) {
...
+	cfg, _, err := w.Next(ctx, 0)
+	require.NoError(err)
+
+	// It is undeniably weird that we're returning json data as raw file contents.
+	// We've very much outgrown the current app config system that returns either
+	// files or env vars. IMO, we should refactor this system to return structured data,
+	// and the caller can decide to use it for files, env vars, or some other purpose.
+	//
+	// For now, we're doubling down on the existing hack wherein the variable system
+	// requests it's for its dynamic config vars to be treated as "files", and it
```
paladin-devops, line 539: Did you mean this?
```suggestion
	// requests its dynamic config vars to be treated as "files", and it
```

internal/config/variables/variables.go
```diff
@@ -237,18 +240,24 @@ func decodeVariableBlock(
...
+			case cty.Number, cty.Bool:
 				diags = append(diags, &hcl.Diagnostic{
 					Severity: hcl.DiagError,
 					Summary:  fmt.Sprintf("type for variable %q must be string for dynamic values", name),
 					Detail: "When using dynamically sourced configuration values, " +
-						"the variable type must be string. You may use explicit " +
-						"type conversion functions such as `tonumber` when using " +
-						"the variable.",
+						"can either be strings, or complex types. If you're unsure of " +
+						"which, consult your configsourcer documentation. If you want " +
```
paladin-devops, line 258: nit
```suggestion
						"which, consult your configsourcer plugin documentation. If you want " +
```

paladin-devops (commented):
Just read through, this looks great! I haven't tested, but had a few minor comments. Can't wait to see this in a use case! 😄

builtin/tfc/config_sourcer.go
izaaklauer: It's included! Just realized that i'm using string formatting in the logger though - should have included k/v pairs at the end instead. That always gets me 😅

builtin/tfc/config_sourcer.go
izaaklauer: Done!

builtin/tfc/config_sourcer.go
izaaklauer: Fixed! Not sure which docs I meant there, but I linked to waypoint-examples.

internal/appconfig/watcher.go
```diff
@@ -852,6 +852,41 @@ func buildAppConfig(
 					}
 				}
 
+			case *sdkpb.ConfigSource_Value_Json:
+				if req.cv.Internal {
+					// We don't yet support using non-string types as internal vals.
+					// Logging and moving along isn't great though - the user
+					// won't have much indication as to why their variable isn't working.
+					L.Warn("Complex-typed outputs cannot be used as internal values. Skipping.", "var name", req.cv.Name)
```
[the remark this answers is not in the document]
izaaklauer: Good point! Done.

builtin/tfc/config_sourcer.go
briancain: Oh yes totally! Yeah update those loggers with the key/value format instead, I forget about that one myself.

PR #4359, opened 2023-01-03
docs: Updating legacy links

ashleemboyer:
Updates a few legacy links that were merged into `main` since #4277


krantzinator (approved):
Thank you!

PR #4362, opened 2023-01-04
Fixed Stanzas example

Eyal-Shalev:
Added missing `"`


briancain (approved):
Good catch, thank you!

PR #4379, opened 2023-01-09
core/api: Add grpc-gateway api

demophoon:
This PR enables gRPC-Gateway on the http port of Waypoint server to add in an HTTP API for interfacing with Waypoint.

Most of this PR is autogenerated from the existing `server.swagger.json` file that is directly generated from our protos.


internal/server/http.go
```diff
@@ -48,8 +50,17 @@ func newHttpServer(grpcServer *grpc.Server, ln net.Listener, opts *options) *htt
 	// gRPC server. This is used by the exec handler.
 	grpcAddr := opts.GRPCListener.Addr().String()
 
+	// Create grpc-gateway muxer
+	grpcMux := runtime.NewServeMux()
+	grpcOpts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
+	err := gen.RegisterWaypointHandlerFromEndpoint(opts.Context, grpcMux, "localhost:9701", grpcOpts)
+	if err != nil {
+		log.Error("Unable to register waypoint grpc gateway service")
```
briancain, line 58: Hrm, I think we'd want to return this error too instead of only logging it?

evanphx (approved):
Innocently small once you remove the generated files!

internal/server/http.go
demophoon: Resolved in `c9a3fd348844a628157b0e0167822bb14764024d`

.changelog/4379.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+api: Enable HTTP API to talk to Waypoint
```
briancain, line 2: We should make this more descriptive, something like

```suggestion
server: Enable gRPC-Gateway on the http port of Waypoint server to add in an HTTP API for interfacing
```

briancain (approved):

PR #4381, opened 2023-01-10
Fixing job heartbeat test

izaaklauer:
Prior to this change, we would cancel our context, but then use that context for a state lookup. This causes the lookup to fail with a context canceled error, which isn't our desired result.


xiaolin-ninja (approved):
Such a simple fix! LGTM.

PR #4382, opened 2023-01-10
Scoped Config Sources Upgrade Fix

### LINKED ISSUE #4360: Vault Configuration Is Broken After 0.10.14
**Describe the bug**
We had a server that was previously running 0.10.3 auto-upgrade to 0.10.5 and all of the applications are failing because there seems to be two different vault configurations.

**Steps to Reproduce**

I was able to replicate locally by querying the 0.10.5 server with a client that was 0.10.3 and one that was 0.10.4 or greater.

```
> waypoint --version
CLI: v0.10.5 (ece0f7541)
Server: v0.10.5

> waypoint config source-get -type=vault                                                                                                                                                                           130 ↵
   KEY  |             VALUE
--------+---------------------------------
  addr  | https://our-old-vault-server
  token | an-old-token


> waypoint --version
CLI: v0.10.3 (136a885b9)
Server: v0.10.5

> waypoint config source-get -type=vault
   KEY  |                                              VALUE
--------+--------------------------------------------------------------------------------------------------
  token | our-current-token
  addr  | https://our-new-vault-server
```

Please include any `waypoint.hcl` files if applicable, as well as a
[GitHub Gist](https://gist.github.com/) of any relevant logs or steps to
reproduce the bug. Running `waypoint` commands with `-v` up to `-vvv` will
include any additional debugging info in the log.

**Expected behavior**
Doing a `waypoint config source-set` should point to one configuration.

Everything seemed fine prior to 0.10.4. A new server and older clients seem to work fine together.

**Additional context**

I've tried resetting and deleting the configuration from both versions of the CLI and have had no luck. If I do a `waypoint config source-set ....` with 0.10.5 CLI, it does not seem to take.

paladin-devops:
This PR updates the Bolt DB implementation of config sources to "upgrade" a pre-v0.10.4 database record from the config_source table. Prior to v0.10.4, the unique identifier for each record in the table was the name of the plugin. This was changed in v0.10.4 with the addition of [scoped config sources](https://github.com/hashicorp/waypoint/pull/4191), and now the ID is a hashed value, whose hash structure is defined by the plugin name, scope, and workspace name.

This PR therefore deletes any "old" records, those whose ID is the name of any known config sourcer plugin, from Bolt DB, and creates new records with the ID correctly hashed. With this fix, users who had config sources prior to v0.10.3 may continue to interact with those config sources, whereas before this commit, they could not, and the config sources were permanently "stuck" in the database as they were before upgrading.

Fixes #4360.

### DIFF
--- a/internal/server/boltdbstate/config_source.go
+++ b/internal/server/boltdbstate/config_source.go
@@ -300,6 +300,10 @@ func (s *State) configSourceIndexSet(txn *memdb.Txn, id []byte, value *pb.Config
 // configSourceIndexInit initializes the config index from persisted data.
 func (s *State) configSourceIndexInit(dbTxn *bolt.Tx, memTxn *memdb.Txn) error {
 	bucket := dbTxn.Bucket(configSourceBucket)
+
+	// Regex to match for the names of any config sourcer plugins v0.10.5 and earlier
+	re := regexp.MustCompile(`aws-ssm|consul|kubernetes|null|packer|terraform-cloud|vault`)
+
 	return bucket.ForEach(func(k, v []byte) error {
 		var value pb.ConfigSource
 		if err := proto.Unmarshal(v, &value); err != nil {
@@ -312,7 +316,6 @@ func (s *State) configSourceIndexInit(dbTxn *bolt.Tx, memTxn *memdb.Txn) error {
 		// is the simplest way for users to upgrade since custom config sourcer
 		// plugins aren't yet supported, as of 1/10/2023.
 		key := string(k)
-		re := regexp.MustCompile(`aws-ssm|consul|kubernetes|null|packer|terraform-cloud|vault`)
 		if re.MatchString(key) {
 			if err := bucket.Delete(k); err != nil {
 				return err
--- a/internal/server/boltdbstate/config_source.go
+++ b/internal/server/boltdbstate/config_source.go
@@ -3,6 +3,7 @@ package boltdbstate
 import (
 	"context"
 	"encoding/binary"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/server/boltdbstate/config_source.go
```diff
@@ -304,10 +305,28 @@ func (s *State) configSourceIndexInit(dbTxn *bolt.Tx, memTxn *memdb.Txn) error {
...
-			return err
-		}
 
+		// If there are any records whose ID is the name of a plugin, pre-v0.10.4
+		// behavior, then we delete that record from the database, since it is
+		// about to be saved with a hashed ID. This is not very elegant, but
+		// is the simplest way for users to upgrade since custom config sourcer
+		// plugins aren't yet supported, as of 1/10/2023.
+		key := string(k)
+		re := regexp.MustCompile(`aws-ssm|consul|kubernetes|null|packer|terraform-cloud|vault`)
```
izaaklauer, line 315: Nit: I think it's good for performance to compile your regex once at startup, and then use it throughout (example here: https://github.com/hashicorp/waypoint/blob/e19fa4dbdc6a80007a82311790a2e29d766f84fd/internal/cli/help.go#L167)

I don't think it really matters here, because I don't think this is in a hot loop, so feel free to disregard in this case.

izaaklauer (approved):

internal/server/boltdbstate/config_source.go
paladin-devops: @izaaklauer thanks for pointing that out! with `bucket.ForEach` we do loop through each record in the bucket, so I'd say it's a "warm" loop 😄 I'm going to set the regex before we enter the loop, but still in this function, since this function is called once at the start up of the server.

cicoyle (approved):
This looks great! The regex enhancement and notes make it super easy to understand why you added this. 🚀

PR #4397, opened 2023-01-13
Backport fix: Service mesh use case - v0.10.x

paladin-devops:


website/data/docs-nav-data.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "title": "Introduction",
+    "title": "What is Waypoint?",
```
cicoyle, line 3: Looks like you need to rebase. There are a few changes in the `docs-nav-data.json` file that are being reverted. I dealt with this with my backport PR. I believe the only changes needed in this file are the use case `title` & `path` being added.

PR #4399, opened 2023-01-13
CLI: DRY up how the CLI shows deployment and release URLs

### LINKED ISSUE #4209: CLI: DRY up how the CLI shows deployment and release URLs

Currently, as a waypoint dev, we have ~4 CLIs that all have the same code to show deployment and release URLs:

- `waypoint up`
- `waypoint deploy`
- `waypoint release`
- `waypoint pipeline run`

https://github.com/hashicorp/waypoint/blob/ad4b66798acb83f6b7276de6c1cae2de1868bacc/internal/cli/up.go#L70-L120

It would be nice if these CLIs simply invoked a function to display these rather than duplicating what we have now.

Leave it as is.

**Additional context**

vitorfhc:
Closes #4209


internal/cli/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if parsedUrl.Scheme == string(scheme) || urlString == "" {
+		return urlString, nil
+	}
+
+	// Avoids adding a scheme if it already has a different one
+	if parsedUrl.Scheme != "" && parsedUrl.Scheme != string(scheme) {
+		err := fmt.Errorf("The URL \"%s\" already has a different scheme: \"%s\"", urlString, scheme)
+		return "", err
+	}
```
vitorfhc, line 31: Not sure if it makes sense

internal/cli/helpers.go
```diff
@@ -0,0 +1,36 @@
+package cli
```
briancain, line 1: We generally try to stay away from file names like `helpers` or `utils`. I would simply call this file for what it is: `internal/cli/url_scheme.go`

internal/cli/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+		return "", err
+	}
+
+	if parsedUrl.Scheme == string(scheme) || urlString == "" {
+		return urlString, nil
+	}
+
+	// Avoids adding a scheme if it already has a different one
+	if parsedUrl.Scheme != "" && parsedUrl.Scheme != string(scheme) {
+		err := fmt.Errorf("The URL \"%s\" already has a different scheme: \"%s\"", urlString, scheme)
```
briancain, line 29:
```suggestion
		err := fmt.Errorf("The URL %q already has a different scheme: %q", urlString, scheme)
```

`%q` is the same thing as `\"%s\"`

internal/cli/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+		return "", err
+	}
+
+	parsedUrl.Scheme = string(scheme)
+	urlWithScheme := fmt.Sprintf("%s", parsedUrl)
```
briancain, line 34: This looks to be a linter failure:

```
internal/cli/helpers.go:34:19: S1025: should use String() instead of fmt.Sprintf (gosimple)
	urlWithScheme := fmt.Sprintf("%s", parsedUrl)
```

vitorfhc:
@briancain I deleted the `closes` on the description since I won't have enough time to DRY up everything stated in issue #4209. I changed the PR to Ready to Review.

internal/cli/url_scheme.go
```diff
@@ -0,0 +1,36 @@
...
+		return "", err
+	}
+
+	if parsedUrl.Scheme == string(scheme) || urlString == "" {
+		return urlString, nil
+	}
+
+	// Avoids adding a scheme if it already has a different one
+	if parsedUrl.Scheme != "" && parsedUrl.Scheme != string(scheme) {
+		err := fmt.Errorf("The URL \"%s\" already has a different scheme: \"%s\"", urlString, scheme)
```
cicoyle, line 29:
```suggestion
		err := fmt.Errorf("The URL %q already has a different scheme: %q", urlString, scheme)
```

internal/cli/url_scheme.go
```diff
@@ -0,0 +1,36 @@
...
+
+func addUrlScheme(urlString string, scheme UrlScheme) (string, error) {
+	parsedUrl, err := url.Parse(urlString)
+	// The url.Parse function may return an error when there's no scheme due to parser ambiguities,
+	// so we need to verify if the error is because the url lacks a scheme
+	if err != nil && parsedUrl.Scheme != "" {
+		return "", err
+	}
+
+	if parsedUrl.Scheme == string(scheme) || urlString == "" {
```
cicoyle, line 23: We might not need the or case because we won't want to return the empty urlString

cicoyle (commented):
This is a great idea! Thanks for working on it. I believe there may be a missing file change for the `waypoint release` case.

internal/cli/url_scheme.go
cicoyle: It might be worth explicitly checking this separately in its own case. And if we do want this check to be an ok case, we might want to move it up before the url parsing. In any case, it should match the current behavior.

internal/cli/url_scheme.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/url"
+)
+
+type UrlScheme string
+
+const (
+	httpsScheme UrlScheme = "https"
+	httpScheme  UrlScheme = "http"
+)
+
```
cicoyle, line 14: Mind adding some documentation on how to use this function and the expected format is for the url string? I think that additional insight would be valuable to have - even a snip it of such examples as a comment.

internal/cli/url_scheme.go
```diff
@@ -0,0 +1,36 @@
...
+func addUrlScheme(urlString string, scheme UrlScheme) (string, error) {
+	parsedUrl, err := url.Parse(urlString)
+	// The url.Parse function may return an error when there's no scheme due to parser ambiguities,
+	// so we need to verify if the error is because the url lacks a scheme
+	if err != nil && parsedUrl.Scheme != "" {
```
cicoyle, line 19: Mind pulling out the err check and delete the rest?
```suggestion
	if err != nil {
```

xiaolin-ninja:
Hi @vitorfhc, are you still working on this PR?

PR #4400, opened 2023-01-13
[website] docs: updates inline alert authoring info

kendallstrautman:
This updates the generated docs for the new [mdx inline alert component](https://github.com/hashicorp/dev-portal/pull/1517). See [docs on dev portal here.](https://developer.hashicorp.com/swingset/components/mdxinlinealert)

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -98,7 +98,7 @@ There is currently a small bug with new page creation - if you create a new page
 
 There are several custom markdown plugins that are available by default that enhance [standard markdown](https://commonmark.org/) to fit our use cases. This set of plugins introduces a couple instances of custom syntax, and a couple specific pitfalls that are not present by default with markdown, detailed below:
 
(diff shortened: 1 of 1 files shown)


zchsh (approved):

cicoyle (approved):
Nice cleanup on the typos! The newer inline alerts look simple and easy to use.

website/README.md
```diff
@@ -98,19 +97,19 @@ There is currently a small bug with new page creation - if you create a new page
...
-  Headlines are fairly self-explanatory, but here's an example of how list items that begin with inline code look.
+  Headlines are fairly self-explanatory, but here's an example of how to list items that begin with inline code look.
 
   ```markdown
   - this is a normal list item
   - `this` is a list item that begins with inline code
   ```
 
-  It's worth noting that _only the inline code at the beginning of the list item_ will cause problems if changed. So if you changed the above markup to...
+  Its worth noting that _only the inline code at the beginning of the list item_ will cause problems if changed. So if you changed the above markup to...
```
briancain, line 112: Oops, sorry for reviewing this post merge, but this change is the wrong usage of `Its`. It should be `It's`. `It's worth noting` -> `It is worth nothing`...

kendallstrautman: Good callout! I should fix that in the source generator or else this will just happen again 😆 I'll do that now

PR #4404, opened 2023-01-14
Makes all Helm actions atomic during server install and upgrade

### LINKED ISSUE #4403: CLI: Implement Helm flags on server install

I was trying to install a server, and this was the output:

```
$ waypoint install -platform=kubernetes -accept-tos
❌ Installing Waypoint Helm chart...
! Error installing server into kubernetes: timed out waiting for the condition
```

The server was deployed successfully even though the installation timed out. It needs to be more accurate.

If I try to rerun the command, it fails because there's a Helm release with the same name. Then you start questioning if it failed or not.

It would be nice to control Helm flags (e.g., `--atomic` or `--wait`). It would be even better to set `atomic` to `true` by default, so we avoid misleading errors. If it fails for any reason, then we uninstall it.

Maybe we could even detect if the error was a timeout and suggest the user increase it using `-helm-timeout`.

We could increase the timeout for Helm action ([code](https://github.com/hashicorp/waypoint/blob/c19bb81db05e5925902ba71bcb7b74055df02b8d/internal/serverinstall/k8s.go#L161)), but I don't think it's a good solution.

There are lots of reasons to customize the Helm behaviour. Someone may want to add a description, use the `--replace`, or skip the CRDs.

**Additional context**

It would be tiring to create each Helm flag in the `install` command. We could do it dynamically.

Any `-helm-<flag>` flag would become a `client.<flag> = flag_value`.

Or we could do it at least for some flags like `--atomic` and `--timeout` at first.

vitorfhc:
Closes #4403


internal/serverinstall/k8s.go
```diff
@@ -461,16 +465,20 @@ func (i *K8sInstaller) Upgrade(
 		i.Config.Namespace = "default"
 	}
 
+	if i.Config.Timeout == 0 {
+		i.Config.Timeout = 300
+	}
```
briancain, line 470: This shouldn't be required, we can set a default over in the flag where it's defined!

internal/serverinstall/k8s.go
```diff
@@ -1053,6 +1068,13 @@ func (i *K8sInstaller) InstallFlags(set *flag.Set) {
 }
 
 func (i *K8sInstaller) UpgradeFlags(set *flag.Set) {
+	set.IntVar(&flag.IntVar{
+		Name:   "k8s-helm-timeout",
+		Target: &i.Config.Timeout,
+		Usage: "Time to wait for the Helm operation. It must be in seconds. " +
+			"The default is 5 minutes.",
```
briancain, line 1075:
```suggestion
		Name:   "k8s-helm-timeout",
		Target: &i.Config.Timeout,
		Default: 300,
		Usage: "Time to wait in seconds for the Helm operation.",
```

internal/serverinstall/k8s.go
```diff
@@ -461,16 +465,20 @@ func (i *K8sInstaller) Upgrade(
...
 	client.DisableHooks = false
 	client.Wait = true
 	client.WaitForJobs = false
 	client.Devel = true
 	client.DependencyUpdate = false
-	client.Timeout = 300 * time.Second
+	client.Timeout = time.Duration(i.Config.Timeout) * time.Second
 	client.Namespace = i.Config.Namespace
-	client.Atomic = false
+	client.Atomic = true
```
briancain, line 481: This should probably be a flag too

internal/serverinstall/k8s.go
```diff
@@ -956,6 +964,13 @@ func (i *K8sInstaller) InstallFlags(set *flag.Set) {
 			"The required version number format is: 'vX.Y.Z'.",
 	})
 
+	set.IntVar(&flag.IntVar{
+		Name:   "k8s-helm-timeout",
+		Target: &i.Config.Timeout,
+		Usage: "Time to wait for the Helm operation. It must be in seconds. " +
+			"The default is 5 minutes.",
```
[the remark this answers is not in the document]
briancain: No need to say "default" when its set! Our flag doc generator will include it automatically.

briancain (changes requested):
Thank you for taking the time to make a pull request for Waypoint!

internal/serverinstall/k8s.go
vitorfhc: Should I keep setting it to `true` as the default value? I was discussing about it with @demophoon on #4403

internal/serverinstall/k8s.go
```diff
@@ -956,6 +956,20 @@ func (i *K8sInstaller) InstallFlags(set *flag.Set) {
...
+		Target:  &i.Config.Timeout,
+		Default: 300,
+		Usage:   "Time to wait in seconds for the Helm operation.",
+	})
+
+	set.BoolVar(&flag.BoolVar{
+		Name:    "k8s-helm-atomic",
+		Target:  &i.Config.Atomic,
+		Default: false,
+		Usage:   "Make the Helm operation atomic.",
```
cicoyle, line 970: Mind expanding this description?

cicoyle: you can snag the output from helm, like:
```
if set, upgrade process rolls back changes made in case of failed upgrade. The --wait flag will be set automatically if --atomic is used
```

internal/serverinstall/k8s.go
briancain: I recommend keeping it false by default, because that's the default value in Helm! Might as well match their behavior. At least with the flag it's configurable.

cicoyle (commented):
I noticed you mentioned hoping to avoid `misleading errors` in your GH issue. Do you have an example of how this PR can resolve some of the issues you were facing?

xiaolin-ninja:
Hi @vitorfhc, are you still working on this PR?
Please `make gen/website-mdx` as well to update CLI documentation as well. :)

Issue #4233, opened 2022-11-29
Add `cors` option to `lambda-function-url` releaser


jgwhite:
**Is your feature request related to a problem? Please describe.**
I want to deploy to AWS Lambda using the lambda-function-url releaser and then then call that function from a different origin using XHR. To do this, I need to enable CORS on the Lambda function URL.

**Describe the solution you'd like**
I’d like a `cors` option for the `lambda-function-url` releaser.

```hcl
release {
  use "lambda-function-url" {
    cors = {
      allow_credentials = false
      allow_headers     = ""
      allow_methods     = ""
      allow_origin      = "*"
      expose_headers    = ""
      max_age           = ""
    }
  }
}
```

**Describe alternatives you've considered**
It might be nice to have some simple presets to use here. `cors_preset = "basic"` or something like that.

jgwhite:
Closed by #4418

PR #4418, opened 2023-01-18
Add CORS options to lambda-function-url releaser

jgwhite:
## Why the change?

Addresses #4233

## What’s the plan?

- [x] Get it working
- [x] Changelog
- [x] Make sure docs are decent

## What does it look like?

https://user-images.githubusercontent.com/34030/214331215-f7ab0256-44ec-47a6-bef6-a2fd4e7ed32e.mov

CleanShot 2023-01-24 at 19 33 11@2x

## How do I test it?

Try adding CORS config to one of the [Lambda examples](https://github.com/hashicorp/waypoint-examples/tree/main/aws/lambda).

Visit the [website preview](https://waypoint-o4b1g2i7b-hashicorp.vercel.app/waypoint/plugins/aws-lambda#lambda-function-url-releasemanager) and verify the docs make sense.

### DIFF
--- a/.changelog/4418.txt
+++ b/.changelog/4418.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/aws: Add CORS configuration to lambda-function-url releaser
+```
--- a/builtin/aws/lambda/function_url/releaser.go
+++ b/builtin/aws/lambda/function_url/releaser.go
@@ -15,14 +15,15 @@ import (
 	"github.com/aws/aws-sdk-go/service/lambda"
 	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/hashicorp/go-hclog"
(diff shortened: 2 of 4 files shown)


jgwhite:
This doesn’t seem to update the function URL CORS config at the moment. More investigation needed.

jgwhite:
But further, but curious behavior when no `cors` block is specified:

```
✓ Running release v28
✓ No permissions need to be updated
! error reading from server: EOF
```

briancain:
> ! error reading from server: EOF

I think that usually means the Waypoint server panicked. If you haven't checked already, you'll find what happened in the server logs I think!

jgwhite:
This is now working nicely, but the docs need some love.

demophoon (approved):

Issue #4195, opened 2022-11-14
waypoint pipeline list panics CLI


pandom:
**Describe the bug**
Stacktrace when issuing waypoint pipeline list on empty pipelines.

**Steps to Reproduce**
Selecting a HCP or Self-Hosted server context run `waypoint pipeline list`

```
➜  nodejs git:(main) ✗ waypoint pipeline list

» Waypoint Pipelines for nomad-nodejs
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/hashicorp/waypoint/internal/cli.(*PipelineListCommand).Run(0x140009594e0, {0x14000c428f0?, 0x16?, 0xffffffffffffffff?})
	/private/tmp/waypoint-20221103-1557-tl8hfu/waypoint-0.10.3/internal/cli/pipeline_list.go:118 +0xc94
github.com/mitchellh/cli.(*CLI).Run(0x1400025e000)
	/Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/mitchellh/cli@v1.1.2/cli.go:262 +0x4a8
github.com/hashicorp/waypoint/internal/cli.Main({0x14000114090?, 0x10216f46c?, 0x104e2ab60?})
	/private/tmp/waypoint-20221103-1557-tl8hfu/waypoint-0.10.3/internal/cli/main.go:127 +0x510
main.main()
	/private/tmp/waypoint-20221103-1557-tl8hfu/waypoint-0.10.3/cmd/waypoint/main.go:14 +0xa8
```

**Expected behavior**
Return an empty list of pipelines

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: v0.10.3
* Waypoint Server Platform and Version: nomad
* Waypoint Plugin: pack

**Additional context**
As discussed on Slack.

briancain:
Hey @pandom - thanks for reporting this issue!

I think there's a couple of fixes for this: We should fix the panic in the CLI to make sure a PipelineRun has a list of jobs before accessing it in the job slice.

Then we should also make sure that the HCP Waypoint server doesn't return any pipelines or pipeline runs if there aren't actually any in the database. It might be returning an empty message without any fields set, which is how I think we reached this part of the CLI.

cicoyle:
This also happens on `waypoint pipeline inspect -name=<name>`

briancain:
[...]

PR #4424, opened 2023-01-19
CLI: Fix panics in `waypoint pipeline list` and `waypoint pipeline inspect`

briancain:
This commit fixes up the CLI to guard against empty PipelineRun Job slices. It also DRYs up some duplicated code in `pipeline inspect`.

Fixes https://github.com/hashicorp/waypoint/issues/4195

### DIFF
--- a/internal/cli/pipeline_inspect.go
+++ b/internal/cli/pipeline_inspect.go
@@ -142,47 +142,76 @@ func (c *PipelineInspectCommand) Run(args []string) int {
 		},
 	}
 
-	s := uint64(c.flagRunSequence)
-	if s > 0 {
-		run, err := c.project.Client().GetPipelineRun(c.Ctx, &pb.GetPipelineRunRequest{
+	var (
+		sha          string
+		msg          string
+		startJobTime string
+		endJobTime   string
+		pipelineRun  *pb.PipelineRun
+	)
+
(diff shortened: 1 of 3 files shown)


internal/cli/pipeline_inspect.go
```diff
@@ -142,47 +142,76 @@ func (c *PipelineInspectCommand) Run(args []string) int {
...
+			Sequence: seq,
 		})
 		if err != nil {
 			c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
 		}
+
+		pipelineRun = runResp.PipelineRun
+	} else {
+		pipelineRun = runs.PipelineRuns[len(runs.PipelineRuns)-1]
```
catsby, line 170: Do we need to guard on `len(runs.PipelineRuns)` being `0` here?

catsby (approved):
👍 looks good, just a question about guarding a possible zero length slice

internal/cli/pipeline_inspect.go
briancain: Good idea! I added a more generic block on this further up in 0495330cb26b59faefb92394882e1f4154a6f27c since most of this code won't make any sense if a pipeline has 0 runs!

PR #4435, opened 2023-01-23
cli: Include full help flags for `runner inspect` command

briancain:
The image below shows the before/after with this change! The flags were missing from the help output.

Screen Shot 2023-01-23 at 12 05 18 PM


paladin-devops (approved):
🆘 🎏

PR #4444, opened 2023-01-25
New make command to generate plugin protos with docker

izaaklauer:
Introduces `make docker/gen/plugins`, which uses the same toolchain as `make docker/gen/server`.

This also runs that command to regenerate protos, getting them all to a consistent state.


catsby:
Are these changes compatible with users running `make gen/plugins` directly? Or do we need to move people away from being able to run that directly and only support the `docker/gen/*` versions?

izaaklauer:
I think this actually fixes `make gen/plugins` too. The `go_opt=plugins=grpc` syntax that this PR changes seems to have been [discontinued in v1.2[](](https://stackoverflow.com/questions/61044883/switch-from-go-out-plugins-to-go-grpc-out-path-problem#:~:text=v1.2%20of%20protoc-gen-go%20doesn%27t%20support%20plugins%20anymore)url), and that's what we're [pinned to in the go.mod](https://github.com/hashicorp/waypoint/blob/38d713bcbcda7e58d21413bab376716d6c6a3ebd/go.mod#L369) that both `make gen/plugins` and `make docker/gen/plugins` use to install protoc-gen-go-grpc.

I'd still vote that we switch away from `make gen/plugins`. In my experience, it's really tough to get the local toolchain set up to use.

builtin/aws/ecs/main.go
```diff
@@ -4,7 +4,7 @@ import (
 	sdk "github.com/hashicorp/waypoint-plugin-sdk"
 )
 
-//go:generate protoc -I ../../../.. -I ../../../thirdparty/proto --go_opt=plugins=grpc --go_out=../../../.. waypoint/builtin/aws/ecs/plugin.proto
+//go:generate protoc -I ../../../.. -I ../../../thirdparty/proto --go_out=../../../.. --go-grpc_out=../../../.. waypoint/builtin/aws/ecs/plugin.proto
```
briancain, line 7: If I remember right, from this PR https://github.com/hashicorp/waypoint/pull/3366, I think we need an additional flag for allowing relative paths to work properly in certain cases? Maybe @krantzinator has more insight into this one.

briancain (approved):
Haven't tried it, but I am in favor of moving to the docker version. Ever since we did this for the server, I haven't ever gone back. Agreed trying to get this toolchain installed is a huge pain, so this looks like a win to me. I'm curious if @krantzinator has any insight though since I know she tried to fix this recently last year too, but otherwise looks good to me!

builtin/aws/ecs/main.go
izaaklauer: I looked into that `source-relative` flag - it looks like it controls where the generated code goes (docs [here](https://developers.google.com/protocol-buffers/docs/reference/go-generated#:~:text=If%20the%20paths%3Dsource_relative%20flag%20is%20specified%2C%20the%20output%20file%20is%20placed%20in%20the%20same%20relative%20directory%20as%20the%20input%20file.%20For%20example%2C%20an%20input%20file%20protos/buzz.proto%20results%20in%20an%20output%20file%20at%20protos/buzz.pb.go.)).

Our generate code is going to the right place now, so I thin it's file! I also tried adding `--go-grpc_opt=paths=source_relative` to a random plugin (k8s apply) and re-ran generation, and there was no effect.

PR #4446, opened 2023-01-25
add support for scaleway container as a service

remyleone:


paladin-devops (commented):
Hey @remyleone, I am excited for your plugin! 

We are working on a new ‘registry-lite’ UI that will make plugins discoverable. That will be released in a couple of weeks! We will be encouraging plugin owners to create a repo for their plugin, with an added metadata file to it to sync it with that UI. We’ll also be working on moving several of the current built-in plugins into their own repos over the coming months, to reflect this new usage pattern. 

Could you create a new repo for this plugin as well? For now, you can list it in the README [here](https://github.com/hashicorp/waypoint-plugin-examples#community-built-plugins). In the meantime, I'll also close this PR since your plugin will be in another repo.

PR #4447, opened 2023-01-25
Add `UI_ListProjects`

jgwhite:
## Why the change?

For the “destroy project” UI we want to show a “destroying” status on appropriate items in the project list page. The existing `ListProjects` method only gives us a list of `Ref.Project` which doesn’t include the status, so we’re proposing this new method which includes the stuff we need.

## What’s the plan?

- [x] Write basic handler test
- [x] Add `UI_ListProjects` to `server.proto` and `gateway.yml`
- [x] `make gen/server`
- [x] Add `ProjectListFull` (or something like that) to the state layer
- [x] Implement `boltdbstate.ProjectCount`
- [x] Implement `UI_ListProjects` handler
- [x] Add API validations
- [x] Decide whether to attempt or defer pagination for boltdb (outcome: defer)

## What does it look like?

CleanShot 2023-02-03 at 14 23 40@2x

## How do I test it?

1. Check out the branch
2. `make bin`
3. Boot the server and connect to it
4. Init some projects
5. From the waypoint directory, run `./contrib/waypoint-grpc/waypoint-grpc.sh UI_ListProjects`
6. Verify all non-empty project-fields are present
7. Ignore pagination, we’ll come back to this later


pkg/server/singleprocess/service_ui_project.go
```diff
@@ -10,6 +10,29 @@ import (
 	serverptypes "github.com/hashicorp/waypoint/pkg/server/ptypes"
 )
 
+func (s *Service) UI_ListProjects(
+	ctx context.Context,
+	req *pb.UI_ListProjectsRequest,
+) (*pb.UI_ListProjectsResponse, error) {
+	// TODO(jgwhite): Handle pagination
```
briancain, line 17: We should also add API validations here to ensure the `req` as the expected parameters set and aren't `nil` which could cause panics in the server.

internal/server/boltdbstate/project.go
```diff
@@ -491,6 +515,44 @@ func (s *State) projectList(
...
+			}
+			result = append(result, project)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return result, &pb.PaginationResponse{}, nil // TODO(jgwhite): pagination for ProjectListFull
```
jgwhite, line 553: This is TODO for the future, not for this PR.

jgwhite:
I’m gonna rejig this a little so we return a list of `UI.ProjectBundle` rather than just `Project`.

internal/server/boltdbstate/project.go
```diff
@@ -257,9 +257,25 @@ func (s *State) ProjectDelete(ctx context.Context, ref *pb.Ref_Project) error {
...
+	defer memTxn.Abort()
+
+	iter, err := memTxn.Get(projectIndexTableName, projectIndexIdIndexName+"_prefix", "")
+	if err != nil {
+		return 0, err
+	}
+
 	var count uint64
-	var err error
-	return count, err // TODO: Implement count for boltdb
```
andrew-hashicorp, line 262: 🤩 - Glad to see this implemented!

jgwhite:
> I’m gonna rejig this a little so we return a list of `UI.ProjectBundle` (or perhaps `UI.GetProjectResponse`) rather than just `Project`.

This is done now. Will give this PR some more testing tomorrow then go ahead with it.

briancain (approved):
Looks great! Just a small comment from me ✨

PR #4465, opened 2023-01-30
maint: Send useful error message on bad token magic string

briancain:
This commit sends some actual useful error messaging to the user in the case where they provided a bad auth token string that did not match the expected token magic during token decoding.


izaaklauer (approved):
image

paladin-devops (approved):
Much more helpful!

PR #4469, opened 2023-01-31
test: Ensure a project and pipeline run is upserted for test

briancain:
SQLState is much stricter about what entries exist in the database, given the nature that it needs to query the database for IDs rather than being ok with references like boltdb. This commit updates these tests to upsert a project and a run.


cicoyle (approved):
Nice!

PR #4473, opened 2023-02-01
Enable custom health checks for ECS plugin

### LINKED ISSUE #3600: Improved health check interface for ECS plugin

We have no control over the health checks of target groups within the ECS plugin

We should be able to:

- Enable/disable health checks
- Set a path we want to use for a health check
- Set the response code the health check expects back from the path

- Not using the ECS plugin, and using something like a Kubernetes option
- Not using Waypoint

I don't think this is an issue of use-case, is an issue of operational maturity. Health checks should not be this basic, hitting only a route URL of a deployment.

**Additional context**
None

### LINKED ISSUE #4455: plugin/ecs: Custom Health Checks
If my app immediately bind to its port, but takes a while to fully become healthy. It has a separate `/healthz` healthcheck endpoint, but I can't configure waypoint to use it!

Currently, waypoint is leaving the target group health check details as the defaults ([here](https://github.com/hashicorp/waypoint/blob/38d713bcbcda7e58d21413bab376716d6c6a3ebd/builtin/aws/ecs/platform.go#L1311-L1318)), which means we get http health checks on `/` on the target port.

AWS has lots of configurability for that target group:

https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html#:~:text=service%20is%20unavailable.-,Health%20check%20settings,-You%20configure%20active

We should allow users to configure those via plugin settings, just like we do for k8s with the probe settings.

**Additional context**

This is currently a moot point, as our plugin releases deployments before they become ready in the target group: https://github.com/hashicorp/waypoint/issues/2163. We should fix that one first.

paladin-devops:
Closes #4455. Co-authored with @cicoyle.

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1341,12 +1341,12 @@ func (p *Platform) resourceTargetGroupCreate(
 		createTargetGroupInput.UnhealthyThresholdCount = aws.Int64(p.config.HealthCheck.UnhealthyThresholdCount)
 	}
 
-	if p.config.HealthCheck.GrpcCode != "" {
-		createTargetGroupInput.Matcher.GrpcCode = aws.String(p.config.HealthCheck.GrpcCode)
+	if p.config.HealthCheck.GRPCCode != "" {
+		createTargetGroupInput.Matcher.GrpcCode = aws.String(p.config.HealthCheck.GRPCCode)
 	}
 
-	if p.config.HealthCheck.HttpCode != "" {
-		createTargetGroupInput.Matcher.HttpCode = aws.String(p.config.HealthCheck.HttpCode)
+	if p.config.HealthCheck.HTTPCode != "" {
+		createTargetGroupInput.Matcher.HttpCode = aws.String(p.config.HealthCheck.HTTPCode)
 	}
 
 	ctg, err := elbsrv.CreateTargetGroupWithContext(ctx, createTargetGroupInput)
@@ -2908,13 +2908,13 @@ type AppHealthCheck struct {
 	// The range is 2–10. The default is 2.
 	UnhealthyThresholdCount int64 `hcl:"unhealthy_threshold_count,optional"`
 
-	// GrpcCode is the gRPC codes to use when checking for a successful response from a target.
+	// GRPCCode is the gRPC codes to use when checking for a successful response from a target.
 	// The default value is 12.
-	GrpcCode string `hcl:"grpc_code,optional"`
+	GRPCCode string `hcl:"grpc_code,optional"`
 
-	// HttpCode is the HTTP codes to use when checking for a successful response from a target.
+	// HTTPCode is the HTTP codes to use when checking for a successful response from a target.
 	// The range is 200 to 599. The default is 200-399.
-	HttpCode string `hcl:"http_code,optional"`
+	HTTPCode string `hcl:"http_code,optional"`
 }
 
 func (p *Platform) Documentation() (*docs.Documentation, error) {
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1341,13 +1341,12 @@ func (p *Platform) resourceTargetGroupCreate(
 		createTargetGroupInput.UnhealthyThresholdCount = aws.Int64(p.config.HealthCheck.UnhealthyThresholdCount)
 	}
 
-	if p.config.HealthCheck.Matcher != nil {
-		if *p.config.HealthCheck.Matcher.GrpcCode != "" || *p.config.HealthCheck.Matcher.HttpCode != "" {
-			createTargetGroupInput.Matcher = &elbv2.Matcher{
-				HttpCode: p.config.HealthCheck.Matcher.HttpCode,
-				GrpcCode: p.config.HealthCheck.Matcher.GrpcCode,
-			}
-		}
+	if p.config.HealthCheck.GrpcCode != "" {
+		createTargetGroupInput.Matcher.GrpcCode = aws.String(p.config.HealthCheck.GrpcCode)
+	}
+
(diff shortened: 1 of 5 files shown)


builtin/aws/ecs/platform.go
```diff
@@ -128,6 +128,13 @@ func (p *Platform) ConfigSet(config interface{}) error {
 		}
 	}
 
+	if c.HealthCheck != nil {
+		if c.HealthCheck.Timeout >= c.HealthCheck.Interval {
+			return status.Errorf(codes.FailedPrecondition, "the health check "+
+				"timeout must be less than the interval")
```
briancain, line 134: We should include the interval and timeout in the error message!

builtin/aws/ecs/platform.go
```diff
@@ -2776,6 +2818,34 @@ type Config struct {
...
+	// The range is 2–10. The default is 5.
+	HealthyThresholdCount int64 `hcl:"healthy_threshold_count,optional"`
+
+	// UnhealthyThresholdCount is the number of consecutive failed health checks required before considering a target unhealthy.
+	// The range is 2–10. The default is 2.
+	UnhealthyThresholdCount int64 `hcl:"unhealthy_threshold_count,optional"`
+
+	// Matcher is the HTTP codes to use when checking for a successful response from a target.
+	// The range is 200 to 599. The default is 200-399.
+	Matcher *elbv2.Matcher `hcl:"matcher,optional"`
```
briancain, line 2848: We should document this config struct in the `Documentation()` func below!

builtin/aws/ecs/platform.go
paladin-devops: Added the doc fields to `Documentation`, but `make gen/website-mdx` didn't generate the updated .mdx files 🤔 still looking into it.

briancain: If I remember right, if it's a new markdown file, you have to manually `@include` it in like: https://github.com/hashicorp/waypoint/blob/main/website/content/plugins/aws-ec2.mdx. Not sure if that's the issue

paladin-devops: The ECS platform plugin docs indeed are included. Even more weirdly, this is passing the CircleCI website-mdx check... maybe it's because all of the fields in the new struct are optional?

builtin/aws/ecs/platform.go
```diff
@@ -3013,6 +3085,38 @@ deploy {
 		}),
 	)
 
+	doc.SetField("health_check",
+		"Health check settings for the app.",
+		docs.SubFields(func(doc *docs.SubFieldDoc) {
+			doc.SetField("protocol",
+				"The protocol for the health check to use.")
```
briancain, line 3092: Not sure if our docs have a special field for this, but it would be nice to say what the valid protocols are!

briancain (approved):

builtin/aws/ecs/platform.go
```diff
@@ -1301,14 +1308,48 @@ func (p *Platform) resourceTargetGroupCreate(
...
+	if p.config.HealthCheck.Timeout != 0 {
+		createTargetGroupInput.HealthCheckTimeoutSeconds = aws.Int64(p.config.HealthCheck.Timeout)
+	}
+
+	if p.config.HealthCheck.Interval != 0 {
+		createTargetGroupInput.HealthCheckIntervalSeconds = aws.Int64(p.config.HealthCheck.Interval)
+	}
+
+	if p.config.HealthCheck.HealthyThresholdCount != 0 {
+		createTargetGroupInput.HealthyThresholdCount = aws.Int64(p.config.HealthCheck.HealthyThresholdCount)
```
catsby, line 1337: Where is this default set? If it's the default that AWS gives if we do not supply a value, then we'll need to set it ourselves here `HealthyThresholdCount == 0`. Otherwise if a user decides to remove their custom value, they might assume that it will revert to `5` because that's what our docs say, but it wont, because it will be `0` in our config and then this block wont execute.

builtin/aws/ecs/platform.go
```diff
@@ -2776,6 +2816,38 @@ type Config struct {
...
+	// The range is 2–10. The default is 5.
+	HealthyThresholdCount int64 `hcl:"healthy_threshold_count,optional"`
+
+	// UnhealthyThresholdCount is the number of consecutive failed health checks required before considering a target unhealthy.
+	// The range is 2–10. The default is 2.
+	UnhealthyThresholdCount int64 `hcl:"unhealthy_threshold_count,optional"`
+
+	// GrpcCode is the gRPC codes to use when checking for a successful response from a target.
+	// The default value is 12.
+	GrpcCode string `hcl:"grpc_code,optional"`
```
catsby, line 2846: nit I think `GRPCCode` is more idiomatic, same for `HTTPCode`

catsby (commented):
minor comments

builtin/aws/ecs/platform.go
```diff
@@ -1301,14 +1308,48 @@ func (p *Platform) resourceTargetGroupCreate(
 
 	state.Port = p.config.ServicePort
 
-	ctg, err := elbsrv.CreateTargetGroupWithContext(ctx, &elbv2.CreateTargetGroupInput{
+	if p.config.HealthCheck.Protocol == "" {
```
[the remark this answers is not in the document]
paladin-devops: We do!

builtin/aws/ecs/platform.go
paladin-devops: This is indeed the default from AWS, and what you're saying definitely makes sense. I updated this to explicitly set the value to 5 if the user doesn't supply a value (the `HealthyThresholdCount` ends up being 0).

catsby (approved):
👍 on the changes after my feedback

Issue #4461, opened 2023-01-28
plugin/ecs: cannot destroy deployments


izaaklauer:
**Describe the bug**
I'm attempting to manually delete the second-to-last waypoint deployment, and I get this error:

```
$ waypoint deployment list -workspace=prod

» acmeapp1
     | ID | PLATFORM |  DETAILS   |    STARTED    |   COMPLETED   |      HEALTH        
-----+----+----------+------------+---------------+---------------+--------------------
  🚀 | 15 | aws-ecs  | artifact:2 | 6 minutes ago | 6 minutes ago | ● - 6 minutes ago  
  ✔  | 14 | aws-ecs  | artifact:2 | 8 minutes ago | 8 minutes ago | ● - 8 minutes ago  

$ waypoint deployment destroy 14

» Destroying deployments for acmeapp1

» 1 deployments will be destroyed.
  Destroying deployment: 01GQV0DYRFQWK6XZN311P4QYMX

» Operation is queued waiting for job "01GQV0X723FB3S7TPQW5W63Y42". Waiting for runner assignment...
  If you interrupt this command, the job will still run in the background.
  Performing operation on "aws-ecs" with runner profile "ecs-01GQJJP4GH4PXEDHNZPY8ZA6W7"

» Cloning data from Git
         URL: https://api.hashicorp.cloud/waypoint/2022-04-21/github/clone/hashicorp/waypoint-learn-terraform-aws-ecs
         Ref: main
        Auth: username/password
  Git Commit: 511720182f674ce9fc06eea5d528c64ffaee682c
   Timestamp: 2023-01-28 01:34:45 +0000 UTC
     Message: better cpu value
  
  mem is probably wrong though!
  
✓ Running deployment destroy v14
❌ Destroying ecs deployment...
✓ Deleted service acmeapp1-01GQV0DYYGQ58SH3MSHEW0H
❌ Deleting target group acmeapp1-01GQV0DYYGQ58SH3MSHEW0H
! Error destroying deployment 14: rpc error: code = Internal desc = failed to destroy all resources for deployment: rpc error: code = Internal desc = failed to
delete target group acmeapp1-01GQV0DYYGQ58SH3MSHEW0H (ARN:
"arn:aws:elasticloadbalancing:us-east-1:863953081260:targetgroup/acmeapp1-01GQV0DYYGQ58SH3MSHEW0H/5aff44d18c6b8299"): ResourceInUse: Target group
[...]

PR #4497, opened 2023-02-07
Update target group destroyer to wait for listener

paladin-devops:
This PR updates the ECS platform destroyer to wait for there to be zero AWS ALB listeners targeting the target group to attempt to delete the target group, for a maximum of 5 minutes.

This does not fully address #4461, because my destroy attempts are still failing on deleting the security group, but it's a start! And deleting that will be fixed in a subsequent PR.

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1339,7 +1339,7 @@ func (p *Platform) resourceTargetGroupDestroy(
 		return nil
 	}
 
-	s := sg.Add("Getting target group details", state.Name)
+	s := sg.Add("Getting details for target group %s", state.Name)
 	defer s.Abort()
 
 	elbsrv := elbv2.New(sess)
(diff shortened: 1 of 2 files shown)


builtin/aws/ecs/platform.go
```diff
@@ -1339,21 +1339,89 @@ func (p *Platform) resourceTargetGroupDestroy(
 		return nil
 	}
 
-	s := sg.Add("Deleting target group %s", state.Name)
+	s := sg.Add("Getting target group details", state.Name)
```
briancain, line 1342: Might be nice to preserve the detail about the target group name here?

builtin/aws/ecs/platform.go
```diff
@@ -1339,21 +1339,89 @@ func (p *Platform) resourceTargetGroupDestroy(
...
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to describe target group %s (ARN: %q): %s", state.Name, state.Arn, err)
+	}
+
+	s.Update("Retrieved target group details")
+	s.Done()
+
+	// There should be only one target group we "loop through" here
+	for _, group := range groups.TargetGroups {
+		s = sg.Add("Checking if target group " + *group.TargetGroupName + " has an active ALB listener")
```
briancain, line 1359: I think having `*group.TargetGroupName` as a `%q` rather than the string addition makes more sense here

builtin/aws/ecs/platform.go
```diff
@@ -1339,21 +1339,89 @@ func (p *Platform) resourceTargetGroupDestroy(
...
+	})
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to describe target group %s (ARN: %q): %s", state.Name, state.Arn, err)
+	}
+
+	s.Update("Retrieved target group details")
+	s.Done()
+
+	// There should be only one target group we "loop through" here
+	for _, group := range groups.TargetGroups {
```
briancain, line 1358: This is quite a big loop! I think it's worth considering if there's any way we could reduce this down from 5 nested for loops 😅

briancain (approved):

PR #4499, opened 2023-02-08
`PlatformReleaser` plugin interface docs

paladin-devops:
This PR adds docs for the [`PlatformReleaser` interface of the plugin SDK](https://github.com/hashicorp/waypoint-plugin-sdk/blob/main/component/component.go#L119-L127)! 😄

### DIFF
--- a/website/content/docs/extending-waypoint/plugin-interfaces/platform-releaser.mdx
+++ b/website/content/docs/extending-waypoint/plugin-interfaces/platform-releaser.mdx
@@ -10,7 +10,7 @@ description: |-
 https://pkg.go.dev/github.com/hashicorp/waypoint-plugin-sdk/component#PlatformReleaser
 
 The `PlatformReleaser` interface is an optional interface that a Platform component
-can implement to provide default [`Release`](/docs/extending-waypoint/plugin-interfaces/release-manager) functionality. This only takes effect
+can implement to provide default [`Release`](/waypoint/docs/extending-waypoint/plugin-interfaces/release-manager) functionality. This only takes effect
 if no release is configured, therefore users cannot set custom configurations
 for default releasers. Releases will happen in a way that is wholly determined
 by the output of the deployment.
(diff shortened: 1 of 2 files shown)


evanphx (approved):

website/content/docs/extending-waypoint/plugin-interfaces/platform-releaser.mdx
```diff
@@ -0,0 +1,58 @@
...
+page_title: 'PlatformReleaser'
+description: |-
+  How to implement the PlatformReleaser component for a Waypoint plugin
+---
+
+# PlatformReleaser
+
+https://pkg.go.dev/github.com/hashicorp/waypoint-plugin-sdk/component#PlatformReleaser
+
+The PlatformReleaser interface is an optional interface that a Platform component
```
catsby, line 12:
```suggestion
The `PlatformReleaser` interface is an optional interface that a Platform component
```

I think interface names like this should have the special font applied. I know we aren't very consistent here, but we should try to be IMO

website/content/docs/extending-waypoint/plugin-interfaces/platform-releaser.mdx
```diff
@@ -0,0 +1,58 @@
...
+can implement to provide default Release functionality. This only takes effect
```
catsby, line 13:
```suggestion
can implement to provide default [`Release`](/docs/extending-waypoint/plugin-interfaces/release-manager) functionality. This only takes effect
```

I think this makes sense to add font and link, but this is just a suggestion
could also link this to

website/content/docs/extending-waypoint/plugin-interfaces/platform-releaser.mdx
```diff
@@ -0,0 +1,58 @@
...
+if no release is configured. Therefore, users cannot set custom configurations
```
catsby, line 14:
```suggestion
if no release is configured, therefore users cannot set custom configurations
```

IMO these read like 3 half-sentances in a row, combining these two keeps the chain of thought going

catsby (approved):
I added some minor suggestions but looks good without them, so do as you wish

PR #4503, opened 2023-02-08
Resource Manager Framework docs updates

paladin-devops:
Updated example to show `DestroyedResourcesResp` in use, and documented that the resource manager takes care of dependencies and an example of how to do that.


briancain (approved):
Noice! ✨

PR #4509, opened 2023-02-10
Wayp 751 UI list pipelines implementation

sdav9375:
Adds UI_ListPipelines implementation, with tests. The bundle creation in `pkg/server/singleprocess/service_ui_pipeline.go` may be moved to a state process at a later date.


jgwhite:
Following this discussion here https://github.com/hashicorp/waypoint/pull/4549#discussion_r1114843427 we’ve decided to introduce a `PipelineRunBundle` message right away, rather than trying to add fields to `PipelineRun`. This has implications for the `last_run` field introduced here.

Proposed change for this RFC:

```diff
  message UI {

    ⋮

    message PipelineBundle {
      Pipeline pipeline = 1;
-     PipelineRun last_run = 2;
+     PipelineRunBundle last_run = 2;
      uint64 total_runs = 3;
    }

+   message PipelineRunBundle {
+     PipelineRun pipeline_run = 1;
+     google.protobuf.Timestamp queue_time = 2;
+     Ref.Application application = 3;
+     Job.DataSource.Ref data_source_ref = 4;
+   }
```

pkg/serverhandler/handlertest/test_service_ui_pipeline.go
```diff
@@ -0,0 +1,95 @@
...
+				Project: appRef.Project,
+			},
+			Pagination: &pb.PaginationRequest{
+				PageSize: 10,
+			},
+		})
+		require.NoError(err)
+		require.NotNil(resp.Pipelines[0].LastRun)
+		require.NotNil(resp.Pipelines[1].LastRun)
+		require.EqualValues(1, resp.Pipelines[0].TotalRuns)
```
briancain, line 93: Maybe another test here would be good, i.e. more than just 1! I like rule of threes when it comes to test.

briancain (approved):
This looks pretty great to me @sdav9375 ! Just 1 piece of feedback on another test, but otherwise I'd say this looks good to go ✨ Well done!

pkg/serverhandler/handlertest/test_service_ui_pipeline.go
sdav9375: Ok I've broken out a third test just for the "list"- let me know if this is what you meant.

pkg/server/singleprocess/service_ui_pipeline.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// Create uninitialized array of pipeline bundles
+	var allPipelines []*pb.UI_PipelineBundle
+
+	// Get list of all pipelines
+	pipelineListResponse, err := s.state(ctx).PipelineList(ctx, req.Project)
+	if err != nil {
+		return nil, hcerr.Externalize(
+			hclog.FromContext(ctx),
```
izaaklauer, line 30: Tiny nit, no need to update this PR: 

I'm a fan of adding `log := hclog.FromContext(ctx)` to the top of these functions, then just using `log` going forward. It's more concise overall, and makes it easier to [add things](https://github.com/hashicorp/go-hclog#create-a-new-logger-with-fixed-keyvalue-pairs) to the logger throughout the function.

pkg/server/singleprocess/service_ui_pipeline.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/hashicorp/go-hclog"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+	"github.com/hashicorp/waypoint/pkg/server/hcerr"
+	serverptypes "github.com/hashicorp/waypoint/pkg/server/ptypes"
+)
+
+func (s *Service) UI_ListPipelines(
```
izaaklauer, line 15: Might be nice to add a note somewhere saying that, while paginating is part of the request, this doesn't yet support pagination and will return everything every time.

pkg/server/singleprocess/service_ui_pipeline.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Create bundles
+	for _, pipeline := range pipelineListResponse {
+		ref := &pb.Ref_Pipeline{
+			Ref: &pb.Ref_Pipeline_Id{
+				Id: pipeline.Id,
+			},
+		}
+		// Get total runs for the pipeline
+		pipelineRunListResponse, err := s.state(ctx).PipelineRunList(ctx, ref)
```
izaaklauer, line 44: I can imagine that this list will be very long for some pipelines, and will eventually make this response too large, even if we respect the request pagination!

Might be nice to limit this to something like the 10 most recent runs (or whatever the UI needs), and document that somewhere. Not sure if that state function supports a limit - if it does, use that - but otherwise you can throw out everything but the 10 most recent runs from the list.

pkg/serverhandler/handlertest/test_service_ui_pipeline.go
```diff
@@ -0,0 +1,107 @@
...
+			JobTemplate: jobTemplate,
+		})
+
+		// Call the method
+		resp, err := client.UI_ListPipelines(ctx, &pb.UI_ListPipelinesRequest{
+			Project: &pb.Ref_Project{
+				Project: appRef.Project,
+			},
+			Pagination: &pb.PaginationRequest{
+				PageSize: 10,
```
izaaklauer, line 99: Might be nice to add a test case for this (ensure page size is 10), but comment it out with a note saying something like "todo: this should pass when we implement pagination"

izaaklauer (changes requested):
Great job @sdav9375 !

pkg/server/singleprocess/service_ui_pipeline.go
sdav9375: We are only pulling all of these to get the total count, and not actually returning all the runs themselves. This is an interim solution. We will eventually have the count as part of the state function.

briancain: One solution for getting total runs could be instead grabbing the sequence number from `PipelineRunGetLatest` instead of counting the total list. The sequence number on a run should match total runs, so the latest sequence should be the same as total runs?

jgwhite: @briancain oooh, that’s a very neat idea!

pkg/serverhandler/handlertest/test_service_ui_pipeline.go
sdav9375: I've added more runs to the test!

izaaklauer (approved):

jgwhite:
🎉

PR #4520, opened 2023-02-13
Check target health before releasing ECS deployment

### LINKED ISSUE #2187: ECS Releasing results in downtime due to lack of deployment warming
Currently, performing an ECS release can result in downtime because the new deployment being targeting is not yet running, and so shunting traffic to it results in the application experiencing downtime.

Perhaps Waypoint should check the status of a deployment before manipulating the release.

paladin-devops:
This PR updates the default releaser for AWS ECS to check that each target in the deployment target group is `healthy` before releasing the deployment (routing 100% of traffic to it). Closes #2187. 

If not all targets are healthy, the release fails with an error, informing the user that the release cannot happen until all targets are healthy.

### DIFF
--- a/.changelog/4520.txt
+++ b/.changelog/4520.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+plugin/ecs: Update ECS releaser to verify deployment health before releasing.
+```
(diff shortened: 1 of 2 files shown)


demophoon (approved):
Looks good to me :D

PR #4523, opened 2023-02-14
Fix key for config sources in BoltDB

paladin-devops:
This PR updates how config source records are stored in BoltDB. Previously the key was a uint64 value converted to a byte slice - this is acceptable for BoltDB and reading from it; however, snapshots expect that the database values are UTF-8 encoded, which uint64 is not. This PR therefore converts the uint64 to a string BEFORE converting it to a byte slice, and that is the new "id" for a record in the config_source bucket.

Prior to this PR, `waypoint server snapshot` would fail, and the server would log an error indicating that an invalid UTF-8 character was written.

### DIFF
--- a/.changelog/4523.txt
+++ b/.changelog/4523.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli/snapshot: Fix server snapshot when a config source is set.
+```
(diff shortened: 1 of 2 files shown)


xiaolin-ninja (approved):
Looks good to me.

briancain (approved):

PR #4530, opened 2023-02-16
plugin/ecs: Allow multiple port to be mapped in sidecars

demophoon:
This commit adds an additional field, `ContainerPorts`, which can be used multiple times to define multiple ports in a sidecar container within ecs

### DIFF
--- a/builtin/aws/ecs/components/platform/parameters.hcl
+++ b/builtin/aws/ecs/components/platform/parameters.hcl
@@ -245,6 +245,13 @@ parameter {
 }
 
 parameter {
+  key         = "sidecar.container_ports"
+  description = "List of container port definitions for the container"
+  type        = "list of ecs.ContainerPort"
+  required    = false
+}
+
+parameter {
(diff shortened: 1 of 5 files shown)


builtin/aws/ecs/components/platform/parameters.hcl
```diff
@@ -244,6 +244,13 @@ parameter {
   required    = false
 }
 
+parameter {
+  key         = "sidecar.container_ports"
+  description = "List of container port definitions for the container"
+  type        = "list of ecs.ContainerPort"
```
paladin-devops, line 250: Is there a way that we can inform the user what fields are needed to satisfy `ContainerPort`?

paladin-devops (approved):
This looks great! I had a non-blocking question about documenting the `ContainerPort` field, but ✅ to adding this improvement!

cicoyle (approved):
Looks great!

briancain:
What's the status of this one @demophoon ?

PR #4546, opened 2023-02-21
maint: Update nix environment

demophoon:


briancain (approved):
I don't use nix but, looks reasonable to me!

PR #4549, opened 2023-02-22
[WAYP-753] PipelineRun changes

sdav9375:


pkg/server/proto/server.proto
```diff
@@ -4926,9 +4926,18 @@ message PipelineRun {
...
   State state = 5;
 
+  // The timestamp of the run.
+  google.protobuf.Timestamp queue_time = 6;
+
+  // The application associated with this run.
+  Ref.Application application = 7;
+
+  // The last non-local ref that was used for this pipeline run.
+  Job.DataSource.Ref data_source_ref = 8;
```
briancain, line 4939: Hrm, all of these fields are already on the job. Is the idea to repeat this information to stop making another request to retrieve job info? This would be a lot of extra data to store in the database since it's already being stored via the job.

briancain: I think my preference, if the intention is to reduce API calls retrieving job info, is to include this information on the `GetPipelineRunResponse` message. Then, when a requester is asking for a pipeline run, the server can compute this information based on the jobs attached to a pipeline run. We wouldn't need to store the duplicate information this way too.

jgwhite: We discussed this a little during RFC review.

If we decide not to put this info directly on `PipelineRun` (and I think the rationale is fair) then we’ll need `UI_ListPipelineRuns` — something like this:

```diff
+ rpc UI_ListPipelineRuns(UI.ListPipelineRunsRequest) UI.ListPipelineRunsResponse;

  ⋮

  message UI {

+   message ListPipelineRunsRequest {
+     Ref.Pipeline pipeline = 1;
+     PaginationRequest pagination = 2;
+   }

+   message ListPipelineRunsResponse {
+     repeated UI.PipelineRunBundle pipeline_run_bundles = 1;
+     PaginationResponse pagination = 2;
+   }

+   message PipelineRunBundle {
+     PipelineRun pipeline_run = 1;
+     google.protobuf.Timestamp queue_time = 6;
+     Ref.Application application = 7;
+     Job.DataSource.Ref data_source_ref = 8;
+   }

  }
```

We were also planning to make changes to `ListPipelineRuns`, so this would tackle two changes at once.

The `UI.PipelineRunBundle` message can also be used in `UI.GetPipelineRunResponse` (when we get to that).

> Is the idea to repeat this information to stop making another request to retrieve job info?

Partly, but also to simplify UI logic.

sdav9375:
Closing this PR as the changes needed are being put in a newly created `PipelineRunBundle` as described in @jgwhite 's comment above. The addition can be seen in the [UI_ListPipelines PR](https://github.com/hashicorp/waypoint/pull/4509)

PR #4560, opened 2023-03-03
Add `gateway.yml` check

jgwhite:
## Why the change?

When adding new RPCs, we kept on forgetting to add them to `gateway.yml`[^1]. This patch adds a simple check to stop us making the same oversight in future.

## What does it look like?

CleanShot 2023-03-03 at 15 35 49@2x

CleanShot 2023-03-03 at 15 41 51@2x

CleanShot 2023-03-03 at 15 47 59@2x

CleanShot 2023-03-03 at 15 48 30@2x

## How do I test it?

Locally:

1. Delete an item from the `skip` list in `scripts/gateway-yml-check.sh` or delete an item from `pkg/server/proto/gateway.yml`
2. Run `./scripts/gateway-yml-check.yml`
3. Verify you see an error reporting that the RPC is missing

GitHub:

1. Make a branch of this PR
2. Delete an item from the `skip` list in `scripts/gateway-yml-check.sh` or delete an item from `pkg/server/proto/gateway.yml`
3. Commit and push
4. Open a PR for your branch
5. Verify the check fails

[^1]: `gateway.yml` declares gRPC->HTTP mappings for grpc-gateway

### DIFF
--- a/.github/workflows/gateway-yml-check.yml
+++ b/.github/workflows/gateway-yml-check.yml
@@ -1,0 +1,17 @@
+name: Check gateway.yml for missing RPCs
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches: [main]
+
+jobs:
+  gateway-yml-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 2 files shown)


izaaklauer (approved):
Pragmatism at it's finest.

PR #4561, opened 2023-03-03
Add `UI_GetDeployment` to `gateway.yml`

jgwhite:
## Why the change?

This has been missing for ages, suggesting we’re not currently using it. Still, doesn’t hurt to plug the gap.

### DIFF
--- a/pkg/server/gen/server.swagger.json
+++ b/pkg/server/gen/server.swagger.json
@@ -5838,6 +5838,69 @@
         ]
       }
     },
+    "/ui/deployment/{ref.id}": {
+      "get": {
+        "summary": "GetDeployment returns a deployment",
+        "operationId": "Waypoint_UI_GetDeployment",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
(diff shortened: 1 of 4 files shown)


pkg/server/gen/server.swagger.json
```diff
@@ -6204,6 +6267,122 @@
         ]
       }
     },
+    "/ui/project/{project.project}/pipelines": {
```
jgwhite, line 6270: Not strictly related to this change. Looks like we forgot to do a final `make gen/server` as part of #4509.

sdav9375: I'm a bit unclear on what happened. We did generate a bunch of stuff, so when was this step supposed to happen?

sdav9375 (approved):
Apart from being a bit unclear on the `pipelines` generated code, LGTM

jgwhite:
@sdav9375 the takeaway (and this was new to me too) is that we need to run `make gen/server` whenever we change `gateway.yml`.

PR #4567, opened 2023-03-10
Add `UI_GetPipelineRun`

jgwhite:
## Why the change?

This API will feed the pipeline run detail page.

## What’s the plan?

- [x] Declare rpc and messages
- [x] Implement handlers

### Any deferred tasks?

- [ ] Add more-efficient state functions

## What does it look like?

Given a run of a valid pipeline ([something like this](https://github.com/jgwhite/waypoint-demos/blob/main/pipelines/waypoint.hcl#L30-L44)), running the command below should produce a big ol’ pile of JSON that looks something like [this gist](https://gist.github.com/jgwhite/b2c8d8b22ec10761633c4d8687fdf577).

```sh
./contrib/waypoint-grpc/waypoint-grpc.sh UI_GetPipelineRun '
{
  "pipeline": {
    "id": "01H06BBZH5BK3M7761PD8WP892"
  },
  "sequence": 2
}
'
```

## How do I test it?

1. Run a pipeline
2. From your waypoint source directory, run the following:
   ```sh
   ./contrib/waypoint-grpc/waypoint-grpc.sh UI_GetPipelineRun '
   {
     "pipeline": {
       "id": "<YOUR-PIPELINE-ID>"
     },
     "sequence": <YOUR-RUN-SEQUENCE>
   }
   '
   ```
3. Verify that you see some lovely deeply-nested JSON

### DIFF
--- a/pkg/server/ptypes/pipeline_ui.go
+++ b/pkg/server/ptypes/pipeline_ui.go
@@ -1,0 +1,482 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package ptypes
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/hashicorp/waypoint/internal/pkg/graph"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+// UI_PipelineRunTreeFromJobs takes a graph of Jobs and StatusReports and
+// returns a specialized tree representation that drives the pipeline-run
+// timeline UI.
+func UI_PipelineRunTreeFromJobs(jobs []*pb.Job, statusReports []*pb.StatusReport) (*pb.UI_PipelineRunTreeNode, error) {
+	processor := newUIPipelineProcessor(jobs, statusReports)
+
+	return processor.run()
+}
+
+// newUIPipelineProcessor creates a new uiPipelineProcessor for the given jobs and status reports
+func newUIPipelineProcessor(jobs []*pb.Job, statusReports []*pb.StatusReport) uiPipelineProcessor {
+	p := uiPipelineProcessor{
+		jobIdx:  make(map[string]*pb.Job),
+		nodeIdx: make(map[string]*pb.UI_PipelineRunTreeNode),
+	}
+
+	// Populate job index
+	for _, job := range jobs {
+		p.jobIdx[job.Id] = job
+	}
+
+	// Populate status report index
+	statusReportIdx := make(map[string]*pb.StatusReport)
+	for _, statusReport := range statusReports {
+		if d := statusReport.GetDeploymentId(); d != "" {
+			statusReportIdx[d] = statusReport
+		} else if r := statusReport.GetReleaseId(); r != "" {
+			statusReportIdx[r] = statusReport
+		}
+	}
+
+	// Populate node index
+	for _, job := range jobs {
+		var step *pb.Pipeline_Step
+		var statusReport *pb.StatusReport
+
+		if stepOp := job.GetPipelineStep(); stepOp != nil {
+			step = stepOp.Step
+		} else {
+			// This job represents a nested or referenced pipeline
+			// invocation. It doesn’t encode all the necessary
+			// information to infer the step details, so we’ll
+			// create a placeholder step and fill in the details as
+			// we learn more from other jobs in the set.
+			step = &pb.Pipeline_Step{
+				Name:      job.Pipeline.Step,
+				DependsOn: []string{},
+				Kind:      &pb.Pipeline_Step_Pipeline_{},
+			}
+		}
+
+		// Look up the latest status report for this node
+		if d := job.Result.GetDeploy().GetDeployment(); d != nil {
+			statusReport = statusReportIdx[d.Id]
+		}
+		if r := job.Result.GetRelease().GetRelease(); r != nil {
+			statusReport = statusReportIdx[r.Id]
+		}
+
+		// Build the node (children will be populated later)
+		node := &pb.UI_PipelineRunTreeNode{
+			Step:               step,
+			Job:                &pb.Ref_Job{Id: job.Id},
+			State:              p.nodeStateFromJob(job),
+			StartTime:          job.AckTime,
+			CompleteTime:       job.CompleteTime,
+			Application:        job.Application,
+			Workspace:          job.Workspace,
+			Result:             job.Result,
+			LatestStatusReport: statusReport,
+			Children: &pb.UI_PipelineRunTreeNode_Children{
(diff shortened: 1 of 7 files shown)


pkg/server/singleprocess/service_ui_pipeline_run.go
```diff
@@ -72,3 +74,96 @@ func (s *Service) UI_ListPipelineRuns(
...
+				log,
+				err,
+				"failed to get jobs for all pipeline steps",
+			)
+		}
+		jobs = append(jobs, job)
+	}
+	metrics.MeasureOperation(ctx, start, "fetch_jobs_for_ui_get_pipeline_run")
+
+	// Fetch latest status report for every deployment and release
```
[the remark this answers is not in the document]
jgwhite: Good idea, I’ll take a look!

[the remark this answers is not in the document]
jgwhite: I gave [`getDeploymentBundle`](https://github.com/hashicorp/waypoint/blob/613b24583d0d361e70ea38e34eda0a09293cd7f6/pkg/server/singleprocess/service_ui_deploy.go#L134) a spin and it wasn’t that much simpler…

```diff
  if d := job.Result.GetDeploy().GetDeployment(); d != nil {
-          sr, err := s.GetLatestStatusReport(ctx, &pb.GetLatestStatusReportRequest{
-                  Application: d.Application,
-                  Workspace:   d.Workspace,
-                  Target: &pb.GetLatestStatusReportRequest_DeploymentId{
-                          DeploymentId: d.Id,
-                  },
-          })
-          if err != nil {
-                  return nil, hcerr.Externalize(
-                          log,
-                          err,
-                          "failed to get latest status report for deployment %q",
-                          d.Id,
-                  )
-          }
+          bundle, err := s.getDeploymentBundle(ctx, d.Application, d.Workspace, d)
+          if err != nil {
+                  return nil, hcerr.Externalize(
+                          hclog.FromContext(ctx),
+                          err,
+                          "error getting deployment bundle for application",
+                          "application",
+                          d.Application.GetApplication(),
+                          "workspace",
+                          d.Workspace.GetWorkspace(),
+                  )
+          }
+          sr := bundle.LatestStatusReport

           if sr != nil {
                   statusReports = append(statusReports, sr)
           }
  }
  ```

Is this what you had in mind, or something different?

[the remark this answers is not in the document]
Comment (briancain) [reply pkg/server/singleprocess/service_ui_pipeline_run.go:113]: Yep! I guess in this case we're only caring about the status report since we already have the deployment, so yeah maybe not too useful of an optimization. Worth a shot!

File: pkg/server/ptypes/pipeline_ui.go [hunk 2]
```diff
@@ -0,0 +1,408 @@
...
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/hashicorp/waypoint/internal/pkg/graph"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+func UI_PipelineRunTreeFromJobs(jobs []*pb.Job, statusReports []*pb.StatusReport) (*pb.UI_PipelineRunTreeNode, error) {
```
briancain, line 14: This could use a doc comment detailing what this func does

pkg/server/ptypes/pipeline_ui.go
```diff
@@ -0,0 +1,408 @@
...
+				continue
+			}
+
+			g.Add(j)
+			g.AddEdge(j, i)
+		}
+	}
+
+	// Remove short cuts from the graph
+	g.TransitiveReduction()
```
briancain, line 103: Can you expand in a doc comment here what this is doing for us?

pkg/server/ptypes/pipeline_ui.go
```diff
@@ -0,0 +1,408 @@
...
+			return nil, fmt.Errorf("could not find root node")
+		}
+
+		rootId = id
+		rootNode = node
+	}
+
+	var process func(string, *pb.UI_PipelineRunTreeNode) error
+
+	injectVirtualNode := func(inputId string, nextIds []string, outputNode *pb.UI_PipelineRunTreeNode) (string, *pb.UI_PipelineRunTreeNode, error) {
```
briancain, line 125: It looks like `injectVirtualNode` and `processSubPipeline` might be nice to have as their own functions rather than defined inside this one, I think it makes sense to abstract them out on their own, what do you think @jgwhite ? This would make it nice to write tests for as well.

pkg/server/ptypes/pipeline_ui.go
jgwhite: Haha, whoops! Yep, I’ll get right on that.

pkg/server/ptypes/pipeline_ui.go
jgwhite: Absolutely, will do.

pkg/server/ptypes/pipeline_ui.go
jgwhite: I’ve refactored this all significantly so there are more discrete functions. I didn’t write tests for `injectVirtualNode` and `processSubPipeline` in the end because I suspect they will change, so I’m treating them as an implementation detail. Keen to hear your guidance on this though.

jgwhite:
@briancain this is ready for another look 👀

pkg/server/ptypes/pipeline_ui.go
briancain: I think that's fair, especially if the function they are used in is tested? There won't be 100% coverage but that seems like a good first pass.

jgwhite: > especially if the function they are used in is tested?

Yeah, I think the public function has a good set of test cases.

pkg/server/ptypes/pipeline_ui.go
```diff
@@ -0,0 +1,479 @@
...
+//
+// After:
+//
+//	A
+//	V
+//	F
+//
+// It returns the new node and assumes the parent function will perform further
+// processing to embed the original children within the new node.
+func (p uiPipelineProcessor) foldBranches(inputId string, outputNode *pb.UI_PipelineRunTreeNode) (string, *pb.UI_PipelineRunTreeNode, error) {
```
briancain, line 255: Just a small style nit, since this func is kinda long, we typically format it to be more condensed. Something like:

```suggestion
func (p uiPipelineProcessor) foldBranches(
	inputId string,
	outputNode *pb.UI_PipelineRunTreeNode,
) (string, *pb.UI_PipelineRunTreeNode, error) {
```

briancain (approved):
Just a few comments from me, but looks great! Hurray protobufs 😄

PR #4595, opened 2023-03-27
circle: Enable errcheck linter

briancain:
No more dropped errs.


briancain:
This one might take some work for ignoring pkgs that aren't ours or cases where we don't specifically care about the dropped err.

PR #4598, opened 2023-03-29
Use a separate context when deleting an instance record

evanphx:


izaaklauer (approved):
Also makes great sense.

PR #4601, opened 2023-03-29
More context fixes, better job logging

evanphx:
Log at info level the proper lifecycle of a job to aid in debugging


izaaklauer (approved):
This will help us diagnose a potential race condition on completing jobs.

PR #4603, opened 2023-03-30
docs: add runner-profile-edit to nav data

BRKalow:
* Adds `runner-profile-edit` to commands nav data
* Updates our content conformance package so the content check validates nav data correctly


briancain (approved):
Thank you for catching the missing nav link!

PR #4606, opened 2023-03-31
core: Add Project Templates to protos

demophoon:
Add RPCs and server stubs to Waypoint for managing project templates


pkg/server/proto/server.proto
```diff
@@ -458,6 +458,18 @@ service Waypoint {
   // project config to the Waypoint database.
   rpc ConfigSyncPipeline(ConfigSyncPipelineRequest) returns (ConfigSyncPipelineResponse);
 
+  //----------------------------------------------------------------------
+  // ProjectTemplate endpoints
+  //----------------------------------------------------------------------
+
+  rpc UpsertProjectTemplate(UpsertProjectTemplateRequest) returns (UpsertProjectTemplateResponse);
```
izaaklauer, line 465: Following the convention, let's add some comments! Those will turn into swagger api docs.

pkg/server/proto/server.proto
```diff
@@ -5094,6 +5114,62 @@ message ConfigSyncPipelineResponse {
   map<string, Ref.Pipeline> synced_pipelines = 1;
 }
 
+/********************************************************************
+* ProjectTemplate
+********************************************************************/
+
+message GetProjectTemplateRequest {
```
izaaklauer, line 5121: Same here - lets add comments to these structs.

pkg/server/proto/server.proto
demophoon: Added!

pkg/server/proto/server.proto
```diff
@@ -5094,6 +5122,89 @@ message ConfigSyncPipelineResponse {
...
+  // is shared between platform engineers and application developers.
+  string expanded_summary = 4;
+
+  // The template which is rendered when a Project is created from a
+  // ProjectTemplate to be shown to application developers.
+  string readme_template = 5;
+
+  // Settings for the Waypoint project that should be set when a project is
+  // created from a ProjectTemplate.
+  WaypointProject waypoint_project = 6;
```
catsby, line 5185: What kind of settings are we expecting here? Forgive me I don't recall them from the RFC

pkg/server/proto/server.proto
```diff
@@ -5094,6 +5122,89 @@ message ConfigSyncPipelineResponse {
...
+  // The Terraform No-Code Module which should be used to provision
+  // infrastructure that is used by the project created from a ProjectTemplate.
+  TerraformNocodeModule terraform_nocode_module = 7;
+
+  // A list of descriptive strings that can be applied to the ProjectTemplate.
+  repeated string tags = 8;
+
+  message WaypointProject {
+    // The templated waypoint.hcl file stored as HCL.
+    bytes waypoint_hcl_template = 1;
```
catsby, line 5196: I'm a bit confused on the naming here, the upper level item is a `ProjectTemplate`, which contains a type `WaypointProject` which I would expect by the name to reference a project in Waypoint, but instead it's a HCL template file?

izaaklauer: Maybe we should add a description to the message that says something like:

```
WaypointProject governs the properties that will be set on the final waypoint Project that we create. Future fields will likely include status report and datasource polling settings.
```

izaaklauer (approved):

pkg/server/proto/server.proto
```diff
@@ -5094,6 +5138,138 @@ message ConfigSyncPipelineResponse {
...
+  // multiple templates
+  string summary = 3;
+
+  // A long summary of what the ProjectTemplate is to be used for. This summary
+  // is shared between platform engineers and application developers.
+  string expanded_summary = 4;
+
+  // The template which is rendered when a Project is created from a
+  // ProjectTemplate to be shown to application developers.
+  string readme_template = 5;
```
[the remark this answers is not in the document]
demophoon: Updated in `cf9263587`

paladin-devops (approved):

PR #4607, opened 2023-03-31
App overview

cicoyle:
This PR intends to:

1. Stub out the `EventPut` functionality to put events (build/deployment/release) into an event table upon upserting the build/deployment/release. Although, this is not actually implemented in boltdb yet, hence why I return `Unimplemented` and the test is in the `Unimplemented` list.
2. Server/State tests for events.
3. The addition of artifact id to some tests that were failing due to the deployment not having the artifact id on it, but this is now necessary for eventPut.
4. The service `UI_ListEvents` func and Stub out the `EventListBundles` functionality to return the list of paginated events. Although, this is not actually implemented in boltdb yet, hence why I return `Unimplemented` and the test is in the `Unimplemented` list.
5. Regenerate the protos with the new event proto updates - you should see 2 new rpcs: `UI_ListEvents` & `INTERNAL_AdditionalMessages` with the corresponding messages + the sorting request needed for pagination.
6. README for what the `pkg/server/ptypes` is intended for.
7. Some typo fixes
9. Server state struct for event


pkg/server/ptypes/README.md
```diff
@@ -0,0 +1,10 @@
+# Ptypes
```
izaaklauer, line 1: Way to be the change you want to see!

pkg/server/proto/server.proto
```diff
@@ -485,6 +485,12 @@ service Waypoint {
 
   // List releases for a given application.
   rpc UI_ListReleases(UI.ListReleasesRequest) returns (UI.ListReleasesResponse);
+
+  //List UI events for a given application.
+  rpc UI_ListEvents(UI.ListEventsRequest) returns (UI.ListEventsResponse);
```
izaaklauer, line 490: I think you need an entry in `gateway.yml` to go along with this

pkg/server/proto/server.proto
```diff
@@ -485,6 +485,12 @@ service Waypoint {
...
+
+  // List UI event schemas (primarily for ensuring types flow through to swagger, then generated by OpenAPI).
```
izaaklauer, line 492: Could you elaborate a little more on what we expect the frontend (or other api clients) to use this for?

Also likely needs an entry in `gateway.yml`

pkg/server/ptypes/build.go
```diff
@@ -10,7 +10,7 @@ import (
 	pb "github.com/hashicorp/waypoint/pkg/server/gen"
 )
 
-// TestBuild returns a valid user for tests.
+// TestBuild returns a valid build for tests.
```
izaaklauer, line 13: Another good catch!

pkg/server/ptypes/event.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/hashicorp/waypoint/internal/pkg/validationext"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+// ValidateUIListEventsRequest
+func ValidateUIListEventsRequest(v *pb.UI_ListEventsRequest) error {
+	return validationext.Error(validation.ValidateStruct(v,
+		validation.Field(&v.Application, validation.Required),
+		validation.Field(&v.Workspace, validation.Required),
```
izaaklauer, line 14: Is workspace really required?

pkg/serverhandler/handlertest/test_service_ui_event.go
```diff
@@ -0,0 +1,157 @@
...
+		_, err = client.UpsertApplication(context.Background(), &pb.UpsertApplicationRequest{
+			Project: refProj,
+			Name:    "a_test",
+		})
+		require.NoError(err)
+
+		// Create build, deployment, release
+		createEvents(client, ctx, t)
+
+		time.Sleep(250 * time.Millisecond)
```
izaaklauer, line 44: Curious: what is the sleep here accomplishing?

izaaklauer (changes requested):
A few missing entries in gateway.yml, but otherwise looks great!

pkg/serverhandler/handlertest/test_service_ui_event.go
cicoyle: My idea for the `time.Sleep` was to have a different Status Start & Complete time for the created resources.

pkg/server/ptypes/event.go
cicoyle: Good catch, probably not.

pkg/server/proto/server.proto
andrew-hashicorp: @izaaklauer We don't expect anything to call this, but by defining an RPC that uses `UI.EventBuild / UI.EventDeployment / UI.EventRelease` protos, protoc-gen-swagger picks them up when generating the swagger. The swagger is then used when the frontend by openapi-generator and will generate TypeScript types for `UI.EventBuild / UI.EventDeployment / UI.EventRelease`.

We can then use these on the frontend when unmarshaling the untyped EventBundle.EventData field.

pkg/serverhandler/handlertest/test_service_ui_event.go
andrew-hashicorp: My understanding is that we emit 2 events for the same operation, one with status.startTime filled in, and one with both status.startTime + status.completeTime filled in. Is that correct?

At the state layer, we then upsert events so the same event gets updated with a completeTime when it's complete?

On the frontend then, depending on when we call UI_ListEvents, we could get an event with either a startTime but no completeTime or both startTime and completeTime, but we would always only get one event per operation?

pkg/server/proto/server.proto
cicoyle: I just added that in.

pkg/server/proto/server.proto
cicoyle: Thanks @andrew-hashicorp for elaborating. I updated the description comment line above to read: `  // List possible event data types.` per my discussion with Andrew. LMK if yall feel like that could be better explained.

izaaklauer: Ah interesting. I suppose it makes sense that the types wouldn't make it all the way through to the final js client if it isn't used by an rpc here.

Can we add that as a comment on this RPC then, and rename it to something that makes it clear potential api callers can ignore this? Something like

izaaklauer (approved):
This looks good overall! Small request for comments and updating naming for rpc endpoints we don't ever expect to call, but that shouldn't block merging this.

pkg/serverhandler/handlertest/test_service_ui_event.go
cicoyle: I believe that is correct. As UI_ListEvents calls the EventListBundles function which will list the events from the db. An event will have a completed time in the status field only after its completed and the second Upsert<Operation> call is made from the runner.

Each Upsert<Build/Deploy/Release> gets called upon the correlative cli commands, that puts the operation in the db, then the runner will call it again to update that record to put in the completed time once its done. EventPut operates in the same way as it gets called in the Upsert<Operation>. This will put an event the first time it is 'started', then once the runner updates the operation record the second time it calls Upsert<Operation> it will update the correlative event record with a new EventTimestamp and eventData containing that completed time.

pkg/server/proto/server.proto
```diff
@@ -652,6 +658,81 @@ message UI {
...
+    // The list of events with related messages prefetched.
+    repeated UI.EventBundle events = 1;
+    PaginationResponse pagination = 2;
+  }
+
+  // An event packaged alongside prefetched related messages.
+  message EventBundle {
+    google.protobuf.Timestamp timestamp = 1;
+    EventType eventType  = 2;
+    bytes eventData = 3;
```
andrew-hashicorp, line 684:
```suggestion
    EventType event_type  = 2;
    bytes event_data = 3;
```
This was a mistake in the RFC (since corrected) - Should be in snake_case to follow the convention established in the rest of `server.proto`.

pkg/server/proto/server.proto
```diff
@@ -652,6 +658,81 @@ message UI {
...
+    string pipelineRun_id = 1;
+    uint64 sequence = 2;
+//    Component component = 3; //aka platform
+//    Ref.Workspace workspace = 4;
+  }
+*/
+  message ListEventSchemasResponse {
+    UI.EventBuild eventBuildSchema = 1;
+    UI.EventDeployment eventDeploymentSchema = 2;
+    UI.EventRelease eventReleaseSchema = 3;
```
andrew-hashicorp, line 733:
```suggestion
    UI.EventBuild event_build_schema = 1;
    UI.EventDeployment event_deployment_schema = 2;
    UI.EventRelease event_release_schema = 3;
```
Same as above - should be snake_case to follow convention.

andrew-hashicorp (changes requested):

andrew-hashicorp (changes requested):

pkg/server/proto/gateway.yml
```diff
@@ -520,3 +520,10 @@ http:
 
   - selector: hashicorp.waypoint.Waypoint.UI_ListPipelineRuns
     get: /ui/project/{pipeline.owner.project.project}/pipelines/{pipeline.owner.pipeline_name}/runs
+
+    # Events
+  - selector: hashicorp.waypoint.Waypoint.UI_ListEvents
+    get: /ui/events/application/{application.application}
```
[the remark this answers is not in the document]
cicoyle: oh really? I updated the endpoint accordingly.

cicoyle:
fixing merge conflicts now.

pkg/server/proto/gateway.yml
```diff
@@ -547,3 +547,13 @@ http:
...
+
+    # Events
+  - selector: hashicorp.waypoint.Waypoint.UI_ListEvents
+    get: /ui/project/{application.project}/application/{application.application}/events
+
+  - selector: hashicorp.waypoint.Waypoint.UI_ListEvents
+    get: /ui/project/{application.project}/application/{application.application}/workspace/{workspace.workspace}/events
+
+  - selector: hashicorp.waypoint.Waypoint.INTERNAL_AdditionalMessages
+    get: /ui/events/event-schemas
```
izaaklauer, line 559: can we modify this name to also be internal additional messages instead of event schemas?

cicoyle: ahh, yes I forgot about that. yeah I'll update that

izaaklauer: After that, merge away!

cicoyle: done

PR #4608, opened 2023-03-31
Remove Plugin MDX content

BrandonRomano:
We've moved over to the new [integrations system](https://developer.hashicorp.com/waypoint/integrations), and these docs files for plugins are no longer needed. 

The motivation for removing these now is because these files are actually getting indexed by our Algolia search as if the old pages still existed.


BrandonRomano:
Wow very useful check ☝️

krantzinator:
Have we talked about updating our [website-mdx check](https://github.com/hashicorp/waypoint/pull/4608/checks?check_run_id=12438352035) for the new system? I believe that check is irrelevant now since docs are auto-generated on changes?

demophoon (approved):

PR #4616, opened 2023-04-10
Include project template in project proto

paladin-devops:
The project proto as well as the proto response for getting a project in the UI is updated in this PR to include a project template. This will enable the UI to display which project template was used to create a project, if any.


demophoon (approved):

pkg/server/proto/server.proto
```diff
@@ -918,6 +919,8 @@ message Project {
 
   ProjectState state = 11;
 
+  Ref.ProjectTemplate project_template = 12;
```
izaaklauer, line 922: Can we add a comment like

```
project_template is a reference to the template that this project was created from, if any.
```

izaaklauer (changes requested):

PR #4618, opened 2023-04-13
api: fix gateway UI paths

jgwhite:
These endpoints are not currently used, so this change has very little practical impact other than keeping things consistent.

### DIFF
--- a/pkg/server/proto/gateway.yml
+++ b/pkg/server/proto/gateway.yml
@@ -487,16 +487,7 @@ http:
 
     # Deployments
   - selector: hashicorp.waypoint.Waypoint.UI_ListDeployments
-    get: /ui/deployments/workspace/{workspace.workspace}
-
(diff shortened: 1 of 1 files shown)


pkg/server/proto/gateway.yml
```diff
@@ -487,16 +487,7 @@ http:
...
-    get: /ui/deployments/workspace/{workspace.workspace}
-
-  - selector: hashicorp.waypoint.Waypoint.UI_ListDeployments
-    get: /ui/deployments/application/{application.application}
-
-  - selector: hashicorp.waypoint.Waypoint.UI_ListDeployments
-    get: /ui/deployments/project/{application.project}
-
-  - selector: hashicorp.waypoint.Waypoint.UI_ListDeployments
-    get: /ui/deployments/state/{physical_state}
```
jgwhite, line 499: Nor do we need to fetch all deployments by their physical state.

andrew-hashicorp (approved):

PR #4619, opened 2023-04-13
Additional install flags for Nomad service address and network mode

gulducat:
The overall goal here is to `waypoint install` into a Nomad node from the outside.  In my case, a Vagrant machine from my laptop, but also should work with other cases like cloud vendors where the `waypoint-server` service needs to be given a public IP, to which the install procedure connects after job submission.

For example, using a `meta.public_ip` that I set manually in my Nomad client config:

```
$ ./waypoint install -accept-tos \
  -platform=nomad \
  -nomad-host-volume=waypoint \
  -nomad-runner-host-volume=waypoint \
  -nomad-service-provider=nomad \             
  -nomad-service-address='${meta.public_ip}' \
  -nomad-network-mode=bridge

✓ Waypoint server ready
...
✓ Runner "static" adopted successfully.
...
Advertise Address: 192.168.56.200:9701
Web UI Address: https://192.168.56.200:9702/
```

Of course `-nomad-service-address=192.168.56.200` would have worked just as well, but the meta is for demonstration, as I imagine that on AWS one might use `'${attr.unique.platform.aws.public-ipv4}'` to set the service address dynamically.

The ~publicly-routable IP at the bottom of the install output (and used during install) comes from the Nomad services:

```
$ nomad service info waypoint-server
Job ID           Address              Tags        Node ID   Alloc ID
waypoint-server  192.168.56.200:9701  [waypoint]  f8ffd891  57e5dcfd
$ nomad service info waypoint-ui
Job ID           Address              Tags        Node ID   Alloc ID
waypoint-server  192.168.56.200:9702  [waypoint]  f8ffd891  57e5dcfd
```

I do wonder though if perhaps separate service(s) could/should be made for this connection during install, vs internal waypoint->waypoint communication?  Do the runner(s) connect to waypoint-server using this service?  I didn't look that far into this part.

### DIFF
--- a/.changelog/4619.txt
+++ b/.changelog/4619.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+cli: new flags for `waypoint install` on Nomad:
+-nomad-service-address and -nomad-network-mode
+```
--- a/internal/serverinstall/nomad.go
+++ b/internal/serverinstall/nomad.go
@@ -41,9 +41,12 @@ type nomadConfig struct {
 	namespace          string            `hcl:"namespace,optional"`
 	serviceAnnotations map[string]string `hcl:"service_annotations,optional"`
 
-	serviceProvider          string   `hcl:"service_provider,optional"`
-	serviceUITags            []string `hcl:"service_ui_tags:optional"`
-	serviceBackendTags       []string `hcl:"service_backend_tags:optional"`
+	serviceProvider    string   `hcl:"service_provider,optional"`
+	serviceUITags      []string `hcl:"service_ui_tags:optional"`
+	serviceBackendTags []string `hcl:"service_backend_tags:optional"`
+	serviceAddress     string   `hcl:"service_address,optional"`
+	networkMode        string   `hcl:"network_mode,optional"`
+
 	consulService            bool     `hcl:"consul_service,optional"`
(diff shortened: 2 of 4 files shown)


krantzinator:
@gulducat thanks so much for doing this! Could you add a changelog entry as well? You can copy one of the ones in `./changelog`, just change the name to match this PR number, and the type should be either `improvement` or `feature`, up to your discretion!

demophoon:
@gulducat This looks great, Tested this out and nomad is giving the cli back the address given for the services! Do you mind if I push up the docs related changes for this change to the PR to make tests happy? (You can also do this by running `make gen/website-mdx` and pushing up the changes)

gulducat:
Thanks folks!

I added a changelog entry and what I think is the relevant subset of `make gen/website-mdx` for my new flags.  Do those look good to y'all?

P.S. FYI gosh the full changeset that gen/website produces is quite a lot!  also I didn't realize at first that it had succeeded well-enough, because the `npx .. format content` command errors "GET https:/\/registry.npmjs.org/next-hashicorp - Not found" .. perhaps I lack some configuration to access it?

demophoon:
There may be something missing for other parts of the website, but what you generated and committed was all that we needed! This all looks good from my point of view now

Thanks

demophoon (approved):

PR #4646, opened 2023-05-02
Declare `UI_GetPipelineRun` and friends

jgwhite:
A precursor to #4567. I still have a few kinks to iron out in #4567 but UI work can proceed against these types.


sdav9375 (approved):
LGTM

PR #4647, opened 2023-05-02
Add get latest deployment

andrew-hashicorp:
This PR adds an RPC, [`GetLatestDeployment`](https://github.com/hashicorp/waypoint/pull/4647/files#diff-c567084dcb9faa5b42fdbf43e940675c02d61325c4c36e3edf2134c47fa8b826R142), that exposes the existing [`DeploymentLatest`](https://github.com/hashicorp/waypoint/blob/28c773676b77670c47e38bc14e1ff305956a437a/pkg/serverstate/serverstate.go#L143) state function.


pkg/server/proto/server.proto
```diff
@@ -137,6 +137,10 @@ service Waypoint {
   // GetDeployment returns a deployment
   rpc GetDeployment(GetDeploymentRequest) returns (Deployment) {}
 
+  // GetLatestDeployment returns the most recent successfully completed build
```
briancain, line 140:
```suggestion
  // GetLatestDeployment returns the most recent successfully completed Deployment
```

You will want to make sure to re-gen the protos here so the doc string is accurate in the generated method!

pkg/server/proto/server.proto
```diff
@@ -137,6 +137,10 @@ service Waypoint {
...
+  // for an app.
+  rpc GetLatestDeployment(GetLatestDeploymentRequest) returns (Deployment) {}
```
briancain, line 142: This should be a `GetLatestDeploymentResponse` (a new proto message). It will mean this endpoint will be much more flexible for any clients consuming this endpoint. Otherwise the moment the response format changes, all clients will break who assume they should be receiving a Deployment proto message.

pkg/server/ptypes/deployment.go
```diff
@@ -81,3 +81,11 @@ func ValidateGetDeploymentRequest(v *pb.GetDeploymentRequest) error {
 		}),
 	))
 }
+
+// TODO: Make this generic w/ build + release?
```
briancain, line 85: I wouldn't! I think it's okay we're extra verbose with these API validation functions.

pkg/server/singleprocess/service_deploy.go
```diff
@@ -123,6 +124,35 @@ func setDeploymentUrlIfNeeded(d *pb.Deployment) {
 	}
 }
 
+// TODO: test
```
briancain, line 127: Yes please!

briancain: Ah, looks like there is a test? Maybe this comment can be removed then.

pkg/serverhandler/handlertest/test_service_deploy.go
```diff
@@ -138,6 +139,77 @@ func TestServiceDeployment_GetDeployment(t *testing.T, factory Factory) {
...
+	client, _ := factory(t)
+
+	var application *pb.Ref_Application
+	var workspace *pb.Ref_Workspace
+
+	var latestSuccessfulDepResp *pb.UpsertDeploymentResponse
+	// Upsert some deployments
+	depNum := []int{1, 2, 3}
+	for index := range depNum {
+		// for i := 0; i < 3; i++ {
```
briancain, line 155: Delete?

briancain (changes requested):

pkg/server/proto/server.proto
andrew-hashicorp: That makes sense.

However, every other `GetLatestXYZ` endpoint simply returns XYZ so I'm wondering if it's confusing for `GetLatestDeployment` to not return `Deployment`?

pkg/server/ptypes/deployment.go
andrew-hashicorp: Sounds good to me!

pkg/server/singleprocess/service_deploy.go
andrew-hashicorp: There's a test, the comment is outdated 😄

pkg/serverhandler/handlertest/test_service_deploy.go
andrew-hashicorp: Good catch 💯

pkg/server/proto/server.proto
briancain: Hm, if the other ones return the bare Deployment proto then I guess we should stick to convention.

The idea with a special response message means it's much more flexible to include other proto messages without breaking all of the consuming clients of the API. If it's only a Deployment proto message, if you ever wanted to add status reports proto message for example, you'd have to go through the entire project and fix the response format to add a new field that was expecting the response type to only be a Deployment proto.

briancain (approved):
Looks good! 🚀

PR #4660, opened 2023-05-03
server: Ensure trigger names are valid paths

briancain:
Prior to this commit, a user could create a bad HTTP path with a trigger name like `../bad`. This commit fixes that behavior by validating the given trigger name matches our expected valid path tokens.

With this fix:

```
brian@ubuntu:waypoint-tetris λ waypoint trigger create -name=../hello -op=build                                                                        ±[main]
! trigger: (name: name cannot contain '../'.).
```

### DIFF
--- a/.changelog/4660.txt
+++ b/.changelog/4660.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+trigger: Ensure trigger Name is only alpha-numeric
+```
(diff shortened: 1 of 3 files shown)


jgwhite (approved):
NICE

paladin-devops (approved):

PR #4661, opened 2023-05-03
Move CRT things into their own dockerfile

catsby:
This PR moves the newly added CRT steps in the Dockerfile to their own file, so they can be referenced/used independently


briancain (approved):
Awesome!! Thank you! 🙏🏻

PR #4665, opened 2023-05-05
Standardizing templating routes

izaaklauer:
Using dashes, like our other endpoints do!

Precedent:
```
/waypoint/2022-02-03/namespace/*/config-source
/waypoint/2022-02-03/namespace/*/runner/on-demand
```

NOTE: while this is a breaking api change, the templating feature is currently unreleased, so we don't need to worry about maintaining backwards compat.


paladin-devops (approved):
💨

PR #4671, opened 2023-05-08
cli: Fixup snapshot CLI when creating a file that might exist

briancain:
Prior to this commit, the init writer for snapshot backups wouldn't validate the existence of the file its creating prior to creating it. This fixes that up by instead calling the OpenFile func with CREATE flags enabled.

Fixes WAYP-1150


catsby (approved):
LGTM for posterity `O_EXCL` will cause this to error if the file already exists

PR #4674, opened 2023-05-08
builtin/exec: Properly return a closer for template rendering

briancain:
Prior to this commit, the closer function for rendering a template was commented out, meaning that the exec plugin wouldn't properly cleanup its mess in `/tmp` while rendering commands. This commit updates that behaviort to return the proper closer func to be called by the caller once they're finished rendering templates.

Fixes WAYP-1164

### DIFF
--- a/builtin/exec/platform.go
+++ b/builtin/exec/platform.go
@@ -152,7 +152,7 @@ func (p *Platform) renderTemplate(tpl *ConfigTemplate, data *tplData) (string, f
 	if err != nil {
 		return "", nil, err
 	}
-	closer := func() {} //os.RemoveAll(td) }
+	closer := func() { os.RemoveAll(td) }
 
 	// Render
 	var path string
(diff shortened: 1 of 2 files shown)


builtin/exec/platform.go
```diff
@@ -152,7 +152,7 @@ func (p *Platform) renderTemplate(tpl *ConfigTemplate, data *tplData) (string, f
 	if err != nil {
 		return "", nil, err
 	}
-	closer := func() {} //os.RemoveAll(td) }
+	closer := func() { os.RemoveAll(td) }
```
catsby, line 155: Note that we don't check the error returned here. I'm not totally sure how we would handle that error anyway though. At least it's in /tmp and will be cleaned up by the OS at some point, which is what we're relying on now anyway by not cleaning up

catsby (approved):

PR #4678, opened 2023-05-08
labeler needs 'write' to add labels

krantzinator:
Per https://github.com/actions/labeler/issues/136#issuecomment-1357839196


briancain (approved):
Ohh, makes sense! 👍🏻

catsby (approved):

PR #4696, opened 2023-05-10
Tests for template-related fields Projects

izaaklauer:
## What changed?

This adds state-layer tests for CRUD operations on the new template-related protobuf fields.

Note that the tests are added to a new "unimplemented" stanza in the state tests. Open-source waypoint does not implement templating. 

## What else changed?

This also sneaks in a related protobuf change, updating the readme fields from `string` type to `bytes` on the `Project` and `ProjectTemplate`, as they represent the contents of whole files.

### DIFF
--- a/pkg/server/proto/server.proto
+++ b/pkg/server/proto/server.proto
@@ -1050,8 +1050,9 @@ message Project {
 
   ProjectState state = 11;
 
-  // readme_markdown is markdown formatted instructions on how to operate the project. This may be populated from a project template.
-  string readme_markdown = 13;
+  // readme_markdown is markdown formatted instructions on how to operate the project.
+  // This may be populated from a project template.
+  bytes readme_markdown = 13;
 
   // project_template is a reference to the template that this project was
   // created from, if any.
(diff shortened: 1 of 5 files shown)


pkg/server/proto/server.proto
```diff
@@ -5334,9 +5335,13 @@ message ProjectTemplate {
...
 
-  // A markdown template which is rendered when a Project is created from a
+  // A markdown text template which is rendered when a Project is created from a
   // ProjectTemplate to be shown to application developers.
-  string readme_markdown_template = 5;
+  // Accepted tokens:
+  // "{{ .ProjectName }}", representing the application developer chosen project name
+  // "{{ .TfcOrgName }}", representing the Terraform Cloud organization name in which
+  //  the no-code module was reified
+  bytes readme_markdown_template = 5;
```
paladin-devops, line 5344: Great comments!

pkg/serverstate/statetest/test_project.go
```diff
@@ -1064,3 +1067,116 @@ func TestProjectCount(t *testing.T, factory Factory, restartF RestartFactory) {
...
+			Ref: &pb.Ref_ProjectTemplate_Id{
+				Id: projectTemplate2.Id,
+			},
+		})
+		//)
+
+		resp, err := s.ProjectGet(ctx, &pb.Ref_Project{Project: name})
+		require.NoError(err)
+
+		require.Empty(resp.ProjectTemplate)
```
paladin-devops, line 1180: Based on this, I'm assuming that our "expected behavior" for project details where the template no longer exists is to still return the project proto, just with an empty template field.

paladin-devops (approved):

pkg/serverstate/statetest/test_project.go
izaaklauer: That should be more explicit, good catch!

izaaklauer: I could imagine that users would appreciate us leaving a breadcrumb here, so they could determine the project was created from a template that no longer exists.

It's tricky for other state implementations though. Imagine this as a foreign-key reference to a project template table. Once the project template record is deleted, a bare primary key on project doesn't give us much!

Looking at it today, I think deleting this require is the right call. Clearing the project template ref isn't something think we care about explicitly, and we can follow up in the future with a better breadcrumb as a separate feature.

PR #4699, opened 2023-05-11
fix(cli): honor runner install -platform arg

jgwhite:
Currently we’re only assigning a value to `runnerPlatform` if `c.platform` is empty, leading to the following bug:

```
$ waypoint runner install -platform=docker
! Error installing runner into "": unsupported platform
```

This patch ensures we use the given value for `c.platform`.

```
$ waypoint runner install -platform=docker
✓ Finished connecting to: host.docker.internal:9701
```

### DIFF
--- a/.changelog/4699.txt
+++ b/.changelog/4699.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cli: Honor runner install -platform arg
+```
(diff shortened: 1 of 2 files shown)


briancain (approved):
Oh noooo, good find 🙈

PR #4701, opened 2023-05-12
fix(builtin): de-dupe various hcl annotation keys

jgwhite:


evanphx (approved):
Makes me think we need to build a little hclvet tool that will detect this for us in CI!

evanphx:
Oh lololol my comment is what #4702 does! AMAZING!!

PR #4702, opened 2023-05-12
duplicate tag check

gregone:
❓


.github/workflows/check-struct-tags.yml
```diff
@@ -0,0 +1,17 @@
+name: Check Repo for duplicate struct tags
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches: [main]
+jobs:
+  gateway-yml-check:
```
jgwhite, line 7:
```suggestion
  check-struct-tags:
```

tools/lint-struct-tags/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func analyzeFile(file string) (bool, error) {
+	fset := token.NewFileSet()
+	node, err := parser.ParseFile(fset, file, nil, parser.ParseComments)
+	duplicatesFound := false
+	if err != nil {
+		return false, err
+	}
```
jgwhite, line 52: **[suggestion, non-blocking]**

```suggestion
	if err != nil {
		return false, err
	}
	duplicatesFound := false
```

This is my bad — I think it’s idiomatic to have the `if err != nil {` right next to the line that produces the error.

jgwhite (approved):
This looks good to me (save for the one suggestion). It’ll have to follow #4701 but approving provisionally.

gregone:
@krantzinator it successfully fails. I updated the PR description, but we should merge this after 4701

jgwhite:
Here’s what a “successful failure” looks like, for future reference:

CleanShot 2023-05-12 at 18 14 39@2x

[The job in question](https://github.com/hashicorp/waypoint/actions/runs/4959607061/jobs/8873948537?pr=4702)

jgwhite:
Rebased this now that #4701 is merged. This should now pass.

evanphx:
Rebased to pickup fix from #4701

evanphx:
Rebased again to try to fix the labeler issue.

briancain (approved):
Nice, great idea!

PR #4705, opened 2023-05-12
Fixing a race condition in GetJobStream

izaaklauer:
## What problem does this solve?

Previously, it was theoretically possible for multiple goroutines to write to `err` at the same time in GetJobStream, which would have caused a panic. 

Failing test output: https://gist.github.com/izaaklauer/ff8b42abc38db7785bf49bc3c9b45731

## How does this solve it?

This PR modifies one of the goroutines in job stream handling to reassign the `err` variable when it uses it, avoiding a concurrent write

## How do I verify the fix?

Run the `TestStream_single` test with the `-race` flag, and verify passing!

### DIFF
--- a/pkg/server/singleprocess/service_job.go
+++ b/pkg/server/singleprocess/service_job.go
@@ -738,13 +738,13 @@ func (s *Service) GetJobStream(
 			// Updated job, requery it
 			ws = memdb.NewWatchSet()
 
-			updatedJobjob, err := s.state(ctx).JobById(ctx, job.Id, ws)
+			updatedJob, err := s.state(ctx).JobById(ctx, job.Id, ws)
 			if err != nil {
 				log.Error("error acquiring job by id", "error", err, "id", req.JobId)
 				errCh <- err
 				return
 			}
-			job = updatedJobjob
+			job = updatedJob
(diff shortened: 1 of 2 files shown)


pkg/server/singleprocess/service_job.go
```diff
@@ -737,12 +737,14 @@ func (s *Service) GetJobStream(
 
 			// Updated job, requery it
 			ws = memdb.NewWatchSet()
-			job, err = s.state(ctx).JobById(ctx, job.Id, ws)
+
+			updatedJobjob, err := s.state(ctx).JobById(ctx, job.Id, ws)
```
jgwhite, line 741: I kinda love the name `updatedJobjob` but could it just be `updatedJob`?

izaaklauer: 🤦

evanphx (approved):
So similar to the fix we did over in the runner bits.

pkg/server/singleprocess/service_job.go
paladin-devops: I almost suggested camel-casing it, thinking that @izaaklauer intended to call it "jobjob" 😆

PR #4707, opened 2023-05-12
Runner token ID check

paladin-devops:
This change in this PR prevents a runner token from being used to generate a new token for a runner other than the runner ID which is encoded in the runner token.

### DIFF
--- a/pkg/server/singleprocess/auth.go
+++ b/pkg/server/singleprocess/auth.go
@@ -533,6 +533,20 @@ func (s *Service) GenerateRunnerToken(
 		}
 	}
 
+	decodedToken := s.decodedTokenFromContext(ctx)
+
+	switch k := decodedToken.Kind.(type) {
+	case *pb.Token_Runner_:
+		// Prevent a runner token from generating a token for a runner with an
+		// ID different from the ID encoded in the runner token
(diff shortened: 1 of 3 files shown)


pkg/server/singleprocess/auth.go
```diff
@@ -533,6 +533,20 @@ func (s *Service) GenerateRunnerToken(
 		}
 	}
 
+	decodedToken := s.decodedTokenFromContext(ctx)
```
xiaolin-ninja, line 536: Can we add a test for this?

pkg/server/singleprocess/auth.go
```diff
@@ -533,6 +533,20 @@ func (s *Service) GenerateRunnerToken(
...
+
+	switch k := decodedToken.Kind.(type) {
+	case *pb.Token_Runner_:
+		// Prevent a runner token from generating a token for a runner with an
+		// ID different from the ID encoded in the runner token
+		if k.Runner.Id != req.Id {
```
xiaolin-ninja, line 542: Looks great to me! We should also check to make sure that `k.Runner.Id` would never be `nil`, and also add a test for this new method.

evanphx (approved):

PR #4708, opened 2023-05-12
Trying out the race detector in CI

izaaklauer:


.github/workflows/go-tests.yml
```diff
@@ -103,6 +103,7 @@ jobs:
             --junitfile "$TEST_RESULTS_DIR/gotestsum-report.xml" -- \
             -tags="$GOTAGS" -p 2 \
             -cover -coverprofile=coverage.txt \
+            -race \
```
izaaklauer, line 106: Critical aspect - this enables the race detector in CI!

Makefile
```diff
@@ -217,7 +217,7 @@ tools: # Install dependencies and tools required to build
 
 .PHONY: test
 test: # Run tests
-	go test ./...
+	go test ./... -race
```
catsby, line 220: Should we have `-race` be the default here, or separate that into it's own Make step? There is a potentially [non-trivial](https://go.dev/doc/articles/race_detector#Runtime_Overheads) overhead cost here, though I fully admit I don't know if it impacts us in practice. 

Looking at other projects like Vault and Consul, they break out the race test into their own make step and I wonder if we should follow suit. Since we're adding it to the CI step above, even if a developer fails to check the race test before opening a PR we should still be covered there. Thoughts?

catsby, line 220: We could also through `$(TESTARGS)` in here while we're at it, so we can easily run individual/filtered tests like the make step below

catsby (commented):
LGTM but I wonder if we shouldn't have `-race` be the default

Makefile
izaaklauer: Link to the consul separate race step for anyone following along: https://github.com/hashicorp/consul/blob/8dee3534927eb50572b003e6335f65e063b36701/.github/workflows/go-tests.yml#L258-L276

I can't think of a compelling benefit for us to break the race test into a separate stage, and run them in parallel with the regular tests. Downsides would be complexity when looking at our CI run, double the chance for flaky tests to fail, and additional GHA cost.

> Since we're adding it to the CI step above, even if a developer fails to check the race test before opening a PR we should still be covered there. Thoughts?

I don't think I follow this! I agree that with this CI check, if a developer makes a PR that includes a test with a race condition, this should catch it as part of the test suite, but i'm not sure how that relates to making the race tests their own parallel CI step.

.github/workflows/go-tests.yml
evanphx: >  cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x

evanphx: Basically, we need to be ready for CI runs to be quite a bit slower. I'd suggest perhaps not running -race to block the PR flow, maybe on merges to main?

evanphx (changes requested):

.github/workflows/go-tests.yml
izaaklauer: In practice, it looks like this brings us from ~9 mins to ~10 mins

CI test summary without -race: https://github.com/hashicorp/waypoint/actions/runs/4991989926
And with -race: https://github.com/hashicorp/waypoint/actions/runs/4994064690?pr=4708

And this isn't the slowest check! The tests currently include a dev windows build that takes 20 minutes.

The downside of running this on main only is that, if someone introduces a race bug, they won't notice until after the merge, which will add some confusion. I think the earlier we catch these race conditions, the less time consuming it will be to fix!

I can imagine us introducing tests in the future though that are much slower in race mode though, and we might not realize that we could save significant time in PR test runs by disabling race. 

With all that in mind, wdyt?

Makefile
catsby: > I can't think of a compelling benefit for us to break the race test into a separate stage, and run them in parallel with the regular tests. Downsides would be complexity when looking at our CI run

My suggestion/comment here is not regarding our CI run(s), it's just on the Makefile step `test`. To my knowledge our CI steps do not run `make test`, but I could be wrong. 

> but i'm not sure how that relates to making the race tests their own parallel CI step.

I considered making that suggestion but decided not to. The line I'm commenting on is just in our `Makefile` and not the `go test` CI step.

.github/workflows/go-tests.yml
evanphx: Ok, good justification. Sounds good to me to use it everywhere.

evanphx (approved):

.github/workflows/go-tests.yml
izaaklauer: Worth a try at least! Will plan to revert if it becomes a problem.

Makefile
izaaklauer: I feel like the implicit promise with `make test` is that "if `make test` passes locally, it should also pass in CI". 

I recognize that we're not living up to that currently (CI ignores the e2e tests and `make tests` doesn't), but I feel like the goal should be to keep it as close to in sync with the CI tests as possible. If we make the race detector tests  a separate make file step, it introduces the potential for a contributor to be confused about why the tests are passing locally and failing in CI.

catsby: > "if make test passes locally, it should also pass in CI"

Generally, yeah that makes sense, with the exception(s) of matrix tests/jobs that test different versions, but that's a technicality ... 

> If we make the race detector tests a separate make file step, it introduces the potential for a contributor to be confused about why the tests are passing locally and failing in CI.

That's fair and valid. If as you mention with Evan above the impact is minimal then I'm fine with `-race` being the default 👍

catsby (approved):
👍

PR #4710, opened 2023-05-12
Fix inconsistent errors in plugins.

teresamychu:


builtin/packer/config_sourcer.go
```diff
@@ -107,7 +107,7 @@ func (cs *ConfigSourcer) read(
 			result.Result = &pb.ConfigSource_Value_Error{
 				Error: status.New(codes.Aborted, err.Error()).Proto(),
 			}
-			return nil, err
+			continue
 		}
 		channelParams.BucketSlug = packerConfig.Bucket
 		channelParams.Slug = packerConfig.Channel
```
izaaklauer, line 113: Five lines down from here (on 118) there's another error that should be appended to result and not returned!

builtin/tfc/config_sourcer.go
```diff
@@ -356,7 +356,8 @@ func (cs *ConfigSourcer) read(
 
 			allOutputsJson, err := json.Marshal(allOutputsMap)
 			if err != nil {
-				return nil, errors.Wrapf(err, "failed marshalling all TFC outputs for org %q workspace %q to json", tfcReq.Organization, tfcReq.Workspace)
+				log.Warn("failed to marshall TFC outputs", errors.Wrapf(err, "failed marshalling all TFC outputs for org %q workspace %q to json", tfcReq.Organization, tfcReq.Workspace))
```
izaaklauer, line 359: Rather than log this , we should append it to result and continue

izaaklauer (changes requested):
A good start!

Left a few notes. It would also be great to add a description to this PR of what's changing and why, and a changelog entry.

builtin/packer/config_sourcer.go
teresamychu: ooooo good catch!

izaaklauer:
Nice! A changelog entry and a PR description and you'll be good to go!

izaaklauer (changes requested):
Nice! A changelog entry and a PR description and you'll be good to go!

PR #4711, opened 2023-05-12
Update Go Vulnerabilities.

teresamychu:


izaaklauer (commented):
The changes look good to me!

It's odd that the tests are failing with linting issues. Could one of these version updates have somehow also updated the linter, or tweaked its settings?

izaaklauer (approved):

izaaklauer (changes requested):
Ah, it looks like one of these modules controls the logic that golang-ci lint uses!

narrowing that down now. We shouldn't merge until CI passes here, or it'll block on main too.

izaaklauer:
I've done just enough legwork now to verify the fix won't be trivial.

I wrote a little script to upgrade the dependencies one-by-one to see which caused golang-ci to fail, and it's `github.com/docker/docker`. My guess is that t looks like upgrading docker brings in https://github.com/AdaLogics/go-fuzz-headers as a dependency, which is making the typecheck linter in golangci-lint reject things.

The weird thing is, I don't think the typechecker is right! We are, in fact, using the imports it's claiming we're not using. 

It's also not just the type checker - the branch doesn't build with the current set of dependencies.

I also tried rolling docker back to `go get github.com/docker/docker@v20.10.7+incompatible`, but that exposed other linting and compilation issues mostly around the k8s packages.

PR #4712, opened 2023-05-12
Do not allow forwarded IP to overwrite request source IP in logs

xiaolin-ninja:
Add `forwardedIP` section to response so that `X-Forwarded-For` request header does not overwrite actual client IP, and validate it is a proper IP address.


izaaklauer (approved):
Could use a PR description and a changelog entry, but otherwise looks great!

PR #4723, opened 2023-05-15
Remove race condition in job test

izaaklauer:
This removes a race condition where we're writing to `job` in one goroutine and reading in another during a test.

Race report: https://github.com/hashicorp/waypoint/actions/runs/4962203124/jobs/8879981994?pr=4708#step:8:387

### DIFF
--- a/pkg/serverstate/statetest/test_job.go
+++ b/pkg/serverstate/statetest/test_job.go
@@ -2681,9 +2681,11 @@ func TestJobHeartbeatOnRestart(t *testing.T, factory Factory, rf RestartFactory)
 		defer s.Close()
 
 		// Verify it exists
-		job, err = s.JobById(ctx, "A", nil)
-		require.NoError(err)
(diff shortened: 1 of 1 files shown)


catsby (approved):
👍

pkg/serverstate/statetest/test_job.go
```diff
@@ -2681,9 +2681,11 @@ func TestJobHeartbeatOnRestart(t *testing.T, factory Factory, rf RestartFactory)
 		defer s.Close()
 
 		// Verify it exists
```
catsby, line 2683:
```suggestion
		// Verify it exists. We use a closure here to avoid a possible race condition with the job object
```

nitpick: it's not immediately obvious that we use a closure here so we can make a new `job` object to avoid the race condition, so just add a comment incase someone finds it odd, or happens to copy/paste this test or parts of it for use in future tests

catsby (approved):

PR #4733, opened 2023-05-19
core/cli: Add project template crud commands to cli

demophoon:
This PR adds in the cli interface to managing project templates within Waypoint.

### DIFF
--- a/internal/cli/template_create.go
+++ b/internal/cli/template_create.go
@@ -14,13 +14,13 @@ import (
 type ProjectTemplateCreateCommand struct {
 	*baseCommand
 
-	flagSummary                 string
-	flagExpandedSummary         string
-	flagReadmeMarkdownTemplate  string
-	flagWaypointHCLTemplatePath string
-	flagTFCNoCodeModuleSource   string
-	flagTFCNoCodeModuleVersion  string
-	flagTags                    []string
+	flagSummary                    string
+	flagExpandedSummary            string
+	flagReadmeMarkdownTemplatePath string
+	flagWaypointHCLTemplatePath    string
+	flagTFCNoCodeModuleSource      string
+	flagTFCNoCodeModuleVersion     string
+	flagTags                       []string
 }
 
 func (c *ProjectTemplateCreateCommand) Run(args []string) int {
@@ -52,8 +52,8 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 		template.ExpandedSummary = c.flagExpandedSummary
 	}
 
-	if c.flagReadmeMarkdownTemplate != "" {
-		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplate)
+	if c.flagReadmeMarkdownTemplatePath != "" {
+		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplatePath)
 		if err == os.ErrNotExist {
 			rmt = []byte("")
 		} else if err != nil {
@@ -121,10 +121,10 @@ func (c *ProjectTemplateCreateCommand) Flags() *flag.Sets {
 		})
 
 		f.StringVar(&flag.StringVar{
-			Name:    "readme-markdown-template",
-			Target:  &c.flagReadmeMarkdownTemplate,
+			Name:    "readme-markdown-template-path",
+			Target:  &c.flagReadmeMarkdownTemplatePath,
 			Default: "",
-			Usage:   "Readme template for projects created from a project template",
+			Usage:   "Path to a markdown readme template for projects created from a project template",
 		})
 
 		f.StringVar(&flag.StringVar{
--- a/internal/cli/template_create.go
+++ b/internal/cli/template_create.go
@@ -54,11 +54,15 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 
 	if c.flagReadmeMarkdownTemplatePath != "" {
 		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplatePath)
-		if err == os.ErrNotExist {
-			c.ui.Output("Readme template file does not exist: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
-			return 1
-		} else if err != nil {
-			c.ui.Output("Unable to read readme.md template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+		if err != nil {
+		  // general error message
+		  errMsg := "Unable to read readme.md template file: %s"
+		  if err == os.ErrNotExist {
+			  // give specific error message if the file does not exist
+			  errMsg = "Readme template file does not exist: %s"
+			} 
+		
+			c.ui.Output(errMsg, clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
 		}
 		template.ReadmeMarkdownTemplate = rmt
--- a/internal/cli/template_create.go
+++ b/internal/cli/template_create.go
@@ -55,7 +55,8 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 	if c.flagReadmeMarkdownTemplatePath != "" {
 		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplatePath)
 		if err == os.ErrNotExist {
-			rmt = []byte("")
+			c.ui.Output("Readme template file does not exist: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
 		} else if err != nil {
 			c.ui.Output("Unable to read readme.md template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
--- a/internal/cli/template_create.go
+++ b/internal/cli/template_create.go
@@ -57,7 +57,7 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 		if err == os.ErrNotExist {
 			rmt = []byte("")
 		} else if err != nil {
-			c.ui.Output("unable to read readme.md template file. %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			c.ui.Output("Unable to read readme.md template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
 		}
 		template.ReadmeMarkdownTemplate = rmt
@@ -66,7 +66,7 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 	if c.flagWaypointHCLTemplatePath != "" {
 		wpt, err := os.ReadFile(c.flagWaypointHCLTemplatePath)
 		if err != nil {
-			c.ui.Output("unable to read waypoint.hcl template file. %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			c.ui.Output("Unable to read waypoint.hcl template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 			return 1
 		}
 		template.WaypointProject = &pb.ProjectTemplate_WaypointProject{
@@ -75,10 +75,10 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 	}
 
 	if c.flagTFCNoCodeModuleSource != "" && c.flagTFCNoCodeModuleVersion == "" {
-		c.ui.Output("Terraform no code module version required", terminal.WithErrorStyle())
+		c.ui.Output("Terraform no code module version required.\n\n"+c.Help(), terminal.WithErrorStyle())
 		return 1
 	} else if c.flagTFCNoCodeModuleSource == "" && c.flagTFCNoCodeModuleVersion != "" {
-		c.ui.Output("Terraform no code module source required", terminal.WithErrorStyle())
+		c.ui.Output("Terraform no code module source required.\n\n"+c.Help(), terminal.WithErrorStyle())
 		return 1
 	} else if c.flagTFCNoCodeModuleSource != "" && c.flagTFCNoCodeModuleVersion != "" {
 		template.TerraformNocodeModule = &pb.ProjectTemplate_TerraformNocodeModule{
@@ -93,7 +93,7 @@ func (c *ProjectTemplateCreateCommand) Run(args []string) int {
 		ProjectTemplate: &template,
 	})
 	if err != nil {
-		c.ui.Output("Error creating project template: %q", clierrors.Humanize(err), terminal.WithErrorStyle())
+		c.ui.Output("Error creating project template: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
 		return 1
 	}
 
--- a/internal/cli/template_create.go
+++ b/internal/cli/template_create.go
@@ -1,0 +1,183 @@
+package cli
+
+import (
+	"os"
+
+	"github.com/posener/complete"
+
+	"github.com/hashicorp/waypoint-plugin-sdk/terminal"
+	"github.com/hashicorp/waypoint/internal/clierrors"
+	"github.com/hashicorp/waypoint/internal/pkg/flag"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+type ProjectTemplateCreateCommand struct {
+	*baseCommand
+
+	flagSummary                 string
+	flagExpandedSummary         string
+	flagReadmeMarkdownTemplate  string
+	flagWaypointHCLTemplatePath string
+	flagTFCNoCodeModuleSource   string
+	flagTFCNoCodeModuleVersion  string
+	flagTags                    []string
+}
+
+func (c *ProjectTemplateCreateCommand) Run(args []string) int {
+	flagSet := c.Flags()
+	if err := c.Init(
+		WithArgs(args),
+		WithFlags(flagSet),
+		WithNoConfig(),
+	); err != nil {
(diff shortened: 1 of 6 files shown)


internal/cli/template_list.go
```diff
@@ -0,0 +1,95 @@
...
+
+	ptr, err := c.project.Client().ListProjectTemplates(ctx, &pb.ListProjectTemplatesRequest{})
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+	templates := ptr.ProjectTemplates
+
+	if len(templates) == 0 {
+		c.ui.Output("No project templates found.")
```
briancain, line 43: I think other CLI commands in Waypoint do this, but maybe we just output nothing in the 0 case?

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+			Id: c.flagID,
+		}
+		name = c.flagID
+	} else if c.flagName != "" {
+		tref.Ref = &pb.Ref_ProjectTemplate_Name{
+			Name: c.flagName,
+		}
+		name = c.flagName
+	} else {
+		c.ui.Output("missing project template name or id", terminal.WithErrorStyle())
```
briancain, line 63:
```suggestion
		c.ui.Output("Missing project template name or id.\n\n"+c.Help(), terminal.WithErrorStyle())
```

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+		template.Summary = c.flagSummary
+	}
+	if c.flagExpandedSummary != "" {
+		template.ExpandedSummary = c.flagExpandedSummary
+	}
+
+	if c.flagReadmeMarkdownTemplate != "" {
+		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplate)
+		if err != nil {
+			c.ui.Output("unable to read readme.md template file. %s", clierrors.Humanize(err), terminal.WithErrorStyle())
```
briancain, line 99:
```suggestion
			c.ui.Output("Unable to read readme.md template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
```

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+			c.ui.Output("unable to read waypoint.hcl template file. %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		}
+		template.WaypointProject = &pb.ProjectTemplate_WaypointProject{
+			WaypointHclTemplate: wpt,
+		}
+	}
+
+	if c.flagTFCNoCodeModuleSource != "" && c.flagTFCNoCodeModuleVersion == "" {
+		c.ui.Output("Terraform no code module version required", terminal.WithErrorStyle())
```
briancain, line 117:
```suggestion
		c.ui.Output("Terraform no code module version required.\n\n"+c.Help(), terminal.WithErrorStyle())
```

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+		return 1
+	} else if c.flagTFCNoCodeModuleSource == "" && c.flagTFCNoCodeModuleVersion != "" {
+		c.ui.Output("Terraform no code module source required", terminal.WithErrorStyle())
```
briancain, line 120:
```suggestion
		c.ui.Output("Terraform no code module source required.\n\n"+c.Help(), terminal.WithErrorStyle())
```

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	template.Tags = c.flagTags
+
+	_, err = c.project.Client().UpdateProjectTemplate(ctx, &pb.UpdateProjectTemplateRequest{
+		ProjectTemplate: template,
+	})
+	if err != nil {
+		c.ui.Output("Error updating project template: %q", clierrors.Humanize(err), terminal.WithErrorStyle())
```
briancain, line 135:
```suggestion
		c.ui.Output("Error updating project template: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
```

briancain (changes requested):
Nice job @demophoon ! Looks great. Most of my comments here are around the output UIs to make them look and feel like the rest of our CLI commands. But well done 😄

briancain (changes requested):
Nice job @demophoon ! Looks great. Most of my comments here are around the output UIs to make them look and feel like the rest of our CLI commands. But well done 😄

internal/cli/template_create.go
```diff
@@ -0,0 +1,183 @@
...
+
+		f.StringVar(&flag.StringVar{
+			Name:    "expanded-summary",
+			Target:  &c.flagExpandedSummary,
+			Default: "",
+			Usage:   "Expanded Summary for the project template",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "readme-markdown-template",
```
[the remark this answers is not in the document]
demophoon: Ah, this should be a file path, i'll update the flag to be clearer. Good catch

internal/cli/template_create.go
```diff
@@ -0,0 +1,183 @@
...
+			Target:  &c.flagWaypointHCLTemplatePath,
+			Default: "",
+			Usage:   "Path to a templated waypoint.hcl file for projects created from a project template",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "tfc-nocode-module-source",
+			Target:  &c.flagTFCNoCodeModuleSource,
+			Default: "",
+			Usage:   "The name of the Terraform no-code module that the template should use to provision infrastructure for Waypoint projects created from the template",
```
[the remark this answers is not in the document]
demophoon: This is a no-code module within Terraform Registry, i'll update to make this clearer

internal/cli/template_create.go
```diff
@@ -0,0 +1,183 @@
...
+		if err != nil {
+			c.ui.Output("Unable to read waypoint.hcl template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		}
+		template.WaypointProject = &pb.ProjectTemplate_WaypointProject{
+			WaypointHclTemplate: wpt,
+		}
+	}
+
+	if c.flagTFCNoCodeModuleSource != "" && c.flagTFCNoCodeModuleVersion == "" {
```
paladin-devops, line 77: Should we also have an `if` case here where both the module source and the No Code module version are empty?

Also, it might be worth allowing the user to save an incomplete project template, where not all of the necessary fields are added. We can chalk that up to a future consideration though, not necessarily for this PR!

paladin-devops (changes requested):
This is looking awesome @demophoon! Had a few small changes in mind, but otherwise approved!

internal/cli/template_create.go
demophoon: These might be two conflicting options. Currently a user can either specify both tfc module and version or nothing at all and both of those options would be valid. I'd argue that if a user is putting in a terraform module source, they should already know the version and vice versa.

paladin-devops (approved):

.changelog/4733.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:feature
+core/cli: Add project templating cli
```
briancain, line 2: We should probably be a bit more explicit here before merging this in, since this is something users will be reading and will be under the *features* heading. Also, this is an HCP only feature so we should mention that too! Maybe a good example to get inspiration from is our 0.10.0 release notes: https://github.com/hashicorp/waypoint/releases/tag/v0.10.0

briancain (approved):

.changelog/4733.txt
briancain: Actually, this seems to be a beta feature, not ready for full usage, so let's remove the changelog for now

briancain (approved):

internal/cli/template_create.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if c.flagReadmeMarkdownTemplatePath != "" {
+		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplatePath)
+		if err == os.ErrNotExist {
+			c.ui.Output("Readme template file does not exist: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		} else if err != nil {
+			c.ui.Output("Unable to read readme.md template file: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+			return 1
+		}
```
catsby, line 63: nit: because each of these errors ends in `return` we should decouple them (no need for `else` or `else if`). This generally improves readability and is more "idiomatic":

From [Control Structures](https://go.dev/doc/effective_go#control-structures)
> This is an example of a common situation where code must guard against a sequence of error conditions. The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements, the resulting code needs no else statements.

catsby: Optional refactor suggestion: 

```suggestion
		if err != nil {
		  // general error message
		  errMsg := "Unable to read readme.md template file: %s"
		  if err == os.ErrNotExist {
			  // give specific error message if the file does not exist
			  errMsg = "Readme template file does not exist: %s"
			} 
		
			c.ui.Output(errMsg, clierrors.Humanize(err), terminal.WithErrorStyle())
			return 1
		}
```

internal/cli/template_delete.go
```diff
@@ -0,0 +1,116 @@
...
+	return 0
+}
+
+func (c *ProjectTemplateDeleteCommand) Flags() *flag.Sets {
+	return c.flagSet(0, func(sets *flag.Sets) {
+		f := sets.NewSet("Command Options")
+
+		f.StringVar(&flag.StringVar{
+			Name:    "name",
+			Target:  &c.flagName,
```
catsby, line 75: It seems strange to me that `name` is both a flag and an argument. It's also not clear to the user which would take precedence (it looks like the `name` argument does), same with `-id`

```
$ waypoint template delete -name=sonic -id=the hedgehog
```

internal/cli/template_inspect.go
```diff
@@ -0,0 +1,165 @@
...
+
+		f.BoolVar(&flag.BoolVar{
+			Name:   "json",
+			Target: &c.flagJson,
+			Usage:  "Output project information as JSON.",
+		})
+
+		f.StringVar(&flag.StringVar{
+			Name:    "name",
+			Target:  &c.flagName,
```
catsby, line 131: same comment about flag and arg for `name`

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+		name = c.flagID
+	} else if c.flagName != "" {
+		tref.Ref = &pb.Ref_ProjectTemplate_Name{
+			Name: c.flagName,
+		}
+		name = c.flagName
+	} else {
+		c.ui.Output("Missing project template name or id.\n\n"+c.Help(), terminal.WithErrorStyle())
+		return 1
+	}
```
catsby, line 65: The `tref` that gets populated with the results of this block is never used, we just always use the `name` value down on line 70

internal/cli/template_create.go
```diff
@@ -0,0 +1,184 @@
...
+		template.ReadmeMarkdownTemplate = rmt
+	}
+
+	if c.flagWaypointHCLTemplatePath != "" {
+		wpt, err := os.ReadFile(c.flagWaypointHCLTemplatePath)
```
catsby, line 72: We check for `os.ErrNotExist` when reading the template path but we don't distinguish here

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+	template.Name = name
+	if c.flagSummary != "" {
+		template.Summary = c.flagSummary
+	}
+	if c.flagExpandedSummary != "" {
+		template.ExpandedSummary = c.flagExpandedSummary
+	}
+
+	if c.flagReadmeMarkdownTemplatePath != "" {
+		rmt, err := os.ReadFile(c.flagReadmeMarkdownTemplatePath)
```
catsby, line 97: In `Create` we check for `os.ErrNotExist` but we don't distinguish here or in the hcltemplate path check below

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+			WaypointHclTemplate: wpt,
+		}
+	}
+
+	if c.flagTFCNoCodeModuleSource != "" && c.flagTFCNoCodeModuleVersion == "" {
+		c.ui.Output("Terraform no code module version required.\n\n"+c.Help(), terminal.WithErrorStyle())
+		return 1
+	} else if c.flagTFCNoCodeModuleSource == "" && c.flagTFCNoCodeModuleVersion != "" {
+		c.ui.Output("Terraform no code module source required.\n\n"+c.Help(), terminal.WithErrorStyle())
+		return 1
```
catsby, line 121: nit: we can/should break these `if` blocks apart, since then end in `return`

internal/cli/template_update.go
```diff
@@ -0,0 +1,233 @@
...
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+	if checkResp.ProjectTemplate == nil {
+		c.ui.Output(
+			"Project template %q does not exist", checkResp.ProjectTemplate.Name,
+			terminal.WithErrorStyle(),
+		)
+		return 1
+	}
```
catsby, line 84: Will this block ever trigger? From what I see if the template doesn't exist the service returns `codes.NotFound` error which would fall under the above error check. Should we move this check or specific error message under the error check above and include it if we hit `NotFound`?

```suggestion
	if err != nil {
	  errMsg := clierrors.Humanize(err)
	  if status.Code(err) == codes.NotFound || checkResp.ProjectTemplate == nil {
			errMsg = fmt.Sprintf("Project template %q does not exist", checkResp.ProjectTemplate.Name),
	  }
		c.ui.Output(errMsg, terminal.WithErrorStyle())
		return 1
	}
```

internal/cli/template_delete.go
```diff
@@ -0,0 +1,116 @@
...
+
+	_, err := c.project.Client().DeleteProjectTemplate(ctx, &pb.DeleteProjectTemplateRequest{
+		ProjectTemplate: &tref,
+	})
+	if err != nil {
+		c.ui.Output("Encountered an error while deleting the project template: %s", clierrors.Humanize(err), terminal.WithErrorStyle())
+		return 1
+	}
+
+	c.ui.Output("Template %q deleted", name, terminal.WithSuccessStyle())
```
catsby, line 64: nit: both `create` and `update` report success with a lowercase `t` for template

```suggestion
	c.ui.Output("template %q deleted", name, terminal.WithSuccessStyle())
```

internal/cli/template_inspect.go
```diff
@@ -0,0 +1,165 @@
...
+	); err != nil {
+		return 1
+	}
+	args = flagSet.Args()
+	ctx := c.Ctx
+
+	name := ""
+	if len(args) == 1 {
+		name = args[0]
+	}
```
catsby, line 40: Should we error if length of args is more than 1? I _think_ that's common with our other commands, but I didn't check them all

internal/cli/template_inspect.go
```diff
@@ -0,0 +1,165 @@
...
+	} else {
+		c.ui.Output("Missing project template name or id.\n\n"+c.Help(), terminal.WithErrorStyle())
+		return 1
+	}
+
+	tr, err := c.project.Client().GetProjectTemplate(ctx, &pb.GetProjectTemplateRequest{
+		ProjectTemplate: &tref,
+	})
+	if err != nil {
+		c.ui.Output(clierrors.Humanize(err), terminal.WithErrorStyle())
```
catsby, line 70: we don't do a `nil` or `not found` style check here like we do in other `Update`, do you think we need to here?

catsby (changes requested):
Mostly nit pick things but there is at least one or two things I would like to see addressed, namely the unused `tref` in `Update`

internal/cli/template_delete.go
demophoon: This is a fair point, I think it makes sense to remove the name flag and just use it as an argument. Throwing an error if both an argument and the `-id` flag are specified. That should clear up precedent ambiguity while removing unnecessary flags.

internal/cli/template_update.go
demophoon: Good catch, updated to match behavior with Create in `37e4c8269`

internal/cli/template_inspect.go
demophoon: Resolved in `8329af09e`.

internal/cli/template_inspect.go
demophoon: Resolved in `8329af09e`.

internal/cli/template_update.go
demophoon: A good catch, looks like we do indeed return a `NotFound` error here. Committing your suggested changes

internal/cli/template_inspect.go
demophoon: Updated to match `Update`

demophoon:
@catsby Thanks for the feedback, I've made changes and resolved all of your comments

catsby (approved):
Thanks for everything in this PR, and thanks for making those changes. 

Godspeed, friend @demophoon 🫡

PR #4081, opened 2022-10-21
plugin/helm: Fix helm namespaces

demophoon:
Fixes #3065

PR #4737, opened 2023-05-23
plugin/helm: use proper namespaces

kmcgovern-apixio:
Using #4081 as a base i've updated the PR and done some local testing with runner installs. Before i would get weird behavior where things like the helm config would get put in default namesapce but report as the "waypoint" namespace (for example). This PR from my testing seems to fix these issues.


builtin/k8s/helm/platform.go
```diff
@@ -127,6 +127,8 @@ func (p *Platform) Deploy(
 		client.Description = ""
 		client.CreateNamespace = p.config.CreateNamespace
 
+		log.Info("install client: %#v", client)
```
xiaolin-ninja, line 130:
```suggestion
		log.Info("installing client: %#v", client)
```

internal/installutil/helm/helm.go
```diff
@@ -85,16 +85,23 @@ func (k *restClientGetter) ToRawKubeConfigLoader() clientcmd.ClientConfig {
...
 }
 
-func ActionInit(log hclog.Logger, kubeConfigPath string, context string) (*action.Configuration, error) {
+func ActionInit(log hclog.Logger, kubeConfigPath string, context string, namespace string) (*action.Configuration, error) {
 	// Get our K8S API
 	_, ns, rc, err := k8s.Clientset(kubeConfigPath, context)
 	if err != nil {
 		return nil, err
 	}
+	if (namespace != "") {
```
xiaolin-ninja, line 102:
```suggestion
	if namespace != "" {
```

xiaolin-ninja (commented):
Thank you for this PR, this looks great! I left a few comments for grammar.

builtin/k8s/helm/platform.go
izaaklauer: This feels debug level to me, not info. I'd also recommend that you pick the specific fields on `client` that you're interested in and log those, to avoid logging irrelevant data (or potentially logging a secret)

Lastly, the logger takes a sequence of key value pairs, not a string with args that are interpolated. Example: https://github.com/hashicorp/go-hclog#emit-an-info-level-message-with-2-keyvalue-pairs

xiaolin-ninja (approved):

demophoon (approved):

PR #4747, opened 2023-05-26
Add “root pipeline” fields to jobs

jgwhite:
## Why the change?

Jobs have a field called `pipeline` of type `Ref.PipelineStep`:

```protobuf
message Job {
  // ...
  Ref.PipelineStep pipeline = 117;
  // ...
}
```

Currently the `Ref.PipelineStep` looks like this:

```protobuf
message PipelineStep {
  string pipeline_id = 1;
  string pipeline_name = 2;
  string step = 3;
  uint64 run_sequence = 4;
}
```

This field allows us to cross-reference a job with a particular step of a particular pipeline run, but does not accommodate nested/referenced pipeline invocations, in which case the step will belong to a different pipeline than the run.

To illustrate, imagine we have this `waypoint.hcl`:

```hcl
pipeline "foo" {
  step "invoke" {
    use "pipeline" {
      project = "my-project"
      name    = "bar"
    }
  }
}

pipeline "bar" {
  step "echo" {
    use "exec" {
      // ...
    }
  }
}
```

When we `waypoint pipeline run foo`, Waypoint schedules a job like this for the `echo` step:

```go
pb.Job{
  // ...
  Pipeline: pb.PipelineStep{
    PipelineId:   "bar",
    PipelineName: "bar",
    Step:         "echo",
    RunSequence:  "1",
  }
}
```

This allows us to relate this job to the `echo` step of the `bar` pipeline. However, if we try to look up run `1` of the `bar` pipeline we will find either nothing or, worse, a different pipeline run.

This change aims to address this by adding additional “root pipeline” fields to `Ref.PipelineStep`:

```diff
  message PipelineStep {
    string pipeline_id = 1;
    string pipeline_name = 2;
    string step = 3;
    uint64 run_sequence = 4;
+   string root_pipeline_id = 5;
+   string root_pipeline_name = 6;
  }
```

Thus, our scheduled job for the echo step looks like this:

```go
pb.Job{
  // ...
  Pipeline: pb.PipelineStep{
    PipelineId:       "bar",
    PipelineName:     "bar",
    Step:             "echo",

    RootPipelineId:   "foo",
    RootPipelineName: "foo",
    RunSequence:      1,
  }
}
```

I’ve grouped these fields to clarify how they relate to one another.

## What’s the plan?

- [x] Introduce “root pipeline” fields
- [x] Populated “root pipeline” fields during `RunPipeline`
- [x] Use “root pipeline” fields in `UI_PipelineRunTree` construction

I’ve not made any changes to the CLI or runner code for this PR. In principle, though, we could use this new information to clarify commands such as `job inspect`.

## How do I test it?

1. Run a pipeline
2. Inspect the job for a nested step with the `-json` flag
3. Verify the `pipeline` data looks correct

For example, when I run [this pipeline](https://github.com/jgwhite/waypoint-demos/blob/a8925b436fadb7ad1f28750a0da8ff7175ba78e6/pipelines/waypoint.hcl#L50-L64) then I see the following:

```sh
$ waypoint pipeline run three-step-referenced
$ waypoint job list
# Find the job for the `bye` step...
$ waypoint job inspect -json 01H1RRTQ2ZK1NF1RFGPKNK5WTD
{
  "id":  "01H1RRTQ2ZK1NF1RFGPKNK5WTD",
  // ...snip...
  "pipeline":  {
    "pipelineId":  "01H1RRT6EQP3ZFJG53B77EH6B0",
    "pipelineName":  "two-step",
    "step":  "bye",
    "runSequence":  "1",
    "rootPipelineId":  "01H1RRT6F0R8NA3CA09HQKWQ9H",
    "rootPipelineName":  "three-step-referenced"
  }
}
```


pkg/server/gen/server.swagger.json
```diff
@@ -721,33 +721,47 @@
...
-            "description": "Pipeline run sequence.",
+            "description": "Run sequence for the root pipeline.",
             "in": "query",
             "required": false,
             "type": "string",
             "format": "uint64"
           },
+          {
+            "name": "pipeline.root_pipeline_id",
+            "description": "Name of the root pipeline (the one invoked in `waypoint pipeline run <pipeline>`).",
```
izaaklauer, line 753: Great descriptions

izaaklauer (approved):

briancain (approved):
Thanks for this! I don't think there are any consequences of changing this behavior 🤔 I don't think anything was relying on this yet, beyond the UI, thus the change. So should be safe!

PR #4748, opened 2023-05-26
Add `gateway.yml` mapping for `GetProjectTemplate` by name

jgwhite:


gregone:
LGTM code-wise. That said, I wonder if the same usage could not be fulfilled with a call to `ListProjectTemplates` with a `name` query param filter?

gregone (approved):

PR #4754, opened 2023-05-31
server,cli: Add explicit `delete` RPC and CLI endpoints for `Config` and `ConfigSourcers` + tests

briancain:
This commit adds an explicit delete endpoint for both `Config` and `ConfigSourcer`s. It introduces the RPC endpoints for each, and additionally adds CLI commands for each. It's built on top of the "Unset" flow that already existed to minimize the footprint of this addition to mostly be a front-end change in the API layer.


evanphx (approved):

briancain:
Fixed a merge conflict in the server proto gen file.

PR #4761, opened 2023-06-02
api: fix duplicate gateway.yml paths

jgwhite:
Turns out this doesn’t work:

```yaml
  - selector: hashicorp.waypoint.Waypoint.SetConfig
    put: /project/config
    body: "*"

  - selector: hashicorp.waypoint.Waypoint.DeleteConfig
    put: /project/config
    body: "*"
```

Because both requests are for `PUT /project/config` and as a result grpc-gateway doesn’t know how to direct the request. protoc-grpc-gateway fails with the following:

```
go generate ./pkg/server
--swagger_out: must not set request body when http method is DELETE except allow_delete_body option is true: DeleteConfig
pkg/server/server.go:6: running "sh": exit status 1
make: *** [gen/server] Error 1
```

This change disambiguates the paths, but at the cost of a rather unintuitive path for the Delete rpcs. We can’t use `delete: /project/config` because grpc-gateway does not permit DELETE request to have a body.

### DIFF
--- a/pkg/server/gen/server.pb.gw.go
+++ b/pkg/server/gen/server.pb.gw.go
@@ -4958,6 +4958,40 @@ func local_request_Waypoint_SetConfig_0(ctx context.Context, marshaler runtime.M
 
 }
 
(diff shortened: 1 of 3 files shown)


jgwhite:
We should probably have a CI step that runs `make gen/server` just to check it still works.

paladin-devops (approved):

PR #4762, opened 2023-06-02
Return pipeline run bundle from `UI_GetPipelineRun`

jgwhite:
## Why the change?

This was an oversight on my part. We need the extra info for the UI.

## What’s the plan?

- [x] Update protos
- [x] Update tests
- [x] Fix tests

## What does it look like?

Given I have run [this pipeline](https://github.com/jgwhite/waypoint-demos/blob/a8925b436fadb7ad1f28750a0da8ff7175ba78e6/pipelines/waypoint.hcl#L23-L31).

When I run:

```sh
$ ./contrib/waypoint-grpc/waypoint-grpc.sh UI_GetPipelineRun '
{
  "pipeline": {
    "owner": {
      "project": {
        "project": "pipelines"
      },
      "pipeline_name": "one-step"
    }
  },
  "sequence": 1
}
'
```

Then I see this:

```json
{
  "pipelineRunBundle": {
    "pipelineRun": {
      "id": "01H2AQSAABY33HNR8MS32V6JW5",
      "sequence": "1",
      "pipeline": {
        "id": "01H2AQRY5WCMXTP17DP7R9DPE5"
      },
      "jobs": [
        {
          "id": "01H2AQSAACRZGX35JGP5BSJD4V"
        }
      ],
      "state": "SUCCESS"
    },
    "queueTime": "2023-06-07T10:52:35.021352377Z",
    "application": {
      "application": "web",
      "project": "pipelines"
    },
    "dataSourceRef": {
      "git": {
        "commit": "a8925b436fadb7ad1f28750a0da8ff7175ba78e6",
        "timestamp": "2023-05-23T21:14:53Z",
        "commitMessage": "Add b\n"
      }
    }
  },
  "rootTreeNode": {
    // ...snip...
  }
}
```

## How do I test it?

Essentially follow the steps above, in other words:

1. Run a pipeline
2. Use `waypoint-grpc` to call `UI_GetPipelineRun`
3. Verify `pipelineRunBundle` looks correct


pkg/server/proto/server.proto
```diff
@@ -592,8 +592,10 @@ message UI {
   }
 
   message GetPipelineRunResponse {
+    // deprecated in favor of pipeline_run_bundle.
     PipelineRun pipeline_run = 1;
```
jgwhite, line 596: **Why no `deprecated_` prefix?**

Though changing the name of a field is backwards-compatible in end-to-end protobuf, it’s not backwards-compatible with Swagger and our generated client library.

andrew-hashicorp (approved):

andrew-hashicorp:
Looks good!

pkg/server/proto/server.proto
jgwhite: > What, really? That's surprising. That kind of breaks the one reason why protobufs are cool and useful for crafting APIs. It should be safe to rename fields in protobufs as long as the number stays the same. If we lose that, then that's a shame 😞

Absolutely. All is good when you stay within gRPC. But unfortunately because we go via Swagger and the OpenAPI generator, we lose that delightful property of protobufs.

> Also if we're deprecating this, then we're gonna need to update everywhere in the system that uses this field. I know much of the CLI is relying on this field existing.

Oh, is the CLI using `UI_GetPipelineRun`?

briancain: Ohhh shoot, I think the git diff wasn't showing me the full picture, but isn't this modifying the GetPipelineRunResponse?

briancain: Ahhh, looks like this is within UI? So maybe not, all good I bet!

jgwhite: Yeah, should just be `UI.GetPipelineRunResponse`.

PR #4763, opened 2023-06-02
Azure Plugin auth fix

### LINKED ISSUE #2589: Azure authentication may fail to attempt CLI auth
**Describe the bug**
Reported initially by https://discuss.hashicorp.com/t/azure-container-instance-deployment-is-failing/31064, waypoint reports the following error when using CLI auth:

```
» Deploying example-nodejs...
! Unable to create subscriptions client: MSI not available
```

From my reading of our authenticate function below, it looks like we try Environment auth, and if it times out, we try CLI auth. 

https://github.com/hashicorp/waypoint/blob/c55a8698c63897cad519e68d7c887eea88b38250/builtin/azure/aci/deployment.go#L27-L64

It looks like we should be logging the `Unable to create subscriptions client` error rather than returning, and continue on to try CLI auth.

paladin-devops:
Update the Azure ACI plugin to attempt CLI auth before erroring, rather than erroring if environment auth failed. Closes #2589.

### DIFF
--- a/.changelog/4763.txt
+++ b/.changelog/4763.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+plugin/azure-aci: Update plugin to attempt CLI auth if environment auth fails.
+```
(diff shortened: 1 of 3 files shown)


briancain (approved):
Looks like this should resolve the issue 👍🏻

evanphx (approved):

PR #4764, opened 2023-06-02
Validate CEB token context DeploymentID

xiaolin-ninja:
Validate the deployment ID associated with the context of the CEB token to ensure that the token is authorized to take actions on the current deployment.

### DIFF
--- a/pkg/server/singleprocess/service_entrypoint.go
+++ b/pkg/server/singleprocess/service_entrypoint.go
@@ -51,6 +51,17 @@ func (s *Service) EntrypointConfig(
 		)
 	}
 
+	if tok := s.decodedTokenFromContext(ctx); tok != nil {
+		if tok.UnusedEntrypoint.DeploymentId != req.DeploymentId {
+			return hcerr.Externalize(
+				log,
+				err,
+				"entrypoint token invalid for this deployment ID", "deployment_id",
+				req.DeploymentId,
+			)
+		}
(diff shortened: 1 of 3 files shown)


pkg/server/singleprocess/service_entrypoint.go
```diff
@@ -26,14 +26,20 @@ import (
...
 func (s *Service) EntrypointConfig(
 	req *pb.EntrypointConfigRequest,
 	srv pb.Waypoint_EntrypointConfigServer,
 ) error {
 	log := hclog.FromContext(srv.Context())
 	ctx := srv.Context()
 
+	// Validate CEB token is valid for this deployment
+	if tok := s.decodedTokenFromContext(ctx); tok != nil {
+		if tok.UnusedEntrypoint != nil && tok.UnusedEntrypoint.DeploymentId != req.DeploymentId {
```
evanphx, line 38: I believe that the `UnusedEntrypoint` field isn't being populated, so I don't think this will ever fire.

evanphx (changes requested):

pkg/server/singleprocess/service_entrypoint.go
evanphx: Oh lol, you're totally right it is. Maybe update the comment over there that it's being used for authz.

evanphx (approved):

PR #4767, opened 2023-06-05
Set reasonable defaults for ECS plugin health check timeout

paladin-devops:
Prior to this PR, depending on the type of health check configured, the default timeout could have been set to be longer than the default interval (the defaults are set differently by AWS depending on how you configure the health check protocol), which is not a valid setting in AWS. This PR sets the default timeout to be lower than the interval - 5 second timeout and 30 second interval.

### DIFF
--- a/builtin/aws/ecs/platform.go
+++ b/builtin/aws/ecs/platform.go
@@ -1349,6 +1349,10 @@ func (p *Platform) resourceTargetGroupCreate(
 			createTargetGroupInput.HealthCheckIntervalSeconds = aws.Int64(p.config.HealthCheck.Interval)
 		}
 
+		if *createTargetGroupInput.HealthCheckIntervalSeconds < *createTargetGroupInput.HealthCheckTimeoutSeconds {
+			return status.Errorf(codes.InvalidArgument, fmt.Sprintf("Health check interval cannot be shorter than the timeout"))
+		}
+
 		if p.config.HealthCheck.HealthyThresholdCount != 0 {
 			createTargetGroupInput.HealthyThresholdCount = aws.Int64(p.config.HealthCheck.HealthyThresholdCount)
 		} else {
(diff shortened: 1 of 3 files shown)


builtin/aws/ecs/platform.go
```diff
@@ -1339,10 +1339,12 @@ func (p *Platform) resourceTargetGroupCreate(
 			createTargetGroupInput.HealthCheckPath = aws.String(p.config.HealthCheck.Path)
 		}
 
+		createTargetGroupInput.HealthCheckTimeoutSeconds = aws.Int64(5)
 		if p.config.HealthCheck.Timeout != 0 {
 			createTargetGroupInput.HealthCheckTimeoutSeconds = aws.Int64(p.config.HealthCheck.Timeout)
 		}
 
+		createTargetGroupInput.HealthCheckIntervalSeconds = aws.Int64(30)
```
catsby, line 1347: Should we verify `HealthCheckTimeoutSeconds` is lower than `HealthCheckIntervalSeconds` or do we just let AWS error there?

catsby (approved):
LGTM just a question over if we should check the values prior to sending to AWS

builtin/aws/ecs/platform.go
paladin-devops: This part of the `ConfigSet` function already handles that for us! The reason why this PR was still needed though is because `ConfigSet` will only check what the user sets in the `waypoint.hcl` - it wasn't checking the values that the AWS Go SDK was setting by default when one wasn't supplied.

https://github.com/hashicorp/waypoint/blob/8acf8277bbaef839f7d6e020d6acbf50c84c3faf/builtin/aws/ecs/platform.go#L134-L139

catsby: 👍

builtin/aws/ecs/platform.go
```diff
@@ -3215,6 +3221,9 @@ deploy {
 
 	doc.SetField("health_check",
 		"Health check settings for the app.",
+		docs.Summary("These settings configure a health check for the application"+
```
briancain, line 3224:
```suggestion
		docs.Summary("These settings configure a health check for the application "+
```

briancain (approved):

builtin/aws/ecs/platform.go
briancain: I think these words will get smooshed together without this space.

paladin-devops: Fixed! 😄

PR #4792, opened 2023-06-08
AWS ECS Runner Uninstall Fix

paladin-devops:
This PR fixes an error which sometimes occurred during a `runner uninstall` for AWS ECS runners:

```terminal
❌ Uninstalling runner...
✓ Finding runner in ECS services...
! InvalidParameter: 1 validation error(s) found.
```

This occurred when no AWS EFS file systems were detected which had a tag `waypoint-runner`, with the value of the runner's ID assigned to it. This PR fixes that by properly looping through all paginated results of AWS EFS file systems to find a match, and later use that match correctly during deletion. If no matching file systems are found, `runner uninstall` will skip attempting to delete it.

There was also a possible resource leak with a `defer` inside of a `for` loop, which has been fixed.

Successful output looks like:

```shell
$ waypoint runner uninstall -platform=ecs -id=ABC123 -ecs-cluster=dev -ecs-region=us-east-1
✓ Runner "ABC123" uninstalled successfully
✓ Runner "ABC123" forgotten on server
✓ Waypoint runner AWS ECS service deleted
✓ Runner file system deleted
```

### DIFF
--- a/internal/runnerinstall/ecs.go
+++ b/internal/runnerinstall/ecs.go
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
 package runnerinstall
 
 import (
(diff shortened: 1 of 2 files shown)


internal/runnerinstall/ecs.go
```diff
@@ -359,89 +356,110 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
 		if err != nil {
 			return err
 		}
-		if marker == *fileSystemsResp.Marker {
+
+		if fileSystemsResp.NextMarker == nil {
 			done = true
+		} else {
+			marker = *fileSystemsResp.NextMarker
+			time.Sleep(5 * time.Second)
```
catsby, line 386: We probably don't need the sleep here since we are not waiting for an installation or otherwise status change

internal/runnerinstall/ecs.go
```diff
@@ -359,89 +356,110 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
 	// uninstall fails
+	s = sg.Add("Deleting runner file system")
 	efsSvc := efs.New(sess)
 	marker := ""
 	done := false
 	var fileSystems []*efs.FileSystemDescription
 	for !done {
-		fileSystemsResp, err := efsSvc.DescribeFileSystems(&efs.DescribeFileSystemsInput{
-			Marker:   aws.String(marker),
+		req := &efs.DescribeFileSystemsInput{
```
catsby, line 371: We could possibly reduce this function by using [DescribeFileSystemsPages](https://docs.aws.amazon.com/sdk-for-go/api/service/efs/#EFS.DescribeFileSystemsPages) to handle the pagination logic for us, so something like this:

```go
var fileSystems []*efs.FileSystemDescription
err := client.DescribeFileSystemsPages(
  &efs.DescribeFileSystemsInput{}, // don't need to set marker. MaxItems also defaults to 100
  func(page *efs.DescribeFileSystemsOutput, lastPage bool) bool {
		fileSystems = append(fileSystems, page.FileSystems...)
		
    // double check that lastPage is the API telling us it's the last page though
    return lastPage 
})
```

catsby (approved):
Some nitpicks but nothing major that I see yet

internal/runnerinstall/ecs.go
```diff
@@ -1,6 +1,3 @@
-// Copyright (c) HashiCorp, Inc.
-// SPDX-License-Identifier: MPL-2.0
-
```
catsby, line 3: I think we need these

catsby (changes requested):
The license header needs to be restored

internal/runnerinstall/ecs.go
paladin-devops: I added this sleep because during testing, I triggered the API's rate limiting. That was before I had this loop working correctly, though so it _shouldn't_ still happen. I will go ahead and remove this.

internal/runnerinstall/ecs.go
```diff
@@ -359,89 +356,110 @@ func (i *ECSRunnerInstaller) Uninstall(ctx context.Context, opts *InstallOpts) e
...
 
-	var fileSystemId *string
-	for _, fileSystem := range fileSystems {
-		// Check if tags match ID, if so then delete things
-		for _, tag := range fileSystem.Tags {
-			if *tag.Key == "runner-id" && *tag.Value == opts.Id {
-				fileSystemId = fileSystem.FileSystemId
-				goto DeleteFileSystem
+	if len(fileSystems) == 0 {
+		s.Update("No file systems detected, skipping deletion")
```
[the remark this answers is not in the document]
paladin-devops: I agree with you here, will update this!

To add a bit of context as to why I did it this way, initially I had both cases of this if statement hitting the same `step.Done()` at the end of this function, just before the return. I moved that inside the `else` case, so the step is completed before the `return nil` is hit when the file system gets deleted, but forgot to apply the same up here.

internal/runnerinstall/ecs.go
paladin-devops: Not sure how I managed to remove those - added back.

internal/runnerinstall/ecs.go
```diff
@@ -411,19 +420,16 @@ DeleteFileSystem:
 	describeMountTargetsResp, err := efsSvc.DescribeMountTargets(&efs.DescribeMountTargetsInput{
 		FileSystemId: fileSystemId,
 	})
-	if err != nil {
-		return err
-	}
```
catsby, line 416: Do we not need to check this error?

catsby (approved):
LGTM thanks for making those changes. Just a question remaining about the removed err check other than that ✅

internal/runnerinstall/ecs.go
paladin-devops: We do! I lost this while moving other things around, but fixed up!

PR #4798, opened 2023-06-12
Nomad runner install CPU and memory flags

paladin-devops:
The Nomad runner install configuration has fields for CPU and memory, but no flags for the user to set them - this PR adds those flags. This fix stems from a [discussion on HashiCorp Discuss](https://discuss.hashicorp.com/t/hcp-waypoint-unable-to-install-runner-on-nomad-client/54759/8), where a user installing a runner had insufficient resources on their Nomad client to run the runner, and I realized I could not suggest setting lower memory or CPU on the runner because there were no CLI flags to support it.

### DIFF
--- a/internal/runnerinstall/nomad.go
+++ b/internal/runnerinstall/nomad.go
@@ -250,6 +250,18 @@ func (i *NomadRunnerInstaller) InstallFlags(set *flag.Set) {
 	})
 
 	set.StringVar(&flag.StringVar{
+		Name:   "nomad-runner-cpu",
+		Target: &i.Config.RunnerResourcesCPU,
+		Usage:  "CPU required to run the runner task in MHz.",
+	})
+
+	set.StringVar(&flag.StringVar{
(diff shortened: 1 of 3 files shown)


internal/runnerinstall/nomad.go
```diff
@@ -249,6 +249,18 @@ func (i *NomadRunnerInstaller) InstallFlags(set *flag.Set) {
...
 
+	set.StringVar(&flag.StringVar{
+		Name:   "nomad-runner-cpu",
+		Target: &i.Config.RunnerResourcesCPU,
+		Usage:  "CPU required to run the runner task in MHz.",
+	})
+
+	set.StringVar(&flag.StringVar{
+		Name:   "nomad-runner-memory",
+		Target: &i.Config.RunnerResourcesMemory,
```
briancain, line 260: Weird, did we accidentally delete these install flags or something? Looks like the target var already existed.

briancain (approved):

PR #4815, opened 2023-06-20
RPC Validators for project templates

izaaklauer:
This adds validators for the project-template related RPC requests.

Note that these aren't used here, because project templates aren't implemented in open-source.


pkg/server/ptypes/project_template.go
```diff
@@ -0,0 +1,117 @@
...
+const (
+	PROJECT_TEMPLATE_ID_LENGTH               = 64
+	PROJECT_TEMPLATE_NAME_LENGTH             = 64
+	PROJECT_TEMPLATE_TAG_LENGTH              = 64
+	PROJECT_TEMPLATE_SUMMARY_LENGTH          = 64
+	PROJECT_TEMPLATE_EXPANDED_SUMMARY_LENGTH = 512
+	PROJECT_TEMPLATE_README_LENGTH           = 1024 ^ 2
+	PROJECT_TEMPLATE_WAYPOINT_HCL_LENGTH     = 1024 ^ 2
+
+	TERRAFORM_NOCODE_MODULE_SOURCE_LENGTH  = 4096
```
gregone, line 22: 😮 wow that seems pretty high

pkg/server/ptypes/project_template.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/hashicorp/waypoint/internal/pkg/validationext"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
+)
+
+const (
+	PROJECT_TEMPLATE_ID_LENGTH               = 64
+	PROJECT_TEMPLATE_NAME_LENGTH             = 64
+	PROJECT_TEMPLATE_TAG_LENGTH              = 64
+	PROJECT_TEMPLATE_SUMMARY_LENGTH          = 64
```
gregone, line 17: The initial design number was 110. Maybe this is too short?

pkg/server/ptypes/project_template.go
izaaklauer: Good question!

TFC gets the name from github, which empirically has a 100 character limit:

Screenshot 2023-06-20 at 5 45 02 PM

izaaklauer: Updated to be bigger. I'm setting it above github's limit, because our main concern is around data storage, and if github slightly increases their limit in the future i'd rather avoid having to change it on our side.

pkg/server/ptypes/project_template.go
izaaklauer: Oh this is a good call - if anything, it'll be longer post base64 encoding!

paladin-devops (approved):

pkg/server/ptypes/project_template_test.go
```diff
@@ -0,0 +1,153 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := ValidateCreateProjectTemplateRequest(tt.req)
+			if err != nil {
+				if !tt.wantErr {
+					t.Errorf("ValidateCreateProjectTemplateRequest() error = %v, wantErr %v", err, tt.wantErr)
+				}
+			}
```
catsby, line 49:
```suggestion
			if err == nil && tt.wantErr {
				t.Errorf("expected error in ValidateCreateProjectTemplateRequest() but got none")
			}

			if err != nil && !tt.wantErr {
				t.Errorf("ValidateCreateProjectTemplateRequest() error = %v, wantErr %v", err, tt.wantErr)
			}

```

catsby: We should check `tt.wantErr` even if `err` is `nil`, or we miss situations where the error is nil but we expected one. 

For example, I cloned locally and added this test expecting it to fail:

```
{
        name: "enforces name to not be empty",
        req: &pb.CreateProjectTemplateRequest{
                ProjectTemplate: &pb.ProjectTemplate{
                        Name: " "},
        },
        wantErr: true,
},
```

However not only does that test case pass, I found that if I changed `wantErr` to `false` it still passes... I'll add that as an optional change. I'm not sure if we should consider that an error. I feel like we should, but I'm not sure if we guard against "empty" strings like that elsewhere

pkg/server/ptypes/project_template_test.go
```diff
@@ -0,0 +1,153 @@
...
+		},
+		{
+			name: "Inherits base validator rules (example: name length)",
+			req: &pb.CreateProjectTemplateRequest{
+				ProjectTemplate: &pb.ProjectTemplate{
+					Name: string(make([]byte, PROJECT_TEMPLATE_NAME_LENGTH+1)),
+				},
+			},
+			wantErr: true,
+		},
```
catsby, line 40:
```suggestion
		},
		{
			name: "enforces name to not be empty",
			req: &pb.CreateProjectTemplateRequest{
				ProjectTemplate: &pb.ProjectTemplate{
					Name: " "},
			},
			wantErr: true,
		},

```

Maybe names shouldn't be non-zero length empty strings? I mention this in another comment, it seems like this should error but I don't know if we guard against this anywhere else ¯\_(ツ)_/¯

pkg/server/ptypes/project_template_test.go
```diff
@@ -0,0 +1,153 @@
...
+					Id: string(make([]byte, PROJECT_TEMPLATE_ID_LENGTH+1)),
+				},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := ValidateUpdateProjectTemplateRequest(tt.req)
+			if err != nil {
```
catsby, line 98: Same change recommendation here, we need to check `tt.wantErr` value here

pkg/server/ptypes/project_template_test.go
```diff
@@ -0,0 +1,153 @@
...
+			pt: &pb.ProjectTemplate{
+				Tags: []string{string(make([]byte, PROJECT_TEMPLATE_TAG_LENGTH+1))},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validation.ValidateStruct(tt.pt, validateProjectTemplateRules(tt.pt)...)
+			if err != nil {
```
catsby, line 146: same `tt.wantErr` check

catsby (changes requested):
We should change the tests to check if `tt.wantErr` is set even if `err` is nil

catsby (approved):

PR #4816, opened 2023-06-20
jobstream: Add additional context for long runner ack times

briancain:
Prior to this commit, we didn't give the user much context for why a job might be stuck. This commit adds more debug help when a job has waited an unusual amount of time to be picked up by a runner. The expected scenario here is that a bad or invalid runner profile was selected, and the runner associated with this profile no longer exists.

There might also be a bug here with our Job timeouts not properly erroring out the job after ~2 minutes since it seems to hang forever in the Queued state waiting for a runner to Ack the job.

Screenshot 2023-06-20 at 2 21 34 PM


jgwhite (approved):
Awesome

PR #4818, opened 2023-06-21
runnerinstall/aws-ecs: Add AddTags permission to ODR IAM policy.

paladin-devops:
The `AddTags` permission is required when Waypoint manages the ALB resource for the AWS ECS plugin. If the user supplies their own ALB, this permission is not needed, because the plugin won't add a tag to indicate that it's managed by Waypoint.

### DIFF
--- a/.changelog/4818.txt
+++ b/.changelog/4818.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+runnerinstall/aws-ecs: Fix ODR policy for AWS ECS runners to enable adding tags
+to an ALB.
+```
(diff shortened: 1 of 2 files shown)


catsby (commented):
> The AWS ECS plugin may need to add tags to an AWS ALB

May?

This is proposed as a bug fix, but it's described as a optional feature. Is there a situation where without this an error will occur? If possible, we should mention that here incase people are out there looking for issues/PRs related to it. It will also help future reviewers or code spelunking to understand why this change was made

izaaklauer (approved):

catsby (approved):

PR #4835, opened 2023-07-06
Add Ons protobuf

paladin-devops:
This PR updates the protocol buffer definition to include CRUD for Add Ons and Add On definitions.

This also deprecates a nested proto message TerraformNocodeModule in favor of using a proto message of the same name that isn't nested, as it is commonly used with Add Ons as well as Project Templates.


pkg/server/proto/server.proto
```diff
@@ -1371,6 +1411,17 @@ message Ref {
       string name = 2;
     }
   }
+
+  message AddOnDefinition {
+    oneof ref {
```
briancain, line 1416: Maybe there's a better name for this other than `ref`? Otherwise, I think the proto will look like `Ref.Ref.<name|id>` which is a little odd, but also we have a lot of odd protobuf nested structures so maybe not the weirdest instance of this

pkg/server/proto/server.proto
```diff
@@ -5420,6 +5472,169 @@ message ProjectTemplate {
...
+// AddOn is an add-on definition rendered for a project
+message AddOn {
+  string id = 1;
+  string name = 2;
+  Ref.Project project = 3;
+  Ref.AddOnDefinition definition = 4;
+  string tf_workspace_name = 5;
+  bytes readme_markdown = 6;
+  repeated string tags = 7;
+  Type type = 8;
```
briancain, line 5611: Add variable doc strings here on these please! It's nice for code docs when interacting with protobuf messages.

pkg/server/proto/server.proto
```diff
@@ -5420,6 +5472,169 @@ message ProjectTemplate {
...
+// AddOnDefinition is the template for an add-on
+message AddOnDefinition {
+  string id = 1;
+  string name = 2;
+  TerraformNocodeModule terraform_nocode_module = 3;
+  string short_summary = 4;
+  string long_summary = 5;
+  bytes readme_markdown_template = 6;
+  repeated string tags = 7;
+  repeated string tf_variable_set_id = 8;
```
briancain, line 5635: Same here, variable docs pls! 🙏🏻

briancain (changes requested):

pkg/server/proto/server.proto
paladin-devops: I renamed it to `identifier`!

pkg/server/proto/server.proto
```diff
@@ -496,10 +496,42 @@ service Waypoint {
...
+  // UpdateAddOnDefinition updates an existing Add On definition
+  rpc UpdateAddOnDefinition(UpdateAddOnDefinitionRequest) returns (UpdateAddOnDefinitionResponse);
+
+  // DeleteAddOnDefinition deletes an existing Add On definition
+  rpc DeleteAddOnDefinition(DeleteAddOnDefinitionRequest) returns (google.protobuf.Empty);
+
+  // CreateAddOn adds an Add On to a project from an Add On definition
+  rpc CreateAddOn(CreateAddOnRequest) returns (CreateAddOnResponse);
+
+  // DestroyAddOn destroys an existing Add On
```
catsby, line 522: Is there a distinction we commonly use to know when to use `Destroy` vs. `Delete` verbs?

pkg/server/proto/server.proto
```diff
@@ -5420,6 +5472,264 @@ message ProjectTemplate {
   }
 }
 
+message TerraformNocodeModule {
+  // The module name for the Terraform no-code module.
```
catsby, line 5476: Is the module name unique, meaning does the "name" include the source repo like `hashicorp/somenocode`? It may be worth noting the expected format here if it's supposed to contain a full repo/name format.

pkg/server/proto/server.proto
```diff
@@ -5420,6 +5472,264 @@ message ProjectTemplate {
...
+  // The type of the Add On
+  AddOnType type = 8;
+}
+
+enum AddOnType {
+  other = 0;
+  database = 1;
+  search = 2;
+  monitor = 3;
+  storage = 4;
```
catsby, line 5617: I believe it's standard to have enums be `UPPERCASE`, which is consistent with other parts of our `server.proto` file usage of enums

- https://protobuf.dev/programming-guides/style/#enums

catsby (commented):
minor comments but seems reasonable

pkg/server/proto/server.proto
briancain: I don't think we have a standard yet, but I like the idea that "destroy" means active resources running need to be destroyed where "delete" is a simple database delete from the Waypoint DB. That's how I've been thinking about it, at least!

paladin-devops: +1 to what @briancain said!

pkg/server/proto/server.proto
paladin-devops: Since the Terraform module source format can take many forms, I've updated this comment since the details of the format are there for each source type.

```
  // The source of the Terraform no-code module.
  // https://developer.hashicorp.com/terraform/language/modules/sources
```

pkg/server/proto/server.proto
paladin-devops: I uppercased all of these, however `OTHER` and `STORAGE` are already in use by [another part of the protobuf definition](https://github.com/hashicorp/waypoint/blob/344b5a4163a9c8490c5325865840c780f7176d8c/pkg/server/proto/server.proto#L1544-L1554). Because the strings are the exact same when uppercased, and they're in the same scope (`hashicorp.Waypoint`), they violate the [rules of enum in protobuf](https://github.com/protocolbuffers/protobuf/issues/5425). For that reason, I've now prefixed all of these with `ADDON_`.

catsby: Good idea 👍

pkg/server/proto/server.proto
catsby: Makes sense 👍

catsby (approved):

PR #4836, opened 2023-07-06
Server state: Add Ons & Add On definitions

paladin-devops:
Add new server state interface methods for Add Ons and Add On definitions. Added to boltdbstate as well, but not implemented.

This PR also contains changes from #4861, for adding support for project-level events.


pkg/serverstate/statetest/test_event.go
```diff
@@ -175,17 +178,118 @@ func TestEvent(t *testing.T, factory Factory, restartF RestartFactory) {
 
 		require.NoError(s.EventPut(ctx, releaseEvent))
 
+		// TODO: Put add-on definition, add-on
```
izaaklauer, line 181: Would be great to see CRUD tests for both add-ons and add-on definitions.

izaaklauer (changes requested):

izaaklauer (approved):

PR #4847, opened 2023-07-18
update ecr-pull to support doing entrypoint injection

kmcgovern-apixio:
This PR makes it so ecr-pull can be used the same way docker-pull can to inject the waypoint entrypoint. Currently this is non-passive because it keeps the same default as `docker-pull` which will inject the entrypoint. disabling the entrypoint injection keeps the previous behavior and does not require the docker or kankio builder (as before)


builtin/docker/plugin.pb.go
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
izaaklauer, line 2: Not sure why these are in this PR, but it's ok!

izaaklauer (changes requested):
Nice improvement! Your approach of invoking the docker pull builder from ecr is clever, and makes sense to me.

I requested one small change, and reran our documentation automation and made a PR to your branch here: https://github.com/kmcgovern-apixio/waypoint/pull/1/files. Once those are in, we can merge this.

builtin/docker/plugin.pb.go
kmcgovern-apixio: originally i had changes to the ecr protobuf but ended up not needing that. this is just a remnant from that!

Issue #4460, opened 2023-01-27
Static runner hanging on aws ecs


izaaklauer:
**Describe the bug**
I have a runner installed on aws ecs using `waypoint runner install`, pointing to the prod HCP waypoint server. 

Currently, every remote operation behaves like this:

```
$ wp deploy

» Deploying acmeapp1...

» Operation is queued waiting for job "01GQTP02MT4PDYE8SCSFSP9CHC". Waiting for runner assignment...
  If you interrupt this command, the job will still run in the background.
```

According to `waypoint job list`, we're waiting for the static runner to take the StartTask job. 

Here are the runner's most recent logs, according to cloudwatch:

```


  | 2023-01-25T18:16:57.441-05:00 | 2023-01-25T23:16:57.441Z [INFO] waypoint.runner.agent.runner: waiting for job assignment
-- | -- | --
  | 2023-01-26T18:11:00.128-05:00 | 2023-01-26T23:11:00.128Z [WARN] waypoint.runner.agent.runner: server down before accepting a job, will reconnect
  | 2023-01-26T18:11:00.128-05:00 | 2023-01-26T23:11:00.128Z [WARN] waypoint.runner.agent.runner: server down before accepting a job, will reconnect
  | 2023-01-26T18:11:00.128-05:00 | 2023-01-26T23:11:00.128Z [WARN] waypoint.runner.agent.runner: server down before accepting a job, will reconnect
  | 2023-01-26T18:11:00.128-05:00 | 2023-01-26T23:11:00.128Z [WARN] waypoint.runner.agent.runner: server down before accepting a job, will reconnect
  | 2023-01-26T18:11:00.128-05:00 | 2023-01-26T23:11:00.128Z [WARN] waypoint.runner.agent.runner: server down before accepting a job, will reconnect
  | 2023-01-26T18:11:00.128-05:00 | 2023-01-26T23:11:00.128Z [WARN] waypoint.runner.agent.runner: server down before accepting a job, will reconnect

```

It's currently 2023-01-27, so it looks like the HCP server went down briefly on `2023-01-26T18:11:00.128-05:00`, and it caused the runner to become stuck. 

I've `tcping'd` the runner's health check port `1234`, and it's still open.

I'd like to get in there and take a thread dump, but it looks like enabling exec on aws ecs is non-trivial, and needs to be set up before the task is launched: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html

[...]

PR #4854, opened 2023-07-21
Not interpreting server NotFound error as server down

izaaklauer:
Fixes https://github.com/hashicorp/waypoint/issues/4460

### How does the runner get stuck?

Here's the failure scenario: 

1: The server and runner are both connected and fine.

2: A new job arrives, and the server decides to assign it to our runner.

3: Our first problem - the server has an issue assigning the job to a runner, and it's a NotFound-type error. We've seen it happen here, while getting config-sourcers. https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/pkg/server/singleprocess/service_runner.go#L661. There's likely another server bug at play here making it semi-regularily return NotFound errors, but fundamentally, returning a NotFound error here is something the server is allowed to do.

4: The runner receives this error, and because it's a NotFound error, it calls `goto RESTART_JOB_STREAM` ([here](https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/runner/accept.go#L283)), which we should only do if we've actually disconnected from the server. This is the bug. Relevant chunk: https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/runner/accept.go#L278-L287

5: The runner follows the goto, and then blocks here forever: https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/runner/accept.go#L224

Why is the runner blocking there, you ask? Here's my best understanding: 

If this were a true disconnect, there is a second goroutine that would also be attempting to reconnect - the recvConfig loop ([here](https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/runner/config.go#L77-L79)). It has sophisticated reconnect logic, and if it were to also have seen the disconnect, it would reconnect, and then increment the runner-wide state version [here](https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/runner/config.go#L313), which would unblock the accept goroutine. @martisah and I verified this through observation.

That isn't happening here, because the recvConfig goroutine hasn't seen an error case, so it never has a reason to modify the runner state.

### How does this fix it?

The primary cause is that the runner's accept goroutine is interpreting any NotFound error from the server as a disconnect event, which it expects all other goroutines will see. That's a bad assumption.

This PR modifies the accept goroutine to exit if it sees a NotFound error, rather than attempt a server reconnect. This raised a new bug: the top-level `AcceptMany` function was interpreting the server sending `NotFound` on the runner job stream as a deregistration event, which is also a bad assumption! The server in fact sends FailedPrecondition ([here](https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/server/boltdbstate/job.go#L516)). 

This fixes that second bug by again interpreting NotFound as a generic error. The AcceptMany NotFound check was introduced to prevent a runner DDOS event ([here](https://github.com/hashicorp/waypoint/commit/d635545a53310ae49dfe8cfc3b29e71edf8b9176)), and this continues to protect against that by moving the sleep before reconnect to the default case.

tl;dr: with this change, if the server responds to the runner with a NotFound error, the Accept loop will log the error, exit, and then be restarted successfully by AcceptMany, ready to accept another job.

### How can I reproduce this bug? 

- Install a runner in ECS
- Create a waypoint project with ODR and configsourcers, pointed to the HCP waypoint server
- Set the static runner's concurrency flag to 1
- Somehow get the server to respond with NotFound on the runner job stream. It's not yet clear the exact steps to reproduce this, but we got there by running some number of `waypoint up`'s, destroying the project, recreating the project, and repeating. 
- At this point, the static runner should log [`server down before accepting a job, will reconnect`](https://github.com/hashicorp/waypoint/blob/918df220a8f3058cac7bcadb3e8ebcd341d35685/internal/runner/accept.go#L196), and then refuse to accept additional jobs. You will notice continued logs from `waypoint.runner.agent.runner.config.watcher`, but none from the accept loop.

### Future considerations

- We intend to push some of the debugging logging we introduced to find this issue
- We should investigate why the server is responding NotFound on the job stream
- We should verify further that runner deregistration still works. In my testing, I observed that with this change, the runner handles one additional job before exiting, but that may be preexisting behavior.


internal/runner/operation_project_destroy.go
```diff
@@ -6,12 +6,13 @@ package runner
...
-	projConfig "github.com/hashicorp/waypoint/internal/config"
-	"github.com/hashicorp/waypoint/internal/core"
-	pb "github.com/hashicorp/waypoint/pkg/server/gen"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	empty "google.golang.org/protobuf/types/known/emptypb"
+
+	projConfig "github.com/hashicorp/waypoint/internal/config"
+	"github.com/hashicorp/waypoint/internal/core"
+	pb "github.com/hashicorp/waypoint/pkg/server/gen"
```
izaaklauer, line 15: Unrelated auto-formatting fix

martisah (approved):

PR #4860, opened 2023-07-27
Set workdir

mikolajsobolewski:
Set the workdir for automate run in gitlab/github


catsby (approved):
I don't see any issue with this change but I'm curious was this omission causing issues somewhere? It would be good to know what problem(s) were caused without this just for understanding

PR #4871, opened 2023-08-08
Client-side CLI changes for OIDC Auth with HCP

xiaolin-ninja:
New command `login -with-hcp` enables users to login to HCP waypoint via the cli.

### DIFF
--- a/internal/cli/login.go
+++ b/internal/cli/login.go
@@ -234,6 +234,66 @@ func (c *LoginCommand) loginToken(ctx context.Context) (string, int) {
 	return convertResp.Token, 0
 }
 
+func (c *LoginCommand) loginHCP(ctx context.Context) (string, int) {
+	// Start our callback server
+	callbackSrv, err := wpoidc.NewCallbackServer()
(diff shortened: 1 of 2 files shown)


internal/cli/login.go
```diff
@@ -467,6 +532,14 @@ func (c *LoginCommand) Flags() *flag.Sets {
 			Usage: "The name of the Kubernetes namespace that has the Waypoint token " +
 				"when using the -from-kubernetes flag.",
 		})
+
+		f.BoolVar(&flag.BoolVar{
+			Name:   "from-hcp",
+			Target: &c.flagHCP,
+			Usage: "Perform the initial authentication after Waypoint is installed. " +
+				"This will enable to user to SSO via HCP." +
```
briancain, line 540:
```suggestion
				"This will enable to user to SSO via HCP. " +
```

briancain (approved):
I got a small comment on the cli help, need to add a space so the two sentences aren't up against each other. Otherwise, makes sense to me! 👍🏻

Issue #4872, opened 2023-08-08
`waypoint runner install -platform=ecs` fails due to policy syntax errors


izaaklauer:
**Describe the bug**

Waypoint runner installation to ECS is currently broken due to a policy syntax error:

```
$ waypoint runner install   -platform=ecs   -server-addr=api.hashicorp.cloud:443   -ecs-runner-image=hashicorp/waypoint -ecs-region=us-east-1
✓ Finished connecting to: api.hashicorp.cloud:443
❌ Installing runner...
✓ Networking setup
✓ EFS ready
✓ Found existing IAM role to use: waypoint-runner-execution-role
❌ Creating IAM task role: waypoint-runner
! Error installing runner: MalformedPolicyDocument: Syntax errors in policy.
  	status code: 400, request id: <redacted>
```

**Steps to Reproduce**
Steps to reproduce the behavior.

Run `waypoint runner install -platform=ecs` against an AWS account that has never had a runner installed before, and is therefore missing the `waypoint-runner` aws role.

**Expected behavior**
Runner installation should work

**Workaround**
Waypoint will still create the `waypoint-runner` role when the runner installation fails. You can add the correct permissions as an inline policy (permissions found here: https://github.com/hashicorp/waypoint/blob/641c6b00366af982f47a7cf5df1693e6c0973739/internal/runnerinstall/ecs.go#L48-L118

Subsequent `runner install` commands should succeed.

**Waypoint Platform Versions**
Additional version and platform information to help triage the issue if
applicable:

* Waypoint CLI Version: v0.11.3
* Waypoint Server Platform and Version: (like `docker`, `nomad`, `kubernetes`) n/a
* Waypoint Plugin: (like `aws/ecs`, `pack`, `azure`) n/a

**Additional context**

PR #4840, opened 2023-07-10
runnerinstall/aws-ecs: Add elasticloadbalancing:DescribeTags permission to ODR IAM policy

izaaklauer:
The DescribeTags permission is required for waypoint to get an ALB by ARN:

This causes failures in `waypoint project destroy` on ECS-backed deployments: 

```
» Destroying deployments for application 'Izaakproject'...
✓ Running deployment destroy v1
✓ Finished destroying ECS deployment
✓ Deleted service Izaakproject-01H50BIGHASH
✓ Deleted ALB Listener
✓ Retrieved target group details
✓ ALB listener for target group is deleted
✓ Target group deleted
These resources were not destroyed for app "Izaakproject":
- task role
- service subnets
- execution role
- route53 record
- log group
- internal security groups
- alb subnets
- cluster

These resources were destroyed for app "Izaakproject":
- service
- alb listener
- external security groups
- application load balancer
- target group
- task definition


» Destroying shared deploy resources for application 'Izaakproject'...
❌ Destroying ecs workspace...
❌ Checking if ALB is managed by Waypoint
! Error destroying project "Izaakproject": rpc error: code = Internal desc = failed to destroy all resources for workspace: rpc error:
code = Internal desc = failed to get ALB with ARN
arn:aws:elasticloadbalancing:us-west-2:<account-id>:loadbalancer/app/izaakproject-dev/awshash tags: AccessDenied: User:
arn:aws:sts::<account-id>:assumed-role/waypoint-runner/a9ef1ce3redacted981fba7 is not authorized to perform:
elasticloadbalancing:DescribeTags because no identity-based policy allows the elasticloadbalancing:DescribeTags action
  	status code: 403, request id: e80eredacted462bd
```

This PR adds the required permission during ecs `runner install`.

PR #4873, opened 2023-08-08
Fix syntax error in ECS policy doc

izaaklauer:
Closes https://github.com/hashicorp/waypoint/issues/4872

Fixes a json syntax error in the ecs inline policy introduced in https://github.com/hashicorp/waypoint/pull/4840 by adding a close quote

### How can I verify this fix?

Run a `waypoint install` against an AWS account without a role named `waypoint-runner`, and verify success:

```
$ ./waypoint runner install   -platform=ecs   -server-addr=api.hashicorp.cloud:443   -ecs-runner-image=hashicorp/waypoint -ecs-region=us-east-1 -ecs-cluster=waypoint
✓ Finished connecting to: api.hashicorp.cloud:443
✓ Runner "01H7BA4XNXE0Y4CEMEHKFQAK3R" installed successfully to ecs
✓ Runner profile "ecs-01H7BA4XNXE0Y4CEMEHKFQAK3R" created successfully.
✓ Networking setup
✓ EFS ready
✓ Found existing IAM role to use: waypoint-runner-execution-role
✓ Created IAM task role: waypoint-runner
✓ Using existing log group
✓ Runner service created
✓ Runner "01H7BA4XNXE0Y4CEMEHKFQAK3R" adopted successfully.
```

### DIFF
--- a/.changelog/4873.txt
+++ b/.changelog/4873.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+runneruninstall/aws-ecs: Fix installing runners in new AWS accounts by fixing an inline policy syntax error.
+```
--- a/internal/runnerinstall/ecs.go
+++ b/internal/runnerinstall/ecs.go
@@ -96,7 +96,7 @@ const odrRolePolicy = `{
         "elasticloadbalancing:DescribeListeners",
         "elasticloadbalancing:DescribeLoadBalancers",
         "elasticloadbalancing:DescribeRules",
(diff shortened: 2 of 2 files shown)


izaaklauer:
Integration tests are failing, but it seems to be transient, and we've observed this happen on other branches before and then pass on main.

Going to try merging for now. If it fails on main, we'll deal with that in a moment.

PR #4879, opened 2023-08-11
Add-on updates

paladin-devops:
Add endpoint to support updating Add-ons.


pkg/serverstate/statetest/test_event.go
```diff
@@ -252,15 +254,49 @@ My favorite add-on README.
 
 		require.NoError(s.EventPut(ctx, addOnCreatedEvent))
 
+		// Update Add-on
+		updatedAddOn, err := s.AddOnUpdate(ctx,
+			&pb.AddOn{
+				Name: "your friendly updated neighborhood add-on",
```
izaaklauer, line 260: 😄

pkg/serverstate/statetest/test_add_on.go
```diff
@@ -218,17 +218,31 @@ My favorite add-on README.
...
+				Name: updatedAddOnName,
+			},
+			&pb.Ref_AddOn{
+				Identifier: &pb.Ref_AddOn_Name{
+					Name: testAddOn.Name,
+				},
+			},
+		)
+		require.NoError(err)
+		require.NotNil(updatedAddOn)
```
izaaklauer, line 233: Can we do another get after the update, to check that the updated properties are being returned?

izaaklauer (changes requested):
Looks good! Request for a test enhancement, but otherwise 🚀

pkg/serverstate/statetest/test_add_on.go
paladin-devops: Added!

catsby (approved):

PR #4882, opened 2023-08-14
Update linux package license to BUSL-1.1

sarahethompson:
This updates the linux package license to BUSL-1.1.
This PR should be merged to main so that the change is included in the next minor release that is cut from main.
License changes are not required on release branches (ongoing patch releases of current minor releases).


evanphx:
Thanks @sarahethompson!

